PR #5, opened 2017-02-14
Improve sessions

dimaryaz:
- Add a session hook that handles authentication errors.
- Stop making API calls just to check the validity of the access token.
- Remove the `QUILT_AUTH_URL` constant.

### DIFF
--- a/quilt/tools/command.py
+++ b/quilt/tools/command.py
@@ -61,7 +61,7 @@ def _save_auth(auth):
 
 def _handle_response(resp, **kwargs):
     if resp.status_code == requests.codes.unauthorized:
-        raise CommandException("Failed to log in. Run `quilt login` again.")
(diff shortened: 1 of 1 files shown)


quilt/tools/command.py
```diff
@@ -60,6 +59,10 @@ def _save_auth(auth):
         os.chmod(file_path, stat.S_IRUSR | stat.S_IWUSR)
         json.dump(auth, fd)
 
+def _handle_response(resp, **kwargs):
+    if resp.status_code == requests.codes.unauthorized:
+        raise CommandException("Failed to log in. Run `quilt login` again.")
```
kevinemoore: Since this exception can happen for reasons other than a login failure, can you change the message to be something more general, like "authentication failed?".

PR #6, opened 2017-02-15
Add install unittests

dimaryaz:
Also fix an uncaught exception when the hash doesn't match.


quilt/test/test_command.py
```diff
@@ -74,22 +66,21 @@ def test_login(self, mock_save, mock_input, mock_open):
 
         mock_open.assert_called_with('%s/login' % command.QUILT_PKG_URL)
 
-        assert responses.calls[0].request.body == "refresh_token=123"
+        assert self.requests_mock.calls[0].request.body == "refresh_token=123"
```
kevinemoore: Can you turn the inline constants into variables, especially where you're checking for matching input/output pairs.

PR #8, opened 2017-02-16
Print out the server's error messages

dimaryaz:
Don't need to handle each error individually anymore.
Also, handle connection failures.

### DIFF
--- a/quilt/tools/command.py
+++ b/quilt/tools/command.py
@@ -62,6 +62,12 @@ def _save_auth(auth):
 def _handle_response(resp, **kwargs):
     if resp.status_code == requests.codes.unauthorized:
         raise CommandException("Authentication failed. Run `quilt login` again.")
+    elif not resp.ok:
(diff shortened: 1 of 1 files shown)


quilt/tools/command.py
```diff
@@ -62,6 +62,12 @@ def _save_auth(auth):
 def _handle_response(resp, **kwargs):
     if resp.status_code == requests.codes.unauthorized:
         raise CommandException("Authentication failed. Run `quilt login` again.")
+    elif not resp.ok:
```
kevinemoore: Are we now exclusively using 200 as the response code for successful operations (i.e. even for deletes, etc. that sometime return other values)?

dimaryaz: It's always been 200; were you thinking about the "no content" codes from nginx?
Regardless, though, `resp.ok` will be true for any 2xx status code.

PR #10, opened 2017-02-22
limit to python < 3.6

akarve:
prevent install from barfing


akarve:
unfortunately `classifiers` doesn't prevent installation. as far as i can tell `classifiers` is an fyi for people who bother to read pypi docs.

akarve:
this is a temporary safety valve until `pytables` works on python 3.6

PR #12, opened 2017-02-22
quilt logout; pylint

akarve:


quilt/tools/command.py
```diff
@@ -143,6 +146,16 @@ def login():
 
     _save_auth(auth)
 
+def logout():
+    """
+    Become anonymous. Useful for testing.
+    """
+    auth_file = os.path.join(BASE_DIR, AUTH_FILE_NAME)
```
dimaryaz: It should also revoke the existing refresh token... Maybe add a TODO for that?

akarve: definitely. i was wondering about that. is there an API call for that?

akarve: unless web and client share a refresh token? in which case it would be weird to log out of the web from the cli.

PR #18, opened 2017-03-01
Readme

kevinemoore:
A new leaf node type, FILE, gives packages a way to include unmodified
files. In this first implementation, build.py only looks for a single
file (README), that is expected to contain a description of the package.

To support raw files, object files are now stored without a
format-specific extension (since raw files are format independent).

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -31,15 +31,12 @@ class Node(object):
     def __init__(self, store, prefix=''):
         self._prefix = prefix
         self._store = store
-        print("NODE: %s %s" % (prefix, store))
 
     def __getattr__(self, name):
         # TODO clean if... up since VALID_NAME_RE no longer allows leading _
-        print("GETATTR NAME=%s" % name)
         if name.startswith('_'):
             raise AttributeError
         path = self._prefix + '/' + name
-        print("GETATTR PATH=%s" % path)
         return self._get_store_obj(path)
 
     def __repr__(self):
@@ -64,7 +61,6 @@ class Node(object):
         self._keys()
         try:
             obj = self._store.get(path)
-            print("OBJ=%s" % obj)
         except KeyError:
             # No such group or table
             raise AttributeError("No such table or group: %s" % path)
@@ -86,7 +82,6 @@ class Node(object):
         """
         keys directly accessible on this object via getattr or .
         """
-        print("KEYS=%s" % self._store.keys(self._prefix))
         return self._store.keys(self._prefix)
 
 class FakeLoader(object):
@@ -95,7 +90,6 @@ class FakeLoader(object):
     """
     def __init__(self, path):
         self._path = path
-        print("FAKELOADER, PATH=%s" % path)
 
     def load_module(self, fullname):
         """
@@ -115,13 +109,11 @@ class PackageLoader(object):
     def __init__(self, path, store):
         self._path = path
         self._store = store
-        print("PKGLOADER, PATH=%s" % path)
 
     def load_module(self, fullname):
         """
         Returns an object that lazily looks up tables and groups.
         """
-        print("LOADING MODULE, FULLNAME=%s" % fullname)
         mod = sys.modules.get(fullname)
         if mod is not None:
             return mod
@@ -131,7 +123,6 @@ class PackageLoader(object):
 
         mod = Node(self._store)
         sys.modules[fullname] = mod
-        print("MOD=%s" % mod)
         return mod
 
 class ModuleFinder(object):
@@ -143,17 +134,12 @@ class ModuleFinder(object):
         """
         Looks up the table based on the module path.
         """
-
-        print("FULLNAME=%s" % fullname)
-        print("PATH=%s" % path)
-        
         if not fullname.startswith(__name__ + '.'):
             # Not a quilt submodule.
             return None
 
         submodule = fullname[len(__name__) + 1:]
         parts = submodule.split('.')
-        print("PARTS=%s" % parts)
 
         if len(parts) == 1:
             for package_dir in PackageStore.find_package_dirs():
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -84,8 +84,7 @@ class Node(object):
         """
         group = self._store.get(self._prefix)
         assert isinstance(group, dict), "{type} {grp}".format(type=type(group), grp=group)
-        assert TYPE_KEY in group
-        group.pop(TYPE_KEY)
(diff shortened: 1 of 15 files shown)


quilt/data.py
```diff
@@ -81,7 +82,12 @@ def _keys(self):
         """
         keys directly accessible on this object via getattr or .
         """
-        return self._store.keys(self._prefix)
+        group = self._store.get(self._prefix)
+        assert isinstance(group, dict), "{type} {grp}".format(type=type(group), grp=group)
+        assert TYPE_KEY in group
+        group.pop(TYPE_KEY)
```
dimaryaz: `del group[TYPE_KEY]`, since you don't care about the return value.

Don't even need the assert above since `del` and `pop` both raise an exception.

quilt/tools/build.py
```diff
@@ -85,11 +89,14 @@ def build_package(username, package, yaml_path):
         raise BuildException("Unable to parse YAML: %s" % yaml_path)
 
     tables = data.get('tables')
-    format = data.get('format', None)
+    format = data.get('format')
+    files = data.get('files')
+    readme = files.get('README') if files else None
     if not isinstance(tables, dict):
```
dimaryaz: We probably want the same check for `files`?

quilt/tools/build.py
```diff
@@ -85,11 +89,14 @@ def build_package(username, package, yaml_path):
         raise BuildException("Unable to parse YAML: %s" % yaml_path)
 
     tables = data.get('tables')
```
dimaryaz: Since we're going to allow packages with no tables (or files), I'd do this:
```
tables = data.get('tables', {})
...
files = data.get('files', {})
readme = files.get('README')
```

quilt/tools/command.py
```diff
@@ -470,32 +470,42 @@ def inspect(package):
...
+                fullname = "/".join([path, name])
+                df = store.get(fullname)
+                assert isinstance(df, pd.DataFrame)
+                info = "shape %s, type \"%s\"" % (df.shape, df.dtypes)
+                print(prefix + name_prefix + ": " + info)
+            elif node_type is NodeType.FILE:
+                fullname = "/".join([path, name])
+                print(prefix + name_prefix + name)
+            else:
+                print(prefix + name_prefix + ": " + str(node))
```
dimaryaz: Should this be an assert?

quilt/tools/store.py
```diff
@@ -105,6 +109,17 @@ def save_df(self, df, name, path, ext, target):
         """
         raise NotImplementedError()
 
+    def save_file(self, srcfile, name, path, target):
+        """
+        Save a (raw) file to the store.
+        """
+        self._find_path_write()
+        filehash = digest_file(srcfile)
+        fullname = name.lstrip('/').replace('/', '.')
```
dimaryaz: What is `fullname` used for? That is, given that contents.json already has nested groups, I was assuming that storing the path wouldn't be necessary.

quilt/tools/store.py
```diff
@@ -105,6 +109,17 @@ def save_df(self, df, name, path, ext, target):
...
+        self._add_to_contents(fullname, filehash, '', path, target)
+        objpath = os.path.join(self._pkg_dir, self.OBJ_DIR, filehash)
+        copyfile(srcfile, objpath)
```
dimaryaz: Should we check if `objpath` already exists, just as a tiny optimization (or not so tiny for GB-size files)?

quilt/tools/store.py
```diff
@@ -161,11 +170,14 @@ def get(self, path):
             ptr = ptr[node]
         node = ptr
 
-        if NodeType(node[TYPE_KEY]) is NodeType.TABLE:
-            hash_list = node['hashes']
-            return self.dataframe(hash_list)
-        else:
+        if NodeType(node[TYPE_KEY]) is NodeType.GROUP:
```
dimaryaz: Make `NodeType(node[TYPE_KEY])` a variable

quilt/tools/store.py
```diff
@@ -298,18 +310,29 @@ def _add_to_contents(self, fullname, objhash, ext, path, target):
...
-                                         q_target=target)
-                          )
+        try:
+            target_type = TargetType(target)
+            if target_type is TargetType.PANDAS:
+                node_type = NodeType.TABLE
+            elif target_type is TargetType.FILE:
+                node_type = NodeType.FILE
+            else:
+                raise NotImplementedError()
```
dimaryaz: Shouldn't happen since `TargetType(target)` will raise a `ValueError`. Maybe make it an `assert False`?

quilt/tools/util.py
```diff
@@ -4,6 +4,8 @@
 
 from appdirs import user_data_dir
 
+from .const import NodeType
```
dimaryaz: Unused?

quilt/tools/store.py
kevinemoore: The store doesn't keep a running path or pointer into the JSON tree so we need to specify the full path when saving each dataframe.

quilt/tools/store.py
kevinemoore: The only way this error happens is if we later add a target type (to the Enum) and forget to implement it here. I like the not implemented for that case, but would you prefer an assert False?

dimaryaz: Yeah; adding it to the enum but not here would be a bug in our code (rather than e.g. a user error - trying to use an unsupported target, etc.), so I think `assert` is more appropriate.

quilt/tools/command.py
```diff
@@ -500,7 +500,7 @@ def _print_node(node, prefix, child_prefix, name, path):
                 fullname = "/".join([path, name])
                 print(prefix + name_prefix + name)
             else:
-                print(prefix + name_prefix + ": " + str(node))
+                raise NotImplementedError("Unsupported NodeType {nt}".format(node_type))
```
dimaryaz: Don't you need `nt=node_type`?

quilt/data.py
```diff
@@ -84,8 +84,7 @@ def _keys(self):
         """
         group = self._store.get(self._prefix)
         assert isinstance(group, dict), "{type} {grp}".format(type=type(group), grp=group)
-        assert TYPE_KEY in group
-        group.pop(TYPE_KEY)
+        del group[TYPE_KEY]
```
dimaryaz: Actually, forgot to ask: is it ok to modify the data here? Does `_store.get` return a new dict each time?

PR #25, opened 2017-03-07
Generate buildfile

kevinemoore:


quilt/test/test_build.py
```diff
@@ -62,3 +62,13 @@ def test_build_parquet(self):
             'Expected dataframes to have same # columns'
         del os.environ["QUILT_PACKAGE_FORMAT"]
         # TODO add more integrity checks, incl. negative test cases
+
+    def test_generate_buildfile(self):
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, 'data')
+        buildfilepath = os.path.join(path, 'build.yml')
```
dimaryaz: You should use the current directory instead - it's a temporary directory that will get removed when the test is done.

quilt/tools/build.py
```diff
@@ -13,6 +14,19 @@ class BuildException(Exception):
     """
     pass
 
+def _pythonize_name(name):
+    safename = re.sub('[^A-Za-z0-9_]+', '_', name)
+    starts_w_number = re.match('^[0-9].*', safename)
+    if starts_w_number:
+        safename = ("a%s" % safename)
```
dimaryaz: "a" doesn't seem right... Though I don't have good ideas.

quilt/tools/build.py
```diff
@@ -13,6 +14,19 @@ class BuildException(Exception):
...
+
+    safename = safename.rstrip('_')
+    safename = safename.lstrip('_')
```
dimaryaz: Just use `strip` instead of `lstrip` and `rstrip`

quilt/tools/build.py
```diff
@@ -13,6 +14,19 @@ class BuildException(Exception):
...
+
+    if not VALID_NAME_RE.match(safename):
+        raise BuildException("Unable to determine a Python-legal name for %s" % name)
```
dimaryaz: Might as well generate a name - `table1`, etc. - and the user can fix it in the `.yml` file

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+                continue
+            try:
+                name, ext = file.split('.')
+                # separate files into tables and raw
+                if ext.lower() in TARGET['pandas']:
+                    tablefiles.append(file)
+                else:
+                    rawfiles.append(file)
+            except ValueError:
+                # File with no extension
```
dimaryaz: Or multiple extensions

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+        path = rel_path.split(os.sep)
+
+        tablefiles = []
+        rawfiles = []
+        for file in files:
+            # skip hidden files
+            if file.startswith('.'):
+                continue
+            try:
+                name, ext = file.split('.')
```
dimaryaz: Use `os.path.splitext` instead. It'll handle no extension or multiple extensions, too.

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
             _build_file(build_dir, store, 'README', rel_path=readme)
+
+def generate_build_file(startpath, outfilename='build.yml'):
+    startbase = os.path.basename(startpath)
+    buildfiles = {startbase : {}}
+    buildtables = {startbase : {}}
+
+    def add_to_buildfiles(path, files):
+        try:
+            safepath = [_pythonize_name(dir) if dir != '.' else '.' for dir in path]
```
dimaryaz: Don't use built-ins like `dir`, `file`, etc. as variables. Even if it's a list comprehension, it'll break the built-in for the rest of the scope (in Python2, at least).

dimaryaz: Why would "." appear there? And does it need to handle "..", too?

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+        except BuildException:
+            warning = "Warning: could not determine a Python-legal name for {path}; skipping."
+            print(warning.format(path=os.sep.join(path)))
```
dimaryaz: `os.path.join`

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+            return
+
+        ptr = buildfiles
+        for dir in path:
+            if dir not in ptr:
+                ptr[dir] = {}
+            ptr = ptr[dir]
```
dimaryaz: I'd use `setdefault` here... Unless you think it's really confusing.

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+        for file in files:
+            fullpath = "/".join(path + [file])
```
dimaryaz: `os.path.join`

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+            try:
+                name, ext = file.split('.')
```
dimaryaz: `os.path.splitext`

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+            except ValueError:
+                # file with no extension
+                name = file
+            ptr[_pythonize_name(name)] = fullpath
+
+    def add_to_buildtables(path, files):
```
dimaryaz: A lot of it seems the same as `add_to_buildfiles`...

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+            ptr[_pythonize_name(name)] = [ext.lower(), fullpath]
+
+    for root, dirs, files in os.walk(startpath):
+        # skip hidden directories
+        for d in dirs:
+            if d.startswith('.'):
+                dirs.remove(d)
+
+        rel_path = os.path.relpath(root, startpath)
+        path = rel_path.split(os.sep)
```
dimaryaz: This might not work on Windows, where both "/" and "\" are separators. There's `os.path.split`, but it only splits once.

quilt/tools/build.py
```diff
@@ -106,3 +120,94 @@ def build_package(username, package, yaml_path):
...
+                rawfiles.append(file)
+
+        if rawfiles:
+            add_to_buildfiles(path, rawfiles)
+
+        if tablefiles:
+            add_to_buildtables(path, tablefiles)
+
+    for contents in [buildfiles, buildtables]:
+        if '.' in contents:
```
dimaryaz: `contents.update(contents.pop('.', {}))`

Though it would be even better to just not add "." in the first place - don't know if that's easy

quilt/tools/command.py
```diff
@@ -534,7 +540,9 @@ def main():
 
     build_p = subparsers.add_parser("build")
     build_p.add_argument("package", type=str, help="Owner/Package Name")
-    build_p.add_argument("path", type=str, help="Path to the Yaml build file")
+    buildpath_group = build_p.add_mutually_exclusive_group(required=True)
+    buildpath_group.add_argument("-d", "--directory", type=str, help="Source file directory")
+    buildpath_group.add_argument("path", type=str, nargs='?', help="Path to the Yaml build file")
```
dimaryaz: Can both `-d` and `path` be specified? Not sure what `nargs` does.

quilt/test/test_build.py
kevinemoore: I don't understand your suggestion. I'm pointing the build file generator at an existing directory (part of the code repository).

quilt/tools/build.py
kevinemoore: Good suggestion, but this method isn't the right place for that logic since we don't know if it's a file, table or other.

quilt/tools/build.py
kevinemoore: .. is filtered out as a 'hidden' directory, otherwise it would show up there too.

quilt/tools/build.py
kevinemoore: os.path.join works great, but the linter doesn't like it. Any suggestions?

quilt/tools/build.py
kevinemoore: That duplication will go away when we move to a single tree build.yml syntax.

quilt/tools/build.py
kevinemoore: Let me know if you think of a way to do it.

quilt/test/test_build.py
dimaryaz: I meant, a unit test shouldn't create or modify anything in the code repository - a temporary directory is a better place for that.

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
         _build_table(build_dir, store, '', tables)
         if readme is not None:
             _build_file(build_dir, store, 'README', rel_path=readme)
+
+def splitext_no_dot(filename):
+    name, ext = os.path.splitext(filename)
+    ext.strip('.')
```
dimaryaz: Remove it?

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+        except BuildException:
+            warning = "Warning: could not determine a Python-legal name for {path}; skipping."
+            print(warning.format(path=os.sep.join(path)))
+            return
+
+        ptr = buildtables
+        for folder in safepath:
+            ptr = ptr.setdefault(folder, {})
+        for file in files:
+            fullpath = "/".join(path + [file])
```
dimaryaz: `os.path.join`

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+    return name, ext.strip('.')
+
+def generate_build_file(startpath, outfilename='build.yml'):
+    startbase = os.path.basename(startpath)
+    buildfiles = {startbase : {}}
+    buildtables = {startbase : {}}
+
+    def add_to_buildfiles(path, files):
+        try:
+            safepath = [_pythonize_name(d) if d != '.' else '.' for d in path]
```
dimaryaz: So how do you end up with a '.'? `path` comes from `os.path.relpath`, right? I can't get it to return anything that starts with "./".

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+                rawfiles.append(file)
+
+        if rawfiles:
+            add_to_buildfiles(path, rawfiles)
+
+        if tablefiles:
+            add_to_buildtables(path, tablefiles)
+
+    def remove_node(node, contents):
+        if node in contents:
```
dimaryaz: Did you try the `pop` and `update` code? Does it not work?

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+            warning = "Warning: could not determine a Python-legal name for {path}; skipping."
+            print(warning.format(path=os.sep.join(path)))
+            return
+
+        ptr = buildfiles
+        for folder in path:
+            ptr = ptr.setdefault(folder, {})
+
+        for file in files:
+            fullpath = os.path.join(*safepath, file)
```
dimaryaz: That's not actually valid - hence the linter error. I guess there isn't really a simple way, but this works:
`os.path.join(os.path.join(*safepath), file)`

or

`os.path.join(*(safepath + [file]))`

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+            name, ext = splitext_no_dot(file)
+            ptr[_pythonize_name(name)] = fullpath
```
dimaryaz: What if `_pythonize_name` raises a `BuildException` here?

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+            return
+
+        ptr = buildtables
+        for folder in safepath:
+            ptr = ptr.setdefault(folder, {})
+        for file in files:
+            fullpath = "/".join(path + [file])
+            name, ext = splitext_no_dot(file)
+            # pythonize name
+            ptr[_pythonize_name(name)] = [ext.lower(), fullpath]
```
dimaryaz: Same here

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+            if file.startswith('.'):
+                continue
+            try:
+                name, ext = splitext_no_dot(file)
+                # separate files into tables and raw
+                if ext.lower() in TARGET['pandas']:
+                    tablefiles.append(file)
+                else:
+                    rawfiles.append(file)
+            except ValueError:
```
dimaryaz: Won't happen anymore

quilt/tools/build.py
```diff
@@ -105,3 +118,95 @@ def build_package(username, package, yaml_path):
...
+                # File with no extension
+                rawfiles.append(file)
+
+        if rawfiles:
+            add_to_buildfiles(path, rawfiles)
```
dimaryaz: I think the logic would be simpler if `add_to_buildfiles`/`add_to_buildtables` took a single file instead of a list of files. That is, I don't see much benefit in building a list of files, then having the helper functions loop over the files.

I guess that would mean calculating `safepath` lots of times... But I'd move it out of the helper functions and put it after `path = rel_path.split(os.sep)`. Then just pass a joined safe path into the helpers.

In fact, even the `ptr = ptr.setdefault(folder, {})` logic could be outside the helper. When doing `os.walk`, you know exactly when you need to create the nested dict.

Anyways, just suggestions - don't need to rewrite everything if you don't agree.

kevinemoore: That would be true if we had the same tree for files and tables (technically we could do that, but we'd have empty nodes, which throws off the build logic). The plan is to move to a single tree in build.yml and this code can get a lot cleaner.

quilt/tools/build.py
```diff
@@ -13,6 +14,18 @@ class BuildException(Exception):
     """
     pass
 
+def _pythonize_name(name):
+    safename = re.sub('[^A-Za-z0-9_]+', '_', name)
```
akarve: is this pattern duplication of `VALID_NAME_RE` that we have elsewhere?

quilt/tools/build.py
```diff
@@ -105,3 +118,85 @@ def build_package(username, package, yaml_path):
...
+        tablefiles = []
+        rawfiles = []
+        for file in files:
+            # skip hidden files
+            if file.startswith('.'):
+                continue
+
+            name, ext = splitext_no_dot(file)
+            # separate files into tables and raw
+            if ext.lower() in TARGET['pandas']:
```
akarve: TARGET keys should probably be read from an enum or at least constants

quilt/tools/build.py
```diff
@@ -105,3 +118,85 @@ def build_package(username, package, yaml_path):
...
+    for contents in [buildfiles, buildtables]:
+        for node in ['.', '..']:
+            if node in contents:
+                for key in contents[node]:
+                    contents[key] = contents[node][key]
+                del contents[node]
+
+    contents = dict(files=buildfiles, tables=buildtables)
+    buildfilepath = os.path.join(startpath, outfilename)
+    with open(buildfilepath, 'w') as outfile:
```
akarve: do we want to warn before clobbering if there's something at buildfilepath?

quilt/tools/store.py
```diff
@@ -21,7 +21,7 @@
 except ImportError:
     SparkSession = None
 
-from .const import FORMAT_HDF5, FORMAT_PARQ, FORMAT_SPARK, NodeType, TargetType
+from .const import FORMAT_HDF5, FORMAT_PARQ, FORMAT_SPARK, NodeType, TargetType, PACKAGE_DIR_NAME
```
akarve: why are some camelCase and others ALL_CAPS?

PR #29, opened 2017-03-09
Fix package browsing

kevinemoore:
Make the top-level of contents a group node.


kevinemoore:
True, we could do that, but we'd break existing packages. Are we still early enough that that's a better option?

dimaryaz:
Probably fine to break... Though you could do both, and add a TODO to remove the code that adds the key when reading the package.

quilt/tools/store.py
```diff
@@ -315,6 +316,7 @@ def _add_to_contents(self, fullname, objhash, ext, path, target):
         leaf = ipath.pop()
 
         ptr = contents
+        ptr.setdefault(TYPE_KEY, NodeType.GROUP.value)
```
dimaryaz: Why not just make sure `get_contents` always returns a dict with the key? If the package doesn't have any objects, this won't run, right?

quilt/tools/store.py
```diff
@@ -129,6 +129,8 @@ def get_contents(self):
         try:
             with open(self._path, 'r') as contents_file:
                 contents = json.load(contents_file)
+                # Make the top-level a GROUP node
+                contents[TYPE_KEY] = NodeType.GROUP.value
```
dimaryaz: Just move this to the end of the function, before `return` - I think that should take care of everything?

quilt/tools/store.py
dimaryaz: This is unnecessary now, right?

PR #31, opened 2017-03-09
ArrowPackageStore

kevinemoore:
Adding a package store class for Parquet packages using the pyarrow library. This also cleans up PackageStore class types by making the package format an Enum.


quilt/test/test_push.py
```diff
@@ -21,14 +21,12 @@ def find_object_hashes(contents):
...
     """
     for key, obj in contents[CHILDREN_KEY].items():
-        print("KEY=%s, OBJ=%s" % (key, obj))
-        if key == TYPE_KEY:
-            continue
         obj_type = NodeType(obj[TYPE_KEY])
         if obj_type is NodeType.TABLE or obj_type is NodeType.FILE:
-            yield from obj["hashes"]
+            hash_it = iter(obj['hashes'])
+            yield next(hash_it)
```
dimaryaz: I'm pretty sure you need a loop here and below? Doesn't this just return the first item?

quilt/tools/store.py
```diff
@@ -490,32 +482,86 @@ def dataframe(self, hash_list):
...
+        # Move the build file to the object store and rename it to
+        # its hash
+        objpath = self._object_path(filehash)
+        os.rename(storepath, objpath)
+
+    def dataframe(self, hash_list):
+        """
+        Creates a DataFrame from a set of objects (identified by hashes).
+        """
+        assert len(hash_list) == 1, "Multi-file DFs not supported in HDF5."
```
dimaryaz: Arrow?

PR #32, opened 2017-03-10
Fix a file naming conflict

dimaryaz:
If the object name is the same as an existing user name or "obj", the build will
fail because it will try to overwrite a directory. Make sure we're using a path
that won't conflict with existing paths.

### DIFF
--- a/quilt/tools/store.py
+++ b/quilt/tools/store.py
@@ -49,6 +49,7 @@ class PackageStore(object):
     PACKAGE_FILE_EXT = '.json'
     BUILD_DIR = 'build'
     OBJ_DIR = 'objs'
+    TMP_OBJ_DIR = 'objs/tmp'
 
     @classmethod
     def find_package_dirs(cls, start='.'):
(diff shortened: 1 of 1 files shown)


quilt/tools/store.py
```diff
@@ -396,13 +396,12 @@ def save_df(self, df, name, path, ext, target):
         """
         self._find_path_write()
         buildfile = name.lstrip('/').replace('/', '.')
-        storepath = os.path.join(self._pkg_dir, buildfile)
+        storepath = self._object_path('.' + buildfile)
```
kevinemoore: This seems to move the storage of temporary files to the CWD. Is that right? I don't think we should do that. If the process gets interrupted, we should try our best to clean up, but if even that fails, it'd be nice if the mess was left in a different directory. Maybe we should have a directory explicitly for builds?

dimaryaz: No, it just moves it from `quilt_packages/` to `quilt_packages/objs/` - that's what `self._object_path` returns.

I added a leading "." to make sure it won't conflict with hashes.

kevinemoore: Oh, that makes sense. It seems cleaner to keep the temporary files in another directory, but it's nice to be able to reuse _object_path. Although it's unlikely to come up very often, we should try to make those temporary files easy to locate and delete in case something goes wrong. The object directory gets big very fast.

PR #34, opened 2017-03-10
Switch from `$type` to `type` and `children`

dimaryaz:


quilt/tools/command.py
```diff
@@ -489,9 +489,9 @@ def _print_children(children, prefix, path):
     def _print_node(node, prefix, child_prefix, name, path):
         name_prefix = u"─ "
         if isinstance(node, dict):
-            node_type = NodeType(node.pop(TYPE_KEY))
+            node_type = NodeType(node["type"])
```
kevinemoore: Shouldn't we still use constants for these? That way the linter will catch errors if we make typos.

quilt/test/test_install.py
```diff
@@ -11,7 +11,7 @@
 from six import assertRaisesRegex
 
 from quilt.tools import command
-from quilt.tools.const import HASH_TYPE, TYPE_KEY, NodeType
+from quilt.tools.const import HASH_TYPE, NodeType
```
akarve: what dictates case for constants? (camel vs all-caps)

quilt/tools/command.py
dimaryaz: Constants are kind of a workaround: the real solution is to use classes, and only use JSON for serialization/deserialization - but that's a bigger change.

PR #52, opened 2017-03-27
Don't create the package file until we're actually saving the contents

dimaryaz:
This makes `init_contents` and `clear_contents` unnecessary.

### DIFF
--- a/quilt/tools/command.py
+++ b/quilt/tools/command.py
@@ -436,11 +436,10 @@ def install(session, package, hash=None, version=None, tag=None):
     if pkghash != hash_contents(response_contents):
         raise CommandException("Mismatched hash. Try again.")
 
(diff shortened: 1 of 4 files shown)


quilt/tools/command.py
```diff
@@ -436,11 +436,10 @@ def install(session, package, hash=None, version=None, tag=None):
     if pkghash != hash_contents(response_contents):
         raise CommandException("Mismatched hash. Try again.")
 
-    pkgobj = store.create_package(owner, pkg, PackageFormat.HDF5)
+    pkgobj = store.create_package(owner, pkg)
```
kevinemoore: An alternative to setting format to the default in Package.__init__ would be to set it in create_package. I think we can assume all packages are created by create_package, but not necessarily by build_package.

PR #58, opened 2017-03-29
Build.yml supports arbitrary keys and kwargs

akarve:


quilt/tools/build.py
```diff
@@ -149,10 +160,7 @@ def _generate_contents(dir_path):
                 nodename, ext = splitext_no_dot(name)
                 ext = ext.lower()
                 rel_path = os.path.relpath(path, startpath)
-                if ext in TARGET['pandas']:
-                    data = [ext, rel_path]
-                else:
-                    data = ['raw', rel_path]
+                data = dict(file=rel_path)
```
kevinemoore: This doesn't include the transform. Shouldn't it also have:
transform=ext? (in the dict)

quilt/test/build.yml
```diff
@@ -1,7 +1,11 @@
 ---
 contents:
   dataframes:
-    csv: [csv, data/10KRows13Cols.csv]
-    tsv: [tsv, data/10KRows13Cols.tsv]
-    xls: [xlsx, data/10KRows13Cols.xlsx]
-  README: [raw, data/README.md]
+    csv: 
+      file: data/10KRows13Cols.csv
```
kevinemoore: OK. I see from the code below that the transform can be inferred from the file extension. This convention might be confusing to users wanting to include raw files. As I understand it, no transform specified means "choose the default", but to explicitly choose to include a file as is, they must specify "transform: raw."

quilt/tools/build.py
kevinemoore: Including the transform as a key here will (I think) make it clearer to users what build is doing to create the package and (hopefully) make it clearer how they can modify the build settings.

quilt/test/build.yml
akarve: your understanding is correct. no transform = infer and fallback to raw. transform = always use transform.

akarve: i looked at specifying a transform through auto-generate carefully and it's simpler not to touch that right now. i would need to pipe more function args to do it properly. i also think that a clean build.yml will give people a better first experience and they can learn about transform later.

dimaryaz: The fact that the dataframe has the same name as the transform might actually confuse people. (And really, it should be the name of the contents.)

As for omitting the transform... I guess I can see both sides of the argument, but I kinda like having to explicitly set it.

quilt/tools/build.py
```diff
@@ -15,6 +18,10 @@ class BuildException(Exception):
     """
     pass
 
+def _is_internal_node(node):
+    # all of an internal nodes children are dicts
+    return all([isinstance(x, dict) for x in node.values()])
```
dimaryaz: Don't need `[` and `]`

quilt/tools/build.py
```diff
@@ -28,46 +35,50 @@ def _pythonize_name(name):
...
+        for child_name, child_table in node.items():
+            if not isinstance(child_name, str) or not VALID_NAME_RE.match(child_name):
+                raise StoreException("Invalid table name: %r" % child_name)
+            _build_node(build_dir, package, name + '/' + child_name, child_table)
+    else: # leaf node
+        rel_path = node.get('file')
+        if not rel_path:
+            raise BuildException("Leaf nodes must define a 'file' key")
         path = os.path.join(build_dir, rel_path)
+        ignore, ext = splitext_no_dot(rel_path)
```
dimaryaz: Use `_` instead of `ignore`. Or `[1]` to only get the `ext`.

quilt/tools/build.py
```diff
@@ -28,46 +35,50 @@ def _pythonize_name(name):
...
+                transform = ext
+            else:
+                transform = 'raw'
 
-        if ext == 'raw':
+        if transform == 'raw':
             print("Copying %s..." % path)
             package.save_file(path, name, rel_path)
         else:
+            user_kwargs = {k : node[k] for k in node.keys() if k not in ('file', 'transform')}
```
dimaryaz: Don't need `.keys()`

quilt/test/build.yml
akarve: agree. backlogging.

PR #61, opened 2017-03-30
Monkey-patch pandas to force the use of old pickle format

dimaryaz:


quilt/tools/package.py
```diff
@@ -161,8 +161,16 @@ def save_df(self, df, name, path, ext, target):
 
         # Serialize DataFrame to chosen format
         if enumformat is PackageFormat.HDF5:
-            with pd.HDFStore(storepath, mode='w') as store:
-                store[self.DF_NAME] = df
+            # HACK: Force the use of old pickle to ensure Python 2/3 compatibility.
+            from pandas.compat import cPickle
+            old_protocol = cPickle.HIGHEST_PROTOCOL
```
akarve: since goal is to restore old_protocol, shouldn't we get it from pandas if possible? e.g. this will break if their source code changes.

akarve: also, out of curiosity, do we have any guarantees of atomicity here?

dimaryaz: Not sure what you mean by "get it from pandas"? Patch the pandas code rather than pickle itself? It's hard to do cause they just use the `HIGHEST_PROTOCOL` variable in `pandas/io/pickle.py`.

As for atomicity: nope. This is definitely fragile, but I don't think there's much we can do.

PR #62, opened 2017-03-31
Avoid building data frames when accessing nodes

kevinemoore:
This fixes a major performance bug that caused Quilt to read all
dataframe children of a group node into memory when the node __repr__
method was invoked. Now, package.get is split into two phases: get,
returns just the node and get_obj(node) returns the actual object
(dataframe, file or group).

### DIFF
--- a/quilt/tools/package.py
+++ b/quilt/tools/package.py
@@ -235,7 +235,8 @@ class Package(object):
 
     def get(self, path):
         """
-        Read a group or object from the store.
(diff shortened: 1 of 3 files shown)


quilt/tools/package.py
```diff
@@ -251,6 +252,16 @@ def get(self, path):
                     owner=self._user,
                     pkg=self._package))
         node = ptr
```
dimaryaz: Just `return ptr`

PR #63, opened 2017-03-31
Keep the package contents in memory

dimaryaz:


quilt/tools/package.py
```diff
@@ -204,34 +205,26 @@ def get_contents(self):
...
-                    contents = RootNode(contents.children, PackageFormat.default.value)
-        except IOError:
-            contents = RootNode(dict(), PackageFormat.default)
-
-        return contents
+        assert self._contents is not None
+        return self._contents
 
-    def clear_contents(self):
+    def load_contents(self):
```
kevinemoore: This should be private and maybe automatically called by the constructor.

quilt/tools/store.py
```diff
@@ -68,13 +68,15 @@ def get_package(self, user, package):
...
+                pkg = Package(user=user,
+                              package=package,
+                              path=path,
+                              pkg_dir=package_dir)
+                pkg.load_contents()
+                return pkg
         return None
 
-    def create_package(self, user, package, pkgformat=PackageFormat.default):
+    def create_package(self, user, package, contents):
```
kevinemoore: Let's have an option to create a package without contents so users can create packages interactively.

PR #68, opened 2017-04-06
Refactor the PackageStore

dimaryaz:
- Each PackageStore now represents a single directory
- Object paths moved from package to store
- Building and installing always creates a store in the current directory

### DIFF
--- a/quilt/tools/store.py
+++ b/quilt/tools/store.py
@@ -5,7 +5,7 @@ import os
 import re
 
 
-from .const import STORE_DIR_NAME
+from .const import PACKAGE_DIR_NAME
 from .core import RootNode
 from .package import Package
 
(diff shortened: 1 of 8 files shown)


quilt/tools/store.py
```diff
@@ -29,11 +29,13 @@ class and its subclasses abstract file formats, file naming and
     PACKAGE_FILE_EXT = '.json'
     BUILD_DIR = 'build'
     OBJ_DIR = 'objs'
+    TMP_OBJ_DIR = 'objs/tmp'
 
-    def __init__(self, startpath='.'):
-        self._start_dir = startpath
+    def __init__(self, store_dir='.'):
```
dimaryaz: Do you think the arg should be the `quilt_packages` directory itself, or its parent? The former seems more logical - but then do we enforce that it has the correct name?

dimaryaz:
Ok, I'll rename it to `package_directories`, and include it in the `store_dir`.

Also, turns out, my PR breaks imports - and somehow, unit tests don't catch that. Oops.

PR #70, opened 2017-04-11
Add unit tests for imports and multiple package directories

dimaryaz:


quilt/test/test_import.py
```diff
@@ -0,0 +1,76 @@
...
+
+        import quilt.data.foo.package
+
+        from quilt.data.foo import package
+        from quilt.data.foo.package import dataframes
+        from quilt.data.foo.package import README
+
+        # Good attributes of imported packages
+
+        package.dataframes
```
akarve: seems useful to do some type introspection here. we know these aren't throwing but how do we know, e.g., that a group is a group, a df is a df, etc.?

quilt/test/test_import.py
```diff
@@ -0,0 +1,76 @@
...
+
+        with self.assertRaises(AttributeError):
+            package.dataframes.foo
+
+        with self.assertRaises(AttributeError):
+            package.dataframes.csv.foo
+
+        # Bad imports
+
+        with self.assertRaises(ImportError):
```
kevinemoore: Should these two cases have different error messages? Can the test harness validate the proper errors are returned?

dimaryaz: Do you mean, bad user vs good user with a bad package name? I don't think that's actually possible: the package loader returns either a success or a failure, with no messages.

PR #74, opened 2017-04-12
Use `GroupNode` and `LeafNode` wrapper objects for both groups and leaf nodes

dimaryaz:
Makes things more consistent, and allows accessing leaf nodes without
having to load the dataframe into memory.

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -18,7 +18,6 @@ import imp
 import os.path
 import sys
 
-from .tools.const import PACKAGE_DIR_NAME
 from .tools.core import GroupNode
 from .tools.package import PackageException
 from .tools.store import PackageStore
@@ -31,12 +30,11 @@ class DataNode(object):
     in HDFStore's `root`.
     """
     def __init__(self, package, prefix=''):
+        assert not prefix.endswith('/')
+
         self._package = package
         self._prefix = prefix
-
-        # Make sure the node exists and is a group.
-        node = self._package.get(prefix)
-        assert isinstance(node, GroupNode)
+        self._node = self._package.get(self._prefix)
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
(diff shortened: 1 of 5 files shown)


quilt/data.py
```diff
@@ -79,6 +72,16 @@ def __dir__(self):
         # https://mail.python.org/pipermail/python-ideas/2011-May/010321.html
         return sorted(set((dir(type(self)) + list(self.__dict__) + self._keys())))
 
+    def _is_group(self):
+        return isinstance(self._node, GroupNode)
+
+    def _is_df(self):
+        return not isinstance(self._node, GroupNode)
+
+    def _df(self):
```
dimaryaz: I don't know if `_df()` is the way to go, though: we already have a `_dfs()` method which returns children paths, so it's confusing. Also, do we consider file nodes to be dataframes, too?

I want it to be concise, so not know what to call it. We could even do just `()`, though that's hacky.

akarve: `_df()` is OK for now. suggest `_dfkeys()` and `_groupkeys()` for the other methods; will need to `README.md` with new APIs

akarve:
looks good but missing one thing: changes to README.md to reflect new APIs

quilt/data.py
```diff
@@ -56,38 +54,43 @@ def __getattr__(self, name):
...
 
     def __dir__(self):
         # https://mail.python.org/pipermail/python-ideas/2011-May/010321.html
         return sorted(set((dir(type(self)) + list(self.__dict__) + self._keys())))
 
-    def _dfs(self):
+    def _is_group(self):
+        return isinstance(self._node, GroupNode)
+
+    def _is_df(self):
```
kevinemoore: We shouldn't report a node as a df unless it's actually a meaningful Pandas DF. Files, e.g., should not be dfs. Do we need a new concept, e.g., LeafNode that specifies merely that a node has no children?

dimaryaz: Yeah, that's what I was wondering. I can rename `df` to `leaf`

quilt/data.py
```diff
@@ -31,12 +30,11 @@ class DataNode(object):
     in HDFStore's `root`.
```
dimaryaz: Oh, yes

akarve:
Maybe DataNode === Leaf. because we only have Data at the leaves. Everything else is really a PackageNode or a GroupNode in my mind.

quilt/data.py
dimaryaz: So technically, `_df()` doesn't make sense, either. `_get()`?

kevinemoore: No, _df() makes sense, but it should only apply to nodes that are DataFrames. It should either not be available or throw an exception for groups or files I would think.

akarve:
blame me for `()`. i asked for it. it's just an alias for `data()`. as a quilt power user i want to type less.

PR #75, opened 2017-04-14
Lazily load parquet libraries

dimaryaz:


quilt/tools/package.py
```diff
@@ -37,9 +20,9 @@
 
 
 class ParquetLib(Enum):
+    SPARK = 'pyspark'
```
kevinemoore: Why change the order here? Not alphabetical?

dimaryaz: I rely on the order later, in `for lib in ParquetLib`; I changed it so `SPARK` is preferred, like before.

PR #77, opened 2017-04-15
quilt log: reverse chrono sort, local time

akarve:
and make pylint happy. er. happi-er.

### DIFF
--- a/quilt/test/test_command.py
+++ b/quilt/test/test_command.py
@@ -13,7 +13,7 @@ from six import assertRaisesRegex
 
 from quilt.tools import command, store
 from .utils import QuiltTestCase, patch
-from ..tools.const import UTCF
+from ..tools.const import SERVER_TIME_F
 
 class CommandTest(QuiltTestCase):
     def test_push_invalid_package(self):
@@ -150,7 +150,7 @@ class CommandTest(QuiltTestCase):
         logs_url = "%s/api/log/%s/%s/" % (command.QUILT_PKG_URL, owner, package)
         resp = dict(logs=[dict(
             hash=pkg_hash,
-            created=str(datetime.now().strftime(UTCF)),
+            created=str(datetime.utcnow().strftime(SERVER_TIME_F)),
             author=owner)])
         print("MOCKING URL=%s" % logs_url)
         self.requests_mock.add(responses.GET, logs_url, json.dumps(resp))
--- a/quilt/test/test_command.py
+++ b/quilt/test/test_command.py
@@ -13,6 +13,7 @@ from six import assertRaisesRegex
 
 from quilt.tools import command, store
 from .utils import QuiltTestCase, patch
+from ..tools.const import UTCF
 
 class CommandTest(QuiltTestCase):
     def test_push_invalid_package(self):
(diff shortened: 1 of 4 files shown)


quilt/tools/util.py
```diff
@@ -70,3 +74,37 @@ def file_to_str(fname):
     with open(fname, 'rU') as f:
         data = f.read()
     return data
+
+
+def make_comparator(func):
```
dimaryaz: Not used?

quilt/tools/const.py
```diff
@@ -10,6 +10,7 @@ class TargetType(Enum):
     PANDAS = 'pandas'
     FILE = 'file'
 
+UTCF = "%Y-%m-%dT%H:%M:%S"
```
dimaryaz: The format is not actually specific to UTC, so I'd call it something else...

(Also, apparently, `strftime` supports `%F`, while `strptime` doesn't - despite the docs saying the formats are identical. WTF...)

quilt/tools/command.py
```diff
@@ -200,13 +200,14 @@ def log(session, package):
...
     )
-
+    # table header
     format_str = "%-64s %-19s %s"
-
-    print(format_str % ("Hash", "Created", "Author"))
-    for entry in response.json()['logs']:
+    print(format_str % ("Hash", "Pushed", "Author"))
+    # table body
+    logs = sorted(response.json()['logs'], key=lambda x: parse_utc(x['created']), reverse=True)
```
dimaryaz: The server already sorts this by time (and by ID, in case of duplicate time), so it's kind of silly to sort it again. I'd rather change the server code - or just reverse the order here.

quilt/test/test_command.py
```diff
@@ -149,7 +150,7 @@ def _mock_logs_list(self, owner, package, pkg_hash):
         logs_url = "%s/api/log/%s/%s/" % (command.QUILT_PKG_URL, owner, package)
         resp = dict(logs=[dict(
             hash=pkg_hash,
-            created=str(datetime.now()),
+            created=str(datetime.now().strftime(UTCF)),
```
dimaryaz: That's actually going to be a local time, not UTC (not that it really matters in a unit test)

quilt/tools/util.py
```diff
@@ -70,3 +74,37 @@ def file_to_str(fname):
...
+    """
+    return datetime.strptime(utc, UTCF)
+
+# http://stackoverflow.com/questions/4770297/python-convert-utc-datetime-string-to-local-datetime
+def utc2local(utc):
+    """
+    convert utc time to local time; beware of stackoverflow magic :-/
+    """
+    epoch = time.mktime(utc.timetuple())
+    offset = datetime.fromtimestamp(epoch) - datetime.utcfromtimestamp(epoch)
```
dimaryaz: Damn, there has to be a way that doesn't involve arithmetic. Dates and times suck...

akarve: the proper way of doing this is even worse. tzinfo objects.

quilt/tools/const.py
```diff
@@ -10,6 +10,7 @@ class TargetType(Enum):
     PANDAS = 'pandas'
     FILE = 'file'
 
+SERVER_TIME_F = "%Y-%m-%dT%H:%M:%S"
```
kevinemoore: Because this acts as a coordination point between client and server, it should go in core.py instead of const.py (so eventually the server could use it to guarantee that it delivers dates in the expected format).

quilt/tools/util.py
kevinemoore: The timezone examples for Python 2 on stackoverflow look pretty ugly so this is probably a good compromise. This example (Python 3 only) though looks cleaner to me:

from datetime import timezone
utc_dt.replace(tzinfo=timezone.utc).astimezone(tz=None)

http://stackoverflow.com/questions/4563272/convert-a-python-utc-datetime-to-a-local-datetime-using-only-python-standard-lib

akarve:
closing in favor of the simpler https://github.com/quiltdata/quilt/pull/79

PR #79, opened 2017-04-18
quilt log prints reverse chrono, handles timestamps instead of datetime

akarve:


quilt/test/test_command.py
```diff
@@ -149,7 +149,7 @@ def _mock_logs_list(self, owner, package, pkg_hash):
         logs_url = "%s/api/log/%s/%s/" % (command.QUILT_PKG_URL, owner, package)
         resp = dict(logs=[dict(
             hash=pkg_hash,
-            created=str(datetime.now()),
+            created=str(time.time()),
```
dimaryaz: The server will be returning a `float`, so no need to `str` it here. (Unless you think floats in JSON are a bad idea; they do lose some precision, just not enough to matter in our case.)

quilt/tools/command.py
```diff
@@ -203,10 +204,11 @@ def log(session, package):
 
     format_str = "%-64s %-19s %s"
 
-    print(format_str % ("Hash", "Created", "Author"))
-    for entry in response.json()['logs']:
-        # TODO: convert "created" to local time.
-        print(format_str % (entry['hash'], entry['created'], entry['author']))
+    print(format_str % ("Hash", "Pushed", "Author"))
+    for entry in reversed(response.json()['logs']):
+        ugly = datetime.fromtimestamp(float(entry['created']))
```
dimaryaz: Same here; `float` should be unnecessary.

PR #80, opened 2017-04-18
Add child keys back to repr, clarify class names

akarve:


quilt/data.py
```diff
@@ -50,7 +50,8 @@ def __hash__(self):
     def __repr__(self):
         finfo = self._package.get_path()[:-len(PackageStore.PACKAGE_FILE_EXT)]
         pinfo = self._prefix
-        return "<%s %r %r>" % (self.__class__.__name__, finfo, pinfo)
+        kinfo = '\n'.join(self._keys()) if hasattr(self, '_keys') else ''
```
dimaryaz: `hasattr` is kinda terrible; just append extra info in the subclass.

README.md
```diff
@@ -62,7 +62,7 @@ Let's see what's in the `wine` package:
 >>> wine.quality.red._keys()
```
dimaryaz: You'll need to update the lines above - at least `wine`, but maybe remove the calls to `._keys()` now that `repr` shows them.

PR #83, opened 2017-04-20
Make `quilt.install` and other commands work

dimaryaz:
- Move `main` to `main.py`, just for clarity
- Create the session automatically
- Import useful functions from `tools.command` into `__init__.py`


akarve:
Merging since Dima is in another time zone.

PR #85, opened 2017-04-21
Command args to generate build file without building the package

kevinemoore:
This addition supports a common pattern: generating a build file for a
collection of sources then editing that build file (e.g. to specify
package format or per-source transforms) before building a package.

### DIFF
--- a/quilt/tools/command.py
+++ b/quilt/tools/command.py
@@ -171,7 +171,19 @@ def logout():
     else:
         print("Already logged out.")
 
-def build(package, path, directory=None, generate=None):
+def generate(directory):
+    """
+    Generate a build-file for quilt build from a directory of
+    source files.
+    """
+    try:
+        buildfilepath = generate_build_file(directory)
+    except BuildException as builderror:
+        raise CommandException(str(builderror))
+
+    print("Generated build-file %s." % (buildfilepath))
+
+def build(package, path, directory=None):
     """
     Compile a Quilt data package
     """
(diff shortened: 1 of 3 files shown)


quilt/tools/command.py
```diff
@@ -581,6 +587,7 @@ def main():
     build_p.add_argument("package", type=str, help="Owner/Package Name")
     buildpath_group = build_p.add_mutually_exclusive_group(required=True)
     buildpath_group.add_argument("-d", "--directory", type=str, help="Source file directory")
```
dimaryaz: `-d` and `-g` now have the same help text, which is kind of weird. Would be nice to have more explanation.

Though I feel like there has to be a better way to organize the options. We now have:
- Generate the build file
- Build the package
- Both of the above

The most "logical" solution would be to have an argument for 1, another argument for 2, or both arguments if you want both. But of course it doesn't work well if we want 2 to be the default with no options.

kevinemoore: I suggested -d <directory> and -g (no arg, generate only). Or, we could make generate (without building) a separate command, quilt generate. Generating the build file actually doesn't need a username/package so maybe that would be clearer.

akarve: Agree `quilt generate -d DIRPATH` And error if it will clobber an existing build for

quilt/tools/command.py
```diff
@@ -171,14 +171,28 @@ def logout():
...
+    source files.
+    """
+    try:
+        buildfilepath = generate_build_file(directory)
+    except BuildException as builderror:
+        raise CommandException(str(builderror))
+
+    print("Generated build-file %s." % (buildfilepath))
+
 def build(package, path, directory=None):
```
akarve: this line blows up `TypeError: build() got an unexpected keyword argument 'auto'`

PR #87, opened 2017-04-21
Use file extension to avoid duplicates

kevinemoore:
Appending the file extension to node names for file nodes if the base name appears more than once in the packages and would otherwise produce a name conflict.

### DIFF
--- a/quilt/tools/build.py
+++ b/quilt/tools/build.py
@@ -54,7 +54,6 @@ def _build_node(build_dir, package, name, node, target='pandas'):
                                      (transform, rel_path, target))
         else: # guess transform if user doesn't provide one
             ignore, ext = splitext_no_dot(rel_path)
-            ext = ext.lower()
             if ext in TARGET[target]:
                 transform = ext
                 print("Inferring 'transform: %s' for %s" % (transform, rel_path))
@@ -168,6 +167,7 @@ def splitext_no_dot(filename):
     without the '.' (e.g., csv instead of .csv)
     """
     name, ext = os.path.splitext(filename)
+    ext = ext.lower()
     ext.strip('.')
     return name, ext.strip('.')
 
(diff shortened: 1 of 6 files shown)


quilt/tools/build.py
```diff
@@ -198,19 +199,26 @@ def _generate_contents(dir_path):
             try:
                 safename = _pythonize_name(nodename)
                 if safename in contents:
-                    print("Warning: duplicate name %r in %s." % (safename, dir_path))
-                    continue
+                    if ext:
+                        safename = "{name}_{ext}".format(name=safename, ext=ext)
```
dimaryaz: Theoretically, _this_ could be a duplicate, too.

quilt/tools/main.py
```diff
@@ -29,10 +29,14 @@ def main():
...
 
+    generate_p = subparsers.add_parser("generate")
+    generate_p.add_argument("directory", help="Source file directory")
+    generate_p.set_defaults(func=command.generate)
+
     build_p = subparsers.add_parser("build")
     build_p.add_argument("package", type=str, help="Owner/Package Name")
     buildpath_group = build_p.add_mutually_exclusive_group(required=True)
-    buildpath_group.add_argument("-d", "--directory", type=str, help="Source file directory")
+    buildpath_group.add_argument("-a", "--auto", type=str, help="Source file directory")
```
dimaryaz: I guess this isn't a new bug, I wonder if `-a` should just be a boolean - and then `path` could be used as either build file path or directory?

quilt/tools/build.py
dimaryaz: Ignoring this one? :)

akarve: do we know that they have same parent directory? if so then OS should enforce unique naming. what am i missing?

quilt/tools/main.py
```diff
@@ -36,7 +36,6 @@ def main():
     build_p = subparsers.add_parser("build")
     build_p.add_argument("package", type=str, help="Owner/Package Name")
     buildpath_group = build_p.add_mutually_exclusive_group(required=True)
-    buildpath_group.add_argument("-a", "--auto", type=str, help="Source file directory")
     buildpath_group.add_argument("path", type=str, nargs='?', help="Path to the Yaml build file")
```
akarve: help needs a bit of love here in light of new automatic behavior

PR #89, opened 2017-04-24
Add pytest --integration option and tests, add install --force

akarve:


quilt/test/test_install.py
```diff
@@ -125,3 +125,15 @@ def _mock_package(self, package, pkg_hash, contents, hashes):
     def _mock_s3(self, pkg_hash, contents):
         s3_url = 'https://example.com/%s' % pkg_hash
         self.requests_mock.add(responses.GET, s3_url, contents)
+
+# execute outside of mock scope so we get real network
```
dimaryaz: Uh, unit tests should never hit real servers. You can add "integration test", but it should be a separate file - not something that runs as a part of `pytest`.

quilt/test/test_install.py
```diff
@@ -125,3 +125,15 @@ def _mock_package(self, package, pkg_hash, contents, hashes):
...
+def test_prod_install():
+    # public package
+    command.install('akarve/days', force=True)
+    from quilt.data.akarve import days
+    df = days.names.data()
+    assert(df.loc[3]['Day'] == 'Wednesday')
+    # TODO clean up after this test b/c tests shouldn't have side-effects
```
dimaryaz: Create a temporary directory, `chdir`, run `install`, remove the directory - same as what `QuiltTestCase` does.

quilt/tools/command.py
```diff
@@ -410,10 +410,13 @@ def install(package, hash=None, version=None, tag=None):
...
-        overwrite = input("Overwrite? (y/n) ")
-        if overwrite.lower() != 'y':
-            return
+        if not force:
+            print("{owner}/{pkg} already installed.".format(owner=owner, pkg=pkg))
+            overwrite = input("Overwrite? (y/n) ")
+            if overwrite.lower() != 'y':
+                return
+        else:
+            pass
```
dimaryaz: ?

quilt/tools/command.py
```diff
@@ -410,10 +410,13 @@ def install(package, hash=None, version=None, tag=None):
     existing_pkg = store.get_package(owner, pkg)
 
     if existing_pkg is not None:
```
dimaryaz: Just make it `if existing_pkg is not None and not force`. Or you could skip the `existing_pkg` lookup if `force=True`.

quilt/test/test_install.py
akarve: how about a `pytest --integration` flag; off by default? and then we can decorate integration tests and skip without it.

quilt/test/test_install.py
akarve: will probably end up splitting QuiltTestCase into two classes; one with file mocking; and subclass with scratch dir.

quilt/tools/command.py
akarve: i'm trying to rush and get article ready so many other things

akarve: :hurtrealbad:

quilt/test/utils.py
```diff
@@ -17,28 +17,36 @@
...
+
         self.session_patcher = patch('quilt.tools.command._get_session', requests.Session)
         self.session_patcher.start()
 
         self.requests_mock = responses.RequestsMock(assert_all_requests_are_fired=False)
         self.requests_mock.start()
 
     def tearDown(self):
-        self.requests_mock.stop()
+        super(QuiltTestCase, self).tearDown()
```
dimaryaz: Just to be safe, let's do teardown in the reverse order - so call `super` at the end.

quilt/test/utils.py
```diff
@@ -17,28 +17,36 @@
 import requests
 import responses
 
-
-class QuiltTestCase(unittest.TestCase):
+class QuiltTestCaseBasic(unittest.TestCase):
```
dimaryaz: I'd prefer `BasicQuiltTestCase`, or something than ends in `TestCase`

PR #94, opened 2017-04-26
Implement some very basic package saving functionality

dimaryaz:
Nodes in `data.py` now have their own package hierarchy instead of wrapping core nodes. This is required if we're going to allow moving them around - possibly even moving/copying from one package to another. The downside here is, the node no longer knows which package or path it has (it could have none or multiple).

There's no user-friendly way to create new nodes; that will come later.

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -40,6 +40,8 @@ class Node(object):
     def __repr__(self):
         return self._class_repr()
 
+    def _to_core_node(self):
+        raise NotImplementedError
 
 class GroupNode(Node):
     """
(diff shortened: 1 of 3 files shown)


quilt/data.py
```diff
@@ -20,86 +20,94 @@
...
-        if isinstance(other, self.__class__):
-            return self._package == other._package and self._prefix == other._prefix
-        return NotImplemented
-
-    def __ne__(self, other):
-        return not self == other
-
-    def __hash__(self):
-        return hash((self._package, self._prefix))
+    def _class_repr(self):
```
dimaryaz: This is now a totally private method - I wonder if I should use `__` or something to distinguish it from "public" methods that start with `_`?

dimaryaz:
Yeah, I keep thinking about the node hierarchies... I want to get rid of the duplication, too. It would work for groups, but package nodes (and I think data nodes, too) need to have pointers to the package itself, so can't really be serialized as JSON. So not sure what to do.

akarve:
guys i'm late to the party, one thing missing: docs to README.md. all new features need that. will file a separate issue.

PR #95, opened 2017-05-03
Make format a property of data nodes rather than the package.

dimaryaz:
Allows us to have mixed-format packages. Required for package editing to work correctly.

(We'll need to update the server first, though.)

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -93,7 +93,7 @@ class PackageNode(GroupNode):
         return "<%s %r>" % (self.__class__.__name__, finfo)
 
     def _to_core_node(self):
-        return core.RootNode(self._core_children(), self._package.get_contents().format)
+        return core.RootNode(self._core_children())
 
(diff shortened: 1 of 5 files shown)


quiltdata:
Yeah the server should warn if the min version is less than some number or doesn't exist. Maybe the header should contain a version string; worth logging the version to mixpanel as well.

dimaryaz:
Oh damn, didn't think about that. So besides adding a version string, I could split this into two changes: 1) understand new format, and 2) generate new format. Then we could release 1 immediately, and wait with 2. Requires more work, though.

akarve:
why did this fail ci/push?

akarve:
ci/push barfed on all platforms

akarve:
@dimaryaz by generate the format do you mean migrate the package? i don't think we need to worry about that. we should 5xx if we can't deliver the the package to the client though.

PR #97, opened 2017-05-05
GroupNode data

kevinemoore:
Access the collective data of all data children at a GroupNode.

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -47,6 +47,11 @@ class GroupNode(Node):
     """
     Represents a group in a package. Allows accessing child objects using the dot notation.
     """
+    def __init__(self, node):
+        super(GroupNode, self).__init__()
+        self._node = node
+
+    
     def __repr__(self):
         pinfo = super(GroupNode, self).__repr__()
         kinfo = '\n'.join(self._keys())
@@ -61,6 +66,12 @@ class GroupNode(Node):
         every child key referencing a dataframe
         """
         return [name for name, child in self._items() if not isinstance(child, GroupNode)]
+
+    def data(self):
+        """
+        Returns the contents of the node: a dataframe or a file path.
+        """
+        return self._package.get_obj([getattr(self, name)._node for name in self._data_keys()])
 
     def _group_keys(self):
         """
@@ -84,9 +95,10 @@ class PackageNode(GroupNode):
     """
     Represents a package.
     """
-    def __init__(self, package):
-        super(PackageNode, self).__init__()
+    def __init__(self, package, node):
+        super(PackageNode, self).__init__(node)
         self._package = package
+        self._node = node
 
     def _class_repr(self):
         finfo = self._package.get_path()[:-len(PackageStore.PACKAGE_FILE_EXT)]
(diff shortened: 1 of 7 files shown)


quilt/data.py
```diff
@@ -62,6 +67,12 @@ def _data_keys(self):
         """
         return [name for name, child in self._items() if not isinstance(child, GroupNode)]
 
+    def data(self):
+        """
+        Returns the contents of the node: a dataframe or a file path.
+        """
+        return self._package.get_obj([getattr(self, name)._node for name in self._data_keys()])
```
dimaryaz: Having to call `getattr` seems silly. You could use `self._items()` and then check the types... Or maybe add a `_data_items()` method.

quilt/data.py
```diff
@@ -84,9 +95,10 @@ class PackageNode(GroupNode):
     """
     Represents a package.
     """
-    def __init__(self, package):
-        super(PackageNode, self).__init__()
+    def __init__(self, package, node):
+        super(PackageNode, self).__init__(node)
         self._package = package
+        self._node = node
```
dimaryaz: Already set by `GroupNode`

quilt/tools/package.py
```diff
@@ -249,6 +249,14 @@ def get_obj(self, node):
 
         if isinstance(node, TableNode):
             return self._dataframe(node.hashes, pkgformat)
+        elif isinstance(node, list):
+            if pkgformat != PackageFormat.PARQUET:
+                raise PackageException("Group data access is only supported for Parquet tables.")
+
+            hash_list = []
```
dimaryaz: You can make this more Pythonic:
```hash_list = [h for child in node for h in child.hashes]```

File: quilt/tools/core.py [hunk 4]
```diff
@@ -37,6 +37,24 @@ def __init__(self, children):
         assert isinstance(children, dict)
         self.children = children
 
+    def preorder_tablenodes(self):
+        stack = []
+        output = []        
+
+        node = self
+
+        while node:
```
dimaryaz: Nit: I'd prefer something like this:

```
stack = [node]
while stack:
  node = stack.pop()
  ...
```
I think that's more standard

quilt/data.py
```diff
@@ -60,6 +82,14 @@ def _data_keys(self):
         """
         return [name for name, child in self._items() if not isinstance(child, GroupNode)]
 
+    def data(self):
```
dimaryaz: What if you have a subgroup named "data"?

quilt/data.py
```diff
@@ -65,6 +65,8 @@ def data(self):
 class GroupNode(DataNode):
     """
     Represents a group in a package. Allows accessing child objects using the dot notation.
+    Warning: calling data on a large dataset may exceed local memory capacity in Python (Only
```
dimaryaz: "calling _data()"?

quilt/tools/package.py
```diff
@@ -217,8 +217,11 @@ def get_obj(self, node):
         if isinstance(node, TableNode):
             return self._dataframe(node.hashes, node.format)
         elif isinstance(node, GroupNode):
-            hash_list = [h for c in node.preorder_tablenodes() for h in c.hashes]
-            return self._dataframe(hash_list, PackageFormat.PARQUET)
+            try:
+                hash_list = [h for c in node.preorder_tablenodes() for h in c.hashes]
+                return self._dataframe(hash_list, PackageFormat.PARQUET)
+            except ValueError:
```
dimaryaz: Which line is throwing the `ValueError`? If possible, keep just that one line inside of `try`, to minimize false positives.

PR #98, opened 2017-05-05
When installing a package, don't redownload existing objects.

dimaryaz:
Verify that their hashes are correct, and skip the download.

Use a temporary file until the download is complete, so we don't end up with
partially downloaded objects.

Sort the objects when downloading: otherwise, without a consistent order,
it looks as if existing objects are being downloaded again.

### DIFF
--- a/quilt/test/test_install.py
+++ b/quilt/test/test_install.py
@@ -112,7 +112,7 @@ class InstallTest(QuiltTestCase):
         self._mock_package('foo/bar', contents_hash, contents, [obj_hash])
         self._mock_s3(obj_hash, tabledata)
 
-        with assertRaisesRegex(self, command.CommandException, "Mismatched hash"):
+        with assertRaisesRegex(self, command.CommandException, "hashes do not match"):
             command.install('foo/bar')
 
         assert not os.path.exists('quilt_packages/foo/bar.json')
@@ -151,7 +151,7 @@ class InstallTest(QuiltTestCase):
 
         self._mock_tag('foo/bar', 'latest', contents_hash)
         self._mock_package('foo/bar', contents_hash, contents, file_hash_list)
-        # Don't mock the first file, since it's not supposed to be downloaded.
+        # Don't mock file0, since it's not supposed to be downloaded.
         self._mock_s3(file_hash_list[1], file_data_list[1])
         self._mock_s3(file_hash_list[2], file_data_list[2])
 
--- a/quilt/test/test_install.py
+++ b/quilt/test/test_install.py
@@ -11,7 +11,16 @@ from six import assertRaisesRegex
 
 from ..tools import command
 from ..tools.const import HASH_TYPE
-from ..tools.core import decode_node, encode_node, hash_contents, GroupNode, TableNode, FileNode
+from ..tools.core import (
+    decode_node,
+    encode_node,
+    hash_contents,
+    FileNode,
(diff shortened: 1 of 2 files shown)


dimaryaz:
Sorry, do you mean, display the hashes that are being downloaded or skipped?

quilt/tools/command.py
```diff
@@ -489,18 +489,27 @@ def install(package, hash=None, version=None, tag=None, force=False):
...
-    for idx, (download_hash, url) in enumerate(iteritems(response_urls)):
-        print("Downloading object %d/%d..." % (idx + 1, total))
+    for idx, (download_hash, url) in enumerate(sorted(iteritems(response_urls))):
+        print("Downloading %s (%d/%d)..." % (download_hash, idx + 1, total))
+
+        local_filename = store.object_path(download_hash)
+        if os.path.exists(local_filename):
+            file_hash = digest_file(local_filename)
+            if file_hash == download_hash:
+                print("Object already exists; skipping.")
```
akarve: "Object" is a loaded word. Can we use "fragment" or "file"?

We should also use quilt vocabulary wherever possible:
"Fragment already installed; skipping"

quilt/tools/command.py
```diff
@@ -513,12 +522,14 @@ def install(package, hash=None, version=None, tag=None, force=False):
                         progress.update(length_remaining - response.raw.length_remaining)
                         length_remaining = response.raw.length_remaining
 
-        file_hash = digest_file(local_filename)
+        file_hash = digest_file(temp_path)
         if file_hash != download_hash:
-            os.remove(local_filename)
+            os.remove(temp_path)
             raise CommandException("Mismatched hash! Expected %s, got %s." %
```
akarve: "Fragment hashes do not match..." We should avoid "!" whenever possible.

quilt/test/test_install.py
```diff
@@ -108,6 +117,51 @@ def test_bad_object_hash(self):
...
+
+        # file1 exists, but has the wrong contents.
+        with open('quilt_packages/objs/{hash}'.format(hash=file_hash_list[1]), 'w') as fd:
+            fd.write("Garbage")
+
+        # file2 does not exist.
+
+        self._mock_tag('foo/bar', 'latest', contents_hash)
+        self._mock_package('foo/bar', contents_hash, contents, file_hash_list)
+        # Don't mock the first file, since it's not supposed to be downloaded.
```
akarve: Minor `# Don't mock file0...`

quilt/test/test_install.py
```diff
@@ -108,6 +117,51 @@ def test_bad_object_hash(self):
...
+
+        # file0 already exists.
+        with open('quilt_packages/objs/{hash}'.format(hash=file_hash_list[0]), 'w') as fd:
+            fd.write(file_data_list[0])
+
+        # file1 exists, but has the wrong contents.
+        with open('quilt_packages/objs/{hash}'.format(hash=file_hash_list[1]), 'w') as fd:
+            fd.write("Garbage")
+
+        # file2 does not exist.
```
akarve: where do you check that file2 would be installed?

dimaryaz: That's the usual case that's already covered by other tests, so I'm not doing anything to test it here. The point of this test is to make sure things don't get downloaded unnecessarily.

PR #104, opened 2017-05-15
Use `command.build` to build a package from an existing one

dimaryaz:
Replaces `PackageNode._save`

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -95,10 +95,6 @@ class PackageNode(GroupNode):
     def _to_core_node(self):
         return core.RootNode(self._core_children())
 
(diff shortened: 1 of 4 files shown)


akarve:
also there are some lingering bugs. basically if i reassign column names it doesn't show up in the data frame. if i reassign cell values it does show up but i cannot persist those changes via build. see here https://github.com/quiltdata/examples/blob/master/dev/modifying%20packages.ipynb

dimaryaz:
Yeah, editing dataframes is not supported yet. That's also why I don't want to document this in the README just yet - it's not quite finished...

PR #105, opened 2017-05-16
A couple README fixes

dimaryaz:
- Remove the hdf5-dev requirement from Linux instructions (no longer necessary), but add ssl-dev.
- Fix a '=' in a build.yml example.
- URLs don't work in code blocks, so remove the silly-looking `[` and `]` in build.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,9 @@ A data package is an abstraction that encapsulates and automates data preparatio
 1. `pip install quilt`
 
 ## Linux
-1. `sudo apt-get install libhdf5-serial-dev`
+1. Install the OpenSSL headers:
+   * Ubuntu: `sudo apt-get install libssl-dev`
+   * Fedora: `sudo dnf install openssl-devel`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,9 @@ A data package is an abstraction that encapsulates and automates data preparatio
 1. `pip install quilt`
 
 ## Linux
-1. `sudo apt-get install libhdf5-serial-dev`
+1. Install the OpenSSL headers:
+   * Ubuntu: `sudo apt-get install libssl-dev`
```
kevinemoore: What will it take to avoid this dependency? These are only needed if users are compiling, right? Can we remove this if everything installs from a wheel?

dimaryaz: The `cryptograpy` package, used by `requests`, only has wheels for Win and Mac: https://pypi.python.org/pypi/cryptography

Honestly, though, running `apt-get install`/`dnf install` is so easy that I wouldn't worry about it.

PR #113, opened 2017-05-29
make integration tests safer, more general

akarve:
Require developer to explicitly supply URL for integration tests. (Now, even if CI goes off the rails, the code doesn't have enough info to run the integration tests.) Allows integration testing against `stage` as well as `prod`.

example: `pytest --integration https://pkg-stage.quiltdata.com`

### DIFF
--- a/quilt/test/test_integration.py
+++ b/quilt/test/test_integration.py
@@ -1,7 +1,7 @@
 """
 Integration test of installation from known public package(s)
 """
-from unittest.mock import patch
+from .utils import patch
 
 import pytest
 
--- a/quilt/test/test_integration.py
+++ b/quilt/test/test_integration.py
@@ -9,15 +9,15 @@ from ..tools import command
 from .utils import BasicQuiltTestCase
 from .integration import skip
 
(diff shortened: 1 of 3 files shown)


quilt/test/test_integration.py
```diff
@@ -3,17 +3,25 @@
...
 @skip
 class IntegrationTest(BasicQuiltTestCase):
+    """only runs if --integration ENV_URL is provided to pytest"""
     def test_prod_install(self):
+        # this is done in preference to os.environ.get since there is no
+        # guarantee that command.QUILT_PKG_URL == os.environ.get['QUILT_PKG_URL']
+        # WARNING: not thread safe
+        old_env = command.QUILT_PKG_URL
+        env = pytest.config.getoption("--integration")
+        command.QUILT_PKG_URL = env
```
dimaryaz: You should use `with patch(...):` here. Simpler, and guarantees that the old value gets restored at the end.

akarve:
CI is failing because unittest.mock does not exist in Python 2.7. I will cook up a solution and resub.

PR #114, opened 2017-05-30
Implement _set for adding files, dataframes, and groups

dimaryaz:


dimaryaz:
Oh, you're right, will do that.

PR #115, opened 2017-06-06
Only upload the objects that the server doesn't have yet

dimaryaz:
Use the new API endpoint to upload the objects before
creating the package, and upload only what is necessary.

### DIFF
--- a/quilt/test/test_push.py
+++ b/quilt/test/test_push.py
@@ -12,17 +12,6 @@ from quilt.tools.core import find_object_hashes
 
 from .utils import QuiltTestCase
 
-def upload_urls(contents):
-    all_hashes = set(find_object_hashes(contents))
-
-    urls = {}
-    for blob_hash in all_hashes:
-        template = "https://example.com/{owner}/{pkg}/{hash}"
-        urls[blob_hash] = template.format(owner='foo',
-                                          pkg='bar',
-                                          hash=blob_hash)
-    return urls
-
 
 class PushTest(QuiltTestCase):
     """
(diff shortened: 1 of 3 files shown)


quilt/test/test_push.py
```diff
@@ -37,24 +26,37 @@ def test_push(self):
...
+        all_hashes = set(find_object_hashes(contents))
+        for blob_hash in all_hashes:
+            head_url = "https://example.com/head/{owner}/{hash}".format(
+                owner='foo', hash=blob_hash)
+            put_url = "https://example.com/put/{owner}/{hash}".format(
+                owner='foo', hash=blob_hash)
+            self._mock_get_blob('foo', blob_hash, head_url, put_url)
+            self._mock_s3(head_url, put_url)
+
+        self._mock_put_package('foo/bar', pkg_hash)
```
akarve: so this is just an end-to-end test of pushing a package? just wondering if it's specific to this PR. given that we now dedupe, it's also slightly suspicous that the test guarantees unique hashes with `set`. since `set` is part of the `command.py.push` logic, wouldn't it be better to rely on its presence there?

quilt/test/test_push.py
```diff
@@ -37,24 +26,37 @@ def test_push(self):
...
+        blob_url = '%s/api/blob/%s/%s' % (command.QUILT_PKG_URL, user, blob_hash)
+        self.requests_mock.add(responses.GET, blob_url, json.dumps(dict(
+            head=s3_head,
+            get='',
+            put=s3_put,
         )))
 
+    def _mock_put_package(self, package, pkg_hash):
+        pkg_url = '%s/api/package/%s/%s' % (command.QUILT_PKG_URL, package, pkg_hash)
+        self.requests_mock.add(responses.PUT, pkg_url, json.dumps(dict()))
```
akarve: all of the PUTs in `command.py.push` take args so wondering why `dict` is empty here.

quilt/tools/command.py
```diff
@@ -306,10 +306,44 @@ def push(package):
...
+
+    obj_hashes = sorted(set(find_object_hashes(pkgobj.get_contents())))
+    total = len(obj_hashes)
+
+    headers = {
+        'Content-Encoding': 'gzip'
+    }
+
+    for idx, obj_hash in enumerate(obj_hashes):
+        print("Uploading %s (%d/%d)..." % (obj_hash, idx + 1, total))
```
akarve: oh do we lose filename feedback to the user here? ideally we'd have that. because people won't recognize hashes. if non-trivial then backlog that feature.

quilt/tools/command.py
```diff
@@ -352,7 +368,6 @@ def push(package):
             hash=pkghash
         ))
     )
-    assert response.ok # other responses handled by _handle_response
 
     url = "https://quiltdata.com/package/%s/%s" % (owner, pkg)
     print("Push complete. Your package is live:\n%s" % url)
```
akarve: minor improvement: "usr/pkg is live:..."

quilt/tools/main.py
```diff
@@ -42,10 +42,8 @@ def main():
 
     push_p = subparsers.add_parser("push")
     push_p.add_argument("package", type=str, help="Owner/Package Name")
-    push_p.set_defaults(func=command.push)
-
-    push_p = subparsers.add_parser("push")
-    push_p.add_argument("package", type=str, help="Owner/Package Name")
+    push_p.add_argument("--reupload", action="store_true",
+                        help="Re-upload all fragments, even the ones the server already has")
```
akarve: "Re-upload all fragments (even if fragment is already in registry)"

quilt/tools/command.py
```diff
@@ -306,10 +306,44 @@ def push(package):
...
+                hash=obj_hash
+            )
+        )
+        contents = response.json()
+
+        if not reupload:
+            s3_response = requests.head(contents['head'])
+            if s3_response.ok:
+                print("Fragment already uploaded; skipping.")
+                continue
```
akarve: i guess python doesn't have named loops... very minor nit: this codeblock is long enough to where a comment would clarify what this `continue` applies to.

quilt/test/test_push.py
dimaryaz: But this is the part that mocks the server responses. It wouldn't make sense to mock the same request twice.

(It's not specific to this PR, though - I just had to update it because of the PR.)

quilt/test/test_push.py
dimaryaz: It's the mock. The `dict()` is the server's response - which is not used by the CLI anymore.

quilt/tools/command.py
dimaryaz: We may be able to get filenames out of the JSON; I'll take a look at that later

PR #125, opened 2017-06-26
Small fixes to make editing packages safer & easier

kevinemoore:
Adding a method to add groups to child nodes and putting a test on
setattr to validate that a GroupNode’s children are valid package nodes.

### DIFF
--- a/quilt/data.py
+++ b/quilt/data.py
@@ -45,7 +45,7 @@ class Node(object):
         if name.startswith('_') or isinstance(value, Node):
             super(Node, self).__setattr__(name, value)
         else:
-            raise ValueError("{val} is not a valid package node".format(val=value))
+            raise AttributeError("{val} is not a valid package node".format(val=value))
 
 class DataNode(Node):
     """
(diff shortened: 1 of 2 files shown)


quilt/data.py
```diff
@@ -41,6 +41,12 @@ def _class_repr(self):
     def __repr__(self):
         return self._class_repr()
 
+    def __setattr__(self, name, value):
+        if name.startswith('_') or isinstance(value, Node):
```
dimaryaz: Is there a reason to allow `_`? Does something break otherwise?

quilt/data.py
```diff
@@ -41,6 +41,12 @@ def _class_repr(self):
...
+            super(Node, self).__setattr__(name, value)
+        else:
+            raise ValueError("{val} is not a valid package node".format(val=value))
```
dimaryaz: Actually, sorry, I was wrong earlier; I think `AttributeError` is the appropriate exception here.

quilt/data.py
kevinemoore: Yes, DataNodes store a reference to the package in _package.

PR #126, opened 2017-06-28
Add support for creating public packages

dimaryaz:
Also test the package creation before uploading all of the objects.

### DIFF
--- a/quilt/tools/main.py
+++ b/quilt/tools/main.py
@@ -42,6 +42,8 @@ def main():
 
     push_p = subparsers.add_parser("push")
     push_p.add_argument("package", type=str, help="Owner/Package Name")
+    push_p.add_argument("--public", action="store_true",
+                        help="Create a public package (not valid if the package already exists")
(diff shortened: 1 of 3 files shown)


quilt/tools/main.py
```diff
@@ -42,6 +42,8 @@ def main():
 
     push_p = subparsers.add_parser("push")
     push_p.add_argument("package", type=str, help="Owner/Package Name")
+    push_p.add_argument("--public", action="store_true",
+                        help="Create a public package (not valid if the package already exists")
```
kevinemoore: The help comment here is a bit confusing. How about, "Create or update a public package (fails if the package exists and is private)."

dimaryaz: I wanted to emphasize "create" cause there's no point in using it for existing packages. But sure, I'll change it.

kevinemoore: Is it worth changing the interface to require --public for all public pushes to be consistent and to emphasize to users that they're making the data public? Or is that overkill?

PR #132, opened 2017-07-11
Handle invalid user/package names better

dimaryaz:


quilt/tools/command.py
```diff
@@ -148,6 +148,12 @@ def _parse_package(name):
             raise ValueError
     except ValueError:
         raise CommandException("Specify package as owner/package_name.")
+
+    try:
+        PackageStore.check_name(owner, pkg)
+    except StoreException as ex:
+        raise CommandException(str(ex))
```
akarve: so the upstream CommandException has a good error message that says "illegal characters" or something?

dimaryaz: The error message stays the same ("Invalid package name: ...") - the only difference is, the `main` method catches `CommandException` and just prints out the message so you don't see the stack trace.

akarve: the message should tell people how to do things right. so "Package name contains illegal characters. Rename package with alphanumerics and underscore..." or whatever. I'm not sure if we support numerics in trailing positions.

dimaryaz: Ok, I'll do that in a separate PR. "Rename package" wouldn't be appropriate if e.g. you're running `quilt install ...`; in that case, you just got the name wrong, and there's nothing to rename.

PR #137, opened 2017-07-14
Test for pyspark more explicitly

dimaryaz:
Otherwise, in Python3, any subsequent exceptions get blamed on the failed import, which is misleading.


dimaryaz:
(I think it used to be that way originally, but I "improved" it later - and it turns out to be a bad idea.)

PR #140, opened 2017-07-19
Add a package delete command

dimaryaz:


quilt/tools/command.py
```diff
@@ -640,6 +640,29 @@ def access_remove(package, user):
 
     session.delete("%s/api/access/%s/%s/%s" % (QUILT_PKG_URL, owner, pkg, user))
 
+def package_delete(package, force=False):
```
akarve: we probably shouldn't support a `force` flag on this

quilt/tools/command.py
```diff
@@ -640,6 +640,29 @@ def access_remove(package, user):
...
+    """
+    Delete a package from the server.
+
+    Irreversibly deletes the package along with its history, tags, versions, etc.
+    """
+    owner, pkg = _parse_package(package)
+
+    if not force:
+        answer = input(
+            "This will irreversibly delete the package and all of its history!\n" +
```
akarve: "Are you sure you want to delete this package and its entire history?"

akarve:
Just tested this. A few issues:
1) Deleting package does not remove package from `quilt_packages`. Not strictly a bug but kinda' confusing.
2) If I `quilt delete akarve/test` and then `quilt push akarve/test` it skips all of the fragments. Leading me to believe that the server-side contents were not deleted. Possible that `stage` has not been updated as of my test, causing this bug.

PR #152, opened 2017-08-02
Implement parallel uploads

dimaryaz:
- Use multiple threads to upload fragments
- Retry on errors
- Display a single progress bar instead of one per fragment

### DIFF
--- a/quilt/tools/command.py
+++ b/quilt/tools/command.py
@@ -13,11 +13,14 @@ import stat
 import subprocess
 import sys
 import time
+from threading import Thread, Lock
 
 from packaging.version import Version
 import pandas as pd
 import pkg_resources
 import requests
+from requests.packages.urllib3.util.retry import Retry
+from requests.adapters import HTTPAdapter
 from six import iteritems, string_types
 from tqdm import tqdm
 
@@ -42,6 +45,8 @@ else:
     AUTH_FILE_NAME = "auth-%.8s.json" % hashlib.md5(QUILT_PKG_URL.encode('utf-8')).hexdigest()
 
 CHUNK_SIZE = 4096
+
+PARALLEL_UPLOADS = 20
 
 VERSION = pkg_resources.require('quilt')[0].version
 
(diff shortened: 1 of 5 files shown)


akarve:
tested on `dogscats`:
```
(dev) apk-mbp-3:dogcats karve$ quilt push akarve/dogscats
Uploading 37394 fragments (855480976 bytes before compression)...
Upload failed: error ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
Upload failed: error ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
Upload failed: error ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
Upload failed: error ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
Upload failed: error ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
 86%|███████████████████████████████████      | 732M/855M [25:24<03:55, 527KB/s]
```

dimaryaz:
The "Connection aborted." shouldn't happen anymore:
- All signed URLs are now retrieved in one request
- Retries were not actually working earlier

quilt/tools/command.py
```diff
@@ -372,41 +381,92 @@ def _push_package(dry_run=False):
             }
         )
 
-    _push_package(dry_run=True)
+    print("Getting a list of upload URLs...")
```
akarve: Suggest: "Fetching upload URLs from the registry..."

quilt/tools/command.py
```diff
@@ -372,41 +381,92 @@ def _push_package(dry_run=False):
             }
         )
 
-    _push_package(dry_run=True)
+    print("Fetching upload URLs from the registry...")
```
kevinemoore: TMI? Is this here because it could take a long time and we want to give feedback to the user?

dimaryaz: Yeah - in case of dogscats, it's like 8MB of URLs. (Compressed, but could still take a while...)

akarve: I specifically requested this because it can take a long time. The cli appears to hang. If streaming solves tis problem I guess we can get rid of it. Otherwise leave it in.

PR #157, opened 2017-08-10
Make it possible to install subpackages

dimaryaz:
Install command now accepts "user/package/group/..." as a package name


dimaryaz:
Oh wait, need to add error handling for missing fragments.

PR #158, opened 2017-08-18
Increment version to 2.7.0

dimaryaz:


dimaryaz:
Ah, didn't realize I had to do that first

kevinemoore:
I'm not sure what the best order is, but it'll be nice if we have GH releases for each major version at least if not minor ones and provide the right link to PyPI.

PR #160, opened 2017-08-29
Split up data.py into data.py and nodes.py

dimaryaz:
Let's keep magic imports separate from node definitions.


dimaryaz:
What would you do with a visitor pattern? In any case, yeah, it would be nice to clean up - if you have ideas how to do that :)

PR #161, opened 2017-09-03
Refactor parsing for greater flexibility in parsing JSON, etc.

akarve:
Invert control (focus on file extensions / transforms) and make pandas secondary so that we can use python native parsers, etc. for new file types

### DIFF
--- a/quilt/tools/build.py
+++ b/quilt/tools/build.py
@@ -2,6 +2,9 @@
 parse build file, serialize package
 """
 from collections import defaultdict
+import importlib
+import sys
+from types import ModuleType
 import os
 import re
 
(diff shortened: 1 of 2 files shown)


quilt/tools/build.py
```diff
@@ -96,22 +100,23 @@ def _file_to_spark_data_frame(ext, path, target, user_kwargs):
...
-        raise BuildException('Unsupported target platform: %s' % target)
-    logic = platform.get(ext)
-    if logic is None:
-        raise BuildException(
-            "Unsupported transform: %s. Try setting a 'transform' key." % ext)
-    fname = logic['attr']
+    logic = PARSERS.get(ext)
+    module_name = logic['module']
+    the_module = None
+    if module_name in sys.modules:
```
dimaryaz: I'm pretty sure it's not necessary to check if the module is already imported

quilt/tools/build.py
```diff
@@ -96,22 +100,23 @@ def _file_to_spark_data_frame(ext, path, target, user_kwargs):
...
+        the_module = sys.modules[module_name]
+    else:
+        the_module = importlib.import_module(module_name)
```
dimaryaz: I don't really like the idea of importing arbitrary modules: e.g., you might run arbitrary code just by building a package. Maybe whitelist module names?

akarve: OK but this is in our code in const.py and if they can modify PARSERS they can modify the whitelist. I can see your objection if the modules are specified in build.yml but that's not the case right?

dimaryaz: Oh whoops, I misunderstood - thought it came from build.yml. That's fine then.

PR #165, opened 2017-09-30
fix unicode encode error on `quilt ls` and `quilt inspect` on Python 2.7

diwu1989:
For python 2.7, you will get a unicode encode error like this
```
  File "/usr/local/bin/quilt", line 11, in <module>
    load_entry_point('quilt==2.7.0', 'console_scripts', 'quilt')()
  File "/usr/local/lib/python2.7/dist-packages/quilt/tools/main.py", line 138, in main
    func(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/quilt/tools/command.py", line 799, in ls
    print("%s%s/%s" % (unicode(prefix), owner, pkg))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```
because print can't handle a unicode

simplest repro:
```py
prefix = u"└── "
print prefix
# vs
print prefix.encode('utf-8')
```

### DIFF
--- a/quilt/tools/command.py
+++ b/quilt/tools/command.py
@@ -876,3 +876,34 @@ def inspect(package):
     pkgobj = PackageStore.find_package(owner, pkg)
     if pkgobj is None:
         raise CommandException("Package {owner}/{pkg} not found.".format(owner=owner, pkg=pkg))
+
+    def _print_children(children, prefix, path):
+        for idx, (name, child) in enumerate(children):
+            if idx == len(children) - 1:
+                new_prefix = u"└─"
+                new_child_prefix = u"  "
+            else:
+                new_prefix = u"├─"
+                new_child_prefix = u"│ "
+            _print_node(child, prefix + new_prefix, prefix + new_child_prefix, name, path)
+
+    def _print_node(node, prefix, child_prefix, name, path):
+        name_prefix = u"─ "
+        if isinstance(node, GroupNode):
+            children = list(node.children.items())
+            if children:
+                name_prefix = u"┬ "
+            print((prefix + name_prefix + name).encode('utf-8'))
+            _print_children(children, child_prefix, path + name)
+        elif isinstance(node, TableNode):
+            df = pkgobj.get_obj(node)
+            assert isinstance(df, pd.DataFrame)
+            info = "shape %s, type \"%s\"" % (df.shape, df.dtypes)
+            print((prefix + name_prefix + ": " + info).encode('utf-8'))
+        elif isinstance(node, FileNode):
+            print((prefix + name_prefix + name).encode('utf-8'))
+        else:
+            assert False, "node=%s type=%s" % (node, type(node))
+
+    print(pkgobj.get_path())
+    _print_children(children=pkgobj.get_contents().children.items(), prefix='', path='')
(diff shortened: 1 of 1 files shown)


diwu1989:
other ways to solve this problem more globally
http://chase-seibert.github.io/blog/2014/01/12/python-unicode-console-output.html

monkey patching the `sys.stdout` seems hacky, so up to you how you want to handle this globally

quilt/tools/command.py
```diff
@@ -876,3 +876,34 @@ def inspect(package):
     pkgobj = PackageStore.find_package(owner, pkg)
     if pkgobj is None:
         raise CommandException("Package {owner}/{pkg} not found.".format(owner=owner, pkg=pkg))
+
+    def _print_children(children, prefix, path):
```
diwu1989: this code below is also causing unicode error when i run `quilt inspect`
the CRAZY thing is that on `master` branch, it looks like someone just gutted this whole functionality, see https://github.com/quiltdata/quilt-compiler/commit/f70d8993d53301b78a5c195467ad0979c9c72a16#diff-e97e2bbedcaad1ab3c091235193f000d

just glad that the version installed from `pip` still had a working version of this command

dimaryaz: Oh wow, I'm pretty sure that wasn't intentional. Good catch.

diwu1989:
I'll try to set up a minimal repro for you, maybe with a simple docker image to demo the bug.
Stay tuned :)

dimaryaz:
Ah, Docker... I bet you're missing the LANG environment variable. This should fix it:
`ENV LANG=C.UTF-8`

(That is, one could argue that `encode('utf-8')` is the proper fix... However, unicode is pretty broken in Python2, while Python3 handles it correctly - so probably not worth adding `encode` to every `print` statement. `LANG` seems like the best workaround.)

diwu1989:
Here's another repro that doesn't seem to be fixed by `env LANG`.
`quilt inspect ... | less` on Python 2
Anything that involves piping the output into another tool seems to break it.

I think the better overall solution is to switch to a more comprehensive CLI framework like http://click.pocoo.org/ that handles normalizing stdout via a `echo()` http://click.pocoo.org/5/utils/#printing-to-stdout

diwu1989:
If you agree that it's better to move to a richer CLI framework, then I will close this PR and open another that moves `main.py` onto `Click` and standardize the stdouts.

You should still just copy & paste back in the content of the original `inspect` command just so that it's not broken on master. The unit test for `inspect` doesn't seem to be that useful as builds pass even when the function basically did nothing.

dimaryaz:
You're right, piping is still broken.

I'll look into click - but you've convinced me that your fix is the right solution at least in the short term, so I'll merge it.

Thanks!

dimaryaz:
Oh no, that actually broke Python3:
```
$ quilt ls
/home/dima/src/quilt/quilt_packages
b'\xe2\x94\x9c\xe2\x94\x80\xe2\x94\x80 dima2/test'
b'\xe2\x94\x9c\xe2\x94\x80\xe2\x94\x80 dima/test'
b'\xe2\x94\x94\xe2\x94\x80\xe2\x94\x80 dima/tmp2'
```

I'll have to revert it, unfortunately. Might have to use click, after all.

diwu1989:
Yeah i suspected something like that could happen, thats why more complex CLI are based on some kind of normalization framework.

Click, clint, docopt, cement, etc... they all have some wrapper code to deal with this kind of edge case

PR #167, opened 2017-10-03
Bring back the `quilt inspect` code

dimaryaz:
It got deleted in f70d8993d53301b78a5c195467ad0979c9c72a16


diwu1989:
In this test https://github.com/quiltdata/quilt-compiler/blob/master/quilt/test/test_command.py#L119-L124
If you add a `patch` for `print()` and then assert that it got called at least once, then future obvious breakage of this kind will be avoided

dimaryaz:
@diwu1989: I'm sure it's actually possible to patch `print`. The right solution might be to use the `capsys` feature of `pytest` - but it appears to only work with standalone test functions, not a test class.

PR #169, opened 2017-10-10
Git build

kevinemoore:
Build a quilt package from a git URL. Clones the remote repo to a temp directory then builds the package locally.

### DIFF
--- a/quilt/test/test_command.py
+++ b/quilt/test/test_command.py
@@ -202,3 +202,12 @@ class CommandTest(QuiltTestCase):
         self.requests_mock.add(responses.DELETE, delete_url, json.dumps(dict()))
 
         command.delete('%s/%s' % (owner, package))
+
+    @patch('quilt.tools.command._rmtree')
+    @patch('quilt.tools.command._clone_git_repo')
+    def test_build_from_git(self, mock_clone_git, mock_rmtree):
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, 'data')
+        mock_clone_git.return_value = path
+        mock_rmtree.return_value = None
+        command.build('user/test', 'https://github.com/quiltdata/testdata.git')
--- a/quilt/test/test_command.py
+++ b/quilt/test/test_command.py
@@ -7,7 +7,9 @@ import os
 import time
 
 import responses
+import shutil
 
+import pandas as pd
 from six import assertRaisesRegex
 
 from quilt.tools import command, store
@@ -203,13 +205,18 @@ class CommandTest(QuiltTestCase):
 
         command.delete('%s/%s' % (owner, package))
 
-    @patch('quilt.tools.command._rmtree')
-    @patch('tempfile.mkdtemp')
-    @patch('quilt.tools.command._clone_git_repo')
-    def test_build_from_git(self, mock_clone_git, mock_temp_file, mock_rmtree):
-        mydir = os.path.dirname(__file__)
-        path = os.path.join(mydir, 'data')
-        mock_clone_git.return_value = None
-        mock_temp_file.return_value = path
-        mock_rmtree.return_value = None
-        command.build('user/test', 'https://github.com/quiltdata/testdata.git')
+    @patch('subprocess.check_call')
+    def test_build_from_git(self, mock_subproc):
+        def mock_git_clone(cmd):
+            srcfile = 'foo.csv'
+            mydir = os.path.dirname(__file__)
+            srcpath = os.path.join(mydir, 'data', srcfile)
+            destpath = os.path.join(cmd[-1], srcfile)
+            shutil.copyfile(srcpath, destpath)
+        
+        with patch('subprocess.check_call', mock_git_clone):
+            command.build('user/test', 'https://github.com/quiltdata/testdata.git')
+
+        from quilt.data.user import test
+        assert hasattr(test, 'foo')
+        assert isinstance(test.foo(), pd.DataFrame)
--- a/quilt/test/test_command.py
+++ b/quilt/test/test_command.py
@@ -204,10 +204,12 @@ class CommandTest(QuiltTestCase):
         command.delete('%s/%s' % (owner, package))
 
     @patch('quilt.tools.command._rmtree')
+    @patch('tempfile.mkdtemp')
     @patch('quilt.tools.command._clone_git_repo')
-    def test_build_from_git(self, mock_clone_git, mock_rmtree):
+    def test_build_from_git(self, mock_clone_git, mock_temp_file, mock_rmtree):
         mydir = os.path.dirname(__file__)
         path = os.path.join(mydir, 'data')
-        mock_clone_git.return_value = path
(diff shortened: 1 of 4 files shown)


quilt/tools/build.py
```diff
@@ -333,3 +334,17 @@ def generate_build_file(startpath, outfilename=DEFAULT_BUILDFILE):
     with open(buildfilepath, 'w') as outfile:
         yaml.dump(contents, outfile, default_flow_style=False)
     return buildfilepath
+
+def clone_git_repo(git_url, destination):
+    cmd = ['git', 'clone', '-q', '--depth=1', git_url, destination]
+    try:
+        proc = subprocess.Popen(cmd, cwd=os.getcwd(), env=os.environ.copy())
```
dimaryaz: I think those are the defaults for `cwd` and `env`, so don't need them?

dimaryaz: Also, I think `subprocess.call` or `subprocess.check_call` is what you want - instead of calling `wait`, etc.

quilt/tools/command.py
```diff
@@ -254,9 +255,18 @@ def build(package, path=None, dry_run=False, env='default'):
     """
     Compile a Quilt data package, either from a build file or an existing package node.
     """
-    # we may have a path, PackageNode, or None
+    # we may have a path, git URL, PackageNode, or None
     if isinstance(path, string_types):
-        build_from_path(package, path, dry_run=dry_run, env=env)
+        # is this a git url?
+        is_git_url = re.match('http[s]?://(?:[\w./~_-])+\.git$', path)
```
dimaryaz: Need an `r` in front of the regex. It happens to work - but in general, backslashes will break things.

Also, I'd move it into a constant and compile it, similar to e.g. `VALID_NAME_RE` in `quilt/tools/store.py`

quilt/tools/command.py
```diff
@@ -254,9 +255,18 @@ def build(package, path=None, dry_run=False, env='default'):
...
+
+        if is_git_url:
+            tmpdir="tmp_git_clone"
```
dimaryaz: `tempfile.mkdtemp`

quilt/tools/command.py
```diff
@@ -254,9 +255,18 @@ def build(package, path=None, dry_run=False, env='default'):
...
+            clone_git_repo(path, tmpdir)
+            build_from_path(package, tmpdir, dry_run=dry_run, env=env)
+            rmtree(tmpdir)
```
dimaryaz: Maybe use a `finally:` block?

quilt/tools/build.py
```diff
@@ -333,3 +334,17 @@ def generate_build_file(startpath, outfilename=DEFAULT_BUILDFILE):
...
+    except Exception as exc:
+        print("Error %s while executing command %s", exc, cmd)
```
dimaryaz: You should also test the error handling :)

```
Error %s while executing command %s [WinError 2] File not found ['git', 'clone', '-q', '--depth=1', 'https://github.com/quiltdata/quilt-registry.git', 'tmp_git_clone']
```

Just raise a `CommandException` with the error message. It'll get printed out without the stack trace.

dimaryaz:
I think you should just patch `subprocess.checked_call` - and have the mock function create a directory with the contents (or copy it from elsewhere, etc.). Then you won't need to patch `rmtree`.

quilt/test/test_command.py
```diff
@@ -202,3 +204,19 @@ def test_delete_confirmed(self, mock_input):
         self.requests_mock.add(responses.DELETE, delete_url, json.dumps(dict()))
 
         command.delete('%s/%s' % (owner, package))
+
+    @patch('subprocess.check_call')
```
dimaryaz: Don't need this anymore

quilt/test/test_command.py
```diff
@@ -202,3 +204,19 @@ def test_delete_confirmed(self, mock_input):
...
+    def test_build_from_git(self, mock_subproc):
+        def mock_git_clone(cmd):
+            srcfile = 'foo.csv'
```
dimaryaz: May want to actually assert that `cmd` looks reasonable

quilt/tools/command.py
```diff
@@ -250,21 +252,45 @@ def check(path=None, env='default'):
     # (if not, then require dry_run=True if files!=None/all)
     build("dry_run/dry_run", path=path, dry_run=True, env=env)
 
+def _clone_git_repo(url, branch, dest):
+    cmd = ['git', 'clone', '-q', '--depth=1']
+    if branch:
+        cmd += ['-b', branch]
+    cmd += [url, dest]
+    proc = subprocess.check_call(cmd)
```
dimaryaz: Don't need `proc = `

quilt/tools/command.py
```diff
@@ -250,21 +252,45 @@ def check(path=None, env='default'):
...
+
+def _rmtree(path):
```
dimaryaz: Don't need this anymore

quilt/tools/command.py
```diff
@@ -250,21 +252,45 @@ def check(path=None, env='default'):
...
     Compile a Quilt data package, either from a build file or an existing package node.
     """
-    # we may have a path, PackageNode, or None
+    # we may have a path, git URL, PackageNode, or None
     if isinstance(path, string_types):
-        build_from_path(package, path, dry_run=dry_run, env=env)
+        # is this a git url?
+        is_git_url = GIT_URL_RE.match(path)
+        if is_git_url:
+            tmpdir=tempfile.mkdtemp()
```
dimaryaz: Spaces around `=`

quilt/tools/command.py
```diff
@@ -41,6 +42,7 @@
 
 DEFAULT_QUILT_PKG_URL = 'https://pkg.quiltdata.com'
 QUILT_PKG_URL = os.environ.get('QUILT_PKG_URL', DEFAULT_QUILT_PKG_URL)
+GIT_URL_RE = re.compile(r'(?P<url>http[s]?://[\w./~_-]+\.git)(?:@(?P<branch>[\w_-]+))?')
```
dimaryaz: You're not actually using the group names, as far as I can tell?

quilt/tools/command.py
```diff
@@ -250,21 +252,45 @@ def check(path=None, env='default'):
...
+        # is this a git url?
+        is_git_url = GIT_URL_RE.match(path)
+        if is_git_url:
+            tmpdir=tempfile.mkdtemp()
+            url, branch = is_git_url.groups()
+            try:
+                _clone_git_repo(url, branch, tmpdir)
+                build_from_path(package, tmpdir, dry_run=dry_run, env=env)
+            except Exception as exc:
+                print("Error %s while cloning %s.", exc, path)
```
dimaryaz: Please actually test this. Need `%` to substitute the values. Also, `raise CommandException` instead of `print` and `raise`.

Actually, add a unit test, too. This might come up easily if e.g. `git` is not installed, or the URL is invalid.

PR #171, opened 2017-10-19
Implement `quilt config`

dimaryaz:
Sets the registry URL. The config can still be overridden using the env var.


dimaryaz:
Ping?

quilt/tools/command.py
```diff
@@ -51,25 +52,80 @@
...
+
+def get_auth_path():
+    url = get_registry_url()
+    if url == DEFAULT_REGISTRY_URL:
+        suffix = ''
+    else:
+        # Store different servers' auth in different files.
+        suffix = "-%.8s" % hashlib.md5(url.encode('utf-8')).hexdigest()
+
+    return os.path.join(BASE_DIR, 'auth%s.json' % suffix)
```
kevinemoore: Shouldn't the config and the login/auth be stored in the same folder? Both seem like user_config rather than user_data.

dimaryaz: Well, it's kinda subjective. I was thinking of auth as "state" (that the user should never touch directly), while the registry URL (and soon analytics switch) is "config" that the user gets to edit.

That said, two different directories might be overkill. If we use one, it's probably better to keep using `BASE_DIR` - it's `user_data_dir`, so somewhat neutral, I guess.

kevinemoore: Let's keep it simple and use a single directory. That'll make it easy for someone, including us, to wipe out our config and auth data and start fresh.

PR #172, opened 2017-10-21
checks lib: add support for check_not_* and check_row_*; bugfixes

asah:
@akarve what do you think of this design for row-checks and boolean-not ?  as usual, tight syntax.


akarve:
Closing and backlogging in asana. Branch has fallen behind.

Issue #166, opened 2017-09-30
DataNode and GroupNode should give access to original names


diwu1989:
`DataNode` should be extended to allow access to the original file name via a `.name` property

It doesn't seem possible to get back the extension of the original file otherwise, and extensions matter for video files.

the `metadata` in the json file seem to already have this information

diwu1989:
The same goes for `GroupNode`, should be storing the original folder name.
This way I can easily write a `export` command that take a quilt package and restore it back 1:1 to the folder hierarchy that it came from.

akarve:
This feature will be handled in #495 and is partially implemented in the existing `quilt export feature`. Closing as duplicate.

Issue #164, opened 2017-09-20
Build destination should default to user home and read from an environment variable


akarve:
Suggest `QUILT_BUILD_PATH`as env vars. See #1 for details.

stillmatic:
Would this supersede the current folder search logic (i.e. search in current folder and all above)?

akarve:
We were discussing this today. Concluded that we shouldn't mess with "look in parent" logic for module resolution?

I updated this issue to only affect where `quilt build` writes the package to, as we've seen users build a package in a random directory and then not be able to import it.

@stillmatic What do you think?

stillmatic:
I think that makes sense and resolves a pain point I had while writing the R package. I had to create a random temp directory (https://github.com/stillmatic/quiltr/blob/master/R/build.R#L28) to ephemerally dump the changed files. This method 'works' but is definitely suboptimal.

Having a fixed `QUILT_BUILD_PATH` would make that part easier. How would you deal with collisions?

akarve:
Collisions would overwrite (as we have today). Any pushed packages are of course tracked by `quilt log` and users can revert with `quilt install`.

akarve:
OK this turns out to be nontrivial with unforseen effects on push and install--as I discovered when testing a "quick fix". We're going to need a more comprehensive fix--likely in the spirit of Docker where there's flat repository for all builds. I'm working on it. CC @salvor7

akarve:
Fixed by new store implementation

PR #175, opened 2017-10-30
Refactor package build interface

kevinemoore:
Design sketch describing a way to support multiple package versions in a flat package store.


quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
...
+BASE_DIR/pkgs/
+```
+
+Stores package manifest files (JSON format) for all package instances. Package manifests are identified by the hash of the package.
+
+```bash
+BASE_DIR/contents.json
+```
+
+Catalog of locally resident packages. The contents manfiest maps identifiers including package name, tags and versions to package instances.
```
dimaryaz: Typo: manifest

quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
...
+Stores binary data objects identified by hash. These objects include compressed raw files and Parquet files. Object hashes are verified and objects are stored only once (deduplication).
+
+```bash
+BASE_DIR/pkgs/
+```
+
+Stores package manifest files (JSON format) for all package instances. Package manifests are identified by the hash of the package.
+
+```bash
+BASE_DIR/contents.json
```
dimaryaz: We should consider using something like a Sqlite database - or at the very least, lock the file. Otherwise, it'll be too easy to corrupt the data by running multiple quilt processes at the same time.

quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
...
+Adds or removes a version from a package instance. Note: once pushed to a Quilt registry, there is no way to remove a version without deleting the package.
+
+```bash
+quilt push <package>
+```
+
+Pushes a package instance to the Quilt registry. Quilt push will record any local tags or versions to the registry.
+
+```bash
+quilt install <package>[:tag:version]
```
dimaryaz: The syntax is not clear to me. How do you specify tag vs version?

quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
...
+    - lookup tags:
+        - list all instances with tags, ordered by tag
+    - #Alternate: order instances by reverse creation date
+    - mark each instance printed in metadata map
+    - list all untagged, unversioned instances for the package
+
+### import
+- eliminate search for quilt_packages directory
+- lookup package hash in contents manifest
+    - by tag: package->tags->tag->hash
```
dimaryaz: Is the plan to support multiple local hashes/tags/versions? What's the syntax for importing a specific hash?

quilt_repository_format.md
kevinemoore: Good point. I'd rather not rely on sqlite and would prefer to keep it both human and machine readable if possible. But, we'll definitely need to have a plan for synchronization.

quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
+# Quilt CLI Local Storage
+
+## Motivation
+The primary motivation for unifying the local package store is that any Python process will be able to access any locally installed Quilt package. A common point of friction for Quilt users has been that packages built on the command line aren't visible inside Jupyter notebooks if the Jupyter server is running in a different directory. Unifiying the local storage by adopting the Docker storage model solves that problem and also enables useful future features like comparing (diff) package versions and tracking revision history locally without pushing to a registry.
```
akarve: can't accurately call this a common point of friction; but it is a point of friction per #164 and #166. not clear how many times this is the problem vs. number of times ipython is executing where it can't find quilt.

quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
...
+```bash
+quilt tag add <hash> <tag>
+quilt tag rm <hash> <tag>
+```
+
+Adds or removes a tag from a package instance.
+
+```bash
+quilt version add <hash> <tag>
+quilt version rm <hash> <tag>
```
akarve: should be `<version>`

quilt_repository_format.md
```diff
@@ -0,0 +1,137 @@
...
+Remove local package instances
+
+### quilt cleanup
+Remove unlinked binary objects (garbage collection)
+
+### quilt diff <pgk1> <pkg2>
+Diff package instances
+
+### build FROM
+Import an existing package as the basis for a new package like Docker's FROM.
```
akarve: also for future:
* build cache
* global de-duplication (we only de-dupe fragments per user)
* side-by-side install and import of multiple versions (not sure how we'd do this with our current python module scheme (@dimaryaz?)

quilt_repository_format.md
kevinemoore: @stillmatic, @salvor7 does this docker-like approach of having a single local package repository that is capable of maintaining multiple package instances simultaneously meet your needs?

quilt/tools/store.py
```diff
@@ -81,14 +96,14 @@ def check_name(cls, user, package):
         if not VALID_NAME_RE.match(package):
             raise StoreException("Invalid package name: %r" % package)
 
+    # TODO: lookup hash in contents based on tag or version
```
asah: comment is confusing (to me)... esp "in contents" ?

quilt/tools/store.py
```diff
@@ -141,14 +149,33 @@ def ls_packages(self):
         """
         List packages in this store.
         """
```
asah: time extract this block into a reusable method/function ?

quilt_repository_format.md
kevinemoore: I wasn't meaning to change the syntax for specifying versions and tags on install.

quilt/tools/store.py
kevinemoore: The implementation could certainly be cleaned up. As of now, it's only used by quilt ls. Can you suggest a name & signature for a better reusable function? Perhaps one that returned a list or iterator of Package objects?

quilt_repository_format.md
stillmatic: I think that having the base directory is useful. That would allow different Quilt clients (e.g. in R) to easily access the package root. This change also helps ensure that you are using the 'right' version of the package, versus having several copies in different folders.

In my specific use-case, a number of users share one storage system (everybody has their own folder under a single root folder, but no real restrictions on what you can do). Having a single Quilt folder rather than individual Quilt folders under each person's namespace would save on space and hassle, if we are reusing packages.

kevinemoore: We've talked about using a shared folder for quilt_packages and I left the hooks in for package lookups to check multiple places. Quick question, what would you want the permissions/security model to be for the shared folder? By default, anyone with access to that folder would be able to read the packages there.

quilt/test/utils.py
```diff
@@ -41,12 +48,17 @@ def setUp(self):
         self.auth_patcher = patch('quilt.tools.command._create_auth', lambda: None)
         self.auth_patcher.start()
 
+        self._store_dir = test_store_dir()
```
dimaryaz: Need to delete the dir in `tearDown` (but test it to make sure it doesn't do anything stupid like delete your home dir :) )

quilt/tools/package.py
```diff
@@ -5,6 +5,7 @@
 from shutil import copyfile, copyfileobj, move, rmtree
 import tempfile
 
+from distutils.dir_util import mkpath
```
dimaryaz: Just use `os.makedirs`

quilt/tools/store.py
```diff
@@ -4,15 +4,24 @@
...
+from .util import BASE_DIR
 
 # start with alpha (_ may clobber attrs), continue with alphanumeric or _
 VALID_NAME_RE = re.compile(r'^[a-zA-Z]\w*$')
 CHUNK_SIZE = 4096
 
+# Helper function to return the default package store path
+def default_store_location():
+    path=os.path.realpath(BASE_DIR)
+    package_dir = os.path.join(path, PACKAGE_DIR_NAME)
```
dimaryaz: Is `BASE_DIR` actually a good location - as opposed to something more user-friendly or discoverable, like `~/quilt_packages`? Given that it's user data, not settings, the user might want it to be more visible.

Examples I see in my home dir: "~/VirtualBox VMs", "~/Microsoft/TypeScript/node_modules/".

(On the other hand, Steam has an 11GB folder in my "~/.local/share/" dir, so they clearly don't have any problems with this...)

quilt/tools/store.py
```diff
@@ -141,14 +176,34 @@ def ls_packages(self):
...
+        packages = []
+        pkgdir = os.path.join(self._path, self.PKG_DIR)
+        for user in os.listdir(pkgdir):
+            for pkg in os.listdir(os.path.join(pkgdir, user)):
+                pkgpath = os.path.join(pkgdir, user, pkg)           
+                pkgmap = {h : None for h in os.listdir(os.path.join(pkgpath, Package.CONTENTS_DIR))}
+                for tag in os.listdir(os.path.join(pkgpath, Package.TAGS_DIR)):
+                    with open(os.path.join(pkgpath, Package.TAGS_DIR, tag), 'r') as tagfile:
+                        pkghash = tagfile.read()
+                        pkgmap[pkghash] = tag
```
dimaryaz: Can there be multiple tags for a given hash?

quilt/tools/command.py
```diff
@@ -1002,9 +1002,8 @@ def ls():                       # pylint:disable=C0103
     for pkg_dir in PackageStore.find_store_dirs():
         print("%s" % pkg_dir)
         packages = PackageStore(pkg_dir).ls_packages()
-        for idx, (owner, pkg) in enumerate(packages):
-            prefix = u"└── " if idx == len(packages) - 1 else u"├── "
-            print("%s%s/%s" % (prefix, owner, pkg))
+        for idx, (package, tag, pkghash) in enumerate(packages):
+            print("{0:30} {1:20} {2}".format(package, tag, pkghash))
```
dimaryaz: If there's no tag, it prints out `None` (which could be a valid tag). Make it an empty string?

quilt/tools/store.py
kevinemoore: Yes. You can have many tags for the same hash.

quilt/tools/store.py
```diff
@@ -30,43 +39,60 @@ class and its subclasses abstract file formats, file naming and
...
+        tmpobjdir = os.path.join(self._path, self.TMP_OBJ_DIR)
+        pkgdir = os.path.join(self._path, self.PKG_DIR)
 
-    def __init__(self, store_dir=PACKAGE_DIR_NAME):
-        assert os.path.basename(os.path.abspath(store_dir)) == PACKAGE_DIR_NAME, \
-            "Unexpected package directory: %s" % store_dir
-        self._path = store_dir
+        if os.path.isdir(self._path):
+            # Verify existing package store is compatible
+            if self.VERSION != self._read_format_version():
```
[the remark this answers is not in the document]
kevinemoore: That’s right. This won’t help for users migrating from one version to the next, but it will ensure that users get notified if the client tries to access a quilt_packages directory that’s incompatible.

quilt/tools/store.py
kevinemoore: I find it pretty convenient myself. The files are named by hashes so it’s not very human-friendly anyway. We could use quilt commands to remove old packages and prune un-connected objects.

quilt/tools/store.py
dimaryaz: But they'll overwrite each other here, won't they?

Also, how would I create new tags? `quilt tag add ...` should probably do that - but it's only creating server-side tags right now.

dimaryaz:
I started getting errors all of a sudden (after Ctrl-C'ing a download):

```
$ quilt install danWebster/sgRNAs
  File "/home/dima/envs/cli3/bin/quilt", line 11, in <module>
    load_entry_point('quilt', 'console_scripts', 'quilt')()
  File "/home/dima/src/quilt/quilt/tools/main.py", line 152, in main
    func(**kwargs)
  File "/home/dima/src/quilt/quilt/tools/command.py", line 780, in install
  File "/home/dima/src/quilt/quilt/tools/store.py", line 138, in get_package
    path=path
  File "/home/dima/src/quilt/quilt/tools/package.py", line 74, in __init__
    contents = self._load_contents()
  File "/home/dima/src/quilt/quilt/tools/package.py", line 81, in _load_contents
    with open (latest_tag, 'r') as tagfile:
FileNotFoundError: [Errno 2] No such file or directory: '/home/dima/.local/share/QuiltCli/quilt_packages/pkgs/danWebster/sgRNAs/tags/latest'
```

dimaryaz:
(The "tags" directory exists, but not "tags/latest".)

kevinemoore:
https://github.com/quiltdata/quilt-docs/pull/1

quilt/tools/store.py
kevinemoore: I've left updating the commands (e.g., tag add) to another PR.

quilt/tools/command.py
```diff
@@ -1040,6 +1040,7 @@ def ls():                       # pylint:disable=C0103
         print("%s" % pkg_dir)
         packages = PackageStore(pkg_dir).ls_packages()
         for idx, (package, tag, pkghash) in enumerate(packages):
+            tag = "" if tag is None else tag
```
dimaryaz: I think this won't happen anymore because of your other changes?

quilt/tools/store.py
```diff
@@ -141,14 +177,38 @@ def ls_packages(self):
...
-            if os.path.isdir(os.path.join(self._path, user))
-            for pkg in os.listdir(os.path.join(self._path, user))
-            if pkg.endswith(self.PACKAGE_FILE_EXT)]
+        packages = []
+        pkgdir = os.path.join(self._path, self.PKG_DIR)
+        for user in os.listdir(pkgdir):
+            for pkg in os.listdir(os.path.join(pkgdir, user)):
+                pkgpath = os.path.join(pkgdir, user, pkg)           
+                pkgmap = {h : [] for h in os.listdir(os.path.join(pkgpath, Package.CONTENTS_DIR))}
+                for tag in os.listdir(os.path.join(pkgpath, Package.TAGS_DIR)):
```
dimaryaz: Just realized something: on Windows and OS X, tags and versions will be case-insensitive. That may be ok - but we should think about it.

kevinemoore: Good point. Case insensitive matching seems fine to me. Ideally we'd have consistent behavior on all systems though.

PR #177, opened 2017-11-06
Send logs for package builds.

dimaryaz:
Can be disabled with `{"disable_analytics": true}` in `config.json` in `BASE_DIR`.


quilt/tools/command.py
```diff
@@ -326,10 +328,45 @@ def _clone_git_repo(url, branch, dest):
...
     """
     Compile a Quilt data package, either from a build file or an existing package node.
     """
+    path_str = str(path) if path else None
+    try:
+        _build_internal(package, path, dry_run, env)
+    except Exception as ex:
+        _log(type='build', package=package, path=path_str, dry_run=dry_run, env=env, error=str(ex))
+        raise
+    _log(type='build', package=package, path=path_str, dry_run=dry_run, env=env)
```
kevinemoore: Package names are potentially very sensitive (sometimes even more than the actual data). We should obscure the real package names and paths at least for private packages. We should probably obscure paths or not send them even for public packages.

akarve: Local path we can leave out but package names need to be in there for Mixpanel events to work properly. In fact the events should support the same fields that current events do. Keep in mind that all traffic is https. If there is still some exposure of the names we can issue a public key for the clients but yes that's more complicated.

kevinemoore: Keep in mind that these are packages that haven't been pushed to Quilt so they might be private packages that users never intend to push. The names and contents of those packages haven't been shared with us so we shouldn't look at them. I think we can safely send package names for packages that have already been pushed.

akarve: Too complicated to track whether or not package has been pushed (requires statefulness to know if package has been pushed) and doesn't solve core problem of diagnosing build failures. We don't touch contents at any point in the logs. Considering we have zero build analytics right now we shouldn't get stuck on trying to micro-optimize this. Everything is encrypted in transit, never publicly logged, and users can actually disable all logging with an environment variable.

akarve: So here's what we can do:
* Cryptographically hash package names
* Don't send build path

dimaryaz: I'm also logging exceptions - and those could contain paths or package names. Do we want to change anything there?

PR #180, opened 2017-11-11
Support build parameters at the group level

akarve:
For build trees with numerous similar child nodes it's critical to support "en masse" args. For example, in the datasets for Wes's new book, he has a directory with hundreds of .txt files (which are actually CSVs) and it's not practical to specify `transform: csv` for each child :)

With this PR users can specify transforms at the group level:
```
group:
  transform: csv
  node:
    file: data/foo.txt # will be parsed as CSV
  node2:
    file: data/foo2.txt # will be parsed as CSV
     ...
```

Updated docs here:
https://docs.quiltdata.com/buildyml.html


akarve:
Investigating Python 2.7 bug (causing CI failures); not ready for review; will resubmit.

quilt/test/test_import.py
```diff
@@ -128,7 +128,7 @@ def test_save(self):
 
         # Modify an existing dataframe
         csv = package1.dataframes2.csv._data()
-        csv.set_value(0, 'Int0', 42)
+        csv.at[0, 'Int0'] = 42
```
dimaryaz: Yay! I was too lazy to fix this.

quilt/tools/build.py
```diff
@@ -143,15 +176,15 @@ def _file_to_spark_data_frame(ext, path, target, user_kwargs):
...
+def _file_to_data_frame(ext, path, target, handler_args):
     _ = target  # TODO: why is this unused?
     logic = PARSERS.get(ext)
     the_module = importlib.import_module(logic['module'])
     if not isinstance(the_module, ModuleType):
         raise BuildException("Missing required module: %s." % logic['module'])
     # allow user to specify handler kwargs and override default kwargs
-    kwargs = dict(logic['kwargs'])
-    kwargs.update(user_kwargs)
+    kwargs = dict(logic['kwargs']).copy()
```
dimaryaz: Don't need `copy()`; `dict(...)` creates a new object.

That is, do either this:
`kwargs = dict(logic['kwargs'])`
or this:
`kwargs = logic['kwargs'].copy()`
I actually don't know if there's a difference, or which one is preferred.

quilt/tools/build.py
```diff
@@ -116,7 +139,17 @@ def _build_node(build_dir, package, name, node, fmt, target='pandas', checks_con
...
-def _file_to_spark_data_frame(ext, path, target, user_kwargs):
+def _remove_keywords(d):
+    """
+    copy the dict, filter_keywords
+
+    Parameters
+    ----------
+    d : dict
+    """
+    return { k:d[k] for k in d if k not in RESERVED }
```
dimaryaz: Not that it really matters here, but this is more Pythonic:

```{ k: v for k, v in iteritems(d) if k not in RESERVED }```

(May need to import `iteritems` from `six`.)

File: quilt/tools/build.py [hunk 4]
```diff
@@ -48,20 +49,40 @@ def _run_checks(dataframe, checks, checks_contents, nodename, rel_path, target,
...
+    """
     if _is_internal_node(node):
-        for child_name, child_table in node.items():
+        # add anything whose value is not a dict to the group_arg stack
+        # TODO: there might be some pandas kwargs that take dictionaries as values
+        # in which case this code will break by treating a kwarg as a package node
+        # NOTE: YAML parsing does not guarantee key order so we have to set all
+        # of the group args in one shot for consistent application to subtrees
+        group_args = ancestor_args.copy()
+        group_args.update({k:node[k] for k in node if type(node[k]) is not dict})
```
dimaryaz: Use `iteritems`

quilt/tools/build.py
```diff
@@ -48,20 +49,40 @@ def _run_checks(dataframe, checks, checks_contents, nodename, rel_path, target,
...
+        groups = {k:node[k] for k in node if type(node[k]) is dict}
```
dimaryaz: Same here

quilt/tools/build.py
```diff
@@ -48,20 +49,40 @@ def _run_checks(dataframe, checks, checks_contents, nodename, rel_path, target,
...
+        for child_name, child_table in groups.items():
```
dimaryaz: `iteritems(groups)`, to use iterators in Python2

quilt/tools/build.py
```diff
@@ -20,8 +20,9 @@
 from . import check_functions as qc            # pylint:disable=W0611
 
 def _is_internal_node(node):
-    # all of an internal nodes children are dicts
-    return all(isinstance(x, dict) for x in node.values())
+    # at least one of an internal nodes children are dicts
+    # some (group args) may not be dicts
+    return any(isinstance(x, dict) for x in node.values())
```
dimaryaz: I forgot how this logic works, but do we care about empty lists? (They will be rejected by `any`.)

dimaryaz: Also, might as well use `itervalues(node)`.

quilt/test/build_group_args.yml
```diff
@@ -0,0 +1,44 @@
...
+        file: data/csv.txt
+      many_tsv:
+        # all of the following should inherit transform: csv
+        one:
+          file: data/tsv.txt
+        two:
+          file: data/tsv.txt
+        three:
+          file: data/tsv.txt
+...
```
dimaryaz: Missing line break

quilt/tools/build.py
akarve: I tested this and `any` doesn't care if the dict child is empty. It will return True. Not sure I understand the concern. BTW if there is an empty dict the leaf node code will throw an exception because it doesn't have a `file` key.

quilt/test/build_group_args.yml
akarve: i actually fixed this and github won't show it

quilt/tools/build.py
dimaryaz: I meant, if `node.values()` returns an empty list: `all([])` is `True`, but `any([])` is `False`. Anyways, if it works, that's fine

PR #181, opened 2017-11-13
short hashes and dependencies quilt.yml

asah:
preliminary PR... would appreciate feedback, not expecting this is final.


quilt/tools/command.py
```diff
@@ -649,6 +650,23 @@ def tag_remove(package, tag):
         )
     )
 
+def install_via_requirements(requirements_str, force=False):
+    """
+    Download multiple Quilt data packages via quilt.xml requirements file.
```
akarve: typo in `quilt.xml`

akarve:
I think this is going in the right direction. A few general questions and comments:
1. Does it make sense to check in some more complex `quilt.yml` files in `/test`? Worth adding an integration test for the same?
1. Do we wan't to call the top-level node `packages` or `dependencies`?
1. Obvious point but worth stating: on the off chance that install order matters (it really shouldn't for data) we should probably guarantee that packages are installed in file order? Since it's a YAML array it should in fact be ordered (dictionaries are not); just calling this out as something to think about.

quilt/tools/store.py
```diff
@@ -161,6 +161,34 @@ def temporary_object_path(self, name):
...
 
+def parse_package_extended(name):
+    hash = version = tag = None
+    try:
+        if ':' in name:
+            name, versioninfo = name.split(':', 1)
+            if ':' in versioninfo:
+                info = versioninfo.split(':', 1)
+                if len(info) == 2:
+                    if info[0][0] == 'v':
```
kevinemoore: This looks a bit overly permissive since you're only checking the first letter. It's probably worth explicitly checking for 't' or 'tag' and 'v' or 'version'.

quilt/tools/command.py
asah: DONE

quilt/tools/store.py
asah: NOCHG: aneesh requested this.

quilt/test/test_install.py
```diff
@@ -101,6 +116,93 @@ def test_install_subpackage(self):
             contents = fd.read()
             assert contents == table_data
 
+    @staticmethod
+    def validate_file(filename, contents, table_hash, table_data):
+        with open('quilt_packages/'+filename) as fd:
+            file_contents = json.load(fd, object_hook=decode_node)
+            assert file_contents == contents
+        with open('quilt_packages/objs/{hash}'.format(hash=table_hash)) as fd:
```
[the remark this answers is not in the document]
asah: NOCHG: this was the way the code was already written and I couldn't find an example... ?  there's an accessor by hash but I didn't see a general purpose file accessor...

quilt/tools/store.py
```diff
@@ -180,6 +209,5 @@ def parse_package(name, allow_subpath=False):
     except StoreException as ex:
         raise CommandException(str(ex))
 
-    if allow_subpath:
-        return owner, pkg, subpath
-    return owner, pkg
+    return owner, pkg, subpath
```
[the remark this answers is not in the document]
asah: DONE

asah:
@akarve response...

Does it make sense to check in some more complex quilt.yml files in /test? Worth adding an integration test for the same?

==> imho no.  The unittests are very detailed and subsume the need for a physical file.


Do we wan't to call the top-level node packages or dependencies?

==> I don't care.   https://www.youtube.com/watch?v=FL7yD-0pqZg


Obvious point but worth stating: on the off chance that install order matters (it really shouldn't for data) we should probably guarantee that packages are installed in file order? Since it's a YAML array it should in fact be ordered (dictionaries are not); just calling this out as something to think about.

==> the current implementation installs in order.  I've extended the unittest to check that the file timestamps are later (!!), which ensures "ordering".

quilt/test/test_install.py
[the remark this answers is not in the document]
asah: (discussed: this is fixed in kevin's forthcoming change)

PR #182, opened 2017-11-13
Find fails on YAML sequences

akarve:
Some pandas args take lists (which become sequences in YAML). `build.py:find` did not work for this case; it would descend the children of the sequence and then blow up because, e.g. in test case below, the children (ints) are not iterable.

```
subgroup:
   # transform and skiprows apply to all children, unless overidden
   transform: tsv
   skiprows: [0]
```


akarve:
will rebase and resub

PR #183, opened 2017-11-14
`find()` fails on YAML sequences

akarve:
Some pandas args take lists  (sequences in YAML). `build.py:find` did not work for this case; it would descend the children of the sequence and then blow up because, e.g. in test case below, the children (ints) are not iterable.

```
subgroup:
   # transform and skiprows apply to all children, unless overidden
   transform: tsv
   skiprows: [0,5,20]
```

### DIFF
--- a/quilt/tools/build.py
+++ b/quilt/tools/build.py
@@ -203,17 +203,17 @@ def build_package(username, package, yaml_path, checks_path=None, dry_run=False,
         only descend iterables
         """
         if(isinstance(value, Iterable)):
-          for k, v in iteritems(value):
-              if k == key:
-                  yield v
-              elif isinstance(v, dict):
-                  for result in find(key, v):
-                      yield result
-              elif isinstance(v, list):
-                  for item in v:
-                      for result in find(key, item):
-                          yield result
-      
+            for k, v in iteritems(value):
+                if k == key:
+                    yield v
+                elif isinstance(v, dict):
+                    for result in find(key, v):
+                        yield result
+                elif isinstance(v, list):
+                    for item in v:
+                        for result in find(key, item):
+                            yield result
+        
     build_data = load_yaml(yaml_path)
     # default to 'checks.yml' if build.yml contents: contains checks, but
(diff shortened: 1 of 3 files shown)


akarve:
@asah this is a proper diff of the fix; yes I checked for callers

quilt/tools/build.py
```diff
@@ -198,16 +198,20 @@ def build_package(username, package, yaml_path, checks_path=None, dry_run=False,
...
-            if k == key:
-                yield v
-            elif isinstance(v, dict):
-                for result in find(key, v):
-                    yield result
-            elif isinstance(v, list):
-                for item in v:
-                    for result in find(key, item):
+        try:
+            vals = iteritems(value)
```
dimaryaz: Leave it as it was, a part of the `for ... in`. Storing the iterator is kind of weird.

quilt/tools/build.py
```diff
@@ -198,16 +198,20 @@ def build_package(username, package, yaml_path, checks_path=None, dry_run=False,
...
+                if k == key:
+                    yield v
+                elif isinstance(v, dict):
+                    for result in find(key, v):
                         yield result
+                elif isinstance(v, list):
+                    for item in v:
+                        for result in find(key, item):
+                            yield result
+        except AttributeError:
```
dimaryaz: This feels hacky, and can hide actual errors. Check if the object is an instance of `collections.Iterable`?

quilt/tools/build.py
```diff
@@ -198,17 +198,21 @@ def build_package(username, package, yaml_path, checks_path=None, dry_run=False,
...
-                for result in find(key, v):
-                    yield result
-            elif isinstance(v, list):
-                for item in v:
-                    for result in find(key, item):
+        """
+        find matching nodes recursively;
+        only descend iterables
+        """
+        if(isinstance(value, Iterable)):
```
dimaryaz: Remove parens and add a space: `if isinstance(...):`

PR #189, opened 2017-11-15
quilt install (no arg) tries to use quilt.yml

asah:
here we go!


asah:
weird... this should already be fixed
<https://github.com/quiltdata/quilt-compiler/commit/539605431fb0c7b273ca2a8089efdc407b1b9bdc>
in master and then pulled into this branch... you sure you sync'd to this
branch?

akarve:
I did `pip install git+https://github.com/quiltdata/quilt-compiler.git@quilt-yml-as-default` which should get me top of tree, right?

asah:
found it -- master was still broken... fixed and merged
into quilt-yml-as-default

try it now.

akarve:
OK tested and ran into a few issues:
1. Something that looks like debug output gets printed: `vgauthier/DynamicPopEstimate: o=vgauthier p=DynamicPopEstimate s=[] h=None v=None t=None`
1. Invalid subpaths don't fail, but install entire package
1. Invalid hashes don't fail, but install something
1. Ambiguous hashes (separate PR if need be) should probably fail instead of installing latest match
1. On over-install I get prompted for every package; maybe there should be a global param to avoid prompts? I guess `--force` would do that? Can backlog/file as separate PR if needed.

Here's my `quilt.yml` with comments:
```
packages:
  - danWebster/sgRNAs/brunello:a972d92 # invalid subpath; just installs entire pkg; should fail 
  - vgauthier/DynamicPopEstimate # OK
  - akarve/sales:h:abcdefgh # invalid hash but installs anyway
  - pearson/ece408:t:latest # OK
```

quiltdata:
nice catches - fixes pushed, pls try again.

subpath: real bug that was hiding this issue.  fixed & extended the test
suite.

invalid hash installs anyway: I couldn't repro this... anyway, I've
extended the test suite.

akarve/sales:h:abcdefgh # invalid hash but installs anyway

==> I could not reproduce this -- I'm getting "could not find hash
abcdefgh for package akarve/sales."

$ quilt install @quilt.yml
akarve/sales:h:123456: o=akarve p=sales s=[] h=123456 v=None t=None
akarve/sales already installed.
Overwrite? (y/n) y
could not find hash 123456 for package akarve/sales.

ᐧ

asah:
ready for re-review.

asah:
@akarve @kevinemoore ready for re-review.

quilt/tools/main.py
```diff
@@ -152,7 +154,10 @@ def check_hash(group, hashstr):
         func(**kwargs)
         return 0
     except command.CommandException as ex:
-        print(ex, file=sys.stderr)
+        if re.search(r'Requirements file quilt.yml not found', str(ex)):
```
kevinemoore: This is a pretty crude hack. How about removing this and just adding a suggestion to run `quilt install -h` in the exception message (when the requirements file isn't found) in command.py line: 785?

asah:
even cruder: we're trying to separate command.py from the quilt commandline
and this creates another dependency.

On Nov 21, 2017 2:45 PM, "Kevin Moore" <notifications@github.com> wrote:

*@kevinemoore* requested changes on this pull request.
------------------------------

In quilt/tools/main.py
<https://github.com/quiltdata/quilt-compiler/pull/189#discussion_r152422999>
:

> @@ -152,7 +154,10 @@ def check_hash(group, hashstr):
         func(**kwargs)
         return 0
     except command.CommandException as ex:
-        print(ex, file=sys.stderr)
+        if re.search(r'Requirements file quilt.yml not found', str(ex)):

This is a pretty crude hack. How about removing this and just adding a
suggestion to run quilt install -h in the exception message (when the
requirements file isn't found) in command.py line: 785?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/quiltdata/quilt-compiler/pull/189#pullrequestreview-78283986>,
or mute the thread

akarve:
OK, just re-tested.

* `--force` works, cool
* Bad hashes and bad subpaths are handled, cool

A few issues:
* Seeing this what looks like debug output and should be removed?`o=vgauthier p=DynamicPopEstimate s=[] h=None v=None t=None`
* Take an example of a known bad hash (or bad subpath) `akarve/sales:h:abcde`; user will be prompted to approve over-write (y/N) only to have that the package install fail. Failure should occur **before** over-write prompt (otherwise user is approving something that will never happen and is invalid)
* Error messages need some tiny tweaking and additional info:
** `could not find hash abcdefgh for package akarve/sales` should be `Invalid hash for package akarve/sales: abcdefgh`
** `Invalid subpath: 'librarie'` should be `Invalid subpath for package foo/bar: librarie`

Good stuff and looking forward to seeing this in production so all of GH can use `quilt.yml` :)

asah:
great catches, I'm on it.  Also, once we sat down 1x1, Kevin and I found a
nice solution to the Bad Hack(tm).  Will fix together.

quilt/tools/main.py
```diff
@@ -96,7 +97,8 @@ def check_hash(group, hashstr):
     tag_remove_p.set_defaults(func=command.tag_remove)
 
     install_p = subparsers.add_parser("install")
-    install_p.add_argument("package", type=str, help="owner/package_name[/path/...]")
+    install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
+                           nargs="?", default="@quilt.yml")
```
kevinemoore: Now that you've made 'quilt.yml' a constant in const.py, you should use that constant here.

quilt/tools/main.py
```diff
@@ -152,7 +154,10 @@ def check_hash(group, hashstr):
         func(**kwargs)
         return 0
     except command.CommandException as ex:
-        print(ex, file=sys.stderr)
+        if str(ex) == command.DEFAULT_QUILT_YML_NOT_FOUND_MSG:
```
kevinemoore: This is much safer than the previous version using regex, but it's still pretty hard to understand this code without knowing the background. IMHO, it would be easier to follow the code if all the logic for install were above where the install args are defined.

quilt/tools/main.py
```diff
@@ -152,7 +154,10 @@ def check_hash(group, hashstr):
...
+            parser.print_help()
+        else:
+            print(ex, file=sys.stderr)
```
kevinemoore: It's a matter for another PR, but we should probably call parser.print_help on all command exceptions, and probably print out the exception message with it.

asah:
@kevinemoore good call!  By always printing help on CommandException, I was able to remove this hack entirely.

quilt/tools/main.py
```diff
@@ -153,6 +155,8 @@ def check_hash(group, hashstr):
         return 0
     except command.CommandException as ex:
         print(ex, file=sys.stderr)
+        print()
+        parser.print_help()
```
kevinemoore: This isn't quite what we want because it prints the top-level help for the quilt command. This will print the specific help for the command the user ran (and triggered the exception):

cmd = kwargs['cmd']
print(subparsers.choices[cmd].format_help())

asah:
If that's literally what you want, feel free to change it, commit and
approve the PR.  Faster than holding my hand above the keyboard.

quilt/tools/command.py
```diff
@@ -783,16 +785,18 @@ def install_via_requirements(requirements_str, force=False):
...
     else:
         yaml_data = yaml.load(requirements_str)
     for pkginfo in yaml_data['packages']:
         owner, pkg, subpath, hash, version, tag = parse_package_extended(pkginfo)
-        print('{}: o={} p={} s={} h={} v={} t={}'.format(
-            pkginfo, owner, pkg, subpath, hash, version, tag))
         package = owner + '/' + pkg
-        if subpath is None:
-            package += '/' + subpath
+        if subpath is not None:
```
asah: fix for bug that @akarve found

akarve:
Approving.

Please backlog an issue to refactor error messages with this note: The design of catching all errors and printing help is undesirable because it obscures the actual message (making it hard to act on the actually error), and doesn't really work for installing from `quilt.yml`.

Here's an example:
```
Invalid hash for package akarve/sales: abcdefgh

usage: quilt install [-h] [-f] [-x HASH | -v VERSION | -t TAG] [package]

positional arguments:
  package               owner/package_name[/path/...] or @filename

optional arguments:
  -h, --help            show this help message and exit
  -f, --force           Overwrite without prompting
  -x HASH, --hash HASH  Package hash
  -v VERSION, --version VERSION
                        Package version
  -t TAG, --tag TAG     Package tag - defaults to 'latest'
```

PR #191, opened 2017-11-17
Add support for resuming downloads of partially-downloaded fragments

dimaryaz:
- Use the "Range" header to get partial content
- Ungzip the content manually instead of letting requests do it
- Set connect/read timeouts
- Update the tests to use bytes instead of strings

### DIFF
--- a/quilt/test/test_install.py
+++ b/quilt/test/test_install.py
@@ -133,7 +133,7 @@ class InstallTest(QuiltTestCase):
 
         with open(os.path.join(teststore.package_path(user, package),
                                Package.CONTENTS_DIR,
-                               contents_hash), 'rb') as fd:
+                               contents_hash), 'r') as fd:
             file_contents = json.load(fd, object_hook=decode_node)
             assert file_contents == contents
 
--- a/quilt/test/test_install.py
+++ b/quilt/test/test_install.py
@@ -81,11 +81,11 @@ class InstallTest(QuiltTestCase):
             file_contents = json.load(fd, object_hook=decode_node)
             assert file_contents == contents
 
-        with open(os.path.join(self._store_dir, 'objs/{hash}'.format(hash=table_hash)), 'rb') as fd:
+        with open(teststore.object_path(objhash=table_hash), 'rb') as fd:
             contents = fd.read()
             assert contents == table_data
 
-        with open(os.path.join(self._store_dir, 'objs/{hash}'.format(hash=file_hash)), 'rb') as fd:
+        with open(teststore.object_path(objhash=file_hash), 'rb') as fd:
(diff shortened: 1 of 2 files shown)


quilt/test/test_install.py
```diff
@@ -75,11 +76,11 @@ def test_install_latest(self):
             file_contents = json.load(fd, object_hook=decode_node)
             assert file_contents == contents
 
-        with open('quilt_packages/objs/{hash}'.format(hash=table_hash)) as fd:
+        with open('quilt_packages/objs/{hash}'.format(hash=table_hash), 'rb') as fd:
```
kevinemoore: Instead of hard-coding the path here in the tests, use a PackageStore object to get the path.

quilt/tools/command.py
```diff
@@ -884,29 +888,55 @@ def install(package, hash=None, version=None, tag=None, force=False):
...
-                    # `requests` will automatically un-gzip the content, as long as
-                    # the 'Content-Encoding: gzip' header is set.
-                    # To report progress, however, we need the length of the original compressed data;
-                    # we use the undocumented but technically public `response.raw.length_remaining`.
-                    for chunk in response.iter_content(chunk_size=CHUNK_SIZE):
-                        if chunk: # filter out keep-alive new chunks
+                    # Fragments have the 'Content-Encoding: gzip' header set to make requests ungzip them
+                    # automatically - but that turned out to be a bad idea because it makes resuming downloads
+                    # impossible.
+                    # HACK: For now, just delete the header. Eventually, update the data in S3.
```
kevinemoore: Do you know of a way to update the data in s3 safely?

quilt/test/test_install.py
dimaryaz: You've already updated it - I'll just have to merge and resolve conflicts.

quilt/tools/command.py
dimaryaz: Yeah, looks like it's doable: https://stackoverflow.com/questions/269840/is-it-possible-to-change-headers-on-an-s3-object-without-downloading-the-entire

quilt/test/test_install.py
```diff
@@ -80,11 +81,11 @@ def test_install_latest(self):
             file_contents = json.load(fd, object_hook=decode_node)
             assert file_contents == contents
 
-        with open(os.path.join(self._store_dir, 'objs/{hash}'.format(hash=table_hash))) as fd:
+        with open(os.path.join(self._store_dir, 'objs/{hash}'.format(hash=table_hash)), 'rb') as fd:
```
kevinemoore: I'd recommend calling: PackageStore.object_path

PR #192, opened 2017-11-18
Handle `quilt generate` for empty directories

akarve:
I wasn't sure if creating and then deleting the `empty` directory in code was the right thing to do, or if the unit test cleans up after itself (which obviously doesn't work since `build.yml` lingers when the test fails). note that git makes it impossible to check in an empty directory.

### DIFF
--- a/quilt/tools/build.py
+++ b/quilt/tools/build.py
@@ -312,13 +312,20 @@ def generate_contents(startpath, outfilename=DEFAULT_BUILDFILE):
             name == outfilename
         )
 
+    def _is_empty_dir(dir_path):
+        return os.path.isdir(dir_path) and not os.listdir(dir_path)
+
     def _generate_contents(dir_path):
         safename_duplicates = defaultdict(list)
-        for name in os.listdir(dir_path):
-            if _ignored_name(name):
-                continue
+        children = os.listdir(dir_path)
+        if not children:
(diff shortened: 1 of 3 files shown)


quilt/tools/build.py
```diff
@@ -312,14 +312,21 @@ def _ignored_name(name):
             name == outfilename
         )
 
+    def _is_empty_dir(dir_path):
+        return os.path.isdir(dir_path) and not os.listdir(dir_path)
```
dimaryaz: Tthat's not good enough: the directory could contain files starting with `.`, `build.yml`, `quilt_packages`, other empty directories - which all get ignored. You should handle this at a higher level - maybe even support empty directories and just allow empty groups.

dimaryaz: Another to support empty groups: I have the following build file with commented-out lines, and it fails:
```
contents:
  README:
    file: README.md
  copy_uniques_level0_unclustered_sample_vectorized:
    file: proc_17_108_copy_uniques_level0_unclustered_sample_vectorized.csv
  unique_comments_text_dupe_count:
#    part1:
#      file: proc_17_108_unique_comments_text_dupe_count.csv
#      nrows: 1500000
#    part2:
#      file: proc_17_108_unique_comments_text_dupe_count.csv
#      skiprows: 1500001
```

akarve:
Superseded by #219

PR #193, opened 2017-11-21
Added CLI testing

eode:
CLI testing, including means to address situations like the recent argparse / command API mismatch.  Although only a couple of tests have been added, it should be quick work to add more.   The testing method should be flexible enough to allow non-breaking changes to command API function signatures without complaining.  However, when an argparse argument or option is changed, test_cli.py will complain, or if there's a signature mismatch between argparse options and the command API.

The following tests are present:
 * Notice new, unknown argparse params.  Includes sub-parser parameters.
 * Notice missing known argparse params.  Includes sub-parser parameters.
 * Notice when tests are missing for discovered argparse params
 * Test for 'config' command<-->API compatibility
 * Test for 'push' command<-->API compatibility


eode:
Oh -- possible issue:  Quilt takes a moment to load, and once a large number of tests have been built, it might slow testing down quite a bit.

quiltdata:
Sorry, can you explain?

Worst case, we make this optional (default-on for Travis, default-off for
developers)

Anyway, neat stuff.

asah:
y I've also seen slow startup: 30+secs is definitely worth optimizing - for
this PR, if it's easy to call main() directly, then go for it.

I've filed an issue in the meantime - looks like a Hard Problem(tm):
https://github.com/quiltdata/quilt-compiler/issues/194

eode:
Yeah, #194 does seem like it could be not-so-quick to fix.

I'll look into speeding things up by skipping the subprocess load-time.  It's slightly less complete, testing-wise, than using an actual call from the CLI but not badly so.  We may never even miss it.

eode:
I have only a few remaining things that might be done here:

* Merge the quilt-yml-as-default branch so I can get the bug from #189 
* Add test for install command (including #189)
* possibly add more tests
* Squash commits into a single commit on a different branch, and (?) unless there's a better way, open a new PR.  This is mainly to reduce tree clutter, and avoid potential conflicts in test/util.py.

Aside from those, this should currently be in pretty good condition.

asah:
perfect - IMHO we can defer more tests until this first commit is
approved.  But if (when) you're blocked, go for it!

eode:
I have only a few remaining things that might be done here:

* Merge the quilt-yml-as-default branch so I can get the bug from #189 
* Add test for install command (including #189)
* possibly add more tests
* Squash commits into a single commit on a different branch, and (?) unless there's a better way, open a new PR.  This is mainly to reduce tree clutter, and avoid potential conflicts in test/util.py.

Aside from that, this branch is in pretty good condition.

eode:
Ok, I'll squish this then, and do a PR for the result -- that'll prevent GIT from cycling through the changes I made (and then removed) from utils.py as it merges, and thus avoid conflicts.

PR #195, opened 2017-11-22
Unit tests for column types

akarve:
@kevinemoore I changed null parsing a bit. Maybe I broke some of the use cases we used `keep_default_na` and `na_values` for? If so can you add a breaking unit test?


akarve:
OK making a note that we may need to pin our pandas or pyarrow version higher in the future; the fact that it's passing CI is preliminary confirmation that the current `setup.py` is adequate.

Issue #194, opened 2017-11-21
slow quilt(1) startup times due to import pyarrow (and pandas?)


asah:
1+ secs to startup quilt, which impacts use as a general purpose *nix tool.

I'm not seeing much on pyarrow but here's some info on pandas:
https://github.com/pandas-dev/pandas/issues/7282
https://github.com/pandas-dev/pandas/issues/16764

PR #198, opened 2017-11-25
CLI testing, squashed, some minor improvements

eode:
This is basically pull request #193, squashed, and with recent changes merged.  It also has a minor bugfix, and has an env-var option "QUILT_TEST_CLI_SUBPROC=True", which, if set, will cause it to use subprocesses for testing, instead of the current (faster) method.


asah:
@dimaryaz after recent breakage, I asked @eode to create a framework for testing the CLI, incl catching if the CLI accidentally changes.  From here, we can add detailed CLI tests for subtle details, e.g. error messages, return codes, etc.  This is a starter projectthis and his first commit.

@eode Dima is lead eng and can get backed up with code reviews.  I'll send you some more projects to stay incl bug fixes.  Thx!

quilt/tools/main.py
```diff
@@ -139,7 +142,16 @@ def check_hash(group, hashstr):
...
+    return parser
+
+
+def main(args=None):
+    """Build and run parser
+
+    :param args: cli args from tests
+    """
+    parser = argument_parser()
+    args = parser.parse_args(args) if args is not None else parser.parse_args()
```
dimaryaz: I think `args = parser.parse_args(args)` should be ok - `None` is already the default value, so it just works.

quilt/test/test_cli.py
```diff
@@ -0,0 +1,652 @@
...
+                new[k] = _mcopy(v) if v else None
+            return new
+        return _mcopy(self)
+
+    def __repr__(self):
+        args = '\n'.join(repr(x) for x in self)
+        return 'ArgparseIntrospector with args:\n' + args
+
+
+class RecursiveMappingWrapper(collections.MutableMapping):
```
dimaryaz: This feels really over-engineered to me, to be honest...

quilt/test/test_cli.py
```diff
@@ -0,0 +1,652 @@
...
+
+# inspect.argspec is deprecated, so
+try:
+    from funcsigs import signature  # python 2.7
+except ImportError:
+    from inspect import signature
+
+
+## "Static" vars
+_TEST_DIR = os.path.abspath(inspect.stack()[0][1])
```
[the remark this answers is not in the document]
eode: ..because __file__ is unreliable in some versions of python.  I can just use __file__, though, if you'd rather.

quilt/test/test_cli.py
```diff
@@ -0,0 +1,652 @@
...
+PACKAGE_DIR = os.path.join('/', *_TEST_DIR.split(os.path.sep)[:-3])
```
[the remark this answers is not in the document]
eode: Yep, I thought I'd fixed that.  Thanks for the catch.

quilt/tools/main.py
eode: Not in the case where blank args are sent, if anyone ever tests what happens when `quilt` is called without args.  ..although, we may never test that.

quilt/tools/command.py
```diff
@@ -67,8 +68,19 @@
...
-
+try:
+    VERSION = pkg_resources.require('quilt')[0].version
+except pkg_resources.DistributionNotFound:
+    # Not installed, so we'll get the version manually.
+    try:
+        mydir = os.path.abspath(stack()[0][1])
+        pkgdir = os.path.join('/', *mydir.split(os.path.sep)[:-3])
+        text = open(os.path.join(pkgdir, 'setup.py')).read()
+        text = text.split('\nQUILT_VERSION = ', 1)[1]
```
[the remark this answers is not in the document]
eode: It is super-hacky, but at the time it was written, the goal was to enable invoking main.py from the tests directory so that pytest runs main.py, without an install, and without creating any new files -- although, some of those restrictions may have been made in part simply as questions on the job description.  Also, this was before I was really familiar with the codebase, and before I had a deeper understanding of what's needed.

Calling from the CLI is technically more complete, but it's significantly slower, and with modifications I've added to main.py, isn't absolutely necessary.  The likelihood that there will be a significant difference between doing a syscall of `['quilt'] + quilt_args` vs calling `main.argument_parser()` and `main.main(args)` is fairly small.

I'd be happy to either store the version in a separate version file and read that, or simply forego actual CLI calls altogether in favor of simulated CLI calls.

quilt/test/test_cli.py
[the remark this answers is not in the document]
eode: Wait, no, I remember now -- I was thinking of using os.path.join(), which I did do -- os.path.split only splits off the last item.  I thought this was cleaner than `os.path.split(os.path.split(os.path.split(_TEST_DIR)[0])[0])[0]`

quilt/test/test_cli.py
```diff
@@ -0,0 +1,652 @@
...
+    """Shows CLI param test coverage based on reporting by test functions.
+
+    Returns a dict with `'missing'` and `'percentage'` keys.
+    """
+    param_tree = RecursiveMappingWrapper(get_current_param_tree())
+    ideal_data = [tuple(v) for v in param_tree.iterpaths()]
+    ideal = set(ideal_data)
+    tested = set(tuple(v) for v in TESTED_PARAMS)
+    # remove covered test cases that aren't in the main argparser
+    tested = tested - tested.difference(ideal)
```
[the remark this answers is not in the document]
eode: Yes, that would more ideally be an &.  >.>

quiltdata:
Please add a *short* one line comment explaining the performance issue.  If
you can make it less hacky, that would great ideally moving the hacky code
from command.py to some file in test/ where hacks are (more) allowed.

quilt/test/test_cli.py
eode: It was a piece of code from my personal library that turns any mapping into a recursively-accessible mapping.  Since it was more than is needed, and we have no need for it elsewhere, I've taken just the snippets we need out, and integrated them into ArgparseIntrospector now.

quilt/test/test_cli.py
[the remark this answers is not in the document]
eode: Fixed.

quilt/test/test_cli.py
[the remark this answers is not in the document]
eode: Fixed.

quilt/test/test_cli.py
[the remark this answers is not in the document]
eode: Fixed.

quilt/tools/command.py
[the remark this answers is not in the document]
eode: command.py and setup.py now read from a new VERSION file.  This seemed the least-hacky way of enabling running quilt when not installed.  If this isn't preferred, we can just remove the ability to run quilt when not installed.

quilt/tools/main.py
eode: I misread what you said there.  Fixed.

eode:
@quiltdata Re: One-line comment about quilt performance issue:  That's issue #194, and I've circumvented it by defaulting to not using an actual CLI call, but using main.argument_parser() and main.main(args) instead.  Commented as #NOTE: in test_cli.py.

eode:
@dimaryaz I think this is ready for a re-review.

compiler/quilt/tools/command.py
```diff
@@ -68,7 +69,18 @@
 
 LOG_TIMEOUT = 3  # 3 seconds
 
-VERSION = pkg_resources.require('quilt')[0].version
+try:
+    VERSION = pkg_resources.require('quilt')[0].version
+except pkg_resources.DistributionNotFound:
+    # Not installed, so we'll get the version manually.
```
dimaryaz: So I still don't understand: why is this needed? When would quilt not be installed?

compiler/setup.py
```diff
@@ -12,9 +14,15 @@ def readme():
     """
     return readme_short
 
+
+def quilt_version():
+    pkgdir = os.path.split(os.path.abspath(__file__))[0]
```
dimaryaz: `pkgdir = os.path.dirname(__file__)`

I don't think you even need absolute path; relative ones work, too.

compiler/setup.py
```diff
@@ -12,9 +14,15 @@ def readme():
...
+    return open(os.path.join(pkgdir, 'VERSION')).read().strip()
```
dimaryaz: Need to use `with open(...):`, so the fd is not leaked.

compiler/quilt/tools/command.py
```diff
@@ -68,7 +69,18 @@
...
+    try:
+        filedir = os.path.split(os.path.abspath(__file__))[0]
+        quiltdir = os.path.split(filedir)[0]
+        pkgdir = os.path.split(quiltdir)[0]
+        VERSION = open(os.path.join(pkgdir, 'VERSION')).read().strip()
```
dimaryaz: This looks similar to the helper function below?

compiler/quilt/test/test_cli.py
```diff
@@ -0,0 +1,589 @@
...
+
+# inspect.argspec is deprecated, so
+try:
+    from funcsigs import signature  # python 2.7
+except ImportError:
+    from inspect import signature
+
+
+## "Static" vars
+_TEST_DIR = os.path.split(os.path.abspath(__file__))[0]
```
dimaryaz: Use `os.path.dirname`

eode: Original code:

@dimaryaz wrote:
> There's os.path.split (which also supports both / and \ on win32)

Was told by you to use os.path.split, over splitting on os.path.sep (which is /, or \ on win32).
Changed to your current preference, and changed other occurrences as well.

compiler/quilt/tools/command.py
eode: @eode wrote:
> It is super-hacky, but at the time it was written, the goal was to enable invoking main.py from the tests directory so that pytest runs main.py, without an install, and without creating any new files -- although, some of those restrictions may have been made in part simply as questions on the job description.

To be more clear:  I was instructed to do so.  It was one of the requirements I was initially given on this specific project.  

However, it may be a requirement that has has since changed, like the requirement on not creating any new files has (a `VERSION` file was permitted).

..in either case, it was what I was instructed to do, and I found a way to do it with the specified restrictions.  If you want that capacity removed, or if you'd like for me to use a different implementation, I'll follow the new requirements you set forth.

As to when quilt wouldn't be installed:  I think it was requested of me in order to fulfil a specific (or potential) testing or CI setup need, but I don't know the reason definitively.

compiler/quilt/tools/command.py
eode: * they both operate when not installed
* `quilt.tools.command` can't reasonably be imported from `setup.py`
* `setup.py` can't be reasonably imported from `command.py`

...can you be more specific with what you'd like me to do here?

compiler/setup.py
eode: Nope, I it's not necessary in this case, just done habitually.  Working with absolute paths isn't a bad habit, though, and it doesn't really obscure what's happening here.  Changed to your preference, anyway, as there's no need for it here.

compiler/setup.py
eode: In what case would the FD be leaked here?

* There's no potential for a circular reference here
* The file is opened and read immediately, then closes automatically when not assigned to a variable.
* The closest possible thing to a leak would be an exception in read(), which would result in a program exit
  * Even if that were the case, the FD would only remain open (held in the stack frame associated with the traceback at sys.last_traceback) until the next exception occurs.
    * Even if another exception didn't occur (unlikely) the FD would only remain open until setup exited, and would have no ill effects.
  * Even if the OS is hard-killing the process (KILL as opposed to INT on POSIX systems), then *all* open files are generally 'leaked', and it's up to the OS to reclaim those -- in any case, that's Python and the OS's job, not ours.

compiler/quilt/tools/command.py
asah: This was me -- at the time, I was cautious about requiring pip install to be run before unittests can pass... but more & more, I'm seeing that pip install is kinda required for even basic development.

So if you want to require this, I'm ok with it.  @dimaryaz gets the final say, of course.

eode: ..sounds good.

dimaryaz: Yeah, let's require quilt to be installed. We already need all of the requirements to be installed - so we're not saving anything by not requiring quilt itself.

eode: Sounds good, I'll excise that code.

eode:
..improved some commnets [sic].  Also, minor bugfix re: reading version using cli during Subprocess mode in test_cli.  Verified that this works with QUILT_TEST_CLI_SUBPROC both True and False.  This should be set to True on CI testing for more thorough testing of the cli.

compiler/quilt/tools/command.py
```diff
@@ -11,6 +11,7 @@
 import json
 import os
 import re
+from inspect import stack
```
dimaryaz: Delete this?

compiler/setup.py
```diff
@@ -1,4 +1,6 @@
 from setuptools import setup, find_packages
+import os
```
dimaryaz: Delete this

compiler/quilt/tools/main.py
```diff
@@ -13,18 +13,23 @@
 from . import command
 from .const import DEFAULT_QUILT_YML
 
+# Importing this monkey-patches `command` if testing the CLI in subprocess mode
+from ..test import test_cli     # pylint:disable=W0611
```
dimaryaz: Importing this in production code feels wrong... I think doing a conditional import and patching would be better. Also, make monkey-patching a helper function instead of a side-effect of import. So something like this:

```
if os.environ.get('QUILT_TEST_CLI_SUBPROC') == "True":
  from ..test import test_cli
  test_cli.monkey_patch(...)
```

eode: Yeah, I originally did something similar.  I'll go back to what I had.  I was mainly just trying to get the explicitly testing-related code out of main.py.

PR #199, opened 2017-11-26
Fixed issue where test_import interferes with test_install.

eode:
Underlying issue: _log() removed response hooks from the requests
session without restoring them when done.

This does not yet cause quilt-yml-as-default to pass all tests, as there are still a couple of issues with test_short_hash (exception in teardown as well as a test failure).

### DIFF
--- a/quilt/test/test_install.py
+++ b/quilt/test/test_install.py
@@ -48,7 +48,7 @@ class InstallTest(QuiltTestCase):
         h.update(file_data)
         file_hash = h.hexdigest()
         return file_data, file_hash
-    
+
     @classmethod
     def make_contents(cls, **args):
         contents = RootNode(dict(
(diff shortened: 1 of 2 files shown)


eode:
Strangely, this passes on travis, but not locally.

eode:
Ok, starting to understand the requests 'response' class mocking, and why there's a teardown issue as well as a test failure.  It provides both the expected and the actual URL that way -- nice engineering.

On my machine (only, apparently, as this branch passed on travis), InstallTest.test_short_hashes fails with the following difference:
expected URL (PUT): https://pkg.quiltdata.com/api/version/foo/bar/1.0
called URL (GET): https://pkg.quiltdata.com/api/log/foo/bar/

..which led me to figure that we were just missing the _mock_log() call, which seems to have been the case.

However, the fact that in commit 6925960, it *didn't* pass locally, but *did* pass on travis -- is that an issue?

asah:
nice!  small thing for the future: to simplify code review, pls put whitespace/style/pylint fixes in separate commit(s) with a simple commit msg e.g. "pylint".  If you're coding both at once, commit the style fixes first e.g. in your editor or running pylint.  If you need, git stash to tuck away the changes... sorry if obvious.

PR #201, opened 2017-11-27
Error msg instead of traceback on bad yaml syntax

eode:


quilt/test/test_command.py
```diff
@@ -242,14 +242,26 @@ def mock_git_clone(cmd):
...
 
         from quilt.data.user import test
         assert hasattr(test, 'foo')
         assert isinstance(test.foo(), pd.DataFrame)
 
+    def test_build_yaml_syntax_error(self):
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, 'data')
+        buildfilepath = os.path.join(path, 'build_bad_syntax.yml')
+        try:
```
asah: can you use self.assertRaises() ?

quilt/tools/command.py
```diff
@@ -410,14 +410,23 @@ def _build_internal(package, path, dry_run, env):
...
                 if os.path.exists(tmpdir):
                     rmtree(tmpdir)
         else:
-            build_from_path(package, path, dry_run=dry_run, env=env)
+            try:
+                build_from_path(package, path, dry_run=dry_run, env=env)
+            except yaml.scanner.ScannerError as ex:
+                message_parts = str(ex).split('\n')
+                message_parts.insert(0, "Syntax error while reading {!r}".format(path))
+                raise CommandException('\n  '.join(message_parts))
```
asah: duplicate code - can you move into build_from_path() ?

eode: I could, however, move that section to a function, and try: except: that, or try:except the whole block, or send the path into build_from_path as a new param "original_path" or some such..

I didn't know if the small amount of duplication was worth shuffling things about for, but I'd be happy to find a way.

asah: ah, I mean move the except block down into build_from_path:

def build_from_path(package, path, dry_run=False, env='default'):
...

    try:
...
   except yaml.scanner.ScannerError as ex:
        message_parts = str(ex).split('\n')
        message_parts.insert(0, "Syntax error while reading {!r}".format(path))
        raise CommandException('\n  '.join(message_parts))
    except BuildException as ex:
        raise CommandException("Failed to build the package: %s" % ex)

eode: Ah, I missed the ability to compile from a directory, and thought that 'build from path' took a specific YML file (only).  I see what it's doing now, and I'll shift that over.

quilt/test/test_command.py
eode: ..will do.

quilt/tools/command.py
eode: I had originally had it there so it would catch the original `path` during git repo pulls, instead of the temp dir sent to build_from_path.

eode:
Ok, this should be ready now.

asah:
pls chg merge to be from master, not quilt-yml-as-default which has been merged.

eode:
Can't do that with this PR, so I've opened #204.

PR #202, opened 2017-11-28
Quilt rm

kevinemoore:
Adding a simple, coarse-grained quilt rm command. This first version removes all instances of a package from the local store and isn’t safe to be overlapped with concurrent quilt builds or installs.


quilt/test/test_command.py
```diff
@@ -306,3 +306,34 @@ def callback(request):
...
+        Test removing a package.
+        """
+        mydir = os.path.dirname(__file__)
+        build_path = os.path.join(mydir, './build_simple.yml')
+        command.build('foo/bar', build_path)
+
+        command.rm('foo/bar', force=True)
+        teststore = store.PackageStore(self._store_dir)
+        assert not os.path.isdir(teststore.package_path('foo', 'bar'))
+
```
asah: pls add corner cases:
- rm on non-existent package
- rm on malformed pkg e.g. with subpkg, empty pkgname

pls ping @eode to add a CLI unittest once this is committed (his CLI unittest framework is in review: https://github.com/quiltdata/quilt-compiler/pull/198

quilt/tools/core.py
```diff
@@ -61,20 +61,16 @@ def __init__(self, children):
         assert isinstance(children, dict)
         self.children = children
 
-    def preorder_tablenodes(self):
+    def preorder(self):
```
asah: do you mind adding a function-comment explaining this func, esp since you're changing core behavior...

quilt/tools/main.py
```diff
@@ -140,6 +140,11 @@ def check_hash(group, hashstr):
     inspect_p.add_argument("package", type=str, help=HANDLE)
     inspect_p.set_defaults(func=command.inspect)
 
+    rm_p = subparsers.add_parser("rm")
+    rm_p.add_argument("package", type=str, help=HANDLE)
+    rm_p.add_argument("-f", "--force", action="store_true", help="Remove without prompting")
+    rm_p.set_defaults(func=command.rm)
+
```
asah: Given a single name, are we sure about "rm" vs "uninstall" which makes it clear that this is local.  Sorry for the late question...

quilt rm/del/delete is ambiguous whether it applies server-side, client-side or both...

also, if you intend to someday add metadata-delete-only, you might want to spec this out now, so the syntax won't be awkward...

quilt/tools/package.py
```diff
@@ -71,21 +71,25 @@ def __init__(self, store, user, package, path, contents=None):
...
+            if not os.path.exists(latest_tag):
+                msg = "Could not find latest tag for package {0}/{1}"
+                raise PackageException(msg.format(self._user, self._package))
+
+            with open (latest_tag, 'r') as tagfile:
+                instance_hash = tagfile.read()
 
         contents_path = os.path.join(self._path, self.CONTENTS_DIR, instance_hash)
+        if not os.path.isfile(contents_path):
+            msg = "{h} is not a valid hash for package {u}/{p}"
```
asah: isn't the std errmsg syntax something more like:

$ grep 'Invalid hash' *py */*py | grep package
tools/command.py:        raise CommandException("Invalid hash for package {owner}/{pkg}: {hash}".format(

pls also avoid one letter var/arg/fmt names... ;-)

quilt/tools/package.py
```diff
@@ -257,7 +261,8 @@ def get_obj(self, node):
             self._check_hashes(node.hashes)
             return self._dataframe(node.hashes, node.format)
         elif isinstance(node, GroupNode):
-            hash_list = [h for c in node.preorder_tablenodes() for h in c.hashes]
+            hash_list = [h for c in node.preorder() if isinstance(c, TableNode)
+                         for h in c.hashes]
```
asah: avoid one letter varnames... e.g. what's 'c' ?

pylint should also complain - and if not, let's fix this ;-) for most cases (i for index var is fine...)

quilt/tools/store.py
```diff
@@ -221,6 +243,34 @@ def temporary_object_path(self, name):
...
+        # TODO: the temporary object directory probably shouldn't
+        # live inside the objects directory. Remove the hard-coded
+        # 'tmp' below.
+        all_obj.remove('tmp')
+
+        for pkg in self.iterpackages():
+            for node in pkg.get_contents().preorder():
+                # TODO: the or below isn't scalable. Add a common baseclass for
+                # File and Table nodes like DataNode in nodes.py.
+                if isinstance(node, TableNode) or isinstance(node, FileNode):
```
asah: python has syntax for multiple checks?
https://stackoverflow.com/questions/33311258/python-check-if-isinstance-any-type-in-list

quilt/test/test_command.py
kevinemoore: Added non-existent package and a sub-package case.

quilt/tools/core.py
kevinemoore: Done.

quilt/tools/main.py
kevinemoore: I think 'rm' is the clearest. It matches Docker and Linux. Uninstall feels like a close second because it matches pip.

quilt/tools/package.py
kevinemoore: Done. "C" stood for child, but it's not spelled out.

quilt/tools/store.py
kevinemoore: Good call. Fixed.

compiler/quilt/tools/store.py
```diff
@@ -230,6 +252,29 @@ def cache_path(self, name):
...
+            for node in pkg.get_contents().preorder():
+                # TODO: the or below isn't scalable. Add a common baseclass for
+                # File and Table nodes like DataNode in nodes.py.
+                if isinstance(node, (FileNode, TableNode)):
+                    for objhash in node.hashes:
+                        all_obj.remove(objhash)
+
+        for obj in all_obj:
+            print("Removing: {0}".format(obj))
+            os.remove(self.object_path(obj))
```
dimaryaz: What happens to the build cache?

compiler/quilt/tools/command.py
```diff
@@ -1129,3 +1129,18 @@ def _print_node(node, prefix, child_prefix, name, path):
...
+    owner, pkg = parse_package(package)
+
+    if not force:
+        confirmed = input("Remove {0}? (y/n)".format(package))
+        if confirmed.lower() != 'y':
+            return
+
+    store = PackageStore()
+    store.remove_package(owner, pkg)
+    store.prune()
```
dimaryaz: I'm not convinced that automatic pruning is the right thing to do:
1) Logically, they're two separate actions. `rm` is per package; pruning is not. You may want to prune even without deleting a single package - e.g., after rebuilding a package a bunch of times.
2) This could have unexpected results: e.g., delete objects from a large download that got interrupted.

(One workaround is to collect all used objects, then delete the package, then collect them again, and only delete the newly-orphaned objects... Though not sure it's worth the effort.)

compiler/quilt/tools/store.py
kevinemoore: Good catch. The safest thing seems like checking that objects exist before using a cache entry. I added that and a unit test to catch this bug.

compiler/quilt/tools/command.py
kevinemoore: I think you're right that we should avoid surprises and not simply call prune as part of rm. I've added an input set to prune that will limit the pruning to only that set of objects. Now, remove collects all the objects referenced in the package that's being deleted and passes that set to prune. That way, only objects from the removed package are eligible for deletion.

compiler/quilt/tools/build.py
```diff
@@ -158,8 +158,9 @@ def _build_node(build_dir, package, name, node, fmt, target='pandas', checks_con
                         cachedobjs = cache_entry['obj_hashes']
                         assert isinstance(cachedobjs, list)
 
-            if cachedobjs:
-                # FIXME: Add already present object to the package
+            # Check to see that cached objects actually exist in the store
+            if cachedobjs and all([os.path.exists(store.object_path(obj)) for obj in cachedobjs]):
```
dimaryaz: Don't need the `[` and `]` - that just creates a list unnecessarily.

compiler/quilt/tools/command.py
```diff
@@ -1129,3 +1129,19 @@ def _print_node(node, prefix, child_prefix, name, path):
...
+
+    if not force:
+        confirmed = input("Remove {0}? (y/n)".format(package))
+        if confirmed.lower() != 'y':
+            return
+
+    store = PackageStore()
+    deleted = store.remove_package(owner, pkg)
+    for obj in deleted:
+        print("Removing: {0}".format(obj))
```
dimaryaz: I think it's kind of weird that `print` happens after the objects get deleted. At first, I actually thought you forgot to call `os.remove` here. What's wrong with printing in the `prune` method?

kevinemoore: We're trying to separate the API from interactive commands so it seems better to move the printing to command.py. I changed "removing" to "removed" in the output to make it clear that it already happened.

PR #203, opened 2017-11-28
Help suppression

eode:
This is a work-in-progress on changing how CLI text is displayed.  It's pretty minimally-invasive, but does touch core.py, so I figured I'd check with you before I go further.  It makes the following changes:

* CommandException gains an optional keyword, `show_help`, which defaults to `True`
* tools.main.main() now skips showing help if it gets a CommandException that has show_help=False.
* Should be 100% backwards compatible, as CommandException previously took no kwargs.
* Puts error messages below the help message, if both are present

As an example, I've changed the CommandException that is raised when a path doesn't exist -- so:
`quilt build user/test fakefile` 
will give the message emitted in command.py, without help, such as 
`fakefile does not exist.`

Points to consider:
* Do we need a standard format for errors?
* Should we default to no-help and only add help if a command sets `show_help=True`?


akarve:
* Yes we should consider a standard format for errors, with a `problem` and a `solution`. Problem string is usually of the form `"Invalid SOMETHING: %s" % x`. `solution` should be a required arg and should be specify how the user can remedy the situation. For now a single string arg for each is sufficient. Keras has some nice [error message formats (scroll down towards end)](https://blog.keras.io/user-experience-design-for-apis.html) if you are inspired to format nicely. Looking for something simple and clean. 
* For now `show_help=True` is a fine default

akarve:
Updated comment to clarify that we should not show help text by default. I'm most cases the command syntax will not be the cause of the exception.

eode:
In the interest of backward compatibility with the `CommandException`s we already have, I'm thinking of making the args for a CommandException be as follows:

`CommandException(problem, solution='', returncode=1, **kwargs)`

* Problem and solution are both stripped of whitespace
* When only the problem is present, it is the __str__() value.
* When problem and solution are present, they are joined by "\n---\n"
* Error code is used as the exit code for the program.  We might want to standardize those, and keep them in 'const.py'.

This seems like a good interim step on moving toward a required `solution` param, and once we have complete coverage there, we can make the `solution` param a required one.

Otherwise, I can go through and make solutions for all existing problems.

eode:
Ok, is this alright with you, @akarve?

kevinemoore:
There's a lot of good ideas here, especially standardizing error messages through helpers, but I'm closing this PR because it's gotten more complicated than it needs to be to solve the problem it's meant to solve.

In all cases except quilt install, the argparser properly detects incorrectly specified arguments and displays the proper help text. I'm closing this PR and proposing fixing the argument specification for install so that it also follows that pattern. Let's plan to revisit standardizing exception messages in a later PR.

PR #204, opened 2017-11-28
Use error msg instead of traceback on bad yaml syntax

eode:
Includes these (squashed) commits:

remove duplication by shifting exc into build_from_path
Use assertRaisesRegex for test_build_yaml_syntax_error
Use assertRaisesRegex from `six` module, not from TestCase superclass.


asah:
Cc @akarve

quilt/test/test_command.py
```diff
@@ -250,6 +250,13 @@ def mock_git_clone(cmd):
         assert hasattr(test, 'foo')
         assert isinstance(test.foo(), pd.DataFrame)
 
+    def test_build_yaml_syntax_error(self):
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, 'data')
+        buildfilepath = os.path.join(path, 'build_bad_syntax.yml')
+        with assertRaisesRegex(self, command.CommandException, 'Syntax error .*'):
```
asah: .* does nothing (pls remove), since assertRaisesRegex is already a substring match and not a whole-string match:

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

quilt/tools/command.py
```diff
@@ -410,6 +410,8 @@ def _build_internal(package, path, dry_run, env):
             try:
                 _clone_git_repo(url, branch, tmpdir)
                 build_from_path(package, tmpdir, dry_run=dry_run, env=env)
+            except CommandException:
+                raise
```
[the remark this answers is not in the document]
eode: Ah.  Yes, but a CommandException would already have a (more specific) error message in it, wouldn't it?

quilt/test/test_command.py
eode: Right.

quilt/tools/command.py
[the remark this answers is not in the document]
eode: Oh, that git clone message includes the caught, more-specific message.  Fixing.

eode:
Ok, @asah, comments have been resolved.

Should I merge the changes from master into this branch, or do another PR on a new branch, squashing changes from this?

asah:
Merge master into this, not a big deal to have lots of commits in a PR as
long as the ultimate diff is small and you squash commits when merging back
into master

eode:
Ok, I've raised the error in build, and with that it gets the "Failed to build the package:" prefix on the error message.

On a side note, can I change the order of the reporting of error and help messages so that error messages are shown last?  ..and perhaps change the output to stderr so there's no race condition between argparse and our error message?

compiler/quilt/test/test_command.py
```diff
@@ -250,6 +250,12 @@ def mock_git_clone(cmd):
         assert hasattr(test, 'foo')
         assert isinstance(test.foo(), pd.DataFrame)
 
+    def test_build_yaml_syntax_error(self):
+        path = os.path.dirname(__file__)
+        buildfilepath = os.path.join(path, 'build_bad_syntax.yml')
+        with assertRaisesRegex(self, command.CommandException, 'Syntax error'):
+            command.build('user/test', buildfilepath)
+
```
kevinemoore: Can we add a unit test for a syntax error in the checks.yml file too?

eode: Definitely.

eode: Ok, that's been pushed.  Notes:

* I was previously working in command.py, and testing in command_test.py.
* Since the exception catch was moved to build.py, I made tests in build_test.py
* I left the command_test tests in, as they do test (slightly) different code.

compiler/quilt/tools/build.py
```diff
@@ -293,21 +293,28 @@ def find(key, value):
                     for item in v:
                         for result in find(key, item):
                             yield result
+
+    try:
+        build_data = load_yaml(yaml_path)
```
kevinemoore: There are two calls to load_yaml in this try/except block that can throw exceptions for syntax errors. I think that means that the user won't be able to tell which file had the bad syntax.

If it's easier/cleaner, feel free to move the try/except into load_yaml (even though it's in core.py).

eode:
I was using the default blurb for the syntax error to avoid spending too much time on this, which doesn't seem to recognize the filename.  Fixed.  The message has also been re-ordered somewhat, since I couldn't use the default blurb, and I think having the problem near the problem marker is more clear.  Let me know if you want changes to how the error is displayed.

eode:
@kevinemoore  Ready for review.

eode:
Done.

PR #206, opened 2017-11-28
[wip] 'quilt --version' command

eode:
This currently produces the following output:

```
# editable, and in a git repo
quilt 2.7.1 (editable) : 6240cf51e13cd9cd96688c26be4e719937062476

# editable, no git repo
quilt 2.7.1 (editable)

# installed version mismatch -- this is wrapped by argparse.
# I don't really like this output, I just haven't settled on anything yet.
Version mismatch! quilt 2.7.3, installed as 2.7.1 (editable):
6240cf51e13cd9cd96688c26be4e719937062476

# non-editable, no version mismatch
quilt 2.7.1
```

Thoughts
* ..get rid of 'version mismatch!' string.
* First two tokens could always be the name and the version.
* Last two tokens are a : and hash if it's in a git repo
* Like pip freeze, this doesn't display the hash from install, but from the currently-active version.
* pip dependency.  :-/  Not needed if git hash isn't needed, or if parsing git CLI is ok.

..any thoughts before I clean up the output and consider it ready to go?


quiltdata:
Assigned to Aneesh for review... Aneesh, feel free to handoff if you're
slammed.

eode:
Unless specified further, I'm just going to leave this as returning the following:

If not installed, show version only:
`quilt 2.7.1`

If installed, include the egg name:
`quilt 2.7.1 (quilt-2.7.1-py3.5)`

asah:
Sounds good to me.

eode:
Hrm.. ..I'd like to make this history a bit cleaner before finishing this PR.

PR #208, opened 2017-11-28
Move the temporary object storage outside of objects

kevinemoore:
Cleaning up the package store by moving temporary objects out of the
directory that holds package objects.


quilt/tools/store.py
```diff
@@ -35,10 +35,10 @@ class and its subclasses abstract file formats, file naming and
     """
     BUILD_DIR = 'build'
     OBJ_DIR = 'objs'
-    TMP_OBJ_DIR = os.path.join('objs', 'tmp')
+    TMP_OBJ_DIR = 'tmp'
```
asah: what about existing Quilt installs?
also, does this create a directory parallel to quilt_packages/ ?  if so, this can be bad juju for filesystem config...

AFAICT, best practice is to make tmpdir locations be user configurable, ditto for storage limits and garbage collection policy (separate projects of course...)

kevinemoore: This is an extremely minor change that moves the temporary objects from: quilt_packages/objs/tmp/
to:
quilt_packages/tmp/

It will affect existing installs and that's why it should be merged in the same release as the local storage changes.

dimaryaz: I'm not even worried about existing installs: `tmp` is for unfinished downloads, so safe to delete.

PR #211, opened 2017-11-29
Created --version command

eode:
When quilt is installed, it produces `quilt <version> (<egg-name>)`, as in:
`quilt 2.7.1 (quilt-2.7.1-py3.5)`

When quilt is not installed, it produces `quilt <version>` only, as in:
`quilt 2.7.1`

This should be ready for merge into master.

### DIFF
--- a/compiler/quilt/tools/main.py
+++ b/compiler/quilt/tools/main.py
@@ -6,12 +6,26 @@ from __future__ import print_function
 
 import argparse
 import sys
+import pkg_resources
+
 import requests
 
 from . import command
 from .const import DEFAULT_QUILT_YML
 
 HANDLE = "owner/packge_name"
+VERSION = command.VERSION
+
+
+
+def get_full_version():
+    try:
+        quilt = pkg_resources.get_distribution('quilt')
+    except pkg_resources.DistributionNotFound:
+        pass
+    else:
+        return "quilt {} ({})".format(VERSION, quilt.egg_name())
+    return "quilt " + VERSION
 
 def main():
     """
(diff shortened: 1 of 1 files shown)


asah:
Once the CLI unit tests are merged pls as one for this command ie make sure it returns something reasonable.  Don't include in this PR.

asah:
@akarve @dimaryaz ready for review

eode:
Tracking which git hash was mentioned.  The only way to do this for a non-editable install is to write a file at install time.  Since a package isn't importable during its own setup (nor afterward, from later in setup.py), it must be done before `setup` is called.  There are two options I've thought of:

1. Add a data file using `MANIFEST.in`, create it from setup.py at installl time, then at load time, load via pkg_resources, then parse it
2. Add a python file to .gitignore, create it from setup.py at install time, then import it at load time

I'm inclined toward 2 as it's pretty straightforward.

In either case, the info could then be loaded at load time, and added to `quilt.__init__`, so it would appear as `quilt.__version__` and `quilt.__githash__` (or whatever other naming convention is preferred).

In the case of a `pip install --editable`, the code in `__init__.py` could prefer the hash loaded directly from `git -C <pkgdir> rev-parse HEAD` or `pip.git.Git.get_revision(<packagedir>)`.

This would make quilt.__version__ and quilt.__githash__ the cannonical version for any particular install, and the related code in command could be replaced with "from .. import __version__ as VERSION".

eode:
..to be clear, if that feature is desired, let me know and I'll add it.

PR #213, opened 2017-11-30
For install command, require `package` arg if default yml not present

eode:
I've used an if/else line here, instead of the more complicated option of creating a custom argparse action.

I'm not certain, but I don't *think* there's a way to make an argument's `nargs` dependent on a function call using argparse itself without making a custom action.  Anyways, this is clean and clear.

### DIFF
--- a/compiler/quilt/tools/main.py
+++ b/compiler/quilt/tools/main.py
@@ -177,8 +177,6 @@ def main():
         return 0
     except command.CommandException as ex:
         print(ex, file=sys.stderr)
-        print()
-        print(subparsers.choices[cmd].format_help())
         return 1
     except requests.exceptions.ConnectionError as ex:
         print("Failed to connect: %s" % ex, file=sys.stderr)
--- a/compiler/quilt/tools/main.py
+++ b/compiler/quilt/tools/main.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 
 import argparse
 import sys
+import os
 import requests
 
 from . import command
@@ -96,8 +97,12 @@ def main():
     tag_remove_p.set_defaults(func=command.tag_remove)
 
     install_p = subparsers.add_parser("install")
-    install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
-                           nargs="?", default="@"+DEFAULT_QUILT_YML)
+    if os.path.exists(DEFAULT_QUILT_YML):
+        install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
+                               nargs="?", default="@"+DEFAULT_QUILT_YML)
+    else:
+        install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
+                               nargs="+")
     install_p.set_defaults(func=command.install)
     install_p.add_argument("-f", "--force", action="store_true", help="Overwrite without prompting")
     install_group = install_p.add_mutually_exclusive_group()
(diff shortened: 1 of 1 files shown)


compiler/quilt/tools/main.py
```diff
@@ -96,8 +97,12 @@ def check_hash(group, hashstr):
...
 
     install_p = subparsers.add_parser("install")
-    install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
-                           nargs="?", default="@"+DEFAULT_QUILT_YML)
+    if os.path.exists(DEFAULT_QUILT_YML):
+        install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
+                               nargs="?", default="@"+DEFAULT_QUILT_YML)
+    else:
+        install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
+                               nargs="+")
```
asah: rofl forehead-smacker :-)

install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename", nargs="?", default=("@"+DEFAULT_QUILT_YML) if os.path.exists(DEFAULT_QUILT_YML) else None)

eode: No problem with a ternary operation, but using that would require two ternary operations:
```
install_p.add_argument("package", type=str, help="owner/package_name[/path/...] or @filename",
       nargs="?" if os.path.exists(DEFAULT_QUILT_YML) else "+",
       default=("@"+DEFAULT_QUILT_YML) if os.path.exists(DEFAULT_QUILT_YML) else None)
```
...would that method still be preferred?

asah:
or *({'nargs':'?', 'default':"@"+DEFAULT_QUILT_YML} if
os.path.exists(DEFAULT_QUILT_YML) else {'nargs':'+'})

eode:
Yeah.. ..pretty convoluted either way.  I have no preference on how it's expressed, as it works in any case.  I just chose what I thought would be easily readable.  Should I use the last example, then?

eode:
In my initial description, I was more curious if there was a way to express it in the argparse api itself, as opposed to with a conditional as I did initially or with the ternary conditional.

compiler/quilt/tools/main.py
kevinemoore: The if/else seems easier to read to me, but I'm ok with either. Either way, please add a comment explaining the behavior.

eode: added comment.

eode:
Egad, I'd love to remove that.  I just mentioned re-ordering it and sending all error output to stderr in #204 --  I thought that printing help as well as errors was Someone's Preferred Way.


..pushed.

eode:
As a side note, the nargs bit would (in theory) have been caught by the CLI tests once they're completed.  That first PR only has one CLI test, aside from more generic and sweeping tests that just watch for changes.  It'll be nice to have the CLI tests done, when they get there.

kevinemoore:
Yes, I'm looking forward to having the CLI tests too! Printing the help every time was my preferred way, but I now see why that's not a good idea.

eode:
@kevinemoore Yeah, I think the spirit of that is right, though -- I really like what has happened with git, for example, over the years.  Now, anything you try to do that you run into a roadblock for, it spews out very useful contextual information.  It's so good, I almost don't miss mercurial.  

..but I definitely don't miss bitbucket.

PR #214, opened 2017-11-30
pylint githook for registry and compiler

mhassan102:
Need to place this file in .git/hooks as pre-commit (locally) to work as commit hook

@asah

### DIFF
--- a/pylint_git_commit_hook.sh
+++ b/pylint_git_commit_hook.sh
@@ -22,10 +22,8 @@ fi
 # Checking pylint version, (easy to parse pip freeze than pylint --version)
 #
 pylint_version=$(pip freeze | grep pylint | awk -F'==' '{ print $2 }')
-if [[ $pylint_version == $PYLINT_VERSION ]]
+if [[ $pylint_version != $PYLINT_VERSION ]]
 then
-	echo "pylint exist with specified version"
-else
 	echo "$0: Pylint version mismatch!"
 	echo "$0: Do pip install pylint==$PYLINT_VERSION"
 	exit 1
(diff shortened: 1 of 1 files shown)


pylint_git_commit_hook.sh
```diff
@@ -0,0 +1,53 @@
...
+	exit 1
+fi
+
+#
+# Checking pylint version, (easy to parse pip freeze than pylint --version)
+#
+pylint_version=$(pip freeze | grep pylint | awk -F'==' '{ print $2 }')
+if [[ $pylint_version == $PYLINT_VERSION ]]
+then
+	echo "pylint exist with specified version"
```
asah: delete this -- not an error.

pylint_git_commit_hook.sh
```diff
@@ -0,0 +1,53 @@
...
+# Checking for specified directories inside repo 
+#
+for i in "registry" "compiler"
+do
+	echo "$0: Checking $i directory.."
+	# purpose of changing directory is to use already existing pylintrc as default pylint config
+	pushd $i && find . -iname "*.py" | xargs pylint $PYLINT_ARGS >> /dev/null
+	if [[ $? -ne 0 ]]
+	then
+  		echo "$0: Pylint failed in $i!"
```
asah: @kevinemoore we need to decide what constitutes a failure... best practice is to cleanup the code and make all warnings/errors fail, then use pylintrc and pylint:disable=NNN comments to hide spurious (safe) warnings.

kevinemoore: That sounds like a good plan to me. Go ahead and make all warnings and errors fail and we'll add pylint:disable as needed.

mhassan102:
@dimaryaz Yes it is working. if you want to use existing solution.

I had quick test it, by adding our desired output i.e. "--enable=W,E --disable=C" directly to parser.add_argument.

PR #215, opened 2017-12-01
command_test methods requiring input() mocking

eode:
I was tasked with mocking tests in command_test that needed to mock input().  I wasn't, however, able to test install() in a reasonable amount of time.  I think it would take me a bit do do that, as it's quite long, and there are a few different code paths that would need to be mocked.

I mean, it could be done, but it seems that it might be better if install() were broken into smaller, more readily-testable chunks.  ..and in any case, it might take a bit more time than you'd really desire for me to spend on it.  ..although, that's up to you.

..meanwhile, here are the others.

__edit:__ Inaccurate PR title -- changed.

### DIFF
--- a/compiler/quilt/test/test_command.py
+++ b/compiler/quilt/test/test_command.py
@@ -45,17 +45,17 @@ class CommandTest(QuiltTestCase):
             ]
         # test general URL setting -- result should match input
         for test_url in test_urls:
-                mock_load_config.return_value = {}
-                mock_input.return_value = test_url
+            mock_load_config.return_value = {}
+            mock_input.return_value = test_url
 
-                command.config()
+            command.config()
 
-                assert mock_input.called
-                mock_input.reset_mock()
+            assert mock_input.called
+            mock_input.reset_mock()
 
-                args, kwargs = mock_save_config.call_args
-                mock_load_config.return_value = args[0] if args else kwargs['cfg']
-                assert test_url == command.get_registry_url()
+            args, kwargs = mock_save_config.call_args
+            mock_load_config.return_value = args[0] if args else kwargs['cfg']
+            assert test_url == command.get_registry_url()
 
     @patch('quilt.tools.command._save_config')
     @patch('quilt.tools.command._load_config')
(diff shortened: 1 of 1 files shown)


eode:
Oh -- I had a question I tagged with #review -- I noticed while setting up the command.config tests that URLs that include a path are rejected.  I.e.,  `http://this.url/has/a/path` unlike `http://this.url`.

Is that intentional, or a bug?

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+    @patch('quilt.tools.command._load_config')
+    @patch('quilt.tools.command.input')
+    def test_config_good_urls(self, mock_input, mock_load_config, mock_save_config):
+        config_prompt = ("Please enter the URL for your custom Quilt registry (ask your administrator),\n" +
+                         "or leave this line blank to use the default registry: ")
+        test_urls = [
+            'https://foo.com',
+            'http://foo.com',
+            'https://foo.bar.net',
+#REVIEW: is this style of URL intentionally invalid?
```
dimaryaz: Yes. It's the host for the API calls, so all paths are hard-coded since they're part of the API. (But, it can be http or https, so I had to use an actual URL, not just a host:port string.)

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+#            'https://foo.bar.net/baz',
+            ]
+        # test general URL setting -- result should match input
+        for test_url in test_urls:
+                mock_load_config.return_value = {}
```
dimaryaz: Indentation. (Run `pylint`?)

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+                mock_input.return_value = test_url
+
+                with assertRaisesRegex(self, command.CommandException, 'Invalid URL'):
+                    command.config()
+
+                mock_input.assert_called_with(config_prompt)
+                assert mock_save_config.call_args is None
```
dimaryaz: `assert_not_called()`

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+                assert command.get_registry_url() != test_url
```
dimaryaz: `!=` is not a very strong assertion; check that it's equal to the original (default?) value instead.

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+
+    @patch('requests.Session.put')
```
dimaryaz: All network requests are already mocked by `QuiltTestCase` and fail by default, so this isn't needed. To actually fake responses, use `self.requests_mock.add(...)`; see other tests for examples.

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
 from .utils import QuiltTestCase, patch
 
 class CommandTest(QuiltTestCase):
+    # no counterpart to command.get_registry_url, so we mock _save_config and _load_config
```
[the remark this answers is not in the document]
eode: There's a get_registry_url in command.py, but no set_registry_url.  Thus, I mocked the _save_config and _load_config commands, instead.  I suppose the comment was unnecessary -- I can remove it.

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+    @patch('quilt.tools.command._save_config')
+    @patch('quilt.tools.command._load_config')
+    @patch('quilt.tools.command.input')
+    def test_config_urls_default(self, mock_input, mock_load_config, mock_save_config):
+        config_prompt = ("Please enter the URL for your custom Quilt registry (ask your administrator),\n" +
```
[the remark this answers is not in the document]
eode: OK by me.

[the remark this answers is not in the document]
eode: Well, I'll have to use 'assert mock_input.called',  as 'mock_input.assert_called()' doesn't seem to work when testing.  It might be that it was added in a later version.

compiler/quilt/test/test_command.py
```diff
@@ -18,6 +18,106 @@
...
+                         "or leave this line blank to use the default registry: ")
+        # test setting default URL with blank string -- result should be default
+        mock_load_config.return_value = {}
+        mock_input.return_value = ''
+        command.config()
+
+        mock_input.assert_called_with(config_prompt)
+
+        args, kwargs = mock_save_config.call_args
+        mock_load_config.return_value = args[0] if args else kwargs['cmd']
```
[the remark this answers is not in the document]
eode: Yep, should be cfg.  Fixed.

compiler/quilt/test/test_command.py
eode: Works for me.  I was thinking of the instance where someone has a reverse proxy to their server or some such, and just wanted to be sure.

compiler/quilt/test/test_command.py
eode: Huh.  I must have something configured wrong in my editor -- it didn't complain about that.

compiler/quilt/test/test_command.py
eode: Sure.

compiler/quilt/test/test_command.py
eode: I've pulled the initial value from get_registry_url() and compared it directly now.

eode:
@dimaryaz The issues you mentioned have been addressed now -- anything else?

PR #219, opened 2017-12-05
Add support for building empty packages, empty groups, refactor pandas kwargs

akarve:
* Added reserved word `kwargs` which is where all panadas/handler args should go (eliminates ambiguity between pandas kwargs and groups)

### DIFF
--- a/compiler/quilt/test/test_build.py
+++ b/compiler/quilt/test/test_build.py
@@ -101,6 +101,17 @@ class BuildTest(QuiltTestCase):
             'Expected column of ints with nulls to deserialize as numeric'
         # TODO add more integrity checks, incl. negative test cases
 
+    def test_build_empty(self):
+        """
+        test building from build_empty.yml
+        """
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, './build_empty.yml')
+        build.build_package('empty', 'pkg', path)
+
+        from quilt.data.empty import pkg
+        assert not pkg._keys(), 'Expected package to be empty'
+
     def test_build_group_args(self):
         """
         test building from build_group_args.yml
--- a/compiler/quilt/test/test_build.py
+++ b/compiler/quilt/test/test_build.py
@@ -122,9 +122,9 @@ class BuildTest(QuiltTestCase):
 
         from quilt.data.groups import pkg
         
-        assert type(pkg.group_a.csv()) is DataFrame, \
+        assert isinstance(pkg.group_a.csv(), DataFrame), \
             'Expected parent `transform: csv` to affect group_a.csv()'
-        assert type(pkg.group_a.tsv()) is DataFrame, \
+        assert isinstance(pkg.group_a.tsv(), DataFrame), \
             'Expected local `transform: tsv` to affect group_a.tsv()'
         # TODO these tests should really test the node type and verify it as a file node
         # but currently both raw files and DFs are DataNode instances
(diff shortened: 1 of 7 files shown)


compiler/quilt/tools/package.py
```diff
@@ -231,6 +231,14 @@ def save_file(self, srcfile, name, path):
         if not os.path.exists(objpath):
             copyfile(srcfile, objpath)
 
+    def save_group(self, name):
+        """
+        Save a (raw) file to the store.
```
dimaryaz: Comment?

compiler/quilt/tools/package.py
```diff
@@ -365,7 +373,9 @@ def _add_to_contents(self, fullname, hashes, ext, path, target, fmt):
 
         try:
             target_type = TargetType(target)
-            if target_type is TargetType.PANDAS:
+            if target_type is TargetType.GROUP:
+                node = GroupNode(dict())
```
dimaryaz: This feels hacky to me... The code above - `ptr = ptr.children.setdefault(...)` - already handles the creation of groups.

compiler/quilt/tools/build.py
```diff
@@ -85,36 +89,47 @@ def _build_node(build_dir, package, name, node, fmt, target='pandas', checks_con
...
-      that affect entire subtrees (e.g. transform: csv for 500 .txt files)
-      Position of definition in file also matters and allows for composition
+      any transform inherited from an ancestor
+      plus any inherited handler kwargs
+      Users can thus define kwargs that affect entire subtrees
+      (e.g. transform: csv for 500 .txt files)
       and overriding of ancestor or peer values.
+      Child transform or kwargs override ancestor k:v pairs.
     """
     if _is_internal_node(node):
```
dimaryaz: I think logically, empty groups should fall into this category. That is, empty shouldn't be a special case.

akarve: OK, suppose we do that. And now groups are handled by the `package.save_file` codepath? I thought about that and it also looked smelly because I would have to make a bunch of arguments that should be required for actual files (e.g. `srcfile`, `[filehash`) in `save_file` optional and that would undermine some necessary argument checks? Or am I missing something and there is an elegant way to reuse `save_file` for this?

compiler/quilt/test/test_build.py
```diff
@@ -100,6 +101,53 @@ def _test_dataframes(self, dataframes):
...
+        """
+        test building from build_group_args.yml
+        """
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, './build_group_args.yml')
+        build.build_package('groups', 'pkg', path)
+
+        from quilt.data.groups import pkg
+        
+        assert type(pkg.group_a.csv()) is DataFrame, \
```
kevinemoore: Use isinstance?

compiler/quilt/test/test_build.py
```diff
@@ -100,6 +101,53 @@ def _test_dataframes(self, dataframes):
...
+            'Expected parent `transform: csv` to affect group_a.csv()'
+        assert type(pkg.group_a.tsv()) is DataFrame, \
```
kevinemoore: Here too?

compiler/quilt/tools/build.py
```diff
@@ -51,9 +51,13 @@ def _path_hash(path, transform, kwargs):
     return digest_string(srcinfo)
 
 def _is_internal_node(node):
```
kevinemoore: Including both functions seems like overkill for this PR since only _is_internal_node is ever used. Keep them if you think they'll both be used in the future, but if not, the code will be easier to read if you collapse the logic into a single function.

compiler/quilt/tools/build.py
akarve: Per discussion with Dima: propose keeping logic as written so that internal nodes are simply descended whereas leaf nodes modify the package structure

compiler/quilt/tools/package.py
akarve: The alternative is to special case `ipath.pop()` above so either way we will require a dedicated codepath for `GROUP` vs `FILE`. It's also nice to keep the last assignment in the routine as is (and not special-case `ipath.pop()`). We could signal `GROUP` with `target=None` but, given that we have (and need) `save_group`, it's probably better to be explicit about Group creation.

compiler/quilt/tools/build.py
akarve: fixed

PR #220, opened 2017-12-05
Add the catalog code

dimaryaz:


dimaryaz:
All of the data comes from environment variables passed into the docker container. They get substituted in `config.js.tmpl`, and the result overwrites the existing `config.js`.

dimaryaz:
Looks like I missed a few files; will send out a new PR instead.

PR #222, opened 2017-12-06
Display the overwrite prompt before looking up tags and versions

dimaryaz:
Prompt early, before doing unnecessary network I/O.

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -826,6 +826,12 @@ def install(package, hash=None, version=None, tag=None, force=False):
     store = PackageStore()
     existing_pkg = store.get_package(owner, pkg)
 
+    if existing_pkg is not None and not force:
+        print("{owner}/{pkg} already installed.".format(owner=owner, pkg=pkg))
+        overwrite = input("Overwrite? (y/n) ")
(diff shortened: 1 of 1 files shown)


compiler/quilt/tools/command.py
```diff
@@ -826,6 +826,12 @@ def install(package, hash=None, version=None, tag=None, force=False):
     store = PackageStore()
     existing_pkg = store.get_package(owner, pkg)
 
+    if existing_pkg is not None and not force:
+        print("{owner}/{pkg} already installed.".format(owner=owner, pkg=pkg))
+        overwrite = input("Overwrite? (y/n) ")
+        if overwrite.lower() != 'y':
+            return
+
```
asah: I seem to recall a UI issue with this... @akarve I think you didn't like this for some reason...

akarve: The issue I want to avoid is this `quilt install usr/pkg --tag NON_EXISTENT_TAG` will prompt "y/N over-write?". I think it makes more sense for the install to fail first, instead of asking for permission for an over-write that will never happen.

dimaryaz: Ah... Makes sense

PR #224, opened 2017-12-06
Implement parallel downloads

dimaryaz:
We lose precise progress bar, though: we don't know the object sizes, so can only count the number of objects.

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -60,6 +60,7 @@ GIT_URL_RE = re.compile(r'(?P<url>http[s]?://[\w./~_-]+\.git)(?:@(?P<branch>[\w_
 CHUNK_SIZE = 4096
 
 PARALLEL_UPLOADS = 20
+PARALLEL_DOWNLOADS = 20
 
 S3_CONNECT_TIMEOUT = 30
 S3_READ_TIMEOUT = 30
@@ -233,6 +234,17 @@ def _clear_session():
     if _session is not None:
         _session.close()
         _session = None
+
+def _create_s3_session():
+    """
+    Creates a session with automatic retries on 5xx errors.
+    """
+    sess = requests.Session()
+    retries = Retry(total=3,
+                    backoff_factor=.5,
+                    status_forcelist=[500, 502, 503, 504])
+    sess.mount('https://', HTTPAdapter(max_retries=retries))
+    return sess
 
 def _open_url(url):
     try:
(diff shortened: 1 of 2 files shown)


asah:
@dimaryaz interesting... appveyor error looks legit - take a look?

dimaryaz:
Oops, yes, broken test. Will fix.

compiler/quilt/tools/command.py
```diff
@@ -879,84 +887,125 @@ def install(package, hash=None, version=None, tag=None, force=False):
 
     pkgobj = store.install_package(owner, pkg, response_contents)
 
-    with requests.Session() as s3_session:
-        total = len(response_urls)
-        for idx, (download_hash, url) in enumerate(sorted(iteritems(response_urls))):
-            print("Downloading %s (%d/%d)..." % (download_hash, idx + 1, total))
+    obj_queue = sorted(iteritems(response_urls), reverse=True)
```
kevinemoore: Curious, why reverse?

compiler/quilt/tools/command.py
```diff
@@ -879,84 +887,125 @@ def install(package, hash=None, version=None, tag=None, force=False):
...
-            if file_hash != download_hash:
-                os.remove(temp_path)
-                raise CommandException("Fragment hashes do not match: expected %s, got %s." %
-                                       (download_hash, file_hash))
+                    if not success:
+                        # We've already printed an error, so not much to do - just move on to the next object.
+                        continue
+
+                    # Ungzip the downloaded fragment.
+                    temp_path = store.temporary_object_path(obj_hash)
```
kevinemoore: Maybe some of this code can move into PackageStore?

compiler/quilt/tools/command.py
dimaryaz: `obj_queue` is actually a stack (just cause I can use a normal list for that) - so I reverse it so it shows up sorted when popping items off the end.

compiler/quilt/tools/command.py
dimaryaz: Which part of it? The ungzipping code?

dimaryaz:
Yeah, it's definitely getting worse - though it's hard to refactor cause there are so many variables involved. (Like, I'd have a helper function that takes 10 params, etc.)

It's particularly bad with nested loops - and I had to add an extra variable just to break out of the inner loop, then continue the outer loop. But anyways, I'll try to do that in a follow-up PR.

PR #228, opened 2017-12-08
better build badges

asah:


akarve:
nice touch. please bump gitter to top though (e.g. that's where airflow, etc. projects have it).

PR #230, opened 2017-12-08
Fix ambiguous hashes

eode:
As long as _match_hash is used, ambiguous hashes will raise a CommandException:

```
$ quilt install example/foo 391e01
Ambiguous hash for package example/foo: 'd91e01' matches the folowing hashes:

d91e01b0cffde1e3c987262620e0050f4377ac55cd000f77b37bae82462e2acd
d91e0143a9e0367b92a8f67b1ce8417adb5b64a5d7df1d0664ad6bd1b94c4b99
```

The server sometimes returns duplicate hashes, which aren't considered ambiguous, even though they both start with the same short-hash string.

### DIFF
--- a/compiler/quilt/test/test_command.py
+++ b/compiler/quilt/test/test_command.py
@@ -131,9 +131,9 @@ class CommandTest(QuiltTestCase):
             registry_url + "/api/log/user/test/",
             json=fake_data
         )
-        # Data will be reversed in _match_hash, so we need to reverse our data here, too
-        fake_data_ambiguous = [entry['hash'] for entry in reversed(fake_data['logs'])
-                               if entry['hash'].startswith(ambiguous_token)]
+        # Ambiguous hashes in _match_hash's exception will be sorted -- sorted here to match.
+        fake_data_ambiguous = sorted(entry['hash'] for entry in fake_data['logs']
+                               if entry['hash'].startswith(ambiguous_token))
         fake_data_regexp = '[\s\S]'.join(fake_data_ambiguous)
         with assertRaisesRegex(self, command.CommandException, fake_data_regexp):
             command._match_hash(session, owner='user', pkg='test', hash='795a7b')
--- a/compiler/quilt/test/test_command.py
+++ b/compiler/quilt/test/test_command.py
@@ -110,6 +110,34 @@ class CommandTest(QuiltTestCase):
         mock_input.return_value = 'n'
         command.version_add('user/test', '2.9.12', 'fabc123')  # should produce no mock network activity
 
+    def test_ambiguous_hash(self):
+        registry_url = command.get_registry_url()
+        session = command._get_session()
+        ambiguous_token = "795a7b"
+        # There should be at least two results that start with the ambiguous_token, plus some non-ambiguous
+        # results in fake_data to test against.
+        fake_data = {'logs': [
+            {'author': 'user', 'created': 1490816524.0,
+             'hash': '885696c6e40613b3c601e95037caf4e43bda58c39f67ab5d5e56beefb3662ff4'},
+            {'author': 'user', 'created': 1490816507.0,
+             'hash': '795a7bc9e40613b3c601e95037caf4e43bda58c39f67ab5d5e56beefb3662ff4'},
+            {'author': 'user', 'created': 1490816473.0,
+             'hash': '795a7bc6e40613b3c601e95037caf4e43bda58c39f67ab5d5e56beefb3662ff4'},
+            {'author': 'user', 'created': 1490816524.0,
+             'hash': '2501a6c6e40a7b355901fc5037caf4e43bda58c39f67ab5d5e56beefb3662ff4'},
+        ]}
+        self.requests_mock.add(
+            responses.GET,
+            registry_url + "/api/log/user/test/",
+            json=fake_data
+        )
+        # Data will be reversed in _match_hash, so we need to reverse our data here, too
+        fake_data_ambiguous = [entry['hash'] for entry in reversed(fake_data['logs'])
+                               if entry['hash'].startswith(ambiguous_token)]
+        fake_data_regexp = '[\s\S]'.join(fake_data_ambiguous)
+        with assertRaisesRegex(self, command.CommandException, fake_data_regexp):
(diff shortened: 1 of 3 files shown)


eode:
Possibly of interest:

We can still enforce a minimum short-hash length, but we technically don't need to anymore.  Not doing so makes it easier to communicate with others verbally when using small datasets:  "Check out f0a" is a lot easier to remember than "Check out f0a3ea".  Otoh, the potential of under-specifying a commit when working on a larger project increases.

If a minimum short-hash length isn't required, I can remove the existing check from `main`.  

If it is required, would it be okay if I move that check into `command._match_hash`?  ..it seems the right place for it.

asah:
IMHO min hash length is worth having because *future* commits can cause
conflicts and you want to reduce the frequency of this, e.g. if
short-hashes are included in code, docs, examples, etc.

eode:
*nod* that's understandable.  Is it ok if I move the check into `command._match_hash()` from `main.py`?

asah:
y pls do that move - short hash resolution is a lower level function...

compiler/quilt/test/test_command.py
```diff
@@ -110,6 +110,34 @@ def test_version_add_declined(self, mock_input):
...
+        ]}
+        self.requests_mock.add(
+            responses.GET,
+            registry_url + "/api/log/user/test/",
+            json=fake_data
+        )
+        # Ambiguous hashes in _match_hash's exception will be sorted -- sorted here to match.
+        fake_data_ambiguous = sorted(entry['hash'] for entry in fake_data['logs']
+                               if entry['hash'].startswith(ambiguous_token))
+        fake_data_regexp = '[\s\S]'.join(fake_data_ambiguous)
```
dimaryaz: Why `[\s\S]`? Isn't it just `.`?

compiler/quilt/tools/command.py
```diff
@@ -259,14 +260,20 @@ def _match_hash(session, owner, pkg, hash, raise_exception=True):
...
 
+    matches = set(entry['hash'] for entry in response.json()['logs']
+                  if entry['hash'].startswith(hash))
+
+    if len(matches) == 1:
+        return matches.pop()
+    if len(matches) > 1:
+        # Sorting for consistency in testing, as well as visual comparison of hashes
+        ambiguous = '\n'.join(sorted(matches))
+        raise CommandException(
```
dimaryaz: It seems like this should also check for `raise_exception`... But it looks like nobody uses that param, so perhaps we should delete it. @asah - you added it, right?

asah: fine to remove it.

compiler/quilt/test/test_command.py
eode: Newlines in the exception.

compiler/quilt/tools/command.py
eode: I thought about that, but then I wasn't sure if it was sensical to return None when multiple hashes were indeed identified, which is pretty much a show-stopper.  I figured anyone using it would notice an exception, but not notice the problem if it was squelched into just returning None.

..removed the raise_exception param.  None is no longer returned, and an exception is raised in all cases where the number of matches isn't 1.

asah:
Pls add a comment explaining this.

compiler/quilt/test/test_command.py
eode: Commented explaining, per @asah.  I also made the match more flexible, in case the CommandException's text changes.

compiler/quilt/test/test_command.py
```diff
@@ -110,6 +110,36 @@ def test_version_add_declined(self, mock_input):
...
+            responses.GET,
+            registry_url + "/api/log/user/test/",
+            json=fake_data
+        )
+        # Ambiguous hashes in _match_hash's exception will be sorted -- sorted here to match.
+        fake_data_ambiguous = sorted(entry['hash'] for entry in fake_data['logs']
+                               if entry['hash'].startswith(ambiguous_token))
+        # this will match each ambiguous hash, in order, separated by anything.
+        # ..it allows for formatting changes in the error, but requires the same order.
+        fake_data_regexp = '[\s\S]+'.join(fake_data_ambiguous)
```
dimaryaz: I'd use something like `(.|\n)` if that works. Otherwise, I guess that's fine. Also, in any case, add an `r` in front of the regex to make sure backslashes aren't interpreted the usual way: `r'(.|\n)'`

compiler/quilt/tools/main.py
```diff
@@ -152,8 +147,7 @@ def check_hash(group, hashstr):
     install_p.set_defaults(func=command.install)
     install_p.add_argument("-f", "--force", action="store_true", help="Overwrite without prompting")
     install_group = install_p.add_mutually_exclusive_group()
-    install_group.add_argument("-x", "--hash", help="Package hash",
-                               type=lambda val: check_hash(install_p, val))
```
dimaryaz: Does this change anything? I'm guessing the existing code fails sooner, during argument parsing.

eode: Yup, removing that and moving its effects into _match_hash have a few nice effects.  The existing code fails sooner, but it's not necessarily preferable. 

The notable effects of the change are:
* more succinct code with no loss of clarity (argparse 'type' isn't hijacked for cleanup)
* removal of code duplication (otherwise, hash needs to be checked in `main` and `command`)
* provides specific choke point for handling hash scrubbing (short->long, lower(), hash lookup, ambiguous hash errors, invalid/missing hash errors)
* clearer error messages (syntax errors belong in argparse, imo, but for a correctly-placed but invalid token, showing argparse help again just seems distracting)

compiler/quilt/test/test_command.py
eode: Sure, it's just a common idiom.  `(.|\n)` looks nicer, though.  '\s\S' and r'\s\S' are identical, but the raw string is better in case anyone changes it at some point and adds a control character in there.

compiler/quilt/tools/main.py
dimaryaz: Ah, ok then

PR #234, opened 2017-12-10
When on the CLI, catch ctrl-c early in execution and exit cleanly

eode:
Normally I'd just do a try: except: block on or in main(), but at load-time, we load a bunch of external modules that take a lot of time, during which ctrl-c will cause an exception that misses that block. ..so, this was added to `quilt/__init__.py`.

Also, sometimes during development we want to be able to see the traceback -- for example, if we're wondering what's taking so long, or what function is causing network activity.  Toward that end, there's now a variable `quilt._DEV_MODE` which enables/disables traceback.

If your first argument is `--dev`, or if the environment has `QUILT_DEV_MODE=True`, tracebacks will still be shown.  Help for `--dev` is suppressed and doesn't show up in `quilt help`, `quilt --help`, etc.

### DIFF
--- a/compiler/quilt/test/test_cli.py
+++ b/compiler/quilt/test/test_cli.py
@@ -627,6 +627,12 @@ class TestCLI(BasicQuiltTestCase):
         TESTED_PARAMS.append(['--dev'])
 
         cmd = ['--dev', 'install', 'user/test']
+        if os.name == 'posix':
+            SIGINT = signal.SIGINT
+        elif os.name == 'nt':
+            SIGINT = signal.CTRL_C_EVENT
+        else:
+            raise ValueError("Unknown OS type: " + os.name)
 
         result = self.execute(cmd)
 
@@ -644,7 +650,7 @@ class TestCLI(BasicQuiltTestCase):
         proc = Popen(cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE, env=no_mock)
         # if interrupt is sent too fast, the files won't even be parsed.
         sleep(3)
-        proc.send_signal(signal.SIGINT)
+        proc.send_signal(SIGINT)
         stdout, stderr = (b.decode() for b in proc.communicate())
         assert 'Traceback' not in stderr
         # Return code should indicate keyboard interrupt
@@ -655,7 +661,7 @@ class TestCLI(BasicQuiltTestCase):
         proc = Popen(cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE, env=no_mock)
         # if interrupt is sent too fast, the files won't even be parsed.
         sleep(3)
-        proc.send_signal(signal.SIGINT)
+        proc.send_signal(SIGINT)
         stdout, stderr = (b.decode() for b in proc.communicate())
         print("\n\n{}\n\n{}\n\n".format(stdout, stderr))
         assert 'Traceback (most recent call last)' in stderr
--- a/compiler/quilt/test/test_cli.py
+++ b/compiler/quilt/test/test_cli.py
@@ -623,50 +623,50 @@ class TestCLI(BasicQuiltTestCase):
         assert result['kwargs']['public'] is True
         assert result['kwargs']['package'] == 'fakeuser/fakepackage'
 
-    def test_cli_option_dev(self):
-        TESTED_PARAMS.append(['--dev'])
-
-        cmd = ['--dev', 'install', 'user/test']
-        if os.name == 'posix':
-            SIGINT = signal.SIGINT
-        elif os.name == 'nt':
-            SIGINT = signal.CTRL_C_EVENT
-        else:
-            raise ValueError("Unknown OS type: " + os.name)
-
-        result = self.execute(cmd)
-
-        # --dev arg was accepted by argparse?
-        assert result['return code'] == 0
-
-        # We need to run a command that blocks.  To do so, I'm disabling the
-        # test mocking of the command module, and executing a command that
-        # blocks waiting for input ('config').
-        no_mock = os.environ.copy()
-        no_mock['QUILT_TEST_CLI_SUBPROC'] = 'false'
-
-        # With no '--dev' arg, the process should exit without a traceback
-        cmd = self.quilt_command + ['config']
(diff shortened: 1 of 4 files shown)


eode:
Ha!  Caught by my own test.

asah:
pls move the comment about try/except vs __init__ into the code, as a comment.  This helps future maintainers.

eode:
I'm not sure if it's ok for me to touch const.py, let me know if this isn't the best place for exit codes.

I've considered that an 'errors.py' might be useful, which could include (all) exceptions and error codes.

quiltdata:
go ahead and touch whatever files you need - we're done with the merge.

ᐧ

compiler/quilt/test/test_cli.py
```diff
@@ -456,37 +467,70 @@ def setUp(self):
...
-                self._main.main(cli_args)
-            except SystemExit as error:
-                result['return code'] = error.args[0] if error.args else 0
-            else:
-                result['return code'] = 0
-            result.update(self.mock_command._result)
-            return result
+            return self.execute_fast(cli_args)
+
+    def execute_cli(self, cli_args):
```
asah: dumb q: I thought [something like this] was already committed as part of the CLI testing?  sorry if dumb q...

eode: Not a dumb question -- that's just the original function, split into two functions, so that I could call the execute_cli() one separately.  I ended up needing to use Popen, anyways, so it was moot.  ..but I left the change in in case some future test has reason to prefer either `execute_cli` or `execute_fast` over using `execute`, which auto-selects cli or fast mode based on environment.

eode:
One sec, I have a between-python-versions (or similar) portability issue that's not passing appveyor.  I'll push a fix momentarily.

eode:
I've pushed that change, but appveyor looks stuck (on a different test).

eode:
I've removed some code to reset appveyor, but it looks like they're backlogged or something similar.  In the mean time, don't merge this branch until I put the code back in and re-test that, even if appveyor passes in the mean time.

eode:
appveyor has made a new test, but at this point at least one test is still hung (https://ci.appveyor.com/project/quiltdata/quilt-compiler/build/1.0.102, second test, at 52 minutes currently).

eode:
Bad news: Appveyor is hung again.  Looks like it can't handle ctrl-c events in testing.
Good news: That means we know why, and can prevent it from hanging by just xfailing the test on windows.

Amusingly, appveyor found a bug, and the fix led to the appveyor hang. 

In any case, the latest commit, 78cfca4, should still be executed on appveyor, but any running tests on this branch before that should be cancelled, if possible -- or, they'll drop off on their own eventually.

The current fix xfails a suspect portion of code.  If that doesn't work, I'll xfail the whole test in another commit.

quiltdata:
xtest is fine for now.  I think we can get a local windows box (or cloud
VM) for faster testing.

eode:
ready to roll.

eode:
* Using the name we're called as to implement traceback suppression is about as hacky as using args.  In fact, a lot of tools do just this in order to "pretend" to be a different program or act differently when called by a different name.  Swiss-army-knife shell tools, for example, or quite a few MTA's out there that act as sendmail when called as sendmail.
* We've previously decided that quilt must be installed when executed, so figuring out valid executable name by entry points is not only not hacky, it's fully supported, and usages of pkg_resources like this are common, both for a program referencing itself, and to add plugins to a program.
* The only cases I can think of where there's a conflict are:
 * Quilt is executed via a different name, by being linked to, in which case tracebacks would show if ctrl-c is used.  Not really a problem.
 * Quilt is imported by another python program named quilt.  At which point, we would suppress tracebacks for both us and them.  Serves them right.  ..and we're already `quilt` on PYPI, which is about as officially owning-the-name as it gets.
 * Quilt is imported by another program that has mangled sys.argv so that 'quilt' is the first argument, at which point we would suppress tracebacks for both us and them.  Serves them right -- if they're changing sys.argv to make it look like they're running as a different program (which they're importing) then they'd better know what they're doing.  ..and, this is pretty unlikely.

The pure and simple of it is, not only is it reasonable and supported, the consequences are minimal if we're incorrect.  On top of that, the most inconvenient case is also the most improbable -- and we can solve it by a line of documentation somewhere about external usage:

"Quilt does not support being imported by a program that spoofs sys.argv[0].  If you do so, it may suppress your tracebacks by replacing the interrupt handler for SIGINT."

..but really, we probably don't even need that, as it's just going to be noise to most people.

In any case, I know of no mechanism for altering the actual entry point script, which is auto-generated by setuptools and/or pip during install, depending on the specifics of the system.  If you know one, please let me know, because I'd like to know, it would be a cool tidbit to be aware of.

Lastly, if we forego the entry-points system altogether (not recommended for more than one reason I'll not go into here), and use our own script, then we also must detect all of the interoperability differences between systems that many talented people in the pip and setuptools contributor pool have already found and fixed.

..and __init__ is the first thing read after the entry point script.

We could, however, implement lazy loading of modules, and move exception handling into a `try: except:` block in a new __main__.py file or some such.

eode:
This still needs to be tested on Windows.  I'll try to check into that when there's less that's pressing.

asah:
ping re windows

eode:
I'll look into it once I've got something useful on quilt export, so maybe the 20th, or I'll start on it on the 23rd.

asah:
@dimaryaz @eode what's the status of this PR?

eode:
It's ready to go but hasn't been tested on Windows, and the specific test has been skipped because it crashes the Appveyor test.  ..although, ironically Appveyor caught a related bug, and the fix led to the Appveyor test skip.

Aside from that, I consider it done, but I can change it if @dimaryaz still wants a change.

If it's here when my current TODO list is done, or if it's bumped up in priority by you @asah , I'll test it on Windows and check that off.

eode:
Well, it'd be a pain to actually get the test working on Windows, and it's probably not worth the effort.  However, I've manually verified that this works correctly.  Unless we want to put additional time into the test re: windows, I'm good with this.  @dimaryaz ?

quiltdata:
I'm good with that (manual testing for now).

ᐧ

PR #235, opened 2017-12-11
QfTF/vfs: DEFAULT_MODULE_MAPPINGS missing module/callable culling

eode:
Ok, right now (since we're working on TF/vfs) the function is set to be verbose and comment about everything it removes from the default module mappings list.  It does the following:

* checks that the module is importable
* checks that the specified function/class exists
* checks that the specified function/class is callable

..rejects anything that doesn't fit, returns a culled list.

### DIFF
--- a/compiler/quilt/vfs.py
+++ b/compiler/quilt/vfs.py
@@ -7,7 +7,7 @@
     ...
     1. Title: Iris Plants Database
     	Updated Sept 21 by C.Blake - Added discrepency information
-    
+
     2. Sourc
     >>> print(open('foo/bar/iris_names').read()[0:100])
     Traceback (most recent call last):
@@ -23,7 +23,7 @@
     >>> print(open('foo/bar/iris_names').read()[0:100])
     1. Title: Iris Plants Database
     	Updated Sept 21 by C.Blake - Added discrepency information
-    
+
     2. Sourc
     >>> quilt.vfs.teardown(patchers)
     >>> print(open('foo/bar/iris_names').read()[0:100])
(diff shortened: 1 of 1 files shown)


compiler/quilt/vfs.py
```diff
@@ -106,31 +106,73 @@ def open_func(filename, mode='r', module=module, *args, **kwargs):
     patcher.start()
     return patcher
 
+
+# Note: To test this, uninstall h5py, then run your quilt_vfs_test.py.
```
asah: easier to just manually set the module mappings, then include a non-existent module ?

compiler/quilt/vfs.py
```diff
@@ -106,31 +106,73 @@ def open_func(filename, mode='r', module=module, *args, **kwargs):
...
-    # for keras/tensorflow
-    ,'h5py': 'File',
+    'builtins': 'open',
+    'bz2': 'BZ2File',
+    'gzip': 'GzipFile',
+    'h5py': 'File',    # for keras/tensorflow
     # manually add tensorflow because it's a heavyweight library
-    #'tensorflow': 'gfile.FastGFile', 'tensorflow': 'gfile.GFile'
+    #'tensorflow': 'gfile.FastGFile', 
+    #'tensorflow': 'gfile.GFile',
```
asah: given scrub_patchmap, can you add tensorflow back in?  or does it break on multi-level mappings?

compiler/quilt/vfs.py
```diff
@@ -106,31 +106,73 @@ def open_func(filename, mode='r', module=module, *args, **kwargs):
...
 def make_mapfunc(pkg, hash=None, version=None, tag=None, force=False,
                  mappings=None, install=False, charmap=DEFAULT_CHAR_MAPPINGS, **kwargs):
     """core support for mapping filepaths to objects in quilt local objs/ datastore.
        TODO: add support for reading/iterating directories, e.g. os.scandir() and friends
     """
-    if len(kwargs) == 0:
-        kwargs = DEFAULT_MODULE_MAPPINGS
+#XXX: Unused
+    # if len(kwargs) == 0:
+    #     kwargs = DEFAULT_MODULE_MAPPINGS
```
asah: obviously, remove...

compiler/quilt/vfs.py
```diff
@@ -106,31 +106,73 @@ def open_func(filename, mode='r', module=module, *args, **kwargs):
...
     if install:
         command.install(pkg, hash=hash, version=version, tag=tag, force=force)
     owner, pkg = parse_package(pkg)
+
     if mappings is None:
         mappings = { ".": "" }  # TODO: test this case
-    pkgname = "quilt.data."+owner+"."+pkg
+#XXX: Unused/debug
+#    pkgname = "quilt.data."+owner+"."+pkg
+
```
asah: obviously, remove...

compiler/quilt/vfs.py
eode: Yep. >.>

compiler/quilt/vfs.py
eode: No prob, didn't know if they were left in there for dev/debug purposes.  ..not my baby, don't want to mess with it too much.  :-)

compiler/quilt/vfs.py
eode: It's not multilevel mappings that cause the problem, it's that tensorflow has improper behavior.

For example, if you sucessfully do:
`import tensorflow`
..then tensorflow exists in your current namespace.

If you successfully do `import tensorflow.python`, then tensorflow should exist in your current namespace, and python should exist in the tensorflow namespace.  ..but it doesn't.  tensorflow.python gives jack.

..however, I think I've found a way around that -- but, that's for tomorrow.

asah: Killer, can't wait.

eode:
This one should be good to go (though appveyor is queueing, possibly from #234 occupying it)

PR #236, opened 2017-12-11
circle-ci: mac-os test

mhassan102:
Testing Circle-CI for MacOS

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -18,20 +18,27 @@ machine:
     - ./mconda.sh -b -p $HOME/miniconda
     - export PATH="$HOME/miniconda/bin:$PATH" && echo $PATH
     - $HOME/miniconda/bin/conda search "^python$"
-    - export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env27 python=2.7.10 pip
-    - export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env34 python=3.4.1 pip
-    - export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env35 python=3.5.1 pip
-    - export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env36 python=3.6.1 pip
+    - for version in 2.7.10 3.4.1 3.5.1 3.6.1
+      do
+        export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env27 python=2.7.10 pip
+      done
+    #- export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env34 python=3.4.1 pip
+    #- export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env35 python=3.5.1 pip
+    #- export PATH="$HOME/miniconda/bin:$PATH"  && conda create --yes -n quilt_env36 python=3.6.1 pip
 
 dependencies:
 
   post:
     - echo -r "Installing dependency"
     #- export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env && pip install pytest --user && python --version 
-    - export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env27 && $INSTALL_COMPILER && source deactivate quilt_env27
+    #- export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env27 && pip install 'tables' ./compiler --user && source deactivate quilt_env27
     #- export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env34 && pip install 'tables' ./compiler --user && source deactivate quilt_env34
     #- export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env35 && pip install 'tables' ./compiler --user && source deactivate quilt_env35
-    #- export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env36 && python --version && source deactivate quilt_env36
+    - for version in 2.7.10 3.4.1 3.5.1 3.6.1
+      do
+        export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env{{ version }} && python --version && source deactivate quilt_env{{version}}
+      done
+    #- export PATH="$HOME/miniconda/bin:$PATH" && source activate quilt_env{{ version }} && python --version && source deactivate quilt_env{{version}}
         
     #- export PATH="$HOME/miniconda/bin:$PATH" && pip install 'tables' ./compiler --user
     #- if [[ "$TEST_REGISTRY" = "true" ]] ; then export PATH="$HOME/miniconda/bin:$PATH" && pip install -r registry/requirements.txt --user  && pip install registry --user ; fi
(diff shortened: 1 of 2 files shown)


mhassan102:
Working on fixing error with installing compiler on mac-os:

pip install --only-binary 'tables' ./compiler
Processing ./compiler
    Complete output from command python setup.py egg_info:
    error in quilt setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/jm/fw86rxds0xn69sk40d18y69m0000gp/T/pip-xbXZn4-build/

pip install --only-binary 'tables' ./compiler returned exit code 1

Action failed: pip install --only-binary 'tables' ./compiler

akarve:
@mhassan102 we should not use sudo when installing pip as that is not how users in the wild will install pip (and it can create problems). Investigate perhaps the --user flag, and/or try without --only-binary

dimaryaz:
In Travis CI, we only install it using Conda (on Mac). We could either do the same here, or use a normal virtual environment.

(And yeah, "--only-binary" was a workaround for pytables compile errors, and hopefully not necessary anymore. Though unlikely to actually make a difference in this particular case.)

mhassan102:
Sure, I am also thinking to test with conda.

And sorry for too many commits, i didn't have Mac machine to debug this locally.

mhassan102:
I am facing still same error(compile) with conda and virtual env.
https://circleci.com/gh/quiltdata/quilt-compiler/54

Do you think, I need to get ensure this on travis(mac) first ? 
(I can test travis mac with my fork repo, no problem)

akarve:
@mhassan102 let's get this to work *without* conda. it will make the builds much faster and then we can take conda out of travis (so that travis can focus on linux). here's how you can do circle without conda: https://circleci.com/docs/1.0/language-python/

lastly, once the basic case is working, add a few major version of xcode to the matrix.

thanks.

mhassan102:
@akarve finally I am able to build circle test correctly with conda after following specs:

1. xcode = 9.0
2. python => (Python 3.6.3 :: Anaconda, Inc.)
3. build time => 02:22  (without registry test)
https://circleci.com/gh/quiltdata/quilt-compiler/66

(export env variables should be use inline, otherwise not working)

yes, we can test with-out conda to check build time

quiltdata:
Congratulations!

mhassan102:
with 'registry test' build time is 03:26 and failing with error:

pymysql.err.OperationalError: (2003, "Can't connect to MySQL server on 'localhost' ([Errno 61] Connection refused)")
../.local/lib/python3.6/site-packages/pymysql/connections.py:963: OperationalError

Are we expecting mysql service hosted locally ?

akarve:
@mhassan102 please skip travis e.g. "[ci skip]" and appveyor on these builds as we're bombing those services with these commits :) thanks

mhassan102:
sure, using '[ci skip]'  it skip all ci(s), and can execute circle only manually

dimaryaz:
MySQL is required for by the registry - but we don't need to test that on Mac. `pytest compiler` should take care of that - looks like you already got it?

The "install_requires" error makes me think that pip is out of date. In Travis, we upgrade pip before installing everything else.

mhassan102:
Yes pytest compiler went fine. 

For multiple python version, circle ci didn't support matrix as like travis or appveyor. I am thinking to apply manual "virtual env" , or using some 3rd party utility like:
https://github.com/michaelcontento/circleci-matrix

akarve:
@mhassan102 let's do the following:
* since we still have conda, just use it for the python version coverage (2.7, 3.4, 3.5, 3.6) and let's get this PR merge.
* Since we're not testing the registry you can probably delete some unnecessary lines from the script

circle.yml
```diff
@@ -9,7 +9,7 @@ machine:
     QUILT_SERVER_CONFIG: dev_config.py
     FLASK_APP: quilt_server
     FLASK_DEBUG: 1
-    INSTALL_COMPILER: "pip install \'tables\' ./compiler --user"
+    INSTALL_COMPILER: pip install 'tables' ./compiler --user
```
asah: why include quotes at all?

mhassan102:
@asah thanks! you can review now and merge.

@akarve Test now include python version (2.7, 3.4, 3.5, 3.6). 
We have test failing on 3.4, rest are seems to be fine
https://circleci.com/gh/quiltdata/quilt/124

akarve:
@mhassan102 thanks, we're almost there. a few things:
1. can you do the  `conda create` and `activate` exactly as we do in `.travis.yml`? we shouldn't use `&&` as i'm not sure if that will suppress `stderr` or `stdout` if one of the conjuncts fails. we also shouldn't need to set `PATH`. you'll have something like this. note you can do multiline YAML comments with `|`:

      conda create --yes -n quilt_env python=$QPYVER pip
      source activate quilt_env
      pip install pytest
1. unless there's a good reason (e.g. circleci doesn't show shell output) we can do with out `echo`.
1. this PR should also include edits that remove the mac compiler tests from `.travis.yaml`
1. you can no submit this as normal (you don't need `[ci skip]`) since we're converging.

thanks for your efforts so far. almost there.

mhassan102:
@akarve you are right regarding stderr and stout suppress,  but hopefully stderr will not be ignore. '&&' will not execute further-  if command on left side has non-zero(stderr) return  status.

Regarding '&&',   I already tried what you ask above (i.e separating lines),  In Mac(on circle), env variable isn't preserve in separate line and we must need to use it as inline. 

I did a test just now , and you can see in build output (https://circleci.com/gh/quiltdata/quilt/126)
"conda command not found"  error right after installing conda and setting env

(We can do it some how using full command paths in each line-  like $HOME/miniconda/bin/conda and $HOME/miniconda/bin/activate if you want)

akarve:
`&&` is fine for now; if this is working as written all that remains is to remove the duplicate tests from `.travis.yml` (we only tested 2.7, 3.5, and 3.6, to your point about 3.4)

mhassan102:
removed Mac and conda from travis.

Now we have build passing on Mac (circle) with 2.4, 3.5, 3.6
But we have failure on windows(appveyor) and linux(travis) with same error:

>       assert contents1 == contents2
E       AssertionError: assert '{\n  "childr...e": "ROOT"\n}' == '{\n  "childre...e": "ROOT"\n}'
E         Skipping 1771 identical leading characters in diff, use -v to show
E         Skipping 250 identical trailing characters in diff, use -v to show
E         -          "1b2eea66a2099b6e04a042ecc25f73d537f6bd04ddf0ebce39f1bb1112d68af4"
E         +          "26cee029b7b8debb8813083af2d17e6ad5b9807d08cb0c1bd573f941629e4457"
compiler/quilt/test/test_import.py:131: AssertionError

dimaryaz:
Ooh, that's a bug in the unittest. We have two different tests that do `import package1` - but Python is totally free to cache the import. I'll send out a fix.

dimaryaz:
https://github.com/quiltdata/quilt/pull/242

mhassan102:
Thanks!, then I think no need to change in this branch. I hope this will resolve after merge.

dimaryaz:
The fix turned out to be https://github.com/quiltdata/quilt/pull/248 - but regardless, it's now merged, so we should be good.

akarve:
@mhassan102 I think we've fixed this issue on `master`; can you merge `master` and push to this branch?

PR #237, opened 2017-12-14
Make sure objects don't get truncated if build is interrupted

dimaryaz:
When adding a file to the store, copy it to a temporary location first, then move it to the real one. Move is atomic, so should guarantee we don't end up with partial content.

### DIFF
--- a/compiler/quilt/tools/package.py
+++ b/compiler/quilt/tools/package.py
@@ -229,7 +229,11 @@ class Package(object):
         self._add_to_contents(fullname, [filehash], '', path, 'file', None)
         objpath = self._store.object_path(filehash)
         if not os.path.exists(objpath):
-            copyfile(srcfile, objpath)
+            # Copy the file to a temporary location first, then move, to make sure we don't end up with
+            # truncated contents if the build gets interrupted.
+            tmppath = self._store.temporary_object_path(filehash)
+            copyfile(srcfile, tmppath)
+            move(tmppath, objpath)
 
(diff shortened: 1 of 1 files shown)


compiler/quilt/tools/package.py
```diff
@@ -229,7 +229,11 @@ def save_file(self, srcfile, name, path):
         self._add_to_contents(fullname, [filehash], '', path, 'file', None)
         objpath = self._store.object_path(filehash)
         if not os.path.exists(objpath):
-            copyfile(srcfile, objpath)
+            # Copy the file to a temporary location first, then move, to make sure we don't end up with
+            # truncated contents if the build gets interrupted.
+            tmppath = self._store.temporary_object_path(filehash)
+            copyfile(srcfile, tmppath)
+            move(tmppath, objpath)
```
asah: performance issue for large-sized packages? (or many files...)

dimaryaz: It's just a move, though - so doesn't involve any copying. (The temp dir and the real one are both inside of the `quilt_packages` dir, so same file system.) The point of using `move` is that it should be quick and atomic.

akarve:
Do we want a unit test for this? If it's a pain to mock we can backlog.

dimaryaz:
Let's backlog it. At the moment, I can't think of a way to test it other than mocking the `copyfile` function - which makes the unittest rely on implementation details a bit too much.

PR #238, opened 2017-12-14
Refactor the way auth info is passed into API routes

dimaryaz:
Currently, the username is the first parameter in API routes, but is also stored as a thread-local, along with email. That's a bit redundant and inconsistent. We'll be adding team info soon which will only make things messier.
So:
- Create an `Auth` class which will encapsulate all of the auth info
- Remove the username parameter and only use the thread-locals.

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -210,10 +210,9 @@ class Auth:
     """
     Info about the user making the API request.
     """
-    def __init__(self, user, email, header):
+    def __init__(self, user, email):
         self.user = user
         self.email = email
-        self.header = header
 
 
 class ApiException(Exception):
@@ -268,7 +267,7 @@ def api(require_login=True, schema=None):
     def innerdec(f):
         @wraps(f)
         def wrapper(*args, **kwargs):
-            g.auth = Auth(PUBLIC, None, None)
+            g.auth = Auth(PUBLIC, None)
 
             user_agent_str = request.headers.get('user-agent', '')
             g.user_agent = httpagentparser.detect(user_agent_str, fill_none=True)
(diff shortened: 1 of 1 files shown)


registry/quilt_server/views.py
```diff
@@ -283,9 +293,11 @@ def wrapper(*args, **kwargs):
 
                     data = resp.json()
                     # TODO(dima): Generalize this.
-                    g.user = data.get('current_user', data.get('login'))
-                    assert g.user
-                    g.email = data['email']
+                    user = data.get('current_user', data.get('login'))
```
akarve: why is there also user data in `login`?

registry/quilt_server/views.py
```diff
@@ -670,8 +683,8 @@ def _generate_preview(node, max_depth=PREVIEW_MAX_DEPTH):
 @app.route('/api/package_preview/<owner>/<package_name>/<package_hash>', methods=['GET'])
 @api(require_login=False)
 @as_json
-def package_preview(auth_user, owner, package_name, package_hash):
-    instance = _get_instance(auth_user, owner, package_name, package_hash)
+def package_preview(owner, package_name, package_hash):
+    instance = _get_instance(g.auth, owner, package_name, package_hash)
```
akarve: you don't mean `g.auth.user`?

registry/quilt_server/views.py
dimaryaz: It's for Django and GitHub - looks like this part of OAuth2 is not standardized. (And GitHub Enterprise even lets you configure the field names.)

We could add a config option, too, though this was easier to start with.

registry/quilt_server/views.py
dimaryaz: The point is to pass all of auth - which will eventually include team info, admin status, etc.

dimaryaz:
We actually do have pretty good unit test coverage for access control, so I'm not worried.

As for the auth header: a better question is, why are we storing it in the first place? It's used by the package invite endpoint, but it could also be looked up again cause it's still part of the request. Anyways, I wanted "Auth" to mean "information about the user - username, email, team, etc." - and the auth header felt out of place.

PR #241, opened 2017-12-15
tf vfs patching (and work on checkpoint reads)

eode:
I'm currently stuck on checkpoint reads.  I mean, I could push it further, but it would take more time, and there are other things that are moving forward more quickly.   When I was saying I was stuck earlier, that was the point I was stuck at, not on the maybe_download replacement.  In any case, when a checkpoint is read, it ends up calling  `tensorflow.python.pywrap_tensorflow.GetMatchingFiles`.  We could possibly read the tensorflow files and mimic/mock the c functionality in some way, but it might be quicker to just require a `quilt export`.

This basically has the following:

* Moved 'STATIC' vars to top
* Made a small 'Patch' class that doesn't bork where `mock` patch does (handle tensorflow bad package importing).
* Changed patch specifications to 'some.module.func' instead of ('some.module', 'func')

When doing patches in working toward checkpoint loading, I wanted a generic means of patching that was pretty sucinct, so with patch(object_path, patch, arg_replacer=<some func> or None), the specified patch can be:
  * a func to replace the specified object
  * a list of params to call some_func(arg) on (usually mapfunc) (wraps specified object)
  * a dict of {'param': <func>} where func(arg) will be called for `param` (wraps specified object)
* When an object wraps an original, its signature should match the original (if possible)

* Broke a lot of code out into smaller functions
* Most of the functions that take a lot of params and end up creating a mapfunc can now take a mapfunc as an arg, so if you call it more than once, the second time you can just pass in mapfunc.


eode:
Closed -- data already pulled by Adam.

PR #244, opened 2017-12-19
Export command (python API only)

eode:
Allows:
```python
from quilt import command

command.export('some/package')    # exports to current dir
command.export('some/package/submodule')   # exports subset to current dir
command.export('some/package', 'somedir')  # exports to 'somedir' (creates if needed)

# sends each path as a list to some_func, which returns a new (or the same) path list.
# specifically, this allows renaming files as needed with tensorflow exports, since we don't retain filenames
command.export('another/package', mapper=some_func)

# rejects export to files that already exist
# doesn't care if the dir exists
# filename conflicts are checked before any writing happens
command.export('some/package', 'somedir')   # we did this earlier
CommandException: Invalid export path: file already exists: '/home/foo/quilt/compiler/somedir/cifar_10/x_6700_meta'
```

Caveats:
 * this hasn't been tested with the tensorflow branch, but there shouldn't be any conflicts.
 * no tests yet
 * filters not implemented yet
 * export to symlinks / links not implemented yet (is this actually a good idea anyways? *::imagines changes happening to files in storage::*)

..that's all, folks.


eode:
You might want to merge master into tensorflow, or the diffs on this will be ugly and it'll be hard to see what work is mine vs others.

PR #245, opened 2017-12-19
TensorFlow example feedback

akarve:
@asah note my comments in markdown (might not render in markdown engine but will show in diffs)


examples/tensorflow/mnist/README.md
```diff
@@ -16,3 +15,4 @@ Extracting /Users/asah/Library/Application Support/QuiltCli/quilt_packages/objs/
 2017-12-14 04:32:38.470310: I tensorflow/core/platform/cpu_feature_guard.cc:137] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX
 0.9135
 ```
+[//]: # I get a slightly different accuracy... which isn't quite reproducibility :)
```
asah: sigh, fundamental to the code... I suppose we can dig into random seeds and the like, though I've heard that even those aren't enough for DNNs... :-(

the good news is that I've found DNNs to be pretty good at statistical repro, i.e. within a small error%.  Worth researching.

In the meantime, I've removed this line to avoid confusion.

PR #248, opened 2017-12-19
Blacklist pyarrow 0.8.* for now, until the unittest is fixed.

dimaryaz:
(Or until pyarrow is fixed)


nickray:
pyarrow 0.9 is out for a while now.

PR #250, opened 2017-12-20
Export command

eode:
I've modified this to use string paths instead of path lists in filter / mapper.

Further comments will have to happen later, I'm late for something here.

### DIFF
--- a/compiler/quilt/test/test_cli.py
+++ b/compiler/quilt/test/test_cli.py
@@ -141,6 +141,9 @@ KNOWN_PARAMS = [
     [0, 'config'],
     [0, 'delete'],
     [0, 'delete', 0],
+    [0, 'export'],
+    [0, 'export', 0],
+    [0, 'export', 1],
     [0, 'generate'],
     [0, 'generate', 0],
     [0, 'help'],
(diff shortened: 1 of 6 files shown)


compiler/quilt/test/test_cli.py
```diff
@@ -488,6 +488,18 @@ def execute(self, cli_args):
             result.update(self.mock_command._result)
             return result
 
+    def _general_execute_tests(self, result, funcname):
```
asah: seems like a subclass is the better approach?

compiler/quilt/test/test_cli.py
```diff
@@ -488,6 +488,18 @@ def execute(self, cli_args):
...
+        """Takes the result from an execution and verifies general conditions
+
+        Convenience method.
+        This may not always be applicable, but it checks a few commond conditions.
```
asah: common

compiler/quilt/tools/command.py
```diff
@@ -1234,24 +1230,32 @@ def update(pkginfo, content):
 #     if pathlist[-1].startswith('x_'):
 #         pathlist[-1] = '-' + pathlist[-1][2:]
 #     return pathlist
-def export(pkginfo, output_path='.', filter=None, filename_mapper=lambda x: x,
-              force=False):
-    """Export package file data.  Note: this is *not* the opposite of import.
+def export(package, output_path='.', filter=lambda x: True, mapper=lambda x: x, force=False):
```
asah: the example mapper function needs to be updated as well.

PR #252, opened 2017-12-22
Implement the teams model and update all endpoints to work with teams.

dimaryaz:
- `Team` and `UserTeam` tables are added, as well as a `Package.team_id` column.
- Routes that deal with packages now have a `/<team_name>/` component. Old routes work as before, but set team name to "public". Clients are expected to start using the new routes.
- API endpoints look up the team and check if the user has access to it before doing anything else.
- DB queries filter by team_id in addition to owner and package name.

There is no way to actually create teams yet (other than directly updating the DB). All testing was done with unit tests.

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -27,10 +27,10 @@ import stripe
 
 from . import app, db
 from .analytics import MIXPANEL_EVENT, mp
-from .const import EMAILREGEX, PaymentPlan, PUBLIC
+from .const import EMAILREGEX, PaymentPlan, PUBLIC, TEAM
 from .core import decode_node, encode_node, find_object_hashes, hash_contents, FileNode, GroupNode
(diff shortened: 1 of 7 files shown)


registry/quilt_server/views.py
```diff
@@ -427,6 +485,21 @@ def _get_or_create_customer():
 def _get_customer_plan(customer):
     return PaymentPlan(customer.subscriptions.data[0].plan.id)
 
+def team_route(rule, defaults={}, **options):
+    """
+    Decorator that adds the route itself and the corresponding legacy non-team route.
+    """
```
kevinemoore: This method is a bit tricky and could use a more detailed explanation in the comment. It wasn't obvious to me why two routes are needed (to support older clients) nor what the different URL patterns look like. Also, I think I get why, but it's worth a quick note on why you create and return the decorator function.

PR #255, opened 2017-12-26
Keep build_from_cache in sync with test/build.yml

kevinemoore:
The test test_build_from_cache contains an assertion that verifies that
the cache is actually filled. The hash key in the cache is generated
from both the source file and kwargs so it needs to be kept in sync
with the build file (test/build.yml).

Fix an indentation bug in test/build.yml.

### DIFF
--- a/compiler/quilt/test/build.yml
+++ b/compiler/quilt/test/build.yml
@@ -2,9 +2,9 @@
 contents:
   dataframes:
     csv: 
-    kwargs:
-      parse_dates: ['Date0']
-    file: data/10KRows13Cols.csv
+      kwargs:
+        parse_dates: ['Date0']
+      file: data/10KRows13Cols.csv
     nulls:
       file: data/nulls.csv
     tsv:
--- a/compiler/quilt/test/test_build.py
+++ b/compiler/quilt/test/test_build.py
@@ -49,7 +49,7 @@ class BuildTest(QuiltTestCase):
 
         # Verify cache contents
         srcpath = os.path.join(mydir, 'data/10KRows13Cols.csv')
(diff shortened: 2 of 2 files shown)


compiler/quilt/test/build.yml
```diff
@@ -2,9 +2,9 @@
 contents:
   dataframes:
     csv: 
-    kwargs:
-      parse_dates: ['Date0']
-    file: data/10KRows13Cols.csv
+      kwargs:
```
dimaryaz: @aneesh just fixed this one - but the other change is still needed I'm guessing.

kevinemoore: This is one way to fix that broken test, but @akarve might want to fix it another way. This PR is targeting his branch, so it's up to him.

akarve: I merged those fixes into `master`. Can you catch this branch up?

compiler/quilt/test/test_build.py
```diff
@@ -49,7 +49,7 @@ def test_build_from_cache(self):
 
         # Verify cache contents
         srcpath = os.path.join(mydir, 'data/10KRows13Cols.csv')
-        path_hash = build._path_hash(srcpath, 'csv', {})
+        path_hash = build._path_hash(srcpath, 'csv',  {'parse_dates': ['Date0']})
```
akarve: this works but it's super fragile; @kevinemoore can you get this info from build.yml? other unit tests will need the logic.

compiler/quilt/test/build.yml
kevinemoore: This PR is merging into cacheBug. It's caught up with that branch.

PR #257, opened 2017-12-26
[trivial] Update test_generate_buildfile, fix missing commas in setup.py

eode:
When a buildfile is generated, it now generates the name 'README_md' instead of 'README'.  Updated test.


eode:
I think I'll just roll these into my 'quilt export' PR.  They're pretty trivial, but interfere with test success there.

PR #259, opened 2018-01-04
 Migrate from MySQL to Postgres 

dimaryaz:
Mostly just moving from MySQL to Postgres data types, functions, etc.

A few functional changes:
- Package contents is stored as JSONB instead of text. It gets serialized/deserialized by SQLAlchemy when accessed.
- Hashes are case sensitive. Shouldn't matter because we either generate them ourselves, or verify a user-supplied hash by comparing it to the one we generated.


registry/migrations/env.py
```diff
@@ -57,7 +57,7 @@ def run_migrations_online():
 
     # this callback is used to prevent an auto-migration from being generated
     # when there are no changes to the schema
-    # reference: http://alembic.readthedocs.org/en/latest/cookbook.html
+    # reference: http://alembic.zzzcomputing.com/en/latest/cookbook.html
```
asah: sorry, why this change?

registry/quilt_server/models.py
```diff
@@ -6,26 +6,20 @@
...
 from . import db
 
-UTF8_BIN = 'utf8_bin'
-UTF8_GENERAL_CI = 'utf8_general_ci'
-
-def CaseSensitiveString(length):
-    return mysql.VARCHAR(length, collation=UTF8_BIN)
-
-USERNAME_TYPE = CaseSensitiveString(64)
+USERNAME_TYPE = db.String(64)
```
asah: fyi in postgres it's trivial to setup a real column-constraint which matches the rules for allowable usernames... CREATE FUNCTION then use in a constraint.  imho add as a TODO.

registry/requirements.txt
```diff
@@ -22,7 +22,7 @@ MarkupSafe==1.0
 mixpanel==4.3.2
 oauthlib==2.0.2
 packaging==16.8
-PyMySQL==0.7.11
+psycopg2==2.7.3.2
```
asah: it'll be interesting to see if we can lock onto this specific version...

registry/migrations/env.py
dimaryaz: I re-generated all of the migrations; I guess the new version of alembic did that. The old page doesn't exist anymore, so this looks right.

registry/quilt_server/models.py
dimaryaz: Eh, we'll have a proper User table and integer IDs eventually

registry/requirements.txt
dimaryaz: Not sure what you mean? This does lock it to this version, right?

quiltdata:
pls add this as a comment, e.g. TODO: we'll have a proper User table and
integer IDs eventually
ᐧ

quiltdata:
nevermind, looks good.
ᐧ

PR #260, opened 2018-01-04
ban public users config option v0

meffij:
could not run unit tests locally, Dima please review


meffij:
Would also like to add tests eventually, need guidance on that

akarve:
@meffij the act of submitting the PR will run the unit tests; looks like travis/pr failed

registry/quilt_server/config.py
```diff
@@ -6,6 +6,13 @@
 
 See `app.config.from_object('...')` in __init__.py.
 """
+import os
+
+BAN_PUBLIC_USERS = os.getenv('QUILT_BAN_PUBLIC_USERS', '')
+if len(BAN_PUBLIC_USERS):
```
dimaryaz: `if BAN_PUBLIC_USERS`

dimaryaz:
Can you change the `api` decorator instead of updating each call to it?

registry/quilt_server/views.py
```diff
@@ -31,6 +31,7 @@
 from .models import (Access, Customer, Instance, Invitation, Log, Package,
                      S3Blob, Tag, UTF8_GENERAL_CI, Version)
 from .schemas import LOG_SCHEMA, PACKAGE_SCHEMA
+from .config import BAN_PUBLIC_USERS
```
dimaryaz: Sorry, one more thing... You should use `app.config` instead of importing it directly. See the code below.

PR #262, opened 2018-01-04
Local store hack to avoid lots of code-rewrite

kevinemoore:
Instead of adding an extra level in the local-store directory
hierarchy, which seems more solid, just prepend the package owner with
the registry/team.

Build and import is working.

To do:
- parse the team:owner/pkg syntax (might be some issue with overuse of
colon for both team and version info)
- connect to separate registries based on team

### DIFF
--- a/compiler/quilt/tools/store.py
+++ b/compiler/quilt/tools/store.py
@@ -12,7 +12,7 @@ from .package import Package, PackageException
 from .util import BASE_DIR
 
 # start with alpha (_ may clobber attrs), continue with alphanumeric or _
-VALID_NAME_RE = re.compile(r'^[a-zA-Z]\w*$')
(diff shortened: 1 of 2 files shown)


compiler/quilt/tools/store.py
```diff
@@ -12,7 +12,7 @@
 from .util import BASE_DIR
 
 # start with alpha (_ may clobber attrs), continue with alphanumeric or _
-VALID_NAME_RE = re.compile(r'^[a-zA-Z]\w*$')
+VALID_NAME_RE = re.compile(r'^[a-zA-Z]\w*(:[a-zA-Z]\w*)?$')
```
meffij: Technically the prefix is the optional part, not the suffix. Just a nitpick

PR #265, opened 2018-01-05
Point flask to a new endpoint for verifying users

kevinemoore:
The old /profiles/<username> endpoint in Django was part of the
old-Quilt app and was disabled. Access add relied on that endpoint to
verify that users exist. This change uses a new user-verifying endpoint
in Django.

### DIFF
--- a/registry/quilt_server/prod_config.py
+++ b/registry/quilt_server/prod_config.py
@@ -18,7 +18,7 @@ OAUTH = dict(
     client_secret=os.environ['OAUTH_CLIENT_SECRET'],
     redirect_url='https://%s/oauth_callback' % REGISTRY_HOST,
     user_api='https://%s/api-root' % OAUTH_API_HOST,
-    profile_api='https://%s/accounts/profile?user=%%s/' % OAUTH_API_HOST,
+    profile_api='https://%s/accounts/profile?user=%%s' % OAUTH_API_HOST,
     have_refresh_token=True
 )
 
--- a/registry/quilt_server/prod_config.py
+++ b/registry/quilt_server/prod_config.py
@@ -18,7 +18,7 @@ OAUTH = dict(
     client_secret=os.environ['OAUTH_CLIENT_SECRET'],
     redirect_url='https://%s/oauth_callback' % REGISTRY_HOST,
     user_api='https://%s/api-root' % OAUTH_API_HOST,
-    profile_api='https://%s/profiles/%%s/' % OAUTH_API_HOST,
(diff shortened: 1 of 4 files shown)


registry/quilt_server/views.py
```diff
@@ -1079,12 +1080,14 @@ def access_put(owner, package_name, user):
                 "Invalid credentials"
                 )
         elif resp.status_code != requests.codes.ok:
+            print(resp.text)
```
dimaryaz: Remove it?

registry/quilt_server/views.py
```diff
@@ -1045,6 +1045,10 @@ def access_put(owner, package_name, user):
             "Only the package owner can grant access"
         )
 
+    django_headers = {
```
dimaryaz: Let's not call it django. Maybe `auth_provider_headers`? `auth_headers`?

registry/quilt_server/views.py
dimaryaz: Remove it?

registry/quilt_server/views.py
```diff
@@ -1079,12 +1080,14 @@ def access_put(owner, package_name, user):
...
                 )
         elif resp.status_code != requests.codes.ok:
+            print(resp.text)
             raise ApiException(requests.codes.server_error, "Server error")
         return dict()
 
     else:
         if user != PUBLIC:
-            resp = requests.get(OAUTH_PROFILE_API % user)
+            resp = requests.get(OAUTH_PROFILE_API % user,
```
dimaryaz: It just occurred to me that user probably needs to be url-encoded. I guess we restrict package names to only Python identifiers, but Django technically allows things like `+`, `@`, etc.

kevinemoore: True, but let's leave that for a different PR. This covers what's needed to get sharing working again and we've got an outstanding issue to restrict the allowed usernames to python-legal names anyway.

registry/quilt_server/prod_config.py
```diff
@@ -18,7 +18,7 @@
     client_secret=os.environ['OAUTH_CLIENT_SECRET'],
     redirect_url='https://%s/oauth_callback' % REGISTRY_HOST,
     user_api='https://%s/api-root' % OAUTH_API_HOST,
-    profile_api='https://%s/profiles/%%s/' % OAUTH_API_HOST,
+    profile_api='https://%s/accounts/profile?user=%%s/' % OAUTH_API_HOST,
```
dimaryaz: Oh no... Trailing slash

PR #266, opened 2018-01-05
Fixes / improvements to export and testing on tensorflow branch

eode:
These are a few improvements you requested @asah.

Since forcing an install every time resets local data and undoes any tensorflow work that has been saved in the local package store, but not uploaded to quilt, I created a small ensure_installed() method that checks if the package is installed, and installs it if not.  That helps prevent the manual interaction on the CLI.

Other bits:
* Removed train-short.py
* 'silent' option to squelch output of export
* faster handling of zero-byte file workaround
* check for writability of output dir
* check for pre-existing subdirs, not just files
* catch all OSErrors (such as DiskFull errors) and reraise as CommandException

* Tests for the above
* Test force=True

Lastly, if you have FileNodes missing `metadata['q_path']` in a package (created by _set before it was modified) it might cause issues now that q_path is required.  If you need to fix a package, you can run command._repair_package('asah/cifar10') (for example).  The problem can cause a package not to be deletable via `quilt rm`.  This should fix that.  If you don't need it, and no-one else has packages touched by the tensorflow branch, you can just delete _repair_package, or let me know and I will.


compiler/quilt/test/test_command.py
```diff
@@ -574,6 +591,54 @@ def test_export(self):
...
+            files = set(f for f in test_dir.glob('*') if f.is_file())
+            # sorted and reversed means files before their containing dirs
+            for path in sorted(test_dir.glob('**/*'), reverse=True):
+                # keep files from root of test_dir
+                if path in files:
+                    continue
+                # remove everything else
+                path.rmdir() if path.is_dir() else path.unlink()
+            # now there are only files in the export root to conflict with.
+            with pytest.raises(command.CommandException, match='file already exists'):
```
asah: Test that force doesn't throw this exception?

compiler/quilt/test/test_command.py
```diff
@@ -574,6 +592,57 @@ def test_export(self):
...
+                command.export(pkg_name, str(test_dir))
+
+            # Test raise when exporting to existing dir structure
```
asah: Ditto re Force

compiler/quilt/tools/command.py
```diff
@@ -1326,17 +1332,85 @@ def iter_filename_map(node):
         # Technically successful, but with nothing to do.
         # package may have no file nodes, or user may have filtered out all applicable targets.
         # -- should we consider it an error and raise?
-        print("No files to export.")
+        if not silent:
```
asah: I don't think we're ready to standardize on silent... For now, just print like other commands.

Then in a separate proposal, slack Dima and request permission to add silent= to all python API commands to disable screen/stdout output.

compiler/quilt/tools/command.py
```diff
@@ -1326,17 +1332,85 @@ def iter_filename_map(node):
...
+            print("No files to export.")
         return
 
+    # ensure output path is writable.  I'd just check stat, but this is fully portable.
+    try:
+        output_path.mkdir(exist_ok=True)  # could be '.'
+        with tempfile.TemporaryFile(dir=str(output_path), prefix="quilt-export-write-test-", suffix='.tmp'):
```
asah: Add a TODO: performance hit due to write amplification

compiler/quilt/tools/command.py
```diff
@@ -1326,17 +1332,85 @@ def iter_filename_map(node):
...
+        if not silent:
+            print('..done.')
+    except OSError as error:
+        commandex = CommandException("Unexpected error during export: " + str(error))
+        commandex.original_error = error
+        raise commandex
+
+
+def _repair_package(packagename):
+    """
```
asah: Currently we're not handling locally data store upgrades and instead asking users to clear between upgrades.  This keeps our code cleaner.

If you disagree, pls ping Kevin fire approval

asah: For approval

examples/tensorflow/tensorflow-cifar-10/train.py
```diff
@@ -9,12 +9,12 @@
...
-import quilt, quilt.vfs
+import quilt.vfs
+
 QUILT_PKG = 'asah/cifar10'
 
-print("Connecting to Quilt...")
-# force=False because otherwise local data in storage is overwritten
-quilt.install(QUILT_PKG, force=False)
+print("Ensuring {} is installed".format(QUILT_PKG))
+quilt.vfs.ensure_installed(QUILT_PKG)
```
asah: As I look at this, I'm wondering why ensure installed isn't just a parameter of vfs.setup() ?

compiler/quilt/tools/command.py
eode: Makes sense.

compiler/quilt/tools/command.py
eode: Will do, but that's a one-time hit, per-export, not in a loop or anything.

examples/tensorflow/tensorflow-cifar-10/train.py
eode: Not a bad idea.  Will do.

compiler/quilt/test/test_command.py
eode: The existing test would raise the CommandException on either of these cases, because it writes over a full prior install with both subdirs and files -- should I just catch that (potential) error and reraise as a Failed() exception via pytest.fail()?

compiler/quilt/tools/command.py
eode: No problem -- the _set bug wasn't released anyways, as nothing used _set before this tensorflow branch -- it was just in case your pkgstore was messed up and you didn't want to delete it.

examples/tensorflow/tensorflow-cifar-10/train.py
eode: Oh -- well, actually, would that be a useful command api addition as well?  It seems people using quilt on the python side might want to have a quick 'install' command that doesn't overwrite their existing stuff, and that kind of convenience is nice in an API.

compiler/quilt/test/test_command.py
asah: sounds like export() needs to be changed?  export(force=True) should not throw an exception imho... it should just overwrite.

sorry if i'm missing somethg

eode: Sorry, to be clear:  The checks you requested already implicitly exist due the the `force=True` test -- because, as you say, quilt export shouldn't raise an exception, and doesn't, currently. 

Both of the checks you noted were to ensure that the non-force version *does* raise those exceptions, but the force version of the test doesn't catch any errors, it just checks to make sure the export worked -- so if any errors are raised, the test fails, as it should.

eode:
Windows security is a PITA re: write permissions testing.  I think the best option is to skip the permissions-write-fail test on windows.  ..unless there are objections.   This is ready for review again, @asah .

eode:
@asah this should be good.

PR #267, opened 2018-01-05
Clean up local store for teams

kevinemoore:
Adding an extra level in the local-store directory tree to separate
packages from different teams/registries.

### DIFF
--- a/compiler/quilt/data.py
+++ b/compiler/quilt/data.py
@@ -135,9 +135,10 @@ class ModuleFinder(object):
                 if os.path.isdir(user_path):
                     return FakeLoader(user_path)
 
-            raise ImportError('Could not find package by user {user!r} named {package!r}.\n  '
-                              'Check the name, or use "quilt install {user}/{package}" to install'
-                              .format(**locals()))
+            raise ImportError('Could not find package.\n  '
+                              'Check the name, or use "quilt install {user1}/{package}" or\n'
+                              '"quilt install {team}:{user2}/<packagename>" to install'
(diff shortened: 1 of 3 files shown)


compiler/quilt/tools/store.py
```diff
@@ -40,7 +39,7 @@ class and its subclasses abstract file formats, file naming and
     TMP_OBJ_DIR = 'tmp'
     PKG_DIR = 'pkgs'
     CACHE_DIR = 'cache'
-    VERSION = '1.2'
+    VERSION = '1.3'
```
dimaryaz: Does it mean, users will have to blow away their `quilt_packages` again?

compiler/quilt/data.py
```diff
@@ -131,17 +131,16 @@ def find_module(fullname, path=None):
...
                 # Try A Team/Other-Registry Path
                 team, user = parts
-                file_path = store.user_path(parts[0], parts[1])
-                if os.path.isdir(file_path):
-                    return FakeLoader(file_path)
+                user_path = store.user_path(team, user)
+                if os.path.isdir(user_path):
+                    return FakeLoader(user_path)
 
             raise ImportError('Could not find package by user {user!r} named {package!r}.\n  '
```
dimaryaz: Can you check that the error message looks right if the package doesn't exist? I think it's currently broken.

compiler/quilt/tools/store.py
kevinemoore: Yes.

PR #268, opened 2018-01-06
team demo -- bugfixes, tests for pkgname parsing and imports

eode:
* extended `test_import.py` to include versions of each test, but with team designations.
  * provides coverage for `from quilt.data.team.user import pkg` and related actions.
  * very possibly overkill on this one, but it was straightforward to modify existing methods
* added `test_store.py` -- provides coverage for `store.parse_package()`

Discovered and squashed:
* `store.parse_package()` permitted bad subpath elements -- e.g.,` 'user/package/subpath/bad!subpath'`.  Fixed.
  * Brought to light a situation where a trailing slash was being generated incorrectly during testing.
* Traced rogue slash to an `is None` check in `command.install_via_requirements()` where a truthiness check was needed.  Fixed.

Also, there are some minor changes -- added team display for `quilt ls` and `quilt build` output, and _prefixed a private function in command.py.

### DIFF
--- a/compiler/quilt/test/test_import.py
+++ b/compiler/quilt/test/test_import.py
@@ -186,26 +186,6 @@ class ImportTest(QuiltTestCase):
         # Imports should find the second package
         from quilt.data.foo.nested import dataframes
 
-    def test_team_multiple_package_dirs(self):
-        # First level
-        mydir = os.path.dirname(__file__)
-        build_path = os.path.join(mydir, './build_simple.yml')
-        command.build('test:bar/nested', build_path)
-
-        # Second level: different package
-        os.mkdir("aaa")
-        os.chdir("aaa")
-        build_path = os.path.join(mydir, './build.yml')
-        command.build('test:bar/nested', build_path)
-
-        # Third level: empty package directory
-        os.mkdir("bbb")
-        os.chdir("bbb")
-        os.mkdir(PACKAGE_DIR_NAME)
-
-        # Imports should find the second package
-        from quilt.data.test.bar.nested import dataframes
-
     def test_save(self):
         mydir = os.path.dirname(__file__)
         build_path = os.path.join(mydir, './build.yml')
--- a/compiler/quilt/test/test_import.py
+++ b/compiler/quilt/test/test_import.py
@@ -74,6 +74,65 @@ class ImportTest(QuiltTestCase):
         with self.assertRaises(ImportError):
             from quilt.data.foo.baz import blah
 
+    def test_team_imports(self):
+        mydir = os.path.dirname(__file__)
+        build_path = os.path.join(mydir, './build.yml')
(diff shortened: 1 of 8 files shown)


eode:
Changes:
* Test CLI for `quilt login` and `quilt login teamname`.
* Test `command.login()` and `command.login_with_token()` to ensure rejection of bad team names
* Fixed acceptance of bad teamnames by `login` and `login_with_token` (and others) via `get_registry_url()`

compiler/quilt/test/test_import.py
```diff
@@ -110,6 +186,26 @@ def test_multiple_package_dirs(self):
         # Imports should find the second package
         from quilt.data.foo.nested import dataframes
 
+    def test_team_multiple_package_dirs(self):
```
kevinemoore: This test and the one it's copying are out-of-date. Fixing the other one might be out of scope for this PR, but there's no point in adding this test because changing the directory where build is run no longer changes the PackageStore directory.

compiler/quilt/test/test_store.py
```diff
@@ -0,0 +1,60 @@
+"""
```
kevinemoore: This test looks good, but I think it makes more sense to move the parse_package helpers out of store.py (they're called in multiple places and aren't related to the PackageStore) than to create this new test file.

eode: That works for me.   Where should I move it, though?  There's no 'test_utils.py', and it seems tools.utils would be a good spot for it.

compiler/quilt/test/test_import.py
eode: Nix the test, or make a better one?

compiler/quilt/test/test_store.py
kevinemoore: On second thought, just leave it as is for now and we can move both the helpers and tests in a separate PR.

compiler/quilt/test/test_import.py
kevinemoore: Yeah, just nix the additional test.

compiler/quilt/test/test_store.py
eode: That sounds good to me.  I was noticing the same thing, btw.

eode:
Re: earlier changes (didn't hit comment):
* Test CLI for `quilt login` and `quilt login teamname`.
* Test `command.login()` and `command.login_with_token()` to ensure rejection of bad team names
* Fixed acceptance of bad teamnames by `login` and `login_with_token` (and others) via `get_registry_url()`

eode:
Waiting for team_demo to test clean.

eode:
@kevinemoore  I've resolved the issues in #277, and this is ready to merge in.  The appveyor test was cancelled, not failed, and if there's an issue, we can fix it inside the team_demo branch, yes?

PR #270, opened 2018-01-07
Teams package gallery shows latest packages

akarve:
Instead of showing hardwired packages on quiltdata.com, show packages from `/api/recent_packages`


catalog/static/config.js
```diff
@@ -19,4 +21,21 @@ if (window.location.hostname === 'quiltdata.com') {
     // userApi: 'https://api.github.com/user',
     // signOutUrl: '/'
   };
+} else {
+  // TODO: This should come from Docker variables on team instances
```
dimaryaz: It already does; this file gets completely overwritten by `config.js.tmpl` in Docker images - so just add the team variables there.

catalog/static/config.js
```diff
@@ -1,12 +1,14 @@
 /* eslint-disable */
-if (window.location.hostname === 'quiltdata.com') {
+const { hostname } = window.location;
```
dimaryaz: This file doesn't get processed by the ES6 compiler, so I don't think we can do this here - unless it's supported by all browsers we care about.

.gitignore
```diff
@@ -4,6 +4,7 @@ dist
 *.egg-info
 *.py[cod]
 tmp_*
+node_modules
```
dimaryaz: It's already in `catalog/.gitignore`, though? You can move it here - but then delete it from the other one.

catalog/app/containers/HomePage/index.js
```diff
@@ -27,11 +28,18 @@ const Styler = styled.div`
 
 export default class HomePage extends React.PureComponent { // eslint-disable-line react/prefer-stateless-function
   render() {
+    const { team = {} } = config;
+    const header = team.name && typeof team.name === 'string' ?
```
dimaryaz: Probably safe to assume that if it's set, it's going to be a string.

catalog/app/containers/HomePage/index.js
```diff
@@ -27,11 +28,18 @@ const Styler = styled.div`
...
+      team.name.toUpperCase() : undefined;
```
dimaryaz: `toUpperCase()` might work in this particular case, but not what we want in general. Maybe we need two separate things: team ID (used in the DNS name, etc.) and a user-friendly team name.

catalog/app/containers/HomePage/index.js
akarve: yeah i am preventing `toUpperCase` from blowing up but it's probably overkill

.gitignore
akarve: Yeah done

PR #271, opened 2018-01-07
Add auth/django database to docker

kevinemoore:
Adding a django migration so that the whole local-auth setup can be run from: docker-compose up


registry/README.md
```diff
@@ -1,10 +1,152 @@
-# Overview
+ Overview
```
dimaryaz: The `#` got lost?

PR #274, opened 2018-01-09
Implement `quilt.team` imports

dimaryaz:
- Refactor the magic imports and use the code for both `quilt.data` and `quilt.team`
- Don't raise `ImportError`s when failing to find a package: that's not actually allowed. Just return `None`.

### DIFF
--- a/compiler/.pylintrc
+++ b/compiler/.pylintrc
@@ -191,7 +191,7 @@ notes=FIXME,XXX,TODO,qballs,QBALLS
 
 [TYPECHECK]
 
-ignored-modules=quilt.data
+ignored-modules=quilt.data,quilt.team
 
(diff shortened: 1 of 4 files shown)


dimaryaz:
The diff is misleading, btw: I renamed `data.py` to `imports.py`, then added a new `data.py` and a `team.py`.

dimaryaz:
I actually thought about that - but almost all of the code in `imports.py` is specific to "magic imports", and not useful for things like `quilt.import(...)`. Instead, we should implement `quilt.import` somewhere in tools, and call it from the magic imports code.

kevinemoore:
OK, the code refactor can wait.

dimaryaz:
Anyways, I'll do that in a different PR. We should also figure out the naming - cause we can't use `import`.

PR #275, opened 2018-01-09
Fix `quilt ls`

akarve:
Only descend directories (instead of files, like `.DS_Store` on Mac OS).

### DIFF
--- a/compiler/quilt/tools/store.py
+++ b/compiler/quilt/tools/store.py
@@ -9,7 +9,7 @@ from shutil import rmtree
 from .const import PACKAGE_DIR_NAME
 from .core import FileNode, RootNode, TableNode, CommandException
 from .package import Package, PackageException
-from .util import BASE_DIR
+from .util import BASE_DIR, sub_dirs
 
 # start with alpha (_ may clobber attrs), continue with alphanumeric or _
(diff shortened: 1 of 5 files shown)


compiler/quilt/tools/store.py
```diff
@@ -205,7 +205,7 @@ def iterpackages(self):
         Return an iterator over all the packages in the PackageStore.
         """
         pkgdir = os.path.join(self._path, self.PKG_DIR)
-        for user in os.listdir(pkgdir):
+        for user in sub_dirs(pkgdir):
             for pkg in os.listdir(os.path.join(pkgdir, user)):
```
dimaryaz: This one, too?

compiler/quilt/tools/util.py
```diff
@@ -81,3 +81,11 @@ def gzip_compress(data):
     with gzip.GzipFile(fileobj=buf, mode='wb') as fd:
         fd.write(data)
     return buf.getvalue()
+
+def sub_dirs(path, visible=True):
```
dimaryaz: I wouldn't even bother with `visible`; just ignore them unconditionally.

akarve:
Handling in #284

PR #282, opened 2018-01-10
Keras/Tensorflow features, virtual files (quilt.vfs), quilt export

asah:


asah:
NOTE: this is not ready for review.

First, teams needs to get merged then vfs.py updated to use the new nodes.py/team.py interface style. Cc @eode

eode:
Apparently my 'master' branch was behind quiltdata/master -- or some such.  In any case, I've resolved the conflicts via github this morning, and am awaiting test completion.

eode:
This still requires VFS updates re: teams.

eode:
From Kevin via Slack:
Quilt for Tensor Flow Feedback

High Level:
- Great stuff! Looks like very useful features beyond Tensor Flow
- Would be very helpful to update the design document with a high-level description of how the features work (e.g., VFS & export) as the code is written

Questions:
- I see licenses for the example code, all Apache 2 I think. I couldn’t tell from my eyeballing the code, do we have proper attribution for all the external code we’ll be incorporating? Is it all Apache 2?
- Looks like the filename and path is now part of the package metadata, is that a new field or was it there before?
- Could/should we have separate fields for: filename, path and maybe extension? I can see us wanting to query those independently at times.
- If it’s new, have you tested whether or not adding path to the metadata requires any change to the JSON schema in the registry?
- Is the mapper really needed now that we can specify file paths in the package metadata?

Suggested Changes:
- Change the command name from “importpkg” to “load”. It’s shorter, clearer and matches the syntax in common use in R.
- Avoid adding _ext to node names for File nodes to be consistent with Table nodes. We can keep the existing code to avoid duplicate node names and use the path/filename in the metadata to recover extensions.

eode:
Responses from me:

@asah - Can you deal with the licensing / attribution requirements?  I can't say I'm too familiar with it -- not enough that I'd want others relying on me in that regard.

### Responses to Kevin (via Slack):
**Filename and path:**
Yup!  They were included beforehand, however `PackageNode._set()` (previously unused) didn't set it -- so that's been updated to include as much metadata as it has access to at the time it's called.  Also, `FileNode` has been updated to require `metadata`, instead of leaving it optional.  I checked with others before doing so, since it's a deep area of the code.

**Separate fields for filename/basename/path/extionsion:** 
I'd be all for that.  It might be nice to expose a _path property that gives/receives a `pathlib.Path` object:
```fn.._path.parent   # parent dir
fn._path.name      # foo.txt
fn._path.stem       # foo
fn._path.suffix      # .txt
fn._path.parts       # ('path', 'parts', 'for', 'foo.txt')
#..etc
```

**JSON registry:**  Pre-existing (no problem)

**Mapper:**  
If you  mean the VFS mapper, it probably still has some uses if a person wants to import/export things in a particular way -- for example, take a subnode from a package and expose it in the current directory as 'tensorflow', even if it's not named 'tensorflow' in the package.

eode:
**Comments regarding the a above conversation:**
- `metadata['q_ext'] isn't set to the filename extension, as this actually specifies the transform used -- how ugly of a process would it be to rename this?
- Regarding separate fields
  - Having an attribute that exposes a `pathlib.Path()` for objects with metadata['q_path'] would be a dead-simple way to get all of the desired functionality.

asah:
taking load/importpkg discussion offline into a designdoc.

eode:
Per conversation with Kevin, renamed 'importpkg' to 'load' -- did that before I saw the above, comment.  Can revert if  needed.

quiltdata:
Keep load () for now, just be careful to update EVERYWHERE incl docs,
examples.

eode:
It looks like renaming the test `test_importpkg_updates` to `test_load_updates` put it in a different order -- after `test_import`.   `test_import` uses the same package name, `foo/package`.  I can rename the test package used to `foo/package5`, but is this indicative of an actual issue?

The order of execution, roughly, is:
```python
In test_import.py:

<other tests>
test_imports
    calls command.build('foo/package', "<path>/build.yml")
    <does testing>
test_load_update
    calls command.build('foo/package', "<path>/build.yml")
    calls command.build('foo/package', package)   # ultimately command.build_from_node
```
I would think that the second build would overwrite the first build, but something is left over in a way that causes the build_from_node to fail.  I don't mind researching this, but I want to make sure that I'm not wasting my time.

dimaryaz:
The package naming in tests has bitten us before... Unfortunately, any name used in imports needs to be unique cause Python caches imports. You can use the name of the test as the user name or package name, to be sure it's unique.

eode:
Great!  I just wanted to make sure it wasn't some other issue.  As long as it's python-reloading that's happening, I'll just rename and move on.

compiler/quilt/test/test_import.py
```diff
@@ -3,6 +3,9 @@
 """
 
 import os
+import tempfile
+import time
+import quilt
```
asah: is this the way we're importing quilt?  or are we using a relative import to ensure we get the same version as the local git clone?

compiler/quilt/test/test_import.py
```diff
@@ -351,6 +355,72 @@ def test_set_non_node_attr(self):
...
+        reloaded_module = quilt.load(newpkgname)
+        assert reloaded_module is not module
+        newpath2 = getattr(reloaded_module, newfilename)()
+        assert 'myfile' not in newpath2
+
+    def test_multiple_updates(self):
+        mydir = os.path.dirname(__file__)
+        build_path = os.path.join(mydir, './build.yml')
+        command.build('foo/package', build_path)
+        from quilt.data.foo import package
```
asah: does quilt.import() avoid caching problems?

compiler/quilt/__init__.py
```diff
@@ -87,8 +87,10 @@ def handle_interrupt(signum, stack):
     build,
     check,
     config,
+    export,
```
asah: please add basic CLI and API docs for quilt.export():
https://github.com/quiltdata/quilt/blob/master/docs/api-python.md
https://github.com/quiltdata/quilt/blob/master/docs/cli.md

note: do not add docs for load() until we resolve the name vs 'import'

compiler/quilt/test/test_build.py
```diff
@@ -169,7 +169,7 @@ def test_generate_buildfile(self):
         assert os.path.exists(buildfilepath)
         build.build_package(None, 'test_generated', 'generated', buildfilepath)
         os.remove(buildfilepath)
-        from quilt.data.test_generated.generated import bad, foo, nuts, README
+        from quilt.data.test_generated.generated import bad, foo, nuts, README_md
```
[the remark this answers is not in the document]
eode: In the final change, it should be README -- per.. ..Kevin?  Good catch.

compiler/quilt/__init__.py
eode: done

compiler/quilt/test/test_cli.py
```diff
@@ -566,13 +578,9 @@ def test_cli_command_config(self):
         result = self.execute(cmd)
 
         # General tests
-        assert result['return code'] == 0
-        assert result['matched'] is True  # func name recognized by MockObject class?
-        assert not result['bind failure']
+        self._general_execute_tests(result, funcname='config')
```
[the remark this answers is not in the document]
eode: done

compiler/quilt/test/test_cli.py
```diff
@@ -731,6 +733,44 @@ def test_cli_command_push(self):
...
+            ])
+
+        ## This section tests for circumstances expected to be rejected by argparse.
+        expect_fail_2_args = [
+            'export'.split(),
+            'export too many args'.split(),
+            ]
+        for args in expect_fail_2_args:
+            assert self.execute(args)['return code'] == 2
+
```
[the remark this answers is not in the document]
eode: *nod* this test only tests the cli structure, and that the correct function is called, and that it matches the function signature for that command function.  The functionality of the export is tested in `test_command.py`, so I added this there.

compiler/quilt/test/test_import.py
eode: Quilt base package can't be imported relatively from a module inside that package.  ..however, there's no need to use the base directly here.  File converted to use relative imports only.

There are a few other test files in general that could use that treatment, but that's probably a PR in itself, and not directly related.  Fixed since we touched this file.

compiler/quilt/test/test_import.py
eode: On cursory inspection, `quilt.load` appears to avoid those issues, yes.

However, this also means that calling `quilt.load()` twice means you get different (but identical) objects, and data isn't persisted unless you save the object using `quilt.build()`.

eode:
Closing, as issues from this PR have been addressed, and code from this will be extracted to a newer, cleaner branch.  I'll open a PR to track progress on that.

PR #284, opened 2018-01-10
Fix `quilt ls`

akarve:
Problem: `os.path.listdir` was being used to enumerate directories and OS file like `.DS_Store` were causing `quilt ls` to vomit.

Solution: Add utility functions for `sub_dirs` and `sub_files`.

### DIFF
--- a/compiler/quilt/test/test_util.py
+++ b/compiler/quilt/test/test_util.py
@@ -1,0 +1,46 @@
+"""
+Tests for quilt.tools.util
+"""
+
+import os
+
+from .utils import QuiltTestCase
+from ..tools.util import children
+
+class UtilTest(QuiltTestCase):
+    def test_children(self):
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, './listdir')
+        everything = os.listdir(path)
+
+        allvisible = children(path, dirs=True, files=True)
+        # ensure invisible files are not returned by default
+
+        for f in allvisible:
+            assert not f.startswith('.'), 'Expected only visible files'
+
+        visiblefiles = children(path, dirs=False, files=True, noinvisible=True)
+        for f in visiblefiles:
+            assert os.path.isfile(os.path.join(path, f)), 'Expected only files'
+            assert not f.startswith('.'), 'Expected only visible files'
+
+        visiblefiles = children(path, dirs=False, files=True, noinvisible=True)
+        for f in visiblefiles:
+            assert os.path.isfile(os.path.join(path, f)), 'Expected only files'
+            assert not f.startswith('.'), 'Expected only visible files'
+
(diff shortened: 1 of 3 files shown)


compiler/quilt/tools/util.py
```diff
@@ -81,3 +81,20 @@ def gzip_compress(data):
...
+def children(path, dirs=True, files=True, noinvisible=True):
+    """
+    Return children of a given path
+    dirs=True => include dirs
+    files=True => include files
+    invisible=True => include files that do begin with .
+    """
+    matches = os.listdir(path)
+    if noinvisible:
+        matches = filter(lambda x: not x.startswith('.'), matches)
```
dimaryaz: Actually, list comprehension - `[x for x in ... if ...]` - is preferred over `filter` and `lambda`.

compiler/quilt/tools/util.py
```diff
@@ -81,3 +81,20 @@ def gzip_compress(data):
     with gzip.GzipFile(fileobj=buf, mode='wb') as fd:
         fd.write(data)
     return buf.getvalue()
+
+def children(path, dirs=True, files=True, noinvisible=True):
```
dimaryaz: Strictly speaking, this isn't 100% what we want. You can't say, "only give me directories" - only "don't give me files". So if the user is crazy enough to create devices, sockets, pipes, broken symlinks, etc., things will still break. But maybe it doesn't matter.

dimaryaz: Something like `(isdir(...) and dirs) or (isfile(...) and files)` should do it.

compiler/quilt/test/test_util.py
```diff
@@ -0,0 +1,31 @@
...
+from .utils import QuiltTestCase
+from ..tools.util import sub_dirs, sub_files
+
+class UtilTest(QuiltTestCase):
+    def test_sub_files(self):
+        mydir = os.path.dirname(__file__)
+        path = os.path.join(mydir, './listdir')
+
+        for f in sub_files(path):
+            assert os.path.isfile(os.path.join(path, f)), 'Expected only files'
```
dimaryaz: One more nitpick: this doesn't really test anything cause it's literally the implementation of the function it's testing. E.g., imagine that `isfile` returns `True` for both regular files and directories; the test won't catch that. What you want is just a hardcoded list, like `assert sub_files(path) == ['foo.txt']`, etc.

akarve: @eode can you handle Dima's feedback above; he's right. the files are in `listdir`.

eode: @dimaryaz -- this more to your liking?

eode: Yup, was on it.

Issue #291, opened 2018-01-18
Undeclared Pandas dependency (?) broke any usage of the CLI


JasonBenn:
Using pip-compile to generate my requirements.txt resulted in `pandas==0.19.2` for my project.

Trying to start up quilt's CLI resulted in:
```
(sourceress) sourceress jason/model-info-amount: quilt
Traceback (most recent call last):
  File "/Users/jasonbenn/.pyenv/versions/sourceress/bin/quilt", line 7, in <module>
    from quilt.tools.main import main
  File "/Users/jasonbenn/.pyenv/versions/3.6.3/envs/sourceress/lib/python3.6/site-packages/quilt/__init__.py", line 83, in <module>
    from .tools.command import (
  File "/Users/jasonbenn/.pyenv/versions/3.6.3/envs/sourceress/lib/python3.6/site-packages/quilt/tools/command.py", line 33, in <module>
    from .build import (build_package, build_package_from_contents, generate_build_file,
  File "/Users/jasonbenn/.pyenv/versions/3.6.3/envs/sourceress/lib/python3.6/site-packages/quilt/tools/build.py", line 11, in <module>
    from pandas.errors import ParserError
ModuleNotFoundError: No module named 'pandas.errors'
```
This `pandas.errors` module exists in 0.22.0 but not 0.19.2.

Where is this project's setup.py or requirements.txt?

akarve:
Hi Jason. The `quilt` client's `setup.py` is here https://github.com/quiltdata/quilt/blob/master/compiler/setup.py

Quilt currently depends on `>=0.19.2`. I see that we do depend on pandas.errors (e.g. in `build.py`), so we'll need to bump that dependency that in the next release.

In the meantime if we can help you get back up and running, we can chat on [gitter](https://gitter.im/quilt-data/Lobby).

JasonBenn:
Thanks -- I fixed it by manually pinning pandas to a higher version.

akarve:
Fixed in #295

PR #293, opened 2018-01-18
Bump pandas dependency

akarve:
Avoid failed imports of pandas.errors per #291


akarve:
Might not be needed given #295

PR #296, opened 2018-01-19
Package list pagination

nl0:
Adds pagination to the package list. WIP.

### TODO:
* [x] styling tweaks + use proper colors and sizes from the design system or smth
* [x] integrate pagination into profile view
* [x] intl
* [x] unit tests

### Open questions:
* [x] is UX okayish? (wording, results-per-page selection)
* [x] where to store pagination state? now the component is a "controlled component" with its state in redux, but maybe it makes sense to use local state instead (it will make it easier to use)
  * locally
* [x] should we make the PackageList paginated by default?
  * yes
* [x] do we need to link url params to the pagination state?
  * no

cc @asah


asah:
@akarve can you take a look?  thanks!

akarve:
@nl0 Hi. Thanks for putting this together. Would you `npm run lint`, fix any errors you see, and submit the results? I will have more detailed feedback later.

nl0:
ok, this one looks ready for review. also @akarve could you please address the open questions?

nl0:
btw I was unable to test the profile view, because after the rebase the app required to run a local registry. I've started it up via docker-compose, but was unable to authenticate. I'll give it another try tomorrow.

nl0:
ok, i've managed to set up the local env properly and test things -- everything looks fine.

catalog/app/components/Pagination/index.js
```diff
@@ -0,0 +1,181 @@
...
+import IconButton from 'material-ui/IconButton';
+import { grey500 } from 'material-ui/styles/colors';
+
+import MIcon from 'components/MIcon';
+
+import { DEFAULT_PER_PAGE, DEFAULT_PER_PAGE_OPTIONS } from './constants';
+import messages from './messages';
+
+
+const UNIT_SIZE = '24px';
```
akarve: minor, why two lines above and below this const?

catalog/app/components/Pagination/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+
+const Controls = styled.div`
+  display: flex;
+  height: ${UNIT_SIZE};
+  justify-content: space-between;
+  margin-top: 16px;
+`;
+
+const Chevron = compose(
```
akarve: curious if `recompose` affects webpack size of JS? you can `npm run build` to see how much bigger it makes the webpack fragments.

catalog/app/components/Pagination/index.js
```diff
@@ -0,0 +1,181 @@
...
+  nextPage,
+  prevPage,
+  perPageOptions,
+  perPage,
+  handlePerPageChange,
+  children,
+}) => (
+  <div>
+    {children({ items: pageItems })}
+    {!!items.length &&
```
akarve: what is intended here? ternary might be clearer

catalog/app/containers/Profile/index.js
```diff
@@ -83,8 +84,8 @@ export class Profile extends React.PureComponent { // eslint-disable-line react/
 
   onDowngrade = () => {
     const { selectedPlan } = this.state;
-    const { dispatch } = this.props;
-    dispatch(updatePlan(selectedPlan));
+    const { updatePlan } = this.props;
+    updatePlan(selectedPlan);
```
akarve: interesting. how does this work without `dispatch`?

catalog/app/containers/SearchResults/index.js
```diff
@@ -29,10 +35,18 @@ export class SearchResults extends React.PureComponent { // eslint-disable-line
...
-          emptyMessage={<FormattedMessage {...messages.empty} />}
-          packages={response.packages}
-        />
+        <Pagination
+          items={response.packages}
+          perPage={perPage}
+          page={page}
+          onPageChange={setPage}
+          onPerPageChange={setPerPage}
+        >{({ items }) =>
```
akarve: hard to read as written; brackets should be on their own line and same for arrow expression args

catalog/app/components/Pagination/index.js
nl0: I'm used to separate major parts of file (imports, constants, logical groups of definitions, classes) with double newlines. I can remove the extra newlines if it's preferred.

catalog/app/components/Pagination/index.js
nl0: recompose already was in the bundle because it's MUI's dependency (that's why i pinned it to the specific (a little outdated) version).

catalog/app/components/Pagination/index.js
nl0: The intention is to only show the controls if items length is non-zero. `!!` is a shorthand for boolean conversion (roughly equivalent to `Boolean(items.length)`). If we omit it, literal `0` is rendered when the length is zero, but if we use it, react renders nothing as intended. Ternary would be a little more noisy, something like this:

```javascript
{items.length === 0 ? null :
  ...
}
```

catalog/app/containers/Profile/index.js
nl0: https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options

> If an object is passed, each function inside it is assumed to be a Redux action creator. An object with the same function names, but with every action creator wrapped into a dispatch call so they may be invoked directly, will be merged into the component’s props.

catalog/app/containers/SearchResults/index.js
nl0: adjusted

nl0:
@akarve 

> Why doesn't IconButton with chevrons highlight, etc. on hover and on click

They do highlight (kinda) on click, but not on hover -- it's default MUI styling / behaviour. I can tweak it to highlight on hover if required.

nl0:
@akarve i've adjusted the code and rebased the branch

catalog/app/containers/Profile/index.js
nl0: btw since i don't use redux stuff for the pagination anymore, should i revert the changes related to `mapDispatchToProps`?

nl0:
> ideally we have some very simply unit tests for the paging state machine

ok, I'll do that

catalog/app/containers/Profile/index.js
akarve: yes

akarve:
@nl0 no need to customize IconButton behavior

nl0:
@akarve i've added some basic tests and i'm going to add some more

nl0:
@akarve so, the story with the tests is quite sad and complicated and boils down to this: we need to upgrade our dependencies, because most of them are quite outdated and don't support features we need and also aren't compatible with other tools. I've implemented basic snapshot testing to safeguard from unintentional changes, but that's it for now -- I was not able to test the behaviour. I propose to postpone this for after we upgrade the dependencies (there are two cards for this in asana).

akarve:
* Yes to `PackageList` pagination by default.
* No to URL params for paging state

akarve:
OK to postpone unit testing until we updated dependencies

nl0:
@akarve done

PR #300, opened 2018-01-20
Add installation tests for team

Kurlov:
Is that right to add team based package as dependency?


Kurlov:
CI failed, need to fix errors.

PR #301, opened 2018-01-21
Fix yml team based packages and add tests

Kurlov:
`parse_package_extended` might be confused by adding team based packages to .yml. This should fix it.
Added tests for team installation and parse_package_extended helper

### DIFF
--- a/compiler/quilt/test/test_install.py
+++ b/compiler/quilt/test/test_install.py
@@ -295,7 +295,7 @@ packages:
             command.install("packages:\n")
         with assertRaisesRegex(self, command.CommandException, "Specify package as"):
             command.install("packages:\n- foo")
-        with assertRaisesRegex(self, command.CommandException, "Invalid versioninfo"):
+        with assertRaisesRegex(self, command.CommandException, "Specify package as"):
             command.install("packages:\n- foo/bar:xxx:bar")
         with assertRaisesRegex(self, Exception, "No such file or directory"):
             self.validate_file('foo', 'bar', contents_hash1, contents1, table_hash1, table_data1)
--- a/compiler/quilt/test/test_install.py
+++ b/compiler/quilt/test/test_install.py
@@ -78,8 +78,7 @@ class InstallTest(QuiltTestCase):
         command.install('foo/bar')
         teststore = PackageStore(self._store_dir)
 
-        with open(os.path.join(teststore.package_path(None, 'foo', 'bar'),
-                               Package.CONTENTS_DIR,
+        with open(os.path.join(teststore.package_path(None, 'foo', 'bar'), Package.CONTENTS_DIR,
                                contents_hash)) as fd:
             file_contents = json.load(fd, object_hook=decode_node)
             assert file_contents == contents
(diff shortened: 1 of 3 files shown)


compiler/quilt/test/test_install.py
```diff
@@ -95,10 +95,12 @@ def test_install_team_latest(self):
         """
         table_data, table_hash = self.make_table_data()
         file_data, file_hash = self.make_file_data()
-        contents, contents_hash = self.make_contents(table=table_hash, file=file_hash)
```
dimaryaz: Actually, these are fine; we're using 110 as the max line length. (it's also set in the .pylintrc)

compiler/quilt/test/test_install.py
```diff
@@ -296,6 +335,18 @@ def test_quilt_yml_unknown_version(self):
         with assertRaisesRegex(self, command.CommandException, "Version 99.99 does not exist"):
             command.install("packages:\n- akarve/sales:v:99.99")
 
+    def test_quilt_yml_unknown_team(self):
+        table_data1, table_hash1 = self.make_table_data('table1')
+        contents1, contents_hash1 = self.make_contents(table1=table_hash1)
+        self._mock_tag('baz/bat', 'latest', contents_hash1, team='unknown')
+        self._mock_package('baz/bat', contents_hash1, '', contents1,
```
dimaryaz: Actually, you don't get a 404 if the team doesn't exist. Different teams are hosted on physically different servers; e.g., "unknown" would use the URL "https://unknown-registry.team.quiltdata.com". In an ideal world, this would result in a DNS error. In our case, it actually points to the wrong server and results in an SSL error.

So I'd suggest either simulating what's actually happening now - make the tag request raise an `SSLError` (I think `responses` library, used by `_mock_tag`, allows raising exceptions.), or just skip this test for now, until we fix the servers.

compiler/quilt/tools/store.py
```diff
@@ -314,10 +314,20 @@ def prune(self, objs=None):
...
 def parse_package_extended(name):
     hash = version = tag = None
+    package = name
     try:
         if ':' in name:
-            name, versioninfo = name.split(':', 1)
-            if ':' in versioninfo:
+            package, versioninfo = name.split(':', 1)
+            # in case name has team, versioninfo contains packagename itself
+            if '/' in versioninfo:
```
dimaryaz: I think it's best to use a regex at this point. Multiple levels of `if`s is getting pretty confusing.

Kurlov: Agree, but there is an error raise in case it confused by parsing tag/version/hash. Thus, I should implement two regexs, one for parsing [team:]owner/package, second for the rest tag/version/hash or ignore it and make only one re and thus one raise exception?

compiler/quilt/test/test_install.py
Kurlov: Got it, thanks!

Kurlov:
@dimaryaz can you please review this again?

compiler/quilt/tools/store.py
```diff
@@ -13,6 +13,7 @@
 
 # start with alpha (_ may clobber attrs), continue with alphanumeric or _
 VALID_NAME_RE = re.compile(r'^[a-zA-Z]\w*$')
+VALID_EXTENDED_PCKG_RE = re.compile(r'[\w+:]?\w+/\w+:(version|tag|hash|v|t|h):(.+)', flags=re.IGNORECASE)
```
dimaryaz: The `[\w+:]` doesn't seem right; did you mean `(\w+:)`?

compiler/quilt/tools/store.py
```diff
@@ -314,31 +315,29 @@ def prune(self, objs=None):
...
-                    elif 'hash'.startswith(info[0]):
-                        # usr/pkg:h:<hash>  usr/pkg:hash:<hash>  etc
-                        hash = info[1]
-                    else:
-                        raise CommandException("Invalid versioninfo: %s." % info)
-                else:
-                    # usr/pkg:hashval
-                    hash = versioninfo
-        team, owner, pkg, subpath = parse_package(name, allow_subpath=True)
+        packinfo = re.findall(VALID_EXTENDED_PCKG_RE, name)
```
dimaryaz: Why `findall`, not `match`? You only want one result.

compiler/quilt/tools/store.py
dimaryaz: Also, probably need `^` and `$`

compiler/quilt/tools/store.py
```diff
@@ -314,31 +315,29 @@ def prune(self, objs=None):
...
+            info = packinfo[0][1]
+            if info_type[0] == 'v':
+                version = info
+            elif info_type[0] == 'h':
+                hash = info
+            else:
+                tag = info
+            # update package to send it for futher parsing without versioninfo
+            offset = len(info_type) + len(info) + 2    # two `:`
+            package = name[:-offset]
```
dimaryaz: Just use parenthesis in the regex and extract it as a group

compiler/quilt/tools/store.py
Kurlov: I just didn't want to extract team, but I  guess it's completely fine, just ignore this group.

Kurlov: I just didn't want to extract team, but I guess it is fine to ignore it.

Kurlov:
Not relevant anymore. #309

PR #302, opened 2018-01-22
TakeAction: hide sign up button if team is set

nl0:
The simplest solution.
I could've kept `TakeAction` dumb by adding a prop to hide the signup button and leaving the decision making to the parent component -- it may make things a little more explicit and future-proof, but requires a little more code.

### DIFF
--- a/catalog/app/components/Feature/index.js
+++ b/catalog/app/components/Feature/index.js
@@ -12,7 +12,7 @@ import strings from './messages';
 const xs = `${breaks.sm - 1}px`;
 
 /* eslint-disable jsx-a11y/no-static-element-interactions */
-const Feature = ({ header, tagline }) => (
+const Feature = ({ header, tagline, signUp }) => (
   <ImageRow backgroundColor="#F6B500" src={background}>
     <Content>
       <h1 className="main">
@@ -21,7 +21,7 @@ const Feature = ({ header, tagline }) => (
       <h2 className="main">
         { tagline }
       </h2>
-      <TakeAction />
+      <TakeAction signUp={signUp} />
       <div className="framer">
         <iframe
           src="https://ghbtns.com/github-btn.html?user=quiltdata&repo=quilt&type=star&count=true&size=large"
@@ -46,6 +46,7 @@ Feature.defaultProps = {
 Feature.propTypes = {
   header: PropTypes.string,
   tagline: PropTypes.string,
+  signUp: PropTypes.bool,
 };
 
 const Content = styled.div`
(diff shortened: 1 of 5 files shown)


nl0:
@akarve @asah review pls

catalog/app/components/TakeAction/index.js
```diff
@@ -20,7 +21,9 @@ const Container = styled.div`
 function TakeAction() {
   return (
     <Container>
-      <RaisedButton href={makeSignInURL()} label="Sign Up" primary />
+      {!cfg.team &&
```
akarve: I recommend  something like this as it's more readable:
```
{ cfg.team ? null : <RaisedButton ... }
```

if the statement gets much longer than it's better (for future PRs) to compute the fragment above the return value and just inline it:
```
const optionalButton = ...
return (
  { optionalButton }
);
```

akarve: Also, per your comment, it is indeed preferable to keep `TakeAction` as a dumb component and pass in a boolean to determine whether to show or hide.

catalog/app/components/Feature/index.js
```diff
@@ -94,6 +95,7 @@ const Content = styled.div`
     h2.main {
       font-size: 1.5em;
     }
+  }
```
nl0: seems like `@media` was not closed properly

catalog/app/components/PaymentDialog/index.js
```diff
@@ -112,7 +112,7 @@ class PaymentDialog extends React.PureComponent { // eslint-disable-line react/p
             />
             <br />
             <br />
-            <Pricing includeSignIn={false} />
+            <Pricing takeAction={false} />
```
nl0: there is no `includeSignIn` prop on the `Pricing` component, so I've introduced `takeAction` prop (`true` by default) to be able to completely hide the `<TakeAction />` there

nl0:
@akarve i've refactored things a little, so review again pls

nl0:
rebased

nl0:
@akarve this one seems quite trivial and safe to merge. review pls

akarve:
Has a linting error:
```
/Users/karve/code/quilt/catalog/app/components/TakeAction/index.js
  32:11  error  'signUp' PropType is defined but prop is never used  react/no-unused-prop-types
```

Also causes stage (non-teams instance) to bomb (attached).

I first merged master... which might have broken things.

screen shot 2018-02-05 at 11 02 50 am

nl0:
@akarve i'll take a look

nl0:
@akarve my bad. Fixed and rebased.

nl0:
@akarve btw this change is not compatible with #326 due to use of PropTypes and maybe lint config changes, so when one is merged, the other one should be adjusted before merging

PR #306, opened 2018-01-23
Unbreak download retries

dimaryaz:
:(


kevinemoore:
How did that break?

dimaryaz:
I messed it up while implementing parallel downloads (3b0b5d38)

compiler/quilt/tools/command.py
```diff
@@ -1026,7 +1026,7 @@ def _worker_thread():
                                 else:
                                     with lock:
                                         tqdm.write("Download failed for %s: %s" % (obj_hash, ex))
-                                break
+                                    break
```
asah: ROFL re "un-break" that's hilarious.

dimaryaz: Damn, that wasn't even intentional... Didn't realize it until you pointed it out

PR #310, opened 2018-01-26
Team crud more tests

Kurlov:
- Added more CRUD tests
- Tune views.py a little to ensure proper status_codes


compiler/quilt/test/test_command.py
```diff
@@ -392,7 +400,36 @@ def test_user_create_no_auth(self):
         with self.assertRaises(command.CommandException):
             command.create_user('bob', 'bob@quitdata.io')
         pass
-    
+
+    def test_user_create_duplicate(self):
```
asah: team create dup?

compiler/quilt/test/test_command.py
```diff
@@ -411,14 +448,40 @@ def test_user_disable_no_auth(self):
...
+    def test_user_disable_empty(self):
+        self._mock_method('disable', status=400, message="Username is not valid")
+        with assertRaisesRegex(self, command.CommandException, "Username is not valid"):
+            command.disable_user('')
+
+        self._mock_method('disable', status=400, team='qux', message="Username is not valid")
+        with assertRaisesRegex(self, command.CommandException, "Username is not valid"):
+            command.disable_user('', team='qux')
+
+    def test_user_disable_unknown(self):
```
asah: pls also test disable a user from a non-existent team
pls also test disable a non-existent user
pls also test disable an already-disabled user
pls also test disable an already-deleted user

at the top of the file, please create a table (in text!) of these cross-product test cases, and then make sure you've got them all covered.

registry/quilt_server/views.py
```diff
@@ -1645,8 +1645,15 @@ def disable_user():
 
     data = request.get_json()
     username = data.get('username')
+    user_regex = re.compile(r"^[^\d\W]\w*\Z", re.UNICODE)
```
asah: post-commit, pls ping @dimaryaz to see how he likes to test things like this, where the regexp could be shared with the core or not:
- if it's shared with the core, then cleaner reuse
- if it's not shared with the core, then we (in theory) also test the regexp

again, IMHO pls commit/merge first then followup about this detail.

registry/quilt_server/views.py
```diff
@@ -1674,6 +1681,13 @@ def delete_user():
...
```
asah: pls share regexp within this file - see above about sharing with the core...

registry/quilt_server/views.py
Kurlov: Ok, got it

compiler/quilt/test/test_command.py
```diff
@@ -19,6 +61,14 @@
 from .utils import QuiltTestCase, patch
 
 class CommandTest(QuiltTestCase):
+    def _mock_method(self, endpoint, status=201, team=None, message="", method=responses.POST, api="users"):
```
dimaryaz: It seems like it's only useful for error responses, since anything else would return more than just `message=...`. I'd call it `_mock_error`. Other things:
- We don't really have an "api" vs "endpoint"; make it just a single "endpoint" param. (Also, "users" isn't good as a default cause it's just one of the many APIs.)
- Make the status default to some error code - or just make it required.

registry/quilt_server/const.py
```diff
@@ -9,7 +9,8 @@
 
 PUBLIC = 'public' # This username is blocked by Quilt signup
 VALID_NAME_RE = re.compile(r'^[a-zA-Z]\w*$')
-EMAILREGEX = re.compile(r'^([^\s@]+)@([^\s@]+)$')
+VALID_USERNAME_RE = re.compile(r"^[^\d\W]\w*\Z", re.UNICODE)
```
dimaryaz: We should be using `VALID_NAME_RE` for usernames. If we currently allow them to start with `_`, then that's a bug.

registry/quilt_server/views.py
```diff
@@ -1603,7 +1606,7 @@ def create_user():
     if resp.status_code == requests.codes.not_found:
         raise ApiException(
             requests.codes.not_found,
-            "Cannot list users"
+            "Cannot create user"
```
dimaryaz: Haha, nice, I think I've run into this earlier.

PR #311, opened 2018-01-26
Add a `flask createdb` command.

dimaryaz:
None of the built-in `flask db` commands support creating the database - but it's nice to have when setting up ECS.

Also, switch from "postgres://" to "postgresql://". This works around a bug in sqlalchemy_utils - but also matches the SQLAlchemy docs.


eode:
This is a minor change, passes tests, and looks clean.  IMO it's mergeable.

PR #312, opened 2018-01-26
`quilt export` command

eode:
This is the `quilt export` command.

It has been pulled out of the `tensorflow_virtual_files` branch, as it is a useful and distinct feature.  There are, however, some supporting tools here that are used there as well, so if something looks excessive, there's a decent chance it's used by that branch as well.

Overview of changes, with usage by export and tensorflow marked:
* add `quilt export` / `command.export()` command  `(tf)`
* add `command.load()` to parse package name and load a fresh copy of a node from disk `(tf, export)`
* add `util.is_identifier()` and `util.is_nodename()` for a centralized place for name validation `(export)`
* `GroupNode` now has `_iteritems` and `_iterpaths` methods that allow recursive iteration of node paths `(export)`
* Existing instances of `VALID_NAME_RE.match(<name>)` changed to `is_nodename(<name>)` `(export)`
* `PackageNode._set()` expanded to include metadata and to accept filenames.  Tests updated.  `(tf, export)`
* `quilt export` cli tests
* `command.export()` tests
* `parse_package_extended` updated to handle teams better
* `parse_package` rejects strings with more than one colon -- e.g., containing version info and team.
* A few minor general bugfixes
* `FileNode` must now include metadata that contains `'q_path'`
* `FileNode` and `DataNode`: if `'q_path'` is present, it must not be absolute
* docs and cli docs updated with usage for `quilt export`


eode:
Currently, the test failure here is re: `parse_package_extended`.

There are two copies, a regex version and a code version.  Any preference on which to take?  The regex version will need to be modified to return the same data as the original, but it's a trivial change.

eode:
The two versions of parse_package_extended return:

```
regex: (<str: team, package, subdirs>, <str or None: version>, <str or None: tag>, <str or None: hash>)
regex example: ("team:user/pkg/sub/node", None, "afc39182cbc9532533", None)
original: [<str: team>, <str: package>, <list(str): subdirs>, <str or None: version>, <str or None: tag>, <str or None: hash>)
original example: ["team", "user", "pkg", ['sub', 'node'], None, "afc39182cbc9532533", None]
```

I ended up going with the new return value, as it's easier to split the resultant package specifier than it is to join the (possible team) + user + package + (possible subdirs).

compiler/quilt/tools/compat.py
```diff
@@ -0,0 +1,63 @@
...
+# Manual backport of a stdlib tool -- avoid extra backport dependency.
+try:
+    from tempfile import TemporaryDirectory
+except ImportError:
+    import tempfile as _tempfile
+    import contextlib as _contextlib
+    from shutil import rmtree as _rmtree
+
+    @_contextlib.contextmanager
+    def TemporaryDirectory(suffix='', prefix='tmp', dir=None):
```
dimaryaz: This seems like overkill; same for chmod_context. They're only used in unit tests, right? We already create a temporary directory for each test - so you can create whatever subdirectories you want, chmod them, etc., and not worry about cleanup.

compiler/quilt/tools/util.py
```diff
@@ -101,3 +106,31 @@ def sub_files(path, invisible=False):
...
+
+
+def is_identifier(string):
+    """Check if string could be a valid python identifier
+
+    :param string: string to be tested
+    :returns: True if string can be a python identifier, False otherwise
+    :rtype: bool
+    """
+    val = re.match(r'^[a-zA-Z_]\w*$', string) and not keyword.iskeyword(string)
```
dimaryaz: Make the regex a global, like the old one - just to avoid recompiling it unnecessarily

compiler/quilt/tools/util.py
```diff
@@ -101,3 +106,31 @@ def sub_files(path, invisible=False):
...
+    return True if val else False
```
dimaryaz: `bool(val)`... or use `bool` or `is not None` or whatever above, so `val` is automatically the right type.

compiler/quilt/tools/core.py
```diff
@@ -73,6 +74,7 @@ def __init__(self, hashes, format, metadata=None):
         assert isinstance(hashes, list)
         assert isinstance(format, string_types), '%r' % format
         assert isinstance(metadata, dict)
+        assert not os.path.isabs(metadata.get('q_path') or '')
```
dimaryaz: `metadata.get('q_path', '')`

registry/quilt_server/core.py
```diff
@@ -86,12 +88,14 @@ def __json__(self):
...
-        if metadata is None:
-            metadata = {}
-
+    def __init__(self, hashes, metadata):
         assert isinstance(hashes, list)
         assert isinstance(metadata, dict)
+        assert metadata.get('q_path')
+        assert not os.path.isabs(metadata['q_path'])
+
+        if not metadata.get('q_path'):
```
dimaryaz: Won't happen cause you already have an assert above

compiler/quilt/tools/compat.py
eode: Ah, ok.  I was confused because `__file__` is still the specific file in the working tree, so if you do:
```
mydir = os.path.abspath(__file__)
<do stuff with mydir>
```
..it affects the working tree.

Re: chmod_context:  Sounds good.

compiler/quilt/tools/core.py
eode: Intentional -- this also handles the case where someone sets `'q_path'` to None, but `metadata.get('q_path', '')` doesn't.

compiler/quilt/tools/util.py
eode: It's cached internally by python, and doesn't get recompiled.  But I can still make it global to avoid the cache lookup.

compiler/quilt/tools/util.py
eode: Check.

registry/quilt_server/core.py
eode: *nod* assert added afterward, and I didn't remove that.  Done.

eode: I'll post these changes tomorrow -- it's late here.

PR #313, opened 2018-01-26
Redirect unauthenticated users to Sign In from all team catalog routes

nl0:
WIP. This PR depends on some changes from #296 and so shouldn't be merged before it.

TODO:

* [x] make sure signing out works properly
* [x] make sure profile didn't break

### DIFF
--- a/catalog/app/components/AuthBar/index.js
+++ b/catalog/app/components/AuthBar/index.js
@@ -50,10 +50,9 @@ export class AuthBar extends React.PureComponent { // eslint-disable-line react/
   }
 
   render() {
-    const { bye, error, name, searchText, signedIn, waiting } = this.props;
+    const { error, name, searchText, signedIn, waiting } = this.props;
     const menu = (
       <UserMenu
-        bye={bye}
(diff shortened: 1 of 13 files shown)


nl0:
@akarve this one is ready for review, however, it may not work out-of-the box in dev environment, because of redirect url mismatch (`catalog:3000` vs `localhost:3000`): the urls are hardcoded in `registry/quilt_server/docker_config.py`. In order to make things work I had to create a custom config for the server, mount it via `volumes` in docker compose file and then use it by passing its name as an env var (`QUILT_SERVER_CONFIG=local_config/config.py`). This line in `docker_config.py` causes troubles: `CATALOG_URL = 'http://localhost:3000'`.

nl0:
also, `static/config.js` needs some tweaking: if `catalog:3000` url used, condition on L11 doesn't work and the app tries to use `https://pkg-stage.quiltdata.com` instead of local instance, so i had to change it to `http://flask:5000`

nl0:
rebased. @akarve ready for review

PR #317, opened 2018-01-26
Test the cli --version command

eode:
trivial, but needed doing.

Also, sorted the imports.

### DIFF
--- a/compiler/quilt/test/test_cli.py
+++ b/compiler/quilt/test/test_cli.py
@@ -807,7 +807,7 @@ def test_cli_command_version_flag(capsys):
     # in python 2, apparently argparse's 'version' handler prints to stderr.
     result = outerr.err if PY2 else outerr.out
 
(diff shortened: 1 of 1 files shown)


eode:
Huh.  Looks like we found a bug.. ..or at least, difference from all other argparse behavior, in argparse for Python 2.

```
python3 foo.py --version   # prints to stdout
python2 foo.py --version   # prints to stderr
```

..doesn't appear configurable, so I worked around it in the test.

asah:
whoa crazy.  stdout is correct fyi.

But this seems weird - docs say stdout by default:
https://github.com/quiltdata/quilt/blob/master/compiler/quilt/tools/main.py#L43
https://docs.python.org/2/library/argparse.html#printing-help
https://docs.python.org/3/library/argparse.html#printing-help

eode:
Yeah, I know.  I wonder if I should file a bug..

PR #318, opened 2018-01-26
Log push/install/preview/delete events

dimaryaz:
Also add admin-only endpoints for viewing the events


dimaryaz:
(This is a subset of `team_crud`; trying to simplify code reviews a bit.)

PR #320, opened 2018-01-27
Add docs for the shared local package store feature

kevinemoore:


docs/shared-store.md
```diff
@@ -0,0 +1,41 @@
...
+
+Groups that share packages across users can save storage and network traffic by installing packages from a shared package directory (e.g. on a network file server).
+
+## Create & populate shared package directory
+
+1. Create a `quilt_packages` on the shared file system. Grant read access to all intended recipients. 
+
+1. Set the `QUILT_PRIMARY_PACKAGE_DIR` to the path for `quilt_packages` in step 1. 
+    ```bash
+    export QUILT_PRIMARY_PACKAGE_DIR=/SHARE_PATH/quilt_packages
```
asah: As always, ideal if truly cut/paste.  Also, "SHARE_PATH" looks like a real path... :-(

Here's what I recommend:

```bash
# pedantic: pick some place that you control and is durable, e.g. not /tmp
export MY_SHARE_PATH=$HOME/shared_quilt_packages
# the actual envariable variable is named QUILT_PRIMARY_PACKAGE_DIR but
# in later examples, we might want to use this path in other ways.
export QUILT_PRIMARY_PACKAGE_DIR=$MY_SHARE_PATH
mkdir -p $MY_SHARE_PATH
chmod o+r $MY_SHARE_PATH
...
quilt build ...
quilt install ...
```

docs/shared-store.md
```diff
@@ -0,0 +1,41 @@
...
+1. Set read permissions on shared directory and sub-directories
+    ```bash
+    chgrp -R sharers "$QUILT_PRIMARY_PACKAGE_DIR"
+    chmod -R g+rx "$QUILT_PRIMARY_PACKAGE_DIR"
+    ```
+
+## Configure clients to read from shared directory 
+1. Each reader should set the following environment variable:
+    ```bash
+    export QUILT_PACKAGE_DIRS=/SHARE_PATH/quilt_packages
```
asah:
```bash
# note that other users might not have SHARE_PATH defined, e.g. either define this in their login/startup scripts (e.g. .bashrc) or hard-code for them... in the same startup/login scripts!
export QUILT_PACKAGE_DIRS=$SHARE_PATH
```

docs/shared-store.md
```diff
@@ -0,0 +1,41 @@
...
+    ```
+2. Readers can can import shared packages as follows
+    example:
+    ```python
+    from quilt.data.USERNAME import PACKAGE
+    ```
+
+## Import precedence
+Quilt will first check `QUILT_PRIMARY` (defaults to the local machine) and then check `QUILT_PACKAGE_DIRS` (if available) when importing a package.
+
```
asah: pls provide an example with multiple dirs
pls provide an example where their local quilt_packages is no longer first!
the MY_SHARE_PATH comes in real handy now!

PR #324, opened 2018-02-01
Meta tags

nl0:
So, I've added the title tag for the package page, but seems like that's all I can do now.

1. It doesn't make sense to set `og:` tags via helmet *without server-side rendering*, because:
   - consumers don't seem to support dynamic meta tag updates (confirmed for facebook and slack, not confirmed for twitter);

   - there's an issue with duplicate / conflicting tags (`react-helmet` doesn't replace tags hardcoded in `index.html`).

2. I don't see suitable data for `description` and `keywords`.

@akarve @asah please review


akarve:
Cool, please add two more tags. We'll improve description once we have access to the README.md contents via API.

```
<meta name="description" content="Quilt data package">
<meta name="author" content={username}>
```

nl0:
@akarve done

PR #327, opened 2018-02-02
Disable user endpoint

meffij:


meffij:
Changing this to ENABLE, which defaults to False

PR #328, opened 2018-02-02
Add admin list package endpoint

meffij:


registry/quilt_server/views.py
```diff
@@ -853,6 +853,35 @@ def user_packages(owner):
...
+    return dict(
+        packages=[
+            dict(
+                name=package.name,
+                is_public=is_public
+            )
+            for package, is_public in packages
+        ]
+    )
+
```
kevinemoore: This looks like it should have a basic unit test.

meffij: done

PR #331, opened 2018-02-02
Make team parameter required for state-altering apis

meffij:


compiler/quilt/tools/command.py
```diff
@@ -1370,6 +1370,9 @@ def delete_user(username, force=False, team=None):
     url = get_registry_url(team)
     resp = session.post('%s/api/users/delete' % url, data=json.dumps({'username':username}))
 
+def delete_user_cli(username, force=False):
```
dimaryaz: Why is this only needed for delete - and not create or disable?

Also, by requiring the team, I meant, it needs to be specified, not `None` - otherwise, that's kinda the same thing.

meffij: because delete also has force as an optional parameter

compiler/quilt/tools/command.py
```diff
@@ -1339,37 +1339,34 @@ def cli_list_users(team=None):
...
         if confirmed.lower() != 'y':
             return
 
-    if team is None:
-        team = _find_logged_in_team()
     session = _get_session(team)
     url = get_registry_url(team)
     resp = session.post('%s/api/users/delete' % url, data=json.dumps({'username':username}))
 
+def delete_user_cli(team, username, force=False):
```
dimaryaz: I think it's not necessary now?

PR #336, opened 2018-02-04
Tests for multi dirs

Kurlov:


compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
...
+
+        # build package into shared directory
+        new_build_dir = 'qqq/www/%s' % PACKAGE_DIR_NAME
+        with patch.dict(os.environ, {'QUILT_PRIMARY_PACKAGE_DIR': new_build_dir}):
+            command.build('baz/pckg', build_path)
+
+        # override package
+        command.build('baz/pckg', simple_build_path)
+
+        # check that package is overrideed
```
asah: overrided

compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
             from quilt.data.foo import multiple1
             assert multiple1.dataframes
 
+             # Check search order again just to make sure it doesn't return same named empty package
+            from quilt.data.foo import multiple1
+            assert multiple1.dataframes
```
dimaryaz: Python caches imports, so this can't fail because it was already imported above.

compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
...
+
+            teststore = PackageStore(self._store_dir)
+            contents = open(os.path.join(teststore.package_path(None, 'foo', 'multiple1'),
+                                          Package.CONTENTS_DIR,
+                                          multiple1._package.get_hash())).read()
+            assert contents
+
+            # bad imports
+            with self.assertRaises(ImportError):
+                from quilt.data.nonexisting import multiple2
```
dimaryaz: We already have tests for bad imports. I guess one test for an import with `QUILT_PACKAGE_DIRS` may be useful - but don't need to test every possible combination below.

compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
...
+                import quilt.data.bad_user.nonexisting
+            with self.assertRaises(ImportError):
+                from quilt.data.foo.dataframes import blah
+            with self.assertRaises(ImportError):
+                from quilt.data.foo.baz import blah
+
+        # add same dir to the search path
+        dirs = 'foo/%s:%s:%s' % (PACKAGE_DIR_NAME, new_build_dir, new_build_dir)
+        with patch.dict(os.environ, {'QUILT_PACKAGE_DIRS': dirs}):
+            from quilt.data.foo import multiple2
```
dimaryaz: Doesn't do anything because the import is cached

compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
...
+
+        # check that invalid folder names fail
+        bad_build_dir = '?\*;:|<>'
+        with patch.dict(os.environ, {'QUILT_PRIMARY_PACKAGE_DIR': bad_build_dir}):
+            with self.assertRaises(AssertionError):
```
dimaryaz: This is basically testing a misuse of the API. We may consider doing something different if the environment variable is bad - we shouldn't really test assertion errors

compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
...
+                command.build('bar/multiple1', simple_build_path)
+        bad_build_dir = ''
+        with patch.dict(os.environ, {'QUILT_PRIMARY_PACKAGE_DIR': bad_build_dir}):
+            with self.assertRaises(AssertionError):
+                command.build('bar/multiple1', simple_build_path)
+
+    def test_team_multiple_package_dirs(self):
```
dimaryaz: No need to test teams with multiple dirs

compiler/quilt/test/test_import.py
```diff
@@ -198,6 +198,130 @@ def test_multiple_package_dirs(self):
...
+        bad_build_dir = '?\*;:|<>'
+        with patch.dict(os.environ, {'QUILT_PRIMARY_PACKAGE_DIR': bad_build_dir}):
+            with self.assertRaises(AssertionError):
+                command.build('test:bar/multiple1', simple_build_path)
+        bad_build_dir = ''
+        with patch.dict(os.environ, {'QUILT_PRIMARY_PACKAGE_DIR': bad_build_dir}):
+            with self.assertRaises(AssertionError):
+                command.build('test:bar/multiple1', simple_build_path)
+
+    def test_multiple_package_dirs_with_subpackage(self):
```
dimaryaz: Not really sure what subpackage means here - but I think this tests pretty much the same thing as `test_multiple_package_dirs`

compiler/quilt/test/test_import.py
asah: i.e. pls delete unnecessary tests.

compiler/quilt/test/test_import.py
asah: @dimaryaz is it not a legit concern that someone would provide a malformed build dir?  and if so, we get the same error (and not e.g. a stack trace)

compiler/quilt/test/test_import.py
asah: @Kurlov pls remove and add a comment ("no value in testing same dir in search path because of python import caching")

compiler/quilt/test/test_import.py
asah: @Kurlov  pls remove and add a comment ("no value in re-checking the search order because of python import caching")

compiler/quilt/test/test_import.py
dimaryaz: @asah: Not quite; my point was, regardless of what `QUILT_PACKAGE_DIRS` might be or any other setting, this particular import would succeed (even when it shouldn't) because of caching.

That said, testing duplicates in `QUILT_PACKAGE_DIRS` is probably not that useful - I can't really think of a bug that might cause that to fail.

compiler/quilt/test/test_import.py
dimaryaz: Sure. In that case, we should change the code to not raise an `AssertionError` and add a test for the new behavior.

PR #337, opened 2018-02-04
Fix docker-compose registry on Windows

Kurlov:
Currently, build fails on windows (Windows 10 version 1709), but with latest node version it works fine.
By default, docker on windows uses `hyper-V` as provider and there is some bug with older node version. Although, I am pretty sure that current `docker-compose.yml` should work fine with other VM providers on windows.

Important to note that I didn't find the way to run with bare
`docker-compose up`
it raises an error:
`ERROR: Top level object in '.\docker-compose.yml' needs to be an object not '<type 'str'>'.`

Instead, I specified .yml file manually, like this:
`docker-compose -f .\docker-compose-local-auth.yml up --build`
and it works perfectly.

I also check this branch on macos, works fine.

### DIFF
--- a/catalog/Dockerfile
+++ b/catalog/Dockerfile
@@ -7,7 +7,8 @@ ENV LANG=C.UTF-8
 RUN apt-get update && apt-get install -y \
     nginx \
     curl \
-    gettext
+    gettext \
+    libssl-dev
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
 RUN apt-get install -y nodejs
 
--- a/registry/README.md
+++ b/registry/README.md
@@ -32,7 +32,7 @@ Instructions for popular operating systems:
 * Windows
      * [docker](https://docs.docker.com/docker-for-windows/install/)
      * [docker compose](https://docs.docker.com/compose/install/#install-compose)
-     * [ ] Test these instructions on Windows
+     * Make sure you run both PowerShell and Docker for Windows as administrator.
 
 
 ##  2) Build and start the containers


catalog/Dockerfile
```diff
@@ -7,8 +7,9 @@ ENV LANG=C.UTF-8
 RUN apt-get update && apt-get install -y \
     nginx \
     curl \
-    gettext
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+    gettext \
+    libssl-dev
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash -
```
meffij: Why are you bumping node version?

Kurlov: docker-compose fails on windows with 8 node version with this `package.json`

meffij: Isn't this a Dockerfile? Isn't it strange that the node version inside the dockerfile would cause OS-specific errors? Can you post the error you're seeing?

Kurlov: Docker for windows uses `hypre-V` by default.
Actually, 8th version of node cause freeze `npm install` on windows. I assumed that it caused by sim-links issue, because `hyper-V + vagrant + npm` failed for me previously and `npm install --no-bin-links` did the trick. So I tried to change .Dockerfile to install without sim-links, but it effected `npm build` tasks. Since there are quite complicated tasks structure inside `package.json`. After little google this issue, I tried update node version and it fix docker build process. If node version upgrade is not suitable solution, I can try upgrade `npm` only, this might resolve an issue too.

meffij: Could you try just updating npm? I don't want to have two Node versions if we don't need to.

PR #346, opened 2018-02-07
Add require_admin API decorator argument

meffij:


registry/quilt_server/views.py
```diff
@@ -335,6 +338,11 @@ def wrapper(*args, **kwargs):
                     assert user
                     email = data['email']
                     is_admin = data.get('is_staff', False)
+                    if require_admin and not is_admin:
```
dimaryaz: You should move this check further down, to just above `f(*args, **kwargs)`, and use `g.auth.is_admin`:
- Having the check inside of a `try` or an `else` clause makes it harder to prove that it will always run, and there isn't some code path that bypasses it.
- `require_admin` is more sensitive than `require_login`. Any endpoint that requires login probably relies on the user and will fail if there's no user. An admin endpoint, on the other hand, will likely succeed. So it's good to be very explicit.
- Relying on `require_login=True` is fragile - some day, the code might get refactored, etc.

(You could move the `if require_login` check, too, though, for consistency.)

meffij: Good point, fixed

registry/quilt_server/views.py
```diff
@@ -347,6 +349,13 @@ def wrapper(*args, **kwargs):
                         raise ApiException(requests.codes.server_error, "Server error")
                 except (ConnectionError, requests.RequestException) as ex:
                     raise ApiException(requests.codes.server_error, "Server error")
+
+            if require_admin and (not g.auth or not g.auth.is_admin):
```
dimaryaz: Don't need to check for `not g.auth` - it always gets set at the top of the function

PR #348, opened 2018-02-08
Core changes for TF and Export

eode:
These are the changes to `core.py` for Tensorflow support and the export command.

Basically, it has these changes:

* Require `FileNode`s to have `metadata` and `metadata['q_path']` at creation time.
* Require the value of `metadata['q_path']` in `FileNode` and `TableNode` to be relative.
* If a `FileNode` is received from the API, loaded from a database, or loaded from a packagestore, and `metadata['q_path']` is None or blank, set it to the token "::FILENAME_MISSING::".
  * This uses ':', which is illegal in Windows filenames and in Quilt node names.
  * We should probably make any illegal Windows filename characters give warnings on Linux/Mac at build time, but that's a problem for another day.

__Considerations:__

* Should I include a test that verifies (at least for local package store) that blank/empty `q_path` values are replaced?
  * This has been manually verified, and the logic is short
    * Thanks to @dimaryaz for thinking ahead when having the server, client, and API all use the same function to decode JSON.  I really appreciate that kind of coding.

### DIFF
--- a/compiler/quilt/tools/compat.py
+++ b/compiler/quilt/tools/compat.py
@@ -12,11 +12,8 @@ import frequently-used objects here for convenience.
 
 import sys as _sys
 
-# XXX: This originally had a few modules, but I've managed to remove the others.
-#      I'm tentatively still keeping compat.py, otherwise the try: except: block ends up in
-#      each module's import stanzas, instead of 'from compat import pathlib'.
 
 if _sys.version_info < (3, 6):
-    from pathlib2 import path
+    import pathlib2 as pathlib
 else:
-    from pathlib import path
+    import pathlib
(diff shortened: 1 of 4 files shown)


eode:
Do you mean just as far as the core nodes go, or also for other functions that create `FileNodes`, and just allow `FileNode`s in general to not have a path?

kevinemoore:
I meant that users and/or functions should have the option to create FileNodes without paths--or, specifically with the path key missing or set to None. I think it's a good idea to include a path by default, e.g., in quilt generate, but I don't see why we need it to be required.

eode:
Ok, this is now basically a trivial change.  Since we're supporting blank `metadata` in `FileNodes`, it's now not a special-case to have the filename missing, and I've removed the associated code.  I'll adjust the `export` branch to just use the node path as a filename if the filename is missing.

eode:
@kevinemoore Ready for re-review.

PR #354, opened 2018-02-09
Admin UI endpoint to list users and associated data

meffij:
Will commit tests when they are ready.

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -1657,6 +1657,7 @@ def list_users_detailed():
                 if user.get('username') in event_results else 0,
             'previews' : event_results[user.get('username')]['previews']
                 if user.get('username') in event_results else 0,
+            'status' : 'active' if user.get('is_active') == True else 'disabled'
             }
         for user in users.get('results')
     }
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -1630,16 +1630,12 @@ def list_users_detailed():
 
     events = (
         db.session.query(Event.user, Event.type, sa.func.count(Event.type))
-        .filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW]))
         .group_by(Event.user, Event.type)
         )
 
-    event_results = defaultdict(lambda: defaultdict(int))
+    event_results = defaultdict(int)
     for event_user, event_type, event_count in events:
-        if event_type == Event.Type.INSTALL:
-            event_results[event_user]["installs"] = event_count
-        elif event_type == Event.Type.PREVIEW:
-            event_results[event_user]["previews"] = event_count
+        event_results[(event_user, event_type)] = event_count
 
     # replicate code from list_users since endpoints aren't callable from each other
     auth_headers = {
@@ -1651,17 +1647,16 @@ def list_users_detailed():
     users = requests.get(user_list_api, headers=auth_headers).json()
 
     results = {
-        user.get('username') : {
-            'packages' : package_counts.get(user.get('username'), 0),
-            'installs' : event_results[user.get('username')]['installs']
-                if user.get('username') in event_results else 0,
-            'previews' : event_results[user.get('username')]['previews']
-                if user.get('username') in event_results else 0,
-            'status' : 'active' if user.get('is_active') == True else 'disabled'
+        user['username'] : {
+            'packages' : package_counts.get(user['username'], 0),
+            'installs' : event_results[(user['username'], Event.Type.INSTALL)],
+            'previews' : event_results[(user['username'], Event.Type.PREVIEW)],
+            'pushes' : event_results[(user['username'], Event.Type.PUSH)],
+            'deletes' : event_results[(user['username'], Event.Type.DELETE)],
+            'status' : 'active' if user['is_active'] else 'disabled'
             }
-        for user in users.get('results')
+        for user in users['results']
     }
-
 
     return results
 
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -1617,6 +1617,53 @@ def list_users():
 
     return resp.json()
 
+@app.route('/api/users/list_detailed', methods=['GET'])
+@api(enabled=ENABLE_USER_ENDPOINTS, require_admin=True)
+@as_json
+def list_users_detailed():
+    package_counts_query = db.session.query(Package.owner, sa.func.count(Package.owner)).group_by(Package.owner)
+    package_counts = {}
+    for user in package_counts_query:
+        package_counts[user[0]] = user[1]
+
+    events = db.session.query(Event.user, Event.type, sa.func.count(Event.type)
+            ).filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW])
+            ).group_by(Event.user, Event.type)
+
+    event_results = {}
+    for item in events:
+        if event_results.get(item[0]) is None:
+            event_results[item[0]] = { 'installs':0, 'previews':0 }
+        if item[1] == 2:
+            event_results[item[0]]["installs"] = event_results[item[0]]["installs"] + item[2]
+        elif item[1] == 3:
+            event_results[item[0]]["previews"] = event_results[item[0]]["previews"] + item[2]
+
+    # replicate code from list_users since endpoints aren't callable from each other
+    auth_headers = {
+        AUTHORIZATION_HEADER: g.auth_header
+    }
+
+    user_list_api = "%s/accounts/users" % QUILT_AUTH_URL
+
+    users = requests.get(user_list_api, headers=auth_headers).json()
+
+    results = {}
+    for user in users.get('results'):
+        results[user.get('username')] = { 'last_seen' : user.get('last_login') }
+
+    for username in results.keys():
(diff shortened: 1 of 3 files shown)


registry/quilt_server/views.py
```diff
@@ -1617,6 +1617,53 @@ def list_users():
...
+    package_counts = {}
+    for user in package_counts_query:
+        package_counts[user[0]] = user[1]
+
+    events = db.session.query(Event.user, Event.type, sa.func.count(Event.type)
+            ).filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW])
+            ).group_by(Event.user, Event.type)
+
+    event_results = {}
+    for item in events:
```
dimaryaz: `for event_user, event_type, event_count in events:`

registry/quilt_server/views.py
```diff
@@ -1617,6 +1617,53 @@ def list_users():
...
+        if event_results.get(item[0]) is None:
```
dimaryaz: Just use `defaultdict(int)`

registry/quilt_server/views.py
```diff
@@ -1617,6 +1617,53 @@ def list_users():
...
+    auth_headers = {
+        AUTHORIZATION_HEADER: g.auth_header
+    }
+
+    user_list_api = "%s/accounts/users" % QUILT_AUTH_URL
+
+    users = requests.get(user_list_api, headers=auth_headers).json()
+
+    results = {}
+    for user in users.get('results'):
```
dimaryaz: Use a dict comprehension, i.e:
```
results = {
  user.get('username'): {...}
  for user in ...
}
```

registry/quilt_server/views.py
```diff
@@ -1617,6 +1617,53 @@ def list_users():
...
+        results[user.get('username')] = { 'last_seen' : user.get('last_login') }
+
+    for username in results.keys():
```
dimaryaz: `for username, result in results.items():`

registry/quilt_server/views.py
```diff
@@ -1617,6 +1617,53 @@ def list_users():
...
+        results[username]['packages'] = package_counts.get(username, 0)
+        if username in event_results:
```
dimaryaz: Use `defaultdict`

registry/quilt_server/views.py
```diff
@@ -1617,6 +1617,53 @@ def list_users():
...
+@api(enabled=ENABLE_USER_ENDPOINTS, require_admin=True)
+@as_json
+def list_users_detailed():
+    package_counts_query = db.session.query(Package.owner, sa.func.count(Package.owner)).group_by(Package.owner)
+    package_counts = {}
+    for user in package_counts_query:
+        package_counts[user[0]] = user[1]
+
+    events = db.session.query(Event.user, Event.type, sa.func.count(Event.type)
+            ).filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW])
```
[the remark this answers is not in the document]
meffij: They're the ones Aneesh wanted. I can do them all if we want

registry/quilt_server/views.py
```diff
@@ -1617,6 +1618,53 @@ def list_users():
...
+        AUTHORIZATION_HEADER: g.auth_header
+    }
+
+    user_list_api = "%s/accounts/users" % QUILT_AUTH_URL
+
+    users = requests.get(user_list_api, headers=auth_headers).json()
+
+    results = {
+        user.get('username') : {
+            'packages' : package_counts.get(user.get('username'), 9),
```
meffij: 9 should be 0 lol

registry/quilt_server/views.py
```diff
@@ -1617,6 +1618,53 @@ def list_users():
...
+    events = (
+        db.session.query(Event.user, Event.type, sa.func.count(Event.type))
+        .filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW]))
+        .group_by(Event.user, Event.type)
+        )
+
+    event_results = defaultdict(lambda: defaultdict(int))
+    for event_user, event_type, event_count in events:
+        if event_type == Event.Type.INSTALL:
+            event_results[event_user]["installs"] = event_count
```
dimaryaz: Wait, you're not actually returning `"installs"` or `"previews"` to the user - you're just looking them up later in the code. So you don't need strings. Or nested dictionaries. Just do:
```
event_results[(event_user, event_type)] = event_count
```

In fact:
```
event_results = {
  (event_user, event_type): event_count
  for ...
}
```

registry/quilt_server/views.py
```diff
@@ -1617,6 +1618,53 @@ def list_users():
...
+def list_users_detailed():
+    package_counts_query = (
+        db.session.query(Package.owner, sa.func.count(Package.owner))
+        .group_by(Package.owner)
+        )
+    package_counts = dict(package_counts_query) 
+
+    events = (
+        db.session.query(Event.user, Event.type, sa.func.count(Event.type))
+        .filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW]))
```
dimaryaz: I'd remove it and just return all of the data - there's such a small cost to it.

registry/quilt_server/views.py
```diff
@@ -1617,6 +1618,53 @@ def list_users():
...
+    auth_headers = {
+        AUTHORIZATION_HEADER: g.auth_header
+    }
+
+    user_list_api = "%s/accounts/users" % QUILT_AUTH_URL
+
+    users = requests.get(user_list_api, headers=auth_headers).json()
+
+    results = {
+        user.get('username') : {
```
dimaryaz: Probably safe to assume it'll have a username, so just `user['username']`?

registry/quilt_server/views.py
```diff
@@ -1617,6 +1618,53 @@ def list_users():
...
+            'packages' : package_counts.get(user.get('username'), 0),
+            'installs' : event_results[user.get('username')]['installs']
+                if user.get('username') in event_results else 0,
+            'previews' : event_results[user.get('username')]['previews']
+                if user.get('username') in event_results else 0,
+            }
+        for user in users.get('results')
```
dimaryaz: Same as above; `users['results']`

registry/quilt_server/views.py
```diff
@@ -1630,16 +1630,12 @@ def list_users_detailed():
 
     events = (
         db.session.query(Event.user, Event.type, sa.func.count(Event.type))
-        .filter(Event.type.in_([Event.Type.INSTALL, Event.Type.PREVIEW]))
         .group_by(Event.user, Event.type)
         )
 
-    event_results = defaultdict(lambda: defaultdict(int))
+    event_results = defaultdict(int)
```
dimaryaz: I guess you don't need `defaultdict` anymore... Just use the dictionary comprehension syntax, like I showed earlier?

(Or did you want to avoid `.get(...)` later? Not really worth it, I'd say - especially given that `package_counts` is a regular dict.)

registry/quilt_server/views.py
```diff
@@ -1651,17 +1647,17 @@ def list_users_detailed():
     users = requests.get(user_list_api, headers=auth_headers).json()
```
dimaryaz: Suggestion: create a `usernames` variable and use it later:
```
usernames = (user['username'] for user in users['results'])
```

(It's a generator, so doesn't even build a list unnecessarily.)

registry/tests/admin_test.py
```diff
@@ -59,3 +59,85 @@ def testListPackagesNotAdmin(self):
...
+        resp = self.app.get(
+            '/api/audit/{usr}/'.format(
+                usr=self.user
+            ),
+            headers={
+                'Authorization': self.user
+            }
+        )
+        data = json.loads(resp.get_data()).get('events')
+        assert(len(data) == 3)
```
dimaryaz: `assert` is an operator, not a function; use a space and no parens.

It can actually cause very subtle bugs:
```
assert False, "This fails"
assert(False, "This does NOT fail")
```

registry/tests/admin_test.py
```diff
@@ -59,3 +59,85 @@ def testListPackagesNotAdmin(self):
...
+        for event in data:
+            assert(event.get('type') == 'PUSH')
```
dimaryaz: Don't need `.get`; you'll get an exception regardless, whether it's an assert or a key error.

registry/tests/admin_test.py
```diff
@@ -59,3 +59,85 @@ def testListPackagesNotAdmin(self):
...
+            '/api/audit/{usr}/'.format(
+                usr=self.user
+            ),
+            headers={
+                'Authorization': self.user
+            }
+        )
+        data = json.loads(resp.get_data()).get('events')
+        assert(len(data) == 4)
+        pass
```
dimaryaz: ?

dimaryaz: I meant, don't need `pass`

registry/quilt_server/views.py
meffij: I think it reads better with defaultdict than with the comprehension

registry/quilt_server/views.py
meffij: What would be the utility of this?

PR #357, opened 2018-02-10
Added new tests for admin API  endpoint

Kurlov:


meffij:
Can you fix the conflict? Then I'll merge

compiler/quilt/tools/command.py
```diff
@@ -1340,6 +1340,14 @@ def list_users(team=None):
     resp = session.get('%s/api/users/list' % url)
     return resp.json()
 
+def list_users_detailed(team=None):
```
meffij: You need to add this to the export list in __init__.py for it to be available in the API.

PR #358, opened 2018-02-11
Added test access commands

Kurlov:


compiler/quilt/test/test_command.py
```diff
@@ -687,6 +700,82 @@ def test_audit_not_admin_package(self):
...
+    def test_access_list(self):
+        self.requests_mock.add(
+            responses.GET,
+            '%s/api/access/foo/bar' % command.get_registry_url(None),
+            status=201,
+            json={
+                'users': ['foo']
+            }
+        )
+        command.access_list('foo/bar')
```
meffij: In general, when you test a successful call, don't just test that an exception is not thrown -- validate the data

meffij: Please apply this feedback to this whole PR, don't worry about anything outside this PR for now

PR #360, opened 2018-02-12
Use a small dataset in most of the tests

dimaryaz:
Unless we're specifically testing large inputs, different data types, caching, etc., there's no need to use a huge dataset.

So:
- Change `build.yml` use small inputs
- Rename the existing `build.yml` to `build_large.yml` and use it in two tests that do thorough checks
- Remove a redundant test for pyarrow (which is the default) and instead just test that the env variables work

This cuts the time to run `pytest` in half.

### DIFF
--- a/compiler/quilt/test/build.yml
+++ b/compiler/quilt/test/build.yml
@@ -2,18 +2,8 @@
 contents:
   dataframes:
     csv: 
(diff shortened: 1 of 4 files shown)


akarve:
yay, faster `pytest`

PR #362, opened 2018-02-12
Package table for Admin UI

meffij:
Not complete. Needs:
- [x] last push timestamp
- [x] tests

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -1796,7 +1796,7 @@ def audit_package(owner, package_name):
 
     return dict(
         events=[dict(
-            created=event.created,
+            created=_utc_datetime_to_ts(event.created),
             user=event.user,
             type=Event.Type(event.type).name,
             package_owner=event.package_owner,
@@ -1817,7 +1817,7 @@ def audit_user(user):
 
     return dict(
         events=[dict(
-            created=event.created,
+            created=_utc_datetime_to_ts(event.created),
             user=event.user,
             type=Event.Type(event.type).name,
             package_owner=event.package_owner,
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -1832,15 +1832,17 @@ def audit_user(user):
 @as_json
 def package_summary():
     events = (
-        db.session.query(Event.package_name, Event.type, sa.func.count(Event.type))
-        .group_by(Event.package_name, Event.type)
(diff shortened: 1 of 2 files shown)


registry/quilt_server/views.py
```diff
@@ -1826,3 +1826,31 @@ def audit_user(user):
...
         ) for event in events]
     )
+
+@app.route('/api/admin/package_summary')
+@api(require_admin=True)
+@as_json
+def package_summary():
+    events = (
+        db.session.query(Event.package_owner, Event.package_name, Event.type, 
+                sa.func.count(Event.type), sa.func.max(Event.created))
```
dimaryaz: Make it aligned with the open parenthesis... or just move the args above to the next line

registry/tests/admin_test.py
```diff
@@ -169,3 +170,42 @@ def testAdminListUserUI(self):
...
+            headers={
+                'Authorization': self.user
+            }
+        )
+        data = json.loads(resp.data.decode('utf8'))
+        assert data['status'] == 200
+        package = data['packages']['{user}/{pkg}'.format(user=self.user, pkg=self.pkg)]
+        now = datetime.datetime.utcnow()
+        last_push = package['pushes']['latest']
+        then = datetime.datetime.strptime(last_push, "%Y-%m-%dT%H:%M:%S.%f")
```
dimaryaz: Actually, we should be returning UNIX timestamps everywhere - just to avoid dealing with parsing, timezones, etc. (Not sure if you've already used dates in other endpoints? If so, you should change those.)

Search for _utc_datetime_to_ts in views.py for examples.

meffij: I don't generate new timestamps, just parse the ones that come out of the database.

dimaryaz: Yeah. But you should return them in the UNIX timestamp format.

meffij: Done, also fixed for audit endpoints

registry/tests/admin_test.py
```diff
@@ -202,7 +202,7 @@ def testAdminPackageUserUI(self):
         package = data['packages']['{user}/{pkg}'.format(user=self.user, pkg=self.pkg)]
         now = datetime.datetime.utcnow()
```
dimaryaz: You can also just do `now = time.time()`, and avoid `datetime` completely

PR #365, opened 2018-02-13
Implement the "always requires auth" catalog config

dimaryaz:


catalog/config.js.tmpl
```diff
@@ -5,5 +5,6 @@ window.__CONFIG = {
   signOutUrl: '$SIGN_OUT_URL',
   team: {
     name: '$TEAM_NAME'
-  }
+  },
+  alwaysRequiresAuth: $ALWAYS_REQUIRES_AUTH
```
akarve: minor: it's about time to alphabetize the JSON keys for `config`

dimaryaz: But it's JavaScript. What if the code depends on the key order?

J/k, sure

PR #372, opened 2018-02-14
Fix package list test

meffij:


registry/tests/admin_test.py
```diff
@@ -200,12 +200,11 @@ def testAdminPackageUserUI(self):
...
         package = data['packages']['{user}/{pkg}'.format(user=self.user, pkg=self.pkg)]
-        now = datetime.datetime.utcnow()
+        now = time.time()
         last_push = package['pushes']['latest']
-        then = datetime.datetime.utcfromtimestamp(last_push)
-        delta = now - then
-        acceptable = datetime.timedelta(minutes=2)
-        assert abs(acceptable) > abs(delta)
+        delta = now - last_push
+        acceptable = 5 * 60 # 5 minutes
```
akarve: Minor: in PEP-8 constants should be upper case https://www.python.org/dev/peps/pep-0008/#constants

PR #381, opened 2018-02-15
Team strings

meffij:
QOL fixes for team compiler users

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -507,7 +507,7 @@ def build(package, path=None, dry_run=False, env='default', force=False):
         answer = input("You're logged in as a team member, but you aren't specifying " +
                        "a team for the package you're currently building. Maybe you meant:\n" +
                        "quilt build {team}:{package}\n" +
-                       "Are you sure you want to continue? (y/N)".format(
+                       "Are you sure you want to continue? (y/N) ".format(
                                 team=logged_in_team, package=package))
         if answer.lower() != 'y':
             return
@@ -979,7 +979,7 @@ def install(package, hash=None, version=None, tag=None, force=False):
                                     "Maybe you meant {team}:{owner}/{pkg}?").format(
                                             owner=owner, pkg=pkg, team=logged_in_team))
         else:
-            raise e
+            raise
 
     assert pkghash is not None
 
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -493,7 +493,7 @@ def _log(team, **kwargs):
     # restore disabled error-handling
     session.hooks['response'] = orig_response_hooks
 
-def build(package, path=None, dry_run=False, env='default'):
+def build(package, path=None, dry_run=False, env='default', force=False):
     """
     Compile a Quilt data package, either from a build file or an existing package node.
 
@@ -503,7 +503,7 @@ def build(package, path=None, dry_run=False, env='default'):
     # TODO: rename 'path' param to 'target'?
     team, _, _ = parse_package(package)
     logged_in_team = _find_logged_in_team()
-    if logged_in_team is not None and team is None:
+    if logged_in_team is not None and team is None and force is False:
         answer = input("You're logged in as a team member, but you aren't specifying " +
                         "a team for the package you're currently building. Did you mean " +
                         "quilt build {team}:{package}? N to continue. (Y/n)".format(
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -973,11 +973,11 @@ def install(package, hash=None, version=None, tag=None, force=False):
             pkghash = _match_hash(session, team, owner, pkg, hash)
     except HTTPResponseException as e:
         logged_in_team = _find_logged_in_team()
-        if team is None and logged_in_team is not None:
-            if e.response.status_code == 404:
-                raise CommandException(("Package {owner}/{pkg} does not exist. " +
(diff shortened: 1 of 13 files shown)


compiler/quilt/tools/command.py
```diff
@@ -497,6 +497,14 @@ def build(package, path=None, dry_run=False, env='default'):
     """
     # TODO: rename 'path' param to 'target'?
     team, _, _ = parse_package(package)
+    logged_in_team = _find_logged_in_team()
+    if logged_in_team is not None and team is None:
+        answer = input("You're logged in as a team member, but you aren't specifying " +
+                        "a team for the package you're currently building. Did you mean " +
+                        "quilt build {team}:{package}? N to continue. (Y/n)".format(
```
dimaryaz: I feel like "y" to continue is the convention; "n" feels weird to be. Also, should probably add a "force" parameter to avoid the prompt.

compiler/quilt/tools/command.py
```diff
@@ -934,28 +942,41 @@ def install(package, hash=None, version=None, tag=None, force=False):
...
-        pkghash = response.json()['hash']
-    else:
-        pkghash = _match_hash(session, team, owner, pkg, hash)
+            pkghash = response.json()['hash']
+        else:
+            pkghash = _match_hash(session, team, owner, pkg, hash)
+    except CommandException as e:
+        logged_in_team = _find_logged_in_team()
+        if team is None and logged_in_team is not None:
+            notfoundstr = ("Package {owner}/{pkg} does not exist " +
```
dimaryaz: This is really fragile. At most, we should check for a 404... But I guess that requires changing or disabling the `_handle_response` function.

compiler/quilt/tools/command.py
meffij: The prompt only appears if the install fails. I will add a force parameter to API, but I think CLI is fine.

compiler/quilt/tools/command.py
```diff
@@ -934,28 +947,39 @@ def install(package, hash=None, version=None, tag=None, force=False):
...
-        pkghash = response.json()['hash']
-    else:
-        pkghash = _match_hash(session, team, owner, pkg, hash)
+            pkghash = response.json()['hash']
+        else:
+            pkghash = _match_hash(session, team, owner, pkg, hash)
+    except HTTPResponseException as e:
+        logged_in_team = _find_logged_in_team()
+        if team is None and logged_in_team is not None:
+            if e.response.status_code == 404:
```
dimaryaz: Should be part of the `if` above - otherwise, you're gonna drop the exception

dimaryaz: Also, you should use `requests.codes.not_found`, for consistency with other code. (In case, you know, someone doesn't know what a 404 is...)

compiler/quilt/tools/command.py
meffij: How would you phrase it to make 'y' the default?

dimaryaz: Something like "Are you sure you want to continue?"

compiler/quilt/tools/command.py
```diff
@@ -934,28 +948,39 @@ def install(package, hash=None, version=None, tag=None, force=False):
...
+            pkghash = _match_hash(session, team, owner, pkg, hash)
+    except HTTPResponseException as e:
+        logged_in_team = _find_logged_in_team()
+        if (team is None and logged_in_team is not None
+                and e.response.status_code == requests.codes.not_found):
+            raise CommandException(("Package {owner}/{pkg} does not exist. " +
+                                    "Maybe you meant {team}:{owner}/{pkg}?").format(
+                                            owner=owner, pkg=pkg, team=logged_in_team))
+        else:
+            raise e
```
dimaryaz: You should do a `raise` with no arguments to preserve the original stack trace.

compiler/quilt/tools/command.py
```diff
@@ -497,6 +502,15 @@ def build(package, path=None, dry_run=False, env='default'):
     """
     # TODO: rename 'path' param to 'target'?
     team, _, _ = parse_package(package)
+    logged_in_team = _find_logged_in_team()
+    if logged_in_team is not None and team is None and force is False:
+        answer = input("You're logged in as a team member, but you aren't specifying " +
+                       "a team for the package you're currently building. Maybe you meant:\n" +
+                       "quilt build {team}:{package}\n" +
+                       "Are you sure you want to continue? (y/N)".format(
```
dimaryaz: Space at the end?

See 582c48315c5d61d073839b01e4f8ba76a7b98854

compiler/quilt/tools/command.py
```diff
@@ -488,7 +493,7 @@ def _log(team, **kwargs):
     # restore disabled error-handling
     session.hooks['response'] = orig_response_hooks
 
-def build(package, path=None, dry_run=False, env='default'):
+def build(package, path=None, dry_run=False, env='default', force=False):
```
dimaryaz: You should probably add a `--force` to main, too - otherwise, you can't set it from the CLI?

meffij: I think CLI should not need force, if the prompt becomes such a pain point then they should log out and log in with their public cloud account. We want to avoid people just jamming --force on everything to not have to deal with prompts.

meffij: I can add it but I left it off intentionally, if you disagree I'll add it.

dimaryaz: ah, that's fine then

PR #384, opened 2018-02-15
Allow admins to delete package regardless of owner

meffij:


meffij:
Do we want this to require some sort of extra flag?

dimaryaz:
TBH, I think this should be a separate endpoint... Otherwise, what about all of the other actions - push, install, share, etc.?

akarve:
agreed on separate endpoint. re: do we want another flag, we want the endpoint equivalent of "are you sure?"... if that is even possible; otherwise we'll just require confirmation in the CLI and frontend.

PR #385, opened 2018-02-15
Export command

eode:
### Summary
Includes `quilt export` and `command.export(...)`

Only handles `FileNode` objects, not `TableNode` objects (minimum for TF support).

### Depends On
~~#536 Minor test improvements~~
~~#537 Build.py uncaught exception fix, other minor misc from export PR~~


kevinemoore:
@eode I think we're conflating the need to specify source files using absolute paths and a specification of the destination of a file for export. IMHO it doesn't make sense to allow absolute paths for export, but that doesn't mean that users shouldn't be able to use absolute paths in build files.

eode:
@kevinemoore  I can definitely agree with that.  My preference is to re-root absolute paths into the export dir (#2 above).

If an exporting user wants the exported files to go into their original location, they can export from that location (or specify that location as the destination).

If someone wants to make their package prettier when exporting, we can add something like this as a feature:
```
contents:
  images:
    base:  "/abs/path"
    pellets: 
      # this would use "/abs/path/images/pellets.png" and export to "<export_dir>/images/pellets.png"
      file: "images/pellets.png"    "yeardata/year01"
```
But in any-case, re-homing absolute files into the export dir, as per option 2 above, or ignoring them completely is fine with me.

quiltdata:
I think we're on the same page; I would adjust the syntax though.

Make file: into a list or dictionary and specify the export mapping (for absolute paths) right there.

eode:
To be clear, *this* PR doesn't include the modifications to how `build` should handle absolute file paths.

As it is, I've re-enabled absolute paths to be added via _set(), and any absolute files encountered during export are re-rooted them into the destination dir for export.

For the record, I still think it's kinda bad form to include absolute file paths in the registry -- at very least, it has some potential to leak things (like usernames) to other quilt users.  Not to significant, but it still makes me squirm.  Again, though, theoretically if we handle that right on the build side, it should be ok, and this way retains backward compatibility.

Re: `to_nodename(s)`, `to_identifier(s)`, and removal of `pythonize(s)`:  My hope is that we can have a central place where we define basic data types (like node names) in a consistent way, both for reference and for any changes we may make in the future.  The `to_foo` existed in the TF branch along with `is_foo()` functions, so I chose those.

Re: Keywords for `to_identifer()`:  I re-added keyword checking on `to_identifier()`, but made it optional.  Once we shift to item notation, we will still need that for backward compatibility.  Names like `global`, `nonlocal`, etc will be converted to `global_`, `nonlocal_`, etc.   Once we switch to item notation as per the design doc for `Node Names and Imports`, this will be a python-specific backward-compatibility option (we have a lot of tests that use `somenode.foo` instead of `somenode['foo']`.

eode:
@kevinemoore This should be ready for review again.

eode:
Oh -- and if we haven't, we may need to add a task re: build file changes to Asana.

compiler/quilt/nodes.py
```diff
@@ -93,6 +99,47 @@ def _add_group(self, groupname):
         child = GroupNode(self._package, core.GroupNode({}))
         setattr(self, groupname, child)
 
+    @classmethod
+    def __iteritems(cls, node, base_path, recursive=False):
```
kevinemoore: Why are you adding these helper methods? It's not obvious (to me) why they're needed and how they're used. If possible, we should try to reuse and possibly extend exiting helpers and APIs instead of duplicating functionality.

compiler/quilt/tools/build.py
```diff
@@ -106,9 +113,9 @@ def _gen_glob_data(dir, pattern, child_table):
         node_table = {} if child_table is None else child_table.copy()
         filepath = filepath.relative_to(dir)
         node_table[RESERVED['file']] = str(filepath)
-        node_name = to_nodename(filepath.stem, invalid=used_names)
+        node_name = _to_nodename(filepath.stem, invalid=used_names)
```
kevinemoore: Can we avoid renaming functions in a PR like this? It'll be much easier to follow the diffs if small, but widespread changes like that happen in their own PRs.

compiler/quilt/tools/build.py
```diff
@@ -277,7 +284,7 @@ def _file_to_spark_data_frame(ext, path, target, handler_args):
         dataframe = reader.load(path)
 
         for col in dataframe.columns:
-            pcol = _pythonize_name(col)
+            pcol = _to_identifier(col)
```
kevinemoore: Another rename?

compiler/quilt/tools/command.py
```diff
@@ -1447,3 +1448,204 @@ def load(pkginfo):
     """functional interface to "from quilt.data.USER import PKG"""
     # TODO: support hashes/versions/etc.
     return _load(pkginfo)[0]
+
+def export(package, output_path='.', filter=lambda x: True, mapper=lambda x: x, force=False):
```
kevinemoore: Is the mapper needed for tensor flow? If not, let's remove it to keep the API as simple as possible.

compiler/quilt/nodes.py
```diff
@@ -124,8 +171,6 @@ def _set(self, path, value, build_dir=''):
         elif isinstance(value, string_types + (bytes,)):
             # bytes -> string for consistency when retrieving metadata
             value = value.decode() if isinstance(value, bytes) else value
-            if os.path.isabs(value):
```
kevinemoore: We shouldn't allow absolute paths in the package. I think the confusion is that we need to allow absolute paths in build files (build.yml). I think the answer is that nodes built from absolute paths should get a null path or a sensible default relative path.

compiler/quilt/nodes.py
eode: Clarity in usage.

* Without `DataNode._datafile`:
  * In order to determine what type of data their `DataNode` has, the user must do one of the following:
    * Call `node._data()`, potentially loading a very large amount of data
    * Know that `quilt.nodes.DataNode._node` is either a `quilt.tools.core.FileNode` or a `quilt.tools.core.TableNode`, and check for what they want
    * The resultant code is less succinct and less clear.
* With `DataNode._datafile`:
 * The user needs to know that if the `DataNode` references file data, it's available as `DataNode._datafile`.
 * The resultant code is a one-liner that is very clear.

```
## Minimal Example Without DataNode._datafile:
from quilt.tools.core import FileNode
from quilt.data.user import mydata

files = {}
for key in mydata._data_keys():
    node = getattr(mydata, key)
    if isinstance(node._node, FileNode):
        files[key] = node._data()

# or, as a comprehension, this *is* fairly compact -- however, it still requires internal knowledge of quilt,
# and isn't easy on the eyes.
files = {key: getattr(mydata, key)._data() for key in mydata._data_keys()
            if isinstance(getattr(mydata, key)._node, FileNode)}
# Not too bad, but it's not exactly clear, either.

## Minimal Example With DataNode._datafile:
from quilt.data.user import mydata

files = {}
for key, node in mydata._items():
    if node._datafile:
        files[key] = node._datafile

# Or, as a comprehension, clear and concise:
files = {key: node._datafile for key, node in mydata._items() if node._datafile}
```

As to `DataNode._iteritems` and `DataNode._iterpaths`:
Mine does recursion.  I'll just add it as an optional parameter to `_items()` and `_keys` instead.

eode: They're recursive, and I was avoiding touching the existing `_items` and `_keys`.  The recursivity is used in export, and is also a good user-facing feature.

I've shifted the code over to `_items()` and `_keys()`.

compiler/quilt/nodes.py
eode: My reasoning:
 * It was decided to permit absolute paths in `core.py`.
 * It was decided that we didn't want to export absolute paths as absolute
 * Absolute paths currently exist in global, and probably in some teams
 * Since the check needs to be done for existing absolute paths on the export side, we may as well do the work there
 * As it is, we just ignore the absolute-ness of paths, and treat it as any other exported path, so `'/etc/profile'` becomes `'etc/profile'`, and `'c:\windows\cmd.exe'` becomes `'windows\cmd.exe'` -- and these files, like any other, are exported to the specified destination, if and only if the target doesn't exist.

Do you still want me to change that?

eode: We could also:

 * scan the existing database(s) and remove absolute paths, or
 * do checking and handling on both on package build and package export (preferably with the same code)

compiler/quilt/nodes.py
```diff
@@ -44,6 +45,11 @@ def __init__(self, package, node, data=None):
     def __call__(self):
         return self._data()
 
+    @property
```
kevinemoore: Let's not bloat nodes.py with export-specific helpers. It'll be cleaner to just make a helper in export instead of adding methods here.

compiler/quilt/nodes.py
```diff
@@ -67,7 +73,18 @@ def __repr__(self):
         data_info = '\n'.join(sorted(self._data_keys()))
         return '%s\n%s%s' % (pinfo, group_info, data_info)
 
-    def _items(self):
+    def _items(self, recursive=False):
```
kevinemoore: We already have a method that iterates through a package tree in the Package class. Can you use that instead? Also, please keep an eye on PR534.

compiler/quilt/nodes.py
```diff
@@ -83,16 +100,46 @@ def _group_keys(self):
         """
         return [name for name, child in self._items() if isinstance(child, GroupNode)]
 
-    def _keys(self):
-        """
-        keys directly accessible on this object via getattr or .
+    def _keys(self, recursive=False, iterator=False):
```
kevinemoore: Same here. Can you use the existing iterator instead of making this recursive?

compiler/quilt/nodes.py
kevinemoore: Let's try to consolidate helper functions like these in the Package class so we don't end up with two full implementations of tree classes that are subtly different.

compiler/quilt/nodes.py
kevinemoore: We can use the same code for checking for abs paths during building and package export, but we probably shouldn't crash if we see an absolute path during export.

compiler/quilt/test/test_cli.py
```diff
@@ -632,23 +635,13 @@ def test_cli_command_login(self):
...
         result = self.execute_with_checks(cmd, funcname='login')
 
-        # Specific tests
-        assert not result['args']
         assert result['kwargs']['team'] is None
 
         # login with team name
         cmd = ['login', 'example_team']
-        result = self.execute(cmd)
+        result = self.execute_with_checks(cmd, funcname='login')
```
kevinemoore: This doesn't look related to export. What are you changing here?

compiler/quilt/test/test_cli.py
```diff
@@ -724,27 +709,52 @@ def test_cli_command_push(self):
 
         # team (without reupload)
         cmd = 'push --reupload --team blah:fakeuser/fakepackage'.split()
-        result = self.execute(cmd)
-
-        # General tests
-        assert result['return code'] == 0
-        assert result['matched'] is True  # func name recognized by MockObject class?
-        assert not result['bind failure']
+        result = self.execute_with_checks(cmd, funcname='push')
```
kevinemoore: And here?

compiler/quilt/test/test_command.py
```diff
@@ -281,8 +283,7 @@ def test_login_invalid_team(self, mock_login_with_token, mock_input, mock_open):
 
         mock_input.return_value = old_refresh_token
 
-        with pytest.raises(command.CommandException,
-                match=TEAM_ID_ERROR):
+        with pytest.raises(command.CommandException, match=TEAM_ID_ERROR):
```
kevinemoore: Are these changes just whitespace?

compiler/quilt/tools/build.py
```diff
@@ -79,14 +80,20 @@ def _get_local_args(node, keys):
 def _is_valid_group(group):
     return isinstance(group, dict) or group is None
 
-def _pythonize_name(name):
-    safename = re.sub('[^A-Za-z0-9]+', '_', name).strip('_')
-
-    if safename and safename[0].isdigit():
-        safename = "n%s" % safename
+def pythonize(name):
```
kevinemoore: Can you move the renaming here to a separate PR. These changes aren't related to export, are they?

compiler/quilt/tools/command.py
```diff
@@ -1507,3 +1511,208 @@ def load(pkginfo):
...
+    The `filter` function takes export paths (without the output path prepended)
+    should return `True` or `False` to indicate the node's inclusion in the export.
+
+    The `mapper` function takes export paths (without the output path prepended),
+    and should alter and return that path with any changes that are desired for the
+    export (if any).
+
+    :param package: package or subpackage name, e.g., user/foo or user/foo/bar
+    :param output_path: distination folder
+    :param filter: function -- takes a node path string, returns True to export
```
kevinemoore: Remove filter and mapper parameters from comment.

compiler/quilt/tools/command.py
```diff
@@ -1507,3 +1511,208 @@ def load(pkginfo):
...
+        files = set(dest for src, dest in export_list)
+        conflict_counter = Counter(files)
+        file_conflicts = [dest for dest, count in conflict_counter.items() if count > 1]
+        verified_file_conflicts = []
+
+        if file_conflicts:
+            # kinda slow, but only happens if a conflict has definitely occurred.
+            for conflict in file_conflicts:
+                matches = [item for item in export_list if item[1] == conflict]
+                # if all are from the same source, it's not really a conflict.
```
kevinemoore: What is checking? Are you testing this to allow re-exporting into the same location?

compiler/quilt/tools/util.py
```diff
@@ -187,24 +193,31 @@ def to_identifier(string):
     >>> to_identifier('9foo') -> 'n9foo'
 
     :param string: string to convert
+    :param permit_keyword: Permit python keywords like "import" and "for"
     :returns: `string`, converted to python identifier if needed
     :rtype: string
     """
-    # Not really useful to expose as a CONSTANT, and python will compile and cache
-    string = re.sub(r'[^0-9a-zA-Z_]', '_', string)
+    # Not really useful to expose as a constant, and python will compile and cache
```
kevinemoore: Is this needed for export? If not, could you put this in a separate PR?

compiler/quilt/nodes.py
eode: Ah.  I thought that would be a useful thing for users, because it tells them whether or not it has a data file without actually loading (possibly gigabytes of) data.

compiler/quilt/nodes.py
eode: This iterates through a node tree, not a package tree, so it works in cases where a user specifies `"user/package/fizz/bang/buzz"`.

..corresponding with Dima about #534 now.

eode: Also -- do you mean in the `quilt.tools.package.Package` class?  I don't see the method for recursively walking the package names/values.

compiler/quilt/nodes.py
eode: Sure, what iterator?  I didn't see one in either `nodes.py` or `package.py`.  Is there a helper somewhere else I should use?

compiler/quilt/nodes.py
eode: I'll remove that, though.

compiler/quilt/nodes.py
eode: We talked on Slack, and you mentioned you were referring to `core.preorder`, but that isn't really applicable to this -- plus, it's a fairly useful public facing feature, as you've noted #534, which is similar (but about values only).

It's probably a good idea to have `Node._keys()`, `Node._values()`, and `Node.items()`, and in this case it's also not reinventing the wheel.

compiler/quilt/tools/util.py
eode: Already done -- that's in #537

compiler/quilt/tools/command.py
eode: Testing this to ensure the package doesn't have two files that export to the same name.  For example, if the user adds the contents of a particular folder to a package, but at different times (and different states) under different node names.  The export locations are still the same, because they came from the same directory tree structure.

compiler/quilt/tools/build.py
eode: Already done. #537.  When that clears, this will go away when I merge Master into this.

compiler/quilt/nodes.py
eode: This was resolved as "nodes.py is going away".

compiler/quilt/nodes.py
eode: Since there is no iterator available, and `nodes.py` is going away, I've moved the code for the iterator to be contained inside the export() function.

compiler/quilt/tools/command.py
```diff
@@ -1224,3 +1226,233 @@ def load(pkginfo):
...
     # TODO: support hashes/versions/etc.
     return _load(pkginfo)[0]
+
+def export(package, output_path='.', force=False):
+    """Export package file data.
+
+    Exports children of specified node to files (if they have file data).
+    Does not export dataframes or other non-file data.
+
+    The `filter` function takes export paths (without the output path prepended)
```
kevinemoore: Remove comment about removed filter and mapper parameters.

eode:
OK, I've got to set a max float size for Pandas CSV output, so I used 256 bit / 78 decimal.

Added roundtrip build/export test.

Now that the other PRs have gone though, this is a lot cleaner now, as well.

Note:  Although this doesn't *add* to `nodes.py` since that's being phased out at some point, it does *use* `nodes.py` -- until such functionality is created for `core.py` nodes.  It seems pointless to write the extra code that would be needed to use `core.py`, as that will need to be written separately, anyways.

Once the `core.py` API is complete and has useful features, we can switch to using that instead.

eode:
As to using existing helpers -- they don't exist yet.  :-)  But, that was discussed earlier, this note is just for posterity.

eode:
Whoops -- yup, was trying a new tool, and missed those.

eode:
Merging -- Kevin's remaining issues have been addressed, and is ok with this now (per slack with Aneesh).

PR #386, opened 2018-02-15
Switch from TIMESTAMP datatype to TIMESTAMPTZ

dimaryaz:
According to pretty much every Postgres expert, timestamp with timezone is the correct way to store timestamps. Postgres doesn't actually _store_ the timezone - just converts everything to UTC first.

Makes the hacky helper function unnecessary, and fixes a unit test (when the current timezone is not UTC).


dimaryaz:
(This doesn't actually affect production since everything is already in UTC, so should be a safe fix.)

PR #391, opened 2018-02-16
Store READMEs in the DB and return them in package previews and search results

dimaryaz:
Storing the README in the DB allows us to access it quickly, so we can return it in search results.


registry/quilt_server/views.py
```diff
@@ -503,6 +506,20 @@ def blob_get(owner, blob_hash):
         put=_generate_presigned_url(S3_PUT_OBJECT, owner, blob_hash),
     )
 
+
+def download_object_preview(owner, obj_hash):
+    resp = s3_client.get_object(
```
kevinemoore: Can we guarantee that we're not reading more than MAX_PREVIEW_SIZE bytes? Maybe only fetch that limit from s3? Maybe add a some padding just in case the compressed file is bigger.

PR #399, opened 2018-02-20
Fix strings returned by registry for quilt access

meffij:


registry/quilt_server/views.py
```diff
@@ -619,9 +619,9 @@ def package_put(owner, package_name, package_hash):
             if team_access is None:
                 raise ApiException(
                     requests.codes.forbidden,
-                    ("%(user)s/%(pkg)s is private. To share it with the team, " +
-                     "run `quilt access add %(user)s/%(pkg)s team`.") %
-                    dict(user=owner, pkg=package_name)
+                    ("%(team)s:%(user)s/%(pkg)s is private. To share it with the team, " +
```
dimaryaz: Ohh. `TeamName` is actually a "friendly" name displayed in the Catalog - not the name used in the CLI. So I guess we'll need a new variable here.

(That is, this is going to be mainly a `quilt.yaml` change. You won't need the `.lower()`, though.)

PR #401, opened 2018-02-21
Prevent self-disabling

meffij:


registry/quilt_server/views.py
```diff
@@ -1809,6 +1809,12 @@ def disable_user():
     username = data.get('username')
     _validate_username(username)
 
+    if g.auth.user == username:
+        raise ApiException(
+            403,
```
dimaryaz: Use `requests.codes.forbidden`, for consistency with everything else.

PR #403, opened 2018-02-21
Use a thread-local session when making requests to the auth server

dimaryaz:
Right now, we're creating a new connection for every auth request, which is pretty inefficient (TCP handshake, TLS, etc.). Sessions allow us to use a connection pool - however, they're not quite thread safe. So, use a session per thread.

(Tested it locally using uwsgi with the production config - and it appears to be doing the right thing.)

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -9,6 +9,7 @@ from datetime import datetime, timedelta, timezone
 from functools import wraps
 import json
 import time
+import threading
 from urllib.parse import urlencode
 
 import boto3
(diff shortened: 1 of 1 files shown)


dimaryaz:
@asah: Good question. In general, we just don't use any shared state. We have:
- DB connection: Flask handles per-request sessions
- Stripe: thread-safe the way we're using it (looking up objects - but not sharing them across threads)
- S3: better be thread-safe cause we're only using URL signing (no network requests)

That said, I'm adding S3 downloading in #391 - and apparently, that's not quite safe anymore. Will fix it.

PR #404, opened 2018-02-21
Teams UI: fit and finish (almost)

nl0:
### Must have

* [x] Bug - user can open multiple settings menus

### Should have

* [x] Display X close button at top of audit dialogs
* [x] User table: display disabled members in list; disable cells; Include "Enable" option under Settings that calls /api/users/enable

### Nice to have
* [x] Reverse chronological sort on all Package audit and Members audit
* [x] Add member - progress gear should be outside of button (to the right, larger)
* [x] Admin cannot disable self (will be reinforced by backend)
* [x] Policies, display tool tip: "Contact support@quiltdata.io to change these settings"
* [x] Move Settings gear to far left, in name column so we have `<GEAR> <NAME>`


nl0:
@akarve all the minor fixes are here, so you can merge this. I'll create separate PRs for routes and pagination.

PR #411, opened 2018-02-23
Fix and clean up admin tests

dimaryaz:
- Fix the reset password test - it wasn't even testing the correct endpoint :-/
- Add schemas to make sure username and email are actually present in the request
- Change the way admins are mocked: make it part of auth instead of a global variable (to make it easier to mock different users)
- Make admin tests more useful: have the admin audit a different user, not himself

### DIFF
--- a/registry/tests/utils.py
+++ b/registry/tests/utils.py
@@ -85,8 +85,8 @@ class QuiltTestCase(TestCase):
                 return (401, {}, "Not logged in")
             else:
                 # Hack to make it easy to simulate admin users:
-                # if the username contains "admin", it's an "admin".
-                is_staff = 'admin' in auth
+                # if the username is "admin", it's an "admin".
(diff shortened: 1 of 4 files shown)


registry/tests/utils.py
```diff
@@ -85,17 +84,17 @@ def cb(request):
             if auth is None:
                 return (401, {}, "Not logged in")
             else:
+                # Hack to make it easy to simulate admin users:
+                # if the username contains "admin", it's an "admin".
+                is_staff = 'admin' in auth
```
kevinemoore: This seems a bit dangerous, or at least confusing. E.g., 'notadmin' would actually be treated as an admin user by that rule. Could we just have a field that contains the admin username? Or, a list of names? Even a match of the exact string 'admin' would seem safer ('admin' == vs 'admin' in auth).

PR #412, opened 2018-02-23
Refactor payments and UI for teams

akarve:


catalog/app/components/PaymentDialog/index.js
```diff
@@ -28,8 +27,14 @@ const style = {
 };
 
 class PaymentDialog extends React.PureComponent { // eslint-disable-line react/prefer-stateless-function
-  state = {
-    showConfirm: false,
+  handleConfirm(curr, next, onDowngrade) {
+    if (next.confirmTitle) {
+      // eslint-disable-next-line no-alert no-restricted-globals
+      const proceed = window.confirm(`${next.confirmTitle}\n${next.confirmBody}`);
```
dimaryaz: `window.confirm`???

oh no you didn't

catalog/app/components/PaymentDialog/messages.js
```diff
@@ -4,24 +4,28 @@ import { defineMessages } from 'react-intl';
...
   already: {
     id: 'app.containers.Profile.alreadyFree',
-    defaultMessage: 'Already on plan',
+    defaultMessage: 'Nothing to do',
   },
   free: {
     id: 'app.containers.Profile.free',
     defaultMessage: 'Free',
   },
-  business_monthly_490: {
```
dimaryaz: I guess it doesn't matter much given that no one is using the business plan... But in general, we shouldn't assume that backend and frontend changes will go out simultaneously.

catalog/app/components/PaymentDialog/messages.js
```diff
@@ -4,24 +4,28 @@ import { defineMessages } from 'react-intl';
...
     id: 'app.containers.Profile.individual_monthly_7',
     defaultMessage: 'Individual',
   },
+  team_monthly_490: {
+    id: 'app.containers.Profile.team_monthly_490',
+    defaultMessage: 'Team',
+  },
+  team_unpaid: {
+    id: 'app.containers.Profile.team_unpaid',
+    defaultMessage: 'Team (30-day trial)',
```
dimaryaz: Hm, Stripe has built-in support for trials. We could just use it, and not even have the "unpaid" plan. I only did it because the paid plan currently requires a credit card.

catalog/app/components/Pricing/index.js
```diff
@@ -2,12 +2,22 @@
...
-import { Tabs, Tab } from 'material-ui/Tabs';
 
 import TakeAction from 'components/TakeAction';
 
 export const width = 900;
 
+const emailBody = `To get started, tell us about your team.%0D%0A
+%0D%0A
+Team size:%0D%0A
+Team id (alphabetical string):%0D%0A
```
dimaryaz: Give an example?

catalog/app/components/Pricing/index.js
```diff
@@ -2,12 +2,22 @@
...
+Admin name:%0D%0A
+Admin username:%0D%0A
+Admin phone number:%0D%0A
```
dimaryaz: Email, too? Might be different from the sender.

catalog/app/components/Pricing/index.js
```diff
@@ -54,70 +59,68 @@ const Styler = styled.div`
...
+            <td>
+              <h2>$7 { perUser }</h2>
+            </td>
+            <td>
+              <h2>$49* { perUser }</h2>
+            </td>
+          </tr>
+          <tr>
+            <td>
+              •&nbsp;Unlimited public packages<br />
```
dimaryaz: If only HTML had a tag to create lists...

catalog/app/components/Pricing/index.js
```diff
@@ -54,70 +59,68 @@ const Styler = styled.div`
...
+            </td>
+            <td>
+              •&nbsp;Unlimited public packages<br />
+              •&nbsp;Up to 1TB of private packages<br />
+            </td>
+            <td>
+              • Unlimited public packages<br />
```
dimaryaz: Do you need `&nbsp;` here, too?

catalog/app/containers/Admin/util.js
```diff
@@ -1,15 +1,29 @@
...
+const toIcon = {
+  packages: 'cloud_upload',
+  previews: 'remove_red_eye',
+  installs: 'cloud_download',
+};
+
+export const formatActivity = (map, activity) => map
+  .filter((key) => key in activity)
+  .map((key) => (
+    <span key={key} title="great">
```
dimaryaz: "great"?

catalog/app/containers/Admin/util.js
```diff
@@ -1,15 +1,29 @@
...
+      <MIcon drop="4px" style={{ fontSize: '170%', opacity: 0.5 }}>
+        {toIcon[key]}
+      </MIcon>&nbsp;{activity[key]}
+      &nbsp;&nbsp;&nbsp;&nbsp;
```
dimaryaz: Sigh

catalog/app/containers/Package/index.js
```diff
@@ -168,14 +170,18 @@ const Install = ({ name, owner }) => (
...
     </Code>
     <p><FormattedMessage {...strings.sell} /></p>
     <Help href={blogManage} />
     <h2><FormattedMessage {...strings.access} /></h2>
     <Tabs>
       <Tab label="Python">
-        <Code>from quilt.data.{owner} import {name}</Code>
+        {
+          config.team ?
+            <Code>from quilt.team.{config.team.name}.{owner} import {name}</Code>
```
dimaryaz: I think this is a "friendly" name - e.g., "Quilt Test", not "quilttest". We should have two different variables.

catalog/app/containers/Profile/constants.js
```diff
@@ -11,8 +13,29 @@ export const UPDATE_PLAN = 'app/Profile/UPDATE_PLAN';
...
+export const PLANS = config.team ? {
+  // team plans
+  team_unpaid: {
+    confirmTitle: 'Cancel team account?',
+    confirmBody: 'Your team will lose access to all data.',
+    cost: 0,
+    menu: 'Team (30-day trial)',
+    rank: 0,
+    statusIcon: 'warning',
+    statusMessage: 'Temporary Service plan. Upgrade to avoid data loss.',
```
dimaryaz: I don't think we should use words like "data loss"... it may result in "customer loss".

catalog/app/containers/Package/index.js
```diff
@@ -168,14 +170,18 @@ const Install = ({ name, owner }) => (
...
+      <Unselectable>$ </Unselectable>quilt install {makeHandle(owner, name)}
     </Code>
     <p><FormattedMessage {...strings.sell} /></p>
     <Help href={blogManage} />
     <h2><FormattedMessage {...strings.access} /></h2>
     <Tabs>
       <Tab label="Python">
-        <Code>from quilt.data.{owner} import {name}</Code>
+        {
+          config.team ?
```
dimaryaz: This happens to work - but if we were to run the stage or prod catalog in ECS, we'd have `team: {name: ''}` just cause of the way we set it in config.js.tmpl.

catalog/app/components/PaymentDialog/index.js
akarve: lol. i actually ripped out a pretty custom dialog to do this. i think it's clearer (scarier) and it is definitely better on mobile.

catalog/app/components/PaymentDialog/messages.js
akarve: sure though in this case there is no way that `business...` could ever be selected by the frontend (or the backend as we never allowed people to elect it). i guess a backend could return it

catalog/app/components/PaymentDialog/messages.js
akarve: hard to depend on stripe; e.g. see the logic i do with `plan.rank`, etc. so unless this is demonstrably fragile i'm hoping to keep a client-side object of all possible plans and meta-data.

catalog/app/components/Pricing/index.js
akarve: what makes this sad is it would take me 20+ min. to get `ol` and `li` looking this good. true story.

catalog/app/components/PaymentDialog/messages.js
dimaryaz: Actually, you _can_ upgrade to business right now.

akarve: you are right. damn. upgrade with you 💳 :p

catalog/app/containers/Package/index.js
akarve: oh yeah. what i'd like to do for now is treat `config.team.name` as the case-sensitive team id. that's how i use it right now. i can rename it to `team.id` if that's better. i don't use the friendly string anywhere. and of course i can fix `config.js.tmpl`. separate PR.

catalog/app/containers/Package/index.js
akarve: that is kinda' dangerous. there are other places in this PR that would break since `{}` is truthy. i can always cue off of `config.team.name` (a bit ugly) or can we unset this if it's `''`? i don't suppose cloud formation has nullable strings?

akarve: fixed in `static/config.js`; we can never have an empty team id and still have a `team` object.

catalog/static/config.js
```diff
@@ -1,10 +1,12 @@
...
+/* config.js */
+
+/* eslint-disable no-underscore-dangle */
 if (window.location.hostname === 'quiltdata.com') {
   window.__CONFIG = {
     api: 'https://pkg.quiltdata.com',
     stripeKey: 'pk_live_aV44tCGHpBZr5FfFCUqbXqid',
     userApi: 'https://app.quiltdata.com/accounts/api-root',
-    signOutUrl: 'https://app.quiltdata.com/accounts/logout?next=%2F'
+    signOutUrl: 'https://app.quiltdata.com/accounts/logout?next=%2F',
```
dimaryaz: Technically, this is wrong :) Probably breaks IE8 (not that we really care)

catalog/static/config.js
```diff
@@ -14,16 +16,45 @@ if (window.location.hostname === 'quiltdata.com') {
...
+for (let i = 0; i < keys.length; i += 1) {
+  const k = keys[i];
+  const v = window.__CONFIG[k];
+  // eslint-disable-next-line valid-typeof
+  if (k in mustHave && typeof v !== mustHave[k]) {
+    throw new Error(`Unexpected config['${k}'}]: ${v}`);
+  }
+}
+
+// if deployment script sets an empty team.id, there is no team
```
dimaryaz: If we're using the deployment script, then this file isn't used... It'll be `config.js.tmpl`.

catalog/static/config.js
akarve: oh yeah `eslint` loves it though; there *might* be a babel transformer for it. we set IE support at > 10 so should be ok.

akarve: actually i'll just fix this

catalog/app/containers/HomePage/index.js
```diff
@@ -29,9 +29,9 @@ const Styler = styled.div`
 export default class HomePage extends React.PureComponent { // eslint-disable-line react/prefer-stateless-function
   render() {
     const { team = {} } = config;
-    const header = team.name ? team.name : undefined;
-    const tagline = team.name ? 'Team data catalog' : undefined;
-    const signUp = !team.name;
+    const header = team.id ? team.id : undefined;
```
dimaryaz: So just `team` now?

catalog/static/config.js
```diff
@@ -13,17 +16,45 @@ if (window.location.hostname === 'quiltdata.com') {
...
+    // signOutUrl: 'http://localhost:5002/accounts/logout?next=%2F',
 
     // GitHub
     // userApi: 'https://api.github.com/user',
-    // signOutUrl: '/'
+    // signOutUrl: '/',
   };
 }
+
+const mustHave = {
```
dimaryaz: All of this really belongs in `constants/config.js`... But I guess it doesn't hurt here.

catalog/app/containers/HomePage/index.js
akarve: this is a special case because we set a default value with destructuring above. so `team` is already `{}`.

catalog/static/config.js
akarve: agree for future if we end up needing this logic from another file

dimaryaz: I meant, we should be checking `config.js.tmpl`, too. Otherwise, it's kind of silly.

akarve: you're right. coming in clean PR.

PR #413, opened 2018-02-24
Implement team payments

dimaryaz:


registry/quilt_server/views.py
```diff
@@ -1556,21 +1570,26 @@ def profile():
...
-        raise ApiException(requests.codes.forbidden, "Not allowed to switch to plan: %r" % plan)
+    if TEAM_ID:
+        if not g.auth.is_admin:
+            raise ApiException(requests.codes.forbidden, "Only the admin can update plans")
+        # Can only switch to TEAM (from TEAM_UNPAID)
+        # if plan != PaymentPlan.TEAM:
+        if plan not in (PaymentPlan.TEAM, PaymentPlan.TEAM_UNPAID):
+            raise ApiException(requests.codes.forbidden, "Can only switch between team plans")
+    else:
+        if plan not in (PaymentPlan.FREE, PaymentPlan.INDIVIDUAL, PaymentPlan.BUSINESS_ADMIN):
```
akarve: I have dropped support from `BUSINESS` plans completely in the frontend (and don't expect to need them).
If requires a migration to drop then I guess we can do that in a clean PR.

registry/quilt_server/views.py
```diff
@@ -1353,7 +1367,7 @@ def access_delete(owner, package_name, user):
             "Cannot revoke the owner's access"
         )
 
-    if HAVE_PAYMENTS and user == PUBLIC:
+    if HAVE_PAYMENTS and not TEAM_ID and user == PUBLIC:
```
akarve: my main feedback for this block is readability. I realize there's no `else` here, but I think `if user == PUBLIC` would be cleaner on a separate line and address why no else in a comment. Everywhere else in this PR (and now in the frontend, too), the plans logic is much clearer. I would also like to see separate groups of `PaymentPlan`s as I think it's clear that we have two disjoint sets and they belong in separate dicts. Not religious about this but if anyone goes back to touch this code they should see to distinct sets of plans.

registry/tests/team_payments_test.py
```diff
@@ -0,0 +1,197 @@
...
+    @mock_customer(plan=PaymentPlan.TEAM_UNPAID, have_credit_card=True)
+    def testUpdatePlanInvalid(self, customer):
+        admin = 'admin'
+        subscription = customer.subscriptions.data[0]
+        subscription.save.return_value = None
+
+        resp = self.app.post(
+            '/api/payments/update_plan',
+            data=dict(
+                plan=PaymentPlan.INDIVIDUAL.value
```
akarve: this is the test case I was most looking for. And, to the point above, you could test the full cross-plan transitions easily if you had e.g. `TeamPlan` and `CloudPlan` in separate dicts.

registry/quilt_server/views.py
dimaryaz: Doesn't really require a migration (it's not like we have any actual business customers) - but I'll do it in a separate PR just to keep things simpler.

Also, technically, we should update the frontend in prod before we update the backend; it's not good to have the frontend try to use features that are no longer there.

registry/quilt_server/views.py
dimaryaz: Sure

registry/tests/team_payments_test.py
dimaryaz: Well, `PaymentPlan` is a type, not a dict, so it really should contain all possible values. (Of course, it's Python, but in a strictly typed language that would be required.)

We could have sets like `TEAM_PLANS` and `CLOUD_PLANS` in addition to the type.

dimaryaz:
Figured out why CI is failing... I had `STRIPE_SECRET_KEY` set locally. Even though I was mocking the `HAVE_PAYMENTS` constant, it had to effect cause it was used in a function decorator, which had taken effect way before the mock.

I guess that means, we can't use anything that's not a true constant in the decorators...

registry/quilt_server/views.py
```diff
@@ -1556,21 +1570,26 @@ def profile():
...
-@api()
+@api(enabled=HAVE_PAYMENTS)
 @as_json
 def payments_update_plan():
-    if not HAVE_PAYMENTS:
-        raise ApiException(requests.codes.not_found, "Payments not enabled")
-
     plan = request.values.get('plan')
     try:
         plan = PaymentPlan(plan)
```
dimaryaz: Actually, things like this are why `PaymentPlan` needs to contain all possible values.

registry/tests/team_payments_test.py
akarve: subtypes?

dimaryaz: Can't even do that: https://docs.python.org/3/library/enum.html#restricted-subclassing-of-enumerations

dimaryaz: Actually, maybe... I'll play around with that

Issue #314, opened 2018-01-26
Package contents cannot have a node named 'file'


JacksonMaxfield:
Attempting to push a package in python on 2.8.4 fails with traceback:

```
Inferring 'transform: id' for README.md
Copying /home/jovyan/data/README.md...
Traceback (most recent call last):
  File "push_package.py", line 3, in <module>
    quilt.build('aicsjackson/aics_11_test', '/home/jovyan/data/build.yml')
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/command.py", line 425, in build
    _build_internal(package, path, dry_run, env)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/command.py", line 450, in _build_internal
    build_from_path(package, path, dry_run=dry_run, env=env)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/command.py", line 522, in build_from_path
    build_package(owner, pkg, path, dry_run=dry_run, env=env)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/build.py", line 319, in build_package
    checks_contents=checks_contents, dry_run=dry_run, env=env)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/build.py", line 345, in build_package_from_contents
    checks_contents=checks_contents, dry_run=dry_run, env=env)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/build.py", line 110, in _build_node
    checks_contents=checks_contents, dry_run=dry_run, env=env, ancestor_args=group_args)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/build.py", line 110, in _build_node
    checks_contents=checks_contents, dry_run=dry_run, env=env, ancestor_args=group_args)
  File "/opt/conda/lib/python3.6/site-packages/quilt/tools/build.py", line 121, in _build_node
    path = os.path.join(build_dir, rel_path)
  File "/opt/conda/lib/python3.6/posixpath.py", line 92, in join
    genericpath._check_arg_types('join', a, *p)
  File "/opt/conda/lib/python3.6/genericpath.py", line 149, in _check_arg_types
    (funcname, s.__class__.__name__)) from None
TypeError: join() argument must be str or bytes, not 'dict'
```

Code:
```
import quilt

quilt.build('aicsjackson/aics_11_test', '/home/jovyan/data/build.yml')
quilt.push('aicsjackson/aics_11_test', public=True)
```

Python: 3.6.3
[...]

PR #414, opened 2018-02-25
Fix package builds with nodes named 'file'

Kurlov:
- [x] Handle files named 'file.*'

- [x] Support arbitrary node names, including reserved ones, e.g. `file`, `kwargs` et.c.

Currently, node parser fails in case there is a file called 'file.*'. I believe it's related to the #314 issue.

@dimaryaz Please, review it.


Kurlov:
Also, I think I have to mention that I leave package name rule unchanged. In the issue, author also named nodes as 'file', not only files itself.

Is that right that user cannot create something like this?
```
---
contents:
  file: # <- wrong name
    file: data/file.csv
``` 
I just want to make sure even though there is pretty straightforward condition.
https://github.com/quiltdata/quilt/blob/8ad645103a0a1c1c282cc17d12c02d06267b7380/compiler/quilt/tools/build.py#L144

compiler/quilt/tools/build.py
```diff
@@ -63,7 +63,7 @@ def _path_hash(path, transform, kwargs):
     return digest_string(srcinfo)
 
 def _is_internal_node(node):
-    is_leaf = not node or node.get(RESERVED['file'])
+    is_leaf = not node or isinstance(node.get(RESERVED['file']), str)
```
dimaryaz: So what exactly is this fixing? The unit test is passing with or without this change.

dimaryaz:
I think the bug is that the user _can_ end up with something like your `file: file: ...` example. If you have `file.csv` and run `quilt generate`, it will do that.

So it's pretty weird, but I think we need to support it.

compiler/quilt/tools/build.py
Kurlov: Currently without this fix, any 'quilt/compiler/quilt/test/data/file.*' will fail two tests in `test_build.py`, because node parser will think that it reached leaf but in reality it's not.

`node` variable indicates kinda current depth level of the build file, and leaf of the build config might be path to the file, for example, `file: data/foo.csv`. Also, before reaching leaf, node contains all parsed files, for example,`node` for `build.yml` right before going one level lower will look similar to this:
`node = {'README': {'file': 'README.md'}, 'bad': {'file': 'bad.csv'}, 'csv': {'file': 'csv.txt'}, ...}`
But what if there will be a file, e.g., called 'file.csv', then node will look like this:
`node = {'README': {'file': 'README.md'}, 'bad': {'file': 'bad.csv'}, 'csv': {'file': 'csv.txt'}, 'file': {'file': 'file.csv'}, ...}`
In this case, node parser will expose it as leaf since there is a key `file`, but it's not actually a leaf. Thus, dict will pass as path to the file, which will fail on `join()` eventually. Actually, exactly this error was occurred in the issue ticket #314.

I am sorry for many details, I just wanted to be as clear as possible.

Kurlov:
> So it's pretty weird, but I think we need to support it.

Ok, then I implement this on this PR.

Thanks

eode:
Renamed, since this has to do with the node name, not the file name.

eode:
Have docs been updated?  I think they currently note that keywords can't be used as node names.

Also, is this recursive?  We should test to ensure that 'file.file', 'kwargs.file' etc are also permitted.

akarve:
Yeah, good catch. This doc should be updated https://github.com/quiltdata/quilt/blob/master/docs/buildyml.md
- [ ] docs

dimaryaz:
Hey guys,

I've played around with keywords, and it looks like there's actually no way to fix it - our build.yml grammar is just ambigous :-/

E.g.:
```
contents:
  group:
    kwargs:
      cols:
        file: foo.csv
    blah:
      file: whatever.csv
```

Is `kwargs` a group with a file named `cols`? Or is `cols` a parameter to `read_csv` that should be applied to `blah` (and any other files under `group`)? And what if I want both?

I think in the short term, we can patch common problems, but we're gonna have to design a proper grammar.

akarve:
Def need a grammar. Is there something like Relax NG for yaml? Anyway above ambiguity goes away if kwargs is reserved.

PR #428, opened 2018-02-28
Teams config.name + integrity tests

akarve:


catalog/config.js.tmpl
```diff
@@ -3,6 +3,6 @@ window.__CONFIG = {
   api: '$REGISTRY_URL',
   signOutUrl: '$SIGN_OUT_URL',
   stripeKey: '$STRIPE_KEY',
-  team: '$TEAM_ID' ? { id: '$TEAM_ID' } : undefined,
+  team: '$TEAM_ID' ? { id: '$TEAM_ID' name: '$TEAM_NAME'} : undefined,
```
meffij: what happens if TEAM_NAME isn't defined?

akarve: that is a non-issue. missing team.name is a soft failure `check(...error=false)`. the catalog will get an empty string. the more serious issue with this line of code is that it was missing a comma. fixed.

meffij: lol I had one job

catalog/app/containers/Admin/index.js
```diff
@@ -29,7 +29,7 @@ import Status from './Status';
 import Policies from './Policies';
 
 
-const teamName = config.team && config.team.name;
+const teamName = config.team ? config.team.name || config.team.id : '?';
```
meffij: why '?'? why not empty string or undefined or something?

catalog/config.js.tmpl
akarve: still it was a good question. the catalog with fall back on `team.id` in other cases. and i actually added a case for that just now in `Admin/index.js`. The dangerous thing about this file, btw, is that we don't lint it, which I will push a fix for.

PR #429, opened 2018-02-28
Improve admin UX

akarve:
* Rearrange and compact UI
* Fix `config.team.name` validation error
* Fix router back bug from search page (for all users)

### DIFF
--- a/catalog/app/components/MIcon/index.js
+++ b/catalog/app/components/MIcon/index.js
@@ -33,7 +33,7 @@ const Icon = styled(mapProps(omit(['spin']))(FontIcon))`
 `;
 
 // eslint-disable-next-line object-curly-newline
-function MIcon({ drop, title, className, ...rest }) {
+function MIcon({ className, drop, title, ...rest }) {
   return (
     <Adjust drop={drop} title={title}>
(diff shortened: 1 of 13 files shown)


catalog/app/components/MIcon/index.js
```diff
@@ -33,7 +33,7 @@ const Icon = styled(mapProps(omit(['spin']))(FontIcon))`
 `;
 
 // eslint-disable-next-line object-curly-newline
```
meffij: https://eslint.org/docs/rules/object-curly-newline
we could decrease the amount of eslint poop considerably with a more reasonable setting, e.g. consistent=true

catalog/app/containers/SearchResults/index.js
```diff
@@ -33,26 +41,41 @@ export class SearchResults extends React.PureComponent { // eslint-disable-line
...
+          )
+        }
         <br />
       </div>
     );
   }
 }
 
 SearchResults.propTypes = {
+  dispatch: PropTypes.func.isRequired,
```
nl0: seems like `dispatch` is unused

PR #431, opened 2018-02-28
Print a useful message if the team does not exist

dimaryaz:


akarve:
CI is not happy. I think the root cause is the `IOException` isn't catching the socket errors. I've seen both `socket.gaierror` and even `CommandException` thrown (the latter I think b/c we are validating team names elsewhere?)

I'm also seeing, on my local machine, `AdminTestCase.testAdminPackageUserUI` fail.

compiler/quilt/tools/command.py
```diff
@@ -394,6 +395,23 @@ def _check_team_login(team):
...
+    try:
+        socket.gethostbyname(hostname)
+    except IOError:
+        try:
+            # Do we have internet?
+            socket.gethostbyname('quiltdata.com')
+        except IOError:
+            message = "Can't find quiltdata.com. Check your internet connection."
+        else:
+            message = "Can't find the registry at %s. Make sure your team name is correct." % hostname
```
akarve: At this point we know that only the `team_id` part of the registry is bad. So we should say something like:
`Unable to connect to registry. Is the team name {team_id} correct?`

dimaryaz:
I totally screwed up this branch; gonna start over.

PR #433, opened 2018-02-28
Check that team is specified with lowercase letters only

meffij:


compiler/quilt/test/test_command.py
```diff
@@ -280,14 +280,16 @@ def test_login_invalid_team(self, mock_login_with_token, mock_input, mock_open):
 
         mock_input.return_value = old_refresh_token
 
-        with pytest.raises(command.CommandException, match='Invalid team name'):
+        with pytest.raises(command.CommandException,
+                match='The team you specified is not a valid team.'):
```
akarve: this string should really be a local constant. it's repeated below.

compiler/quilt/tools/command.py
```diff
@@ -394,12 +394,22 @@ def _check_team_login(team):
                 "Can't log in as a public user; log out from team %r first." % existing_team
             )
 
+team_regex = re.compile('^[a-z]+$')
+def _check_team_id(team):
+    if team is not None and team_regex.match(team) is None:
+        raise CommandException(
+            "The team you specified is not a valid team. Teams are specified " +
```
akarve: Brevity is clarity: `Invalid team name: {team}. Lowercase letters only.`

akarve: It's critical to report back the invalid name so people know where to fix the error.

PR #449, opened 2018-03-02
Don't expire objects on commit

dimaryaz:
Turns out, by default, SQLAlchemy expires all objects on commit, and causes them to be re-queried if they're accessed again.
It's particularly bad in case of `package_get`: we commit an audit event, then access blob sizes. If we have 10,000 blobs, SQLAlchemy will issue 10,000 individual queries to look up the size of each one.

We don't depend on the objects getting updated when they're commited, to let's disable this "feature".


kevinemoore:
Unless I'm misunderstanding something, I don't think this will fix the main issue in the package get case. Won't it be much faster to replace the 10k individual queries with a single one?

PR #453, opened 2018-03-02
Testing: Re-add build tests for bad file, bad transform

eode:
* Re-added test for bad file (removed when hdf5 test changed)
* Re added test for bad transform (removed when hdf5 test changed)
* Fix `AttributeError` raised instead of `BuildException` when `_is_internal_node()` encounters an incorrect type
* Note: this fixes Asana issue "Error message needs improvement" as a side-effect.


compiler/quilt/tools/build.py
```diff
@@ -63,7 +63,10 @@ def _path_hash(path, transform, kwargs):
     return digest_string(srcinfo)
 
 def _is_internal_node(node):
-    is_leaf = not node or node.get(RESERVED['file'])
+    try:
+        is_leaf = not node or node.get(RESERVED['file'])
+    except AttributeError:
+        raise BuildException("Invalid yml for quilt: expected node data, but got {!r} instead".format(node))
```
akarve: ~~for quilt~~
~~but~~

akarve: So we want: `Invalid syntax for <FILE>: expected node data, got {!r}".format(node))`

akarve:
@eode OK to merge after above string change and conflict resolution

eode:
This now also fixes a couple of issues that were added recently (both introduced when we started permitting nodes that were named identically to `RESERVED` words):

* `'kwargs'` and other `RESERVED` words may end up as a node
* Invalid node info is ignored rather than causing an error

..these were encountered when merging the recent changes.

eode:
Approved / Aneesh (slack)

PR #454, opened 2018-03-02
Return the total package size in package_preview endpoint.

dimaryaz:


dimaryaz:
Hm, I'll make it `total_size_uncompressed`; I think bytes is the only thing that makes sense for size. For # of objects, we can call it "object_count" or whatever.

PR #457, opened 2018-03-03
Add is_team and is_public to package_preview

meffij:
- [ ] unit tests

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -886,6 +886,11 @@ def package_preview(owner, package_name, package_hash):
     instance = _get_instance(g.auth, owner, package_name, package_hash)
     assert isinstance(instance.contents, RootNode)
 
+    package = _get_package(g.auth, owner, package_name)
+    users = [access.user for access in package.access]
+    is_public = PUBLIC in users
+    is_team = TEAM in users
+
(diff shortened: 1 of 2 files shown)


registry/quilt_server/views.py
```diff
@@ -886,6 +886,11 @@ def package_preview(owner, package_name, package_hash):
     instance = _get_instance(g.auth, owner, package_name, package_hash)
     assert isinstance(instance.contents, RootNode)
 
+    package = _get_package(g.auth, owner, package_name)
```
dimaryaz: `instance.package`

dimaryaz: Also, see the `packages` query in `search()`; you can try to change `_get_instance` to return the instance as well as `is_team` and `is_public`

registry/quilt_server/views.py
```diff
@@ -883,7 +883,20 @@ def _generate_preview(node, max_depth=PREVIEW_MAX_DEPTH):
 @api(require_login=False)
 @as_json
 def package_preview(owner, package_name, package_hash):
-    instance = _get_instance(g.auth, owner, package_name, package_hash)
+    (instance, is_public, is_team) = (
```
dimaryaz: This will fail if there are no results

PR #458, opened 2018-03-03
Show package size on package page

akarve:


catalog/app/containers/Package/index.js
```diff
@@ -199,7 +201,24 @@ Install.propTypes = {
   owner: PropTypes.string.isRequired,
 };
 
-const UpdateInfo = ({ author, date, version }) => (
+function readableBytes(bytes) {
+  const mb = bytes / 1000000;
+  const fracStr = (mb - Math.floor(mb)).toFixed(6);
```
dimaryaz: Move it inside the `if`? In fact, if `mb < 1`, you don't need to subtract anything.

catalog/app/containers/Package/index.js
```diff
@@ -199,7 +201,24 @@ Install.propTypes = {
...
+  if (mb < 1) {
+    // show fractional MB to 3 digits
+    return `${fracStr} MB`;
+  }
+  // only show whole MB
+  return `${numberToCommaString(Math.floor(mb))} MB`;
```
dimaryaz: `floor` or `round`?

catalog/app/utils/string.js
```diff
@@ -10,6 +10,20 @@ export function makeMatcher(exp, flags = 'i') {
   return (s) => re.test(s);
 }
 
+export function numberToCommaString(num) {
+  const digs = num.toString().split('');
+  return arrayToCommaString(digs);
+}
+
+function arrayToCommaString(digs) {
```
dimaryaz: Uhh... `digs.toLocaleString("en-US")`?

PR #470, opened 2018-03-07
Test script to populate a registry with READMEs

meffij:


akarve:
Let's move this to a separate scripts repo

PR #479, opened 2018-03-09
Add timeout to build logging

meffij:


compiler/quilt/tools/command.py
```diff
@@ -510,7 +511,7 @@ def _log(team, **kwargs):
     if cfg.get('disable_analytics'):
         return
 
-    session = _get_session(team)
+    session = _get_session(team, timeout=LOG_TIMEOUT)
```
dimaryaz: Need to catch the exception. Move this into the try/except?

compiler/quilt/tools/command.py
```diff
@@ -510,12 +511,12 @@ def _log(team, **kwargs):
     if cfg.get('disable_analytics'):
         return
 
-    session = _get_session(team)
+    try:
+        session = _get_session(team, timeout=LOG_TIMEOUT)
```
dimaryaz: You also need to modify the `finally` block to check if `session` is set. Anyways, make sure to actually test the timeouts - just set it to 0.001 or something.

meffij: Tested and works

compiler/quilt/tools/command.py
```diff
@@ -212,9 +212,10 @@ def config():
     global _registry_url
     _registry_url = None
 
-def _update_auth(team, refresh_token):
+def _update_auth(team, refresh_token, timeout=None):
     response = requests.post("%s/api/token" % get_registry_url(team), data=dict(
-        refresh_token=refresh_token
+        refresh_token=refresh_token,
+        timeout=timeout
```
dimaryaz: Wait a minute... You're passing it as a POST parameter. There's no way that can work.

PR #483, opened 2018-03-10
Use Raven on npm (instead of script)

akarve:
* Strip useless `onReject`
* Lint


akarve:
For non-cache case it saves a script load. It's now part of our JS bundle with very little impact on chunk size.

PR #484, opened 2018-03-10
Make audit CLI output nice

meffij:


compiler/quilt/tools/command.py
```diff
@@ -1377,6 +1377,30 @@ def list_users(team=None):
     resp = session.get('%s/api/users/list' % url)
     return resp.json()
 
+def _print_table(table, padding=2):
```
dimaryaz: This whole function can actually be done in like three lines:
```
cols_width = [max(len(word) for word in col) for col in zip(*table)]
for row in table:
  print((" " * padding).join(word.ljust(width) for word, width in zip(row, cols_width)))

File: compiler/quilt/tools/command.py [hunk 2]
```diff
@@ -1444,7 +1468,23 @@ def audit(user_or_package):
...
+    team = _find_logged_in_team()
+    teamstr = '%s:' % team
+    rows = [['Time', 'User', 'Package', 'Type']]
+    for item in events:
+        time = item.get('created')
+        pretty_time = datetime.fromtimestamp(time).strftime("%Y-%m-%d %H:%M:%S")
+        user = item.get('user')
+        pkg = '%s%s/%s' % (teamstr, item.get('package_owner'), item.get('package_name'))
+        t = item.get('type')
+        i = list(map(str, (pretty_time, user, pkg, t)))
```
dimaryaz: That's kinda hacky... Just do the same thing as in `_cli_list_users`

PR #486, opened 2018-03-12
Community profile

meffij:


docs/CONTRIBUTING
```diff
@@ -0,0 +1,8 @@
+### Getting Started
+See `registry/README.md` for instructions to run a local registry. Quilt runs in both Python 2.7 and Python 3.
```
akarve: please break out into getting started for catalog, registry, and compiler (not just registry)

PR #487, opened 2018-03-13
Add homepage calls to action

akarve:
And update AICS package on homepage


catalog/app/components/CallToAction/tests/index.test.js
```diff
@@ -0,0 +1,10 @@
+// import React from 'react';
+// import { shallow } from 'enzyme';
+
+// import CallToAction from '../index';
+
+describe('<CallToAction />', () => {
+  it('Expect to have unit tests specified', () => {
+    expect(true).toEqual(false);
```
meffij: what

PR #489, opened 2018-03-13
Add contents and packages checks for .yml  [Quick review]

Kurlov:
This PR add a BuildException raise in case neither `contents` nor `packages` node provided

### DIFF
--- a/compiler/quilt/test/build_no_contents_node.yml
+++ b/compiler/quilt/test/build_no_contents_node.yml
@@ -1,2 +1,3 @@
 does_not:
   contain: contents.node
+  contents: foo.txt
--- a/compiler/quilt/test/build_no_contents_node.yml
+++ b/compiler/quilt/test/build_no_contents_node.yml
@@ -1,0 +1,2 @@
+does_not:
+  contain: contents.node
--- a/compiler/quilt/test/test_build.py
+++ b/compiler/quilt/test/test_build.py
@@ -266,7 +266,7 @@ class BuildTest(QuiltTestCase):
 
     def test_build_no_contents_node(self):
         """
-        Attempt to build a yml file without contents node nor packages
+        Attempt to build a yml file without contents node
         """
         mydir = os.path.dirname(__file__)
         path = os.path.join(mydir, './build_no_contents_node.yml')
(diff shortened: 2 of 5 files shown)


compiler/quilt/tools/build.py
```diff
@@ -527,6 +527,8 @@ def load_yaml(filename, optional=False):
         data = fd.read()
     try:
         res = yaml.load(data)
+        if not any(k in res.keys() for k in ['contents', 'packages']):
+            raise BuildException('"contents" or "packages" node must be specified')
```
akarve: Let's break this out into two cases:
* `Error in FILENAME: missing 'contents' node`
* `Error in FILENAME: missing 'pacakges' node`

akarve: Also note, 'packages' is an entirely different use case. That's only for `quilt.yml`.

compiler/quilt/test/build_no_contents_node.yml
```diff
@@ -0,0 +1,2 @@
+does_not:
+  contain: contents.node
```
akarve: for proper test coverage we should have a build file without 'contents', a `quilt.yml` file without 'package'

Kurlov: `packages` exception tests in `test_command.py:test_install_dependencies_from_invalid_file` where temporary `.yml` created.

Kurlov:
I split checks for `packages` and `contents`. I think lack of `packages` is more about `command.install` error, thus it's `CommandException`. 
Please, tell me if I get it wrong way.

compiler/quilt/test/test_build.py
```diff
@@ -264,6 +264,16 @@ def test_build_yaml_syntax_error(self):
         with assertRaisesRegex(self, build.BuildException, r'Bad yaml syntax.*build_bad_syntax\.yml'):
             build.build_package(None, 'test_syntax_error', PACKAGE, path)
 
+    def test_build_no_contents_node(self):
+        """
+        Attempt to build a yml file without contents node nor packages
```
eode: Docstring inaccurate:  You tested 'packages' elsewhere

PR #490, opened 2018-03-14
Search the node names and hashes in the contents JSON

dimaryaz:


registry/scripts/backfill_keywords_tsv.py
```diff
@@ -0,0 +1,33 @@
...
+        db.session.query(Instance, Package.owner, Package.name)
+        .join(Instance.package)
+        .options(undefer('contents'))
+        .filter(Instance.keywords_tsv.is_(None))
+    )
+
+    for instance, owner, name in rows:
+        instance.keywords_tsv = keywords_tsvector(owner, name, instance.contents)
+
+    db.session.commit()
```
asah: probably fine for our scale, but JFYI in the future you might want to commit() after every N records, say N=1000 or so.

dimaryaz: Yep... But we only have 612 rows right now :)

PR #502, opened 2018-03-19
Fix GroupNode's preorder method

dimaryaz:
Currently, it's:
1) returning nodes in the wrong order
2) not even guaranteeing any particular order cause dictionaries are unordered

Fix that, and also make it return a generator instead of a list.

### DIFF
--- a/compiler/quilt/test/test_core.py
+++ b/compiler/quilt/test/test_core.py
@@ -1,0 +1,37 @@
+"""
+Test core.py
+"""
+import os
+
+import quilt
+from ..tools.core import GroupNode, RootNode, FileNode
+from .utils import QuiltTestCase
+
+class CoreTest(QuiltTestCase):
+    def test_compiler_registry_identical(self):
+        """
+        Verify that the compiler and the registry have the same core.py.
+        Not the best way to accomplish this, but oh well...
+        """
+        quilt_dir = os.path.dirname(quilt.__file__)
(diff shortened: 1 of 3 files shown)


compiler/quilt/tools/core.py
```diff
@@ -52,18 +51,14 @@ def __init__(self, children):
     def preorder(self):
         """
         Performs a pre-order walk of the package tree starting at this node.
-        It returns a list of the nodes in the order visited.
+        It returns a generator of the nodes in the order visited.
```
kevinemoore: Good idea to have these methods return a generator.

compiler/quilt/tools/core.py
```diff
@@ -52,18 +51,14 @@ def __init__(self, children):
...
         while stack:
             node = stack.pop()
-            for child in itervalues(node.children):
-                output.append(child)
-                if isinstance(child, GroupNode):
-                    stack.append(child)
-        return output
+            yield node
+            if isinstance(node, GroupNode):
+                stack.extend(child for key, child in sorted(iteritems(node.children), reverse=True))
```
kevinemoore: Sorting is overkill for pre-order.

compiler/quilt/test/test_core.py
```diff
@@ -0,0 +1,37 @@
+"""
```
kevinemoore: This doesn't seem like the right way to enforce that the registry and compiler are using the same core.py. How about a commit hook? Or, better yet, having a single file and import it into both.

PR #504, opened 2018-03-19
Document _set

meffij:
Document _set method of PackageNode

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -117,7 +117,7 @@ In [9]: wine._group_keys()
 Out[9]: ['raw', 'tables']
 ```
 ### Editing a package
-* `NODE._set(path, value)` sets a child node. `path` is an array of strings, one for each level of the tree. `value` is the new value. If it's a Pandas dataframe, it will be serialized. A string will be interpreted as a path to a file that contains the data to be packaged.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -116,3 +116,16 @@ Out[8]: ['README']
 In [9]: wine._group_keys()
 Out[9]: ['raw', 'tables']
 ```
+### Editing a package
+* `NODE._set(path, value)` sets a child node. `path` is an array of strings, one for each level of the tree. `value` is the new value. If it's a Pandas dataframe, it will be serialized. A string will be interpreted as a path to a file that contains the data to be packaged.
```
akarve: use CAPS for meta-syntax

docs/api.md
```diff
@@ -116,3 +116,16 @@ Out[8]: ['README']
...
+
+#### Example
+```
+import pandas as pd
+import quilt
+quilt.build('USER/PKG') # create new, empty packckage
+from quilt.data.USER import PKG as pkg
+pkg._set(['data'], pd.DataFrame(data=[1, 2, 3]))
```
akarve: include example of file blob, there's one here: https://github.com/quiltdata/examples/blob/master/Examples.ipynb

probably a good idea to link out to that repo for more examples

docs/api.md
```diff
@@ -116,3 +116,16 @@ Out[8]: ['README']
 In [9]: wine._group_keys()
 Out[9]: ['raw', 'tables']
 ```
+### Editing a package
```
akarve: Match tense/plurality of other sections: `Editing package contents`

PR #506, opened 2018-03-21
Improve the keywords backfill script

dimaryaz:
- Show what's being processed
- Commit every 100 rows

### DIFF
--- a/registry/scripts/backfill_keywords_tsv.py
+++ b/registry/scripts/backfill_keywords_tsv.py
@@ -6,11 +6,10 @@ Backfills s3_blob.preview by downloading the contents from S3.
 
 import sys
 
-import sqlalchemy as sa
 from sqlalchemy.orm import undefer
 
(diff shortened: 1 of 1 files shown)


registry/scripts/backfill_keywords_tsv.py
```diff
@@ -21,8 +21,11 @@ def main(argv):
         .filter(Instance.keywords_tsv.is_(None))
     )
 
-    for instance, owner, name in rows:
+    for idx, (instance, owner, name) in enumerate(rows):
+        print("%s/%s/%s" % (owner, name, instance.hash))
```
akarve: extremely minor: if you use `%s/%s:%s` it will be consistent with install syntax. i guess we won't need to install from the logs.

asah:
looks good re committing every 100 updates... I assume y'all are comfortable with the consistency issues (i.e. it's idempotent? so just re-run...)

Issue #427, opened 2018-02-27
Summary Stats for Packages


JacksonMaxfield:
A majority of my packages are quite large so the largest desire is for there to be at least an indication of package size before install on the Quilt registry page.

If this is the current package stats area:
```
Latest update

date
    1/10/2018, 2:07:10 PM
author
    @aicsjackson
version
    2aed0783417118de00913d2ffe7be9968c881eb990993d6f5fdb957d0f49354
```

Would be nice to also see items like:
```
Latest update

date
    1/10/2018, 2:07:10 PM
author
    @aicsjackson
version
    2aed0783417118de00913d2ffe7be9968c881eb990993d6f5fdb957d0f49354
size
    200 GB
files
    tiff | 400
    json | 1600
```

Would allow for better insight into the package contents prior to downloading.

akarve:
@dimaryaz would you add `size` and `leaf_node_count` to `/api/package_preview`?
@JacksonMaxfield counting file types is a bit tricky since some nodes (e.g. in-memory data frames pushed into a package) don't have types.

akarve:
Actually, if possible let's do `size` (total package size) and a dict of file_types (deuced from `q_ext`). I guess we can have `__NO_EXT__` key or something like that for pure dataframes or weird files without a `.EXT`.

akarve:
Sizes are live in production. File types are next. Example: https://quiltdata.com/package/aicsjackson/prod/

akarve:
Complete as of  `master` 6933bab. Will be released to production in next release.

PR #508, opened 2018-03-21
Show file stats, nice sizes

akarve:


akarve:
Resolves #427

catalog/app/containers/Package/index.js
```diff
@@ -224,20 +225,49 @@ Install.propTypes = {
   owner: PropTypes.string.isRequired,
 };
 
+const sizes = [
```
dimaryaz: This feels overly complicated/verbose... Just do this? https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript?answertab=active#tab-top

(Replace 1024 with 1000 or whatever.)

catalog/app/containers/Package/index.js
```diff
@@ -224,20 +225,49 @@ Install.propTypes = {
...
+  { lim: 3, str: 'B' },
+  { lim: 6, str: 'kB' },
```
dimaryaz: Is `k` intentionally lowercase?

catalog/app/containers/Package/index.js
```diff
@@ -224,20 +225,49 @@ Install.propTypes = {
...
+    display: inline-block;
+    width: 5em;
+  }
+`;
+function readableExtensions(fileCounts = {}) {
+  const keys = Object.keys(fileCounts);
+  keys.sort();
+  return keys.map((k) => {
+    const count = numberToCommaString(fileCounts[k]);
+    return <Line><span>.{k || 'None'}</span>{count}</Line>;
```
dimaryaz: Probably don't want `.None`... But actually, if I use `pathlib`, it'll give me extensions with leading dots - so I'll just return those to you.

catalog/app/containers/Package/index.js
akarve: yeah i double-checked

catalog/app/containers/Package/index.js
akarve: apparently `**` is faster than `Math.pow` (linter doesn't like `Math.pow` either). i think my solution should be the SO answer lol. seemed clean to me :) what we actually need are unit tests and @nl0 is working on those. unless this code is really bad would prefer to leave as is.

akarve: anyway, `**` is not really the issue. if you prefer i can take the log and jump right to the correct units if stepping through the array is hard to read.

dimaryaz: I didn't even mean `**` vs `pow` or `log`. I meant, the `lim: 3`, `lim: 6`, etc. are unnecessary - you can just use an array?

akarve: yeah, done

catalog/app/containers/Package/index.js
```diff
@@ -224,20 +223,44 @@ Install.propTypes = {
...
-  const mb = bytes / 1000000;
-  if (mb < 1) {
-    const fracStr = (mb - Math.floor(mb)).toFixed(6);
-    // show fractional MB to 6 sig-digits
-    return `${fracStr} MB`;
+  if (Number.isInteger(bytes) && bytes > -1) {
+    // https://en.wikipedia.org/wiki/Kilobyte
+    const sizes = ['', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'];
+    const log = bytes === 0 ? 0 : Math.log10(bytes);
+    let index = Math.floor(log / 3);
```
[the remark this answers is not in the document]
akarve: yeah, `min` is nicer

PR #509, opened 2018-03-21
Upgrade SQLAlchemy and use the new "pre-ping" option

dimaryaz:
Currently, if the DB connection goes away (DB gets restarted, network settings change, etc.), SQLAlchemy will still try to use the stale connection, fail, and remove the stale connection out of the pool. However, because we have 16 registry processes, we can expect 16 exceptions (each one causing a 500 error) when this happens.

SQLAlchemy has a new option to "pre-ping" a connection (using a `SELECT 1` query) before trying to use it. It seems to have negligible overhead, so let's use it.

Of course, upgrading SQLAlchemy is a slightly risky change.

### DIFF
--- a/registry/quilt_server/__init__.py
+++ b/registry/quilt_server/__init__.py
@@ -26,11 +26,14 @@ app.config.from_envvar('QUILT_SERVER_CONFIG')
 class QuiltSQLAlchemy(SQLAlchemy):
     def apply_driver_hacks(self, app, info, options):
         """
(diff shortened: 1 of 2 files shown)


tlbjr:
SQLA pre ping docs: http://docs.sqlalchemy.org/en/latest/core/pooling.html#dealing-with-disconnects
As an aside, I anxiously awaited SQLA 1.2 for just this feature. The workarounds pre-1.2 were less elegant.

PR #510, opened 2018-03-21
Use an existing helper function for finding hashes instead of preorder

dimaryaz:
Preorder is unnecessary, and the existing function makes things much simpler.


dimaryaz:
Actually, I'll just do everything in one PR; splitting it up only makes things harder to test.

PR #520, opened 2018-03-26
Fix the "preorder" method and use it in "find_object_hashes"

dimaryaz:
- Fix node order in group dataframes: sort child nodes alphabetically and add a unit test
- Use the "find_object_hashes" in a few places instead of calling "preorder" directly
- Delete a redundant teams test: it wasn't testing anything new


dimaryaz:
Oh, true, `preorder` actually makes sense again. (In my first, non-method version it was only returning the leaf nodes - but now it returns everything again, so it is in fact a preorder.)

PR #521, opened 2018-03-26
Fix pipes inside code snippets

meffij:
Fixes formatting in docs


meffij:
￨
The above character is used as a homograph of the pipe character, which has a syntactic purpose of delimiting tables. Since GitBook and GitHub handle pipes in code snippets differently, we use a homograph to sidestep the inconsistency.

PR #523, opened 2018-03-27
Document remaining underscore methods

meffij:


docs/api.md
```diff
@@ -116,8 +117,10 @@ Out[8]: ['README']
 In [9]: wine._group_keys()
 Out[9]: ['raw', 'tables']
 ```
+
 ### Editing Package Contents
-* `NODE._set(PATH, VALUE)` sets a child node. `PATH` is an array of strings, one for each level of the tree. `VALUE` is the new value. If it's a Pandas dataframe, it will be serialized. A string will be interpreted as a path to a file that contains the data to be packaged.
+* `PACKAGENODE._set(PATH, VALUE)` sets a child node. `PATH` is an array of strings, one for each level of the tree. `VALUE` is the new value. If it's a Pandas dataframe, it will be serialized. A string will be interpreted as a path to a file that contains the data to be packaged.
```
akarve: specify "local path to a file... of any format. (Common columnar formats will be transformed into CSVs. All other file formats, e.g. images, will be copied as is.)

docs/api.md
```diff
@@ -119,7 +119,7 @@ Out[9]: ['raw', 'tables']
 ```
 
 ### Editing Package Contents
-* `PACKAGENODE._set(PATH, VALUE)` sets a child node. `PATH` is an array of strings, one for each level of the tree. `VALUE` is the new value. If it's a Pandas dataframe, it will be serialized. A string will be interpreted as a path to a file that contains the data to be packaged.
+* `PACKAGENODE._set(PATH, VALUE)` sets a child node. `PATH` is an array of strings, one for each level of the tree. `VALUE` is the new value. If it's a Pandas dataframe, it will be serialized. A string will be interpreted as a path to a file that contains the data to be packaged (common columnar formats will be transformed into CSVs. All other file formats, e.g. images, will be copied as-is).
```
akarve: Almost. New sentence:
(Common columnar formats will be transformed into data frames. All other file formats, e.g. images, will be copied as-is).

PR #530, opened 2018-03-30
Better logs

meffij:
TODO: 
- [x] CLI augmentation to display tags + versions

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -956,6 +956,13 @@ def package_preview(owner, package_name, package_hash):
     (instance, is_public, is_team) = result
     assert isinstance(instance.contents, RootNode)
 
+    log_count = (
+        db.session.query(
+            sa.func.count(Log.instance_id)
+        )
+        .filter(Log.instance_id==instance.id)
+    ).first()[0]
+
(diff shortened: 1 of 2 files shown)


registry/quilt_server/views.py
```diff
@@ -956,6 +956,13 @@ def package_preview(owner, package_name, package_hash):
     (instance, is_public, is_team) = result
     assert isinstance(instance.contents, RootNode)
 
+    log_count = (
+        db.session.query(
+            sa.func.count(Log.instance_id)
+        )
+        .filter(Log.instance_id==instance.id)
```
dimaryaz: Don't think this does what you want: this will count how many times a given hash has been updated (which means, no-op or metadata-only updates). I'd assume you want the count for the package?

dimaryaz: (Also, you can do `.filter_by(...)` here - a bit cleaner.)

registry/quilt_server/views.py
```diff
@@ -956,6 +956,13 @@ def package_preview(owner, package_name, package_hash):
...
+    ).first()[0]
```
dimaryaz: Use the `.one()` selector instead.

PR #533, opened 2018-03-31
Fix update timestampts on instances

dimaryaz:
- Stop using SQLAlchemy's `onupdate` hook; it's not worth it - just update the timestamp manually
- Add a script to fix incorrect timestamps

### DIFF
--- a/registry/scripts/fix_updated_at.py
+++ b/registry/scripts/fix_updated_at.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+
+"""
+Rebuilds the full-text search index (to fix the language).
+"""
+
+import sys
+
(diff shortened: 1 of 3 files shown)


registry/scripts/fix_updated_at.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+Rebuilds the full-text search index (to fix the language).
+"""
+
+import sys
+
+from quilt_server import db
+
+def main(argv):
+    result = db.engine.execute('''
```
akarve: i guess batching is overkill even for prod?

dimaryaz: Yeah... It's still under 1000 rows.

Issue #494, opened 2018-03-15
Group Nodes as Iterators


JacksonMaxfield:
Currently when I want to loop through a list of nodes contained in a parent node I have to do something like this:

```python
for node_name in pkg.group_node._group_keys():
   actual_node = getattr(pkg.group_node, node_name)
   # do actual work with the node here
```

A nice touch would be to make GroupNode, and I guess PackageNode as well, work like iterators and when I want to loop through something it gives me the node instead of my having to do a roundabout `getattr`.

```python
for node in pkg.group_node:
   # do work
```

dimaryaz:
An iterator sounds like a good idea - but I think it should return all child nodes, not just the ones that are also groups?

That means, the code would look like this:
```
for node in pkg.group_node:
  if isinstance(node, quilt.nodes.GroupNode):
    ...
```

If `isinstance` feels too verbose, we could add some helper functions to nodes - e.g. `_is_group()`?

JacksonMaxfield:
Sorry about my original wording. I agree it should return all child nodes, not just group nodes.

JacksonMaxfield:
@dimaryaz and @akarve on this thread another nice addition would be an `items()` call, depending on the implementation of iterators I just got [this working](https://github.com/AllenCellModeling/QuiltLoader/blob/master/QuiltLoader.py#L128) also as a side note, if the implementations in there look good I can find out where to place them in Quilt Nodes and add them. Just wanted feedback before I toss them in.

akarve:
Resolved in #649.

PR #534, opened 2018-04-01
GroupNode iter

kurlov:
### Iteration

Add iteration over GroupNode. Relates to #494 

Example:
```
from quilt.data.test import group_node
        for node in group_node:
            assert isinstance(node, GroupNode)
```

### DIFF
--- a/compiler/quilt/nodes.py
+++ b/compiler/quilt/nodes.py
@@ -67,6 +67,10 @@ class GroupNode(DataNode):
         data_info = '\n'.join(sorted(self._data_keys()))
         return '%s\n%s%s' % (pinfo, group_info, data_info)
 
+    def __iter__(self):
+        for child in self._group_keys():
+            yield getattr(self, child)
+
     def _items(self):
         return ((name, child) for name, child in iteritems(self.__dict__)
                 if not name.startswith('_'))
--- a/compiler/quilt/test/test_import.py
+++ b/compiler/quilt/test/test_import.py
@@ -41,6 +41,13 @@ class ImportTest(QuiltTestCase):
         assert set(dataframes._group_keys()) == set()
         assert set(dataframes._data_keys()) == {'csv', 'nulls'}
 
+        # this should contain elements
+        for node in package:
+            assert node
+        # iter over an empty GroupNode
+        for node in dataframes:
+            assert not node
+
         assert isinstance(README(), string_types)
         assert isinstance(README._data(), string_types)
         assert isinstance(dataframes.csv(), pd.DataFrame)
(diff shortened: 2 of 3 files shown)


compiler/quilt/nodes.py
```diff
@@ -67,6 +67,10 @@ def __repr__(self):
         data_info = '\n'.join(sorted(self._data_keys()))
         return '%s\n%s%s' % (pinfo, group_info, data_info)
 
+    def __iter__(self):
+        for child in self._group_keys():
```
dimaryaz: We should actually return all nodes, not just groups. (I've commented on the bug report, too.)

compiler/quilt/test/test_import.py
```diff
@@ -41,6 +41,13 @@ def test_imports(self):
         assert set(dataframes._group_keys()) == set()
         assert set(dataframes._data_keys()) == {'csv', 'nulls'}
 
+        # this should contain elements
+        for node in package:
+            assert node
```
dimaryaz: This doesn't actually check that package contains elements; it won't run at all if the iterator doesn't return anything.

Do something like `assert list(package)` instead.

compiler/quilt/test/test_import.py
```diff
@@ -41,6 +41,13 @@ def test_imports(self):
...
+        # iter over an empty GroupNode
+        for node in dataframes:
+            assert not node
```
dimaryaz: This sort of words... But `assert not list(dataframes)` would be better.

eode:
Hrm.  In #385 , I've already extended 'items' in a way that's similar, because I figured this would be needed at some point.  Looks like slowness of the merge of that made this necessary here..

API I used there:

```python
from quilt.data.user import package

for name, node in package._items():
    # do things

# or, if recursivity is required:
for name, node in package._items(recursive=True):
    # do things
```

Kevin noted the existence of this to me over in the Export PR.

eode:
Mine does *not* distinguish node types, it simply returns all node types.

kurlov:
I updated PR, although I am not sure is it relevant due to #385 . On the other hand , I guess it's nice to have   implicit `__iter__` based iteration. 

@dimaryaz Please, review changes.

Thanks!

akarve:
@aicsjackson does this meet #494?

JacksonMaxfield:
This works well! Thanks.

akarve:
@kurlov Please document the new features in `/docs` and add to this PR

eode:
@JacksonMaxfield We're going to have a significant API rewrite soon, which also involves some internal code deprecation.  My code was voted off the island, as it uses the current `quilt.nodes` module, which is planned to go the way of the dinosaur.  That's not to say **a** `quilt.nodes` module won't exist, just that that particular implementation is planned to be replaced.

PR #537, opened 2018-04-03
build.py uncaught exception fix, other minor misc from export pr

eode:
### Summary
Part of export PR, part of TensorFlow branch.

Updates to `is_nodename`, `to_nodename`, `is_identifier`, and `to_identifier`, and usage in `build.py`, toward the end of a centralized place to define what a Node Name is (separately from what a Python Identifier is).

At risk of being pedantic -- Node Names and Python Identifiers aren't the same thing.  What constitutes a Node Name may change, but what constitutes a Python Identifier really doesn't.  The changes to build.py in this PR are a start for handling changes to Node Name definition in a graceful way that doesn't interfere with things that still need to be Python Identifiers.  Case in point -- as we switch to `node['item']` notation, what constitutes a Node Name is changing.  Anything that uses "is_nodename" or "to_nodename" won't need to change, or will need less change, because it can be changed directly in `util.py`.

build.py:
- `_pythonize_name()`:  Now uses `to_identifier` instead of a separate implementation that then checks `is_nodename()`
- `to_nodename()`: catch ValueError and reraise as BuildException

util.py:
- Import ordering and minor whitespace fix
- docstring fixes and updates
- Prevent Python keywords for `to/is_nodename()`in an easily-removable way, since we can't handle them in actual Node Names (yet)
- Added TODO: items for changes needed once `node['item']` notation is working and keywords can be permitted.
- minor fixes and improvements, duplication avoidance, speed improvement during deduplication..


compiler/quilt/tools/util.py
```diff
@@ -166,19 +169,21 @@ def is_nodename(string):
...
+    # * Must not be a python keyword  (limitation of current implementation)
     # * Must not start with an underscore
     if string.startswith('_'):
         return False
-    return bool(PYTHON_IDENTIFIER_RE.match(string))
+    return is_identifier(string, permit_keyword=False)
 
 
-def to_identifier(string):
+def to_identifier(string, permit_keyword=False, strip_underscores=False):
```
dimaryaz: Are we ever planning to using both `permit_keyword=False` and `permit_keyword=True`?

If I understand correctly, it's always `False` right now, and you're planning to switch it to always `True` later? If so, it shouldn't be a parameter - just hardcode the implementation, and then later change it.

compiler/quilt/tools/build.py
```diff
@@ -80,13 +81,19 @@ def _is_valid_group(group):
...
+    """Convert `name` to a Python Identifier, but raise BuildException on failure."""
+    try:
+        safename = to_identifier(name, strip_underscores=True)
+    except ValueError as error:
+        raise BuildException("Invalid name: " + str(error))
+    return safename
 
-    if not is_nodename(safename):
-        raise BuildException("Unable to determine a Python-legal name for %r" % name)
+def to_nodename(name, invalid=None):
```
dimaryaz: I'm worried that having both `util.to_nodename` and `build.to_nodename` is going to create confusion. I'd make this one "private" - i.e., call it `_to_nodename`, or even something more obscure.

eode: `_to_nodename` is what I did originally, but I got called out by Kevin for doing it the other way, because it then involved renaming existing usages -- but, that was in the other (larger) PR, and I think it was just TMI there.  I'll switch it to `to_nodename_reraise` or something similar.

compiler/quilt/tools/util.py
eode: Hrm.. ..perhaps not.  I'll hardcode it.

compiler/quilt/tools/build.py
```diff
@@ -80,13 +80,19 @@ def _is_valid_group(group):
...
+    """Convert `name` to a Python Identifier, but raise BuildException on failure."""
+    try:
+        safename = to_identifier(name, strip_underscores=True)
+    except ValueError as error:
+        raise BuildException("Invalid name: " + str(error))
+    return safename
 
-    if not is_nodename(safename):
-        raise BuildException("Unable to determine a Python-legal name for %r" % name)
+def to_nodename_reraise(name, invalid=None):
```
dimaryaz: This one still bothers me... We shouldn't have to wrap our own helper functions just to raise a different exception. Maybe we should add a generic exception that we catch in `main`?
(@kevinemoore, @akarve - any thoughts?)

compiler/quilt/tools/util.py
```diff
@@ -153,8 +153,8 @@ def is_identifier(string):
     :returns: True if string can be a python identifier, False otherwise
     :rtype: bool
     """
-    val = PYTHON_IDENTIFIER_RE.match(string) and not keyword.iskeyword(string)
-    return True if val else False
+    matched = PYTHON_IDENTIFIER_RE.match(string)
+    return bool(matched and not keyword.iskeyword(string))
```
dimaryaz: Make it `bool(matched) and ...` - since the rest is already boolean

compiler/quilt/tools/build.py
eode: We could put a `QuiltException` in `constants.py` and have it be a superclass of `BuildException` and `CommandException`.  ..then `main.py` catches the `QuiltException` rather than `CommandException` as it does now.

If we want to take that further, we could say:
* File `X` raises `XException`s
* `XException`s are all `QuiltExceptions`
* The first argument to `QuiltException` should be a user-friendly (as possible) message.
* `main.py` catches `QuiltException` and displays the message (instead of doing so with `CommandException`.

akarve: Sounds reasonable to me.

akarve: To clarify I mean the three subclasses of `QuiltException` are fine for now. I don't think we need the `XExceptions`.

eode: Although, that may be better done separately -- I could do it as a PR just following the merge of this and Export into master.

eode: Sounds good.  I'll include these changes in this PR, and use a `QuiltException` for exceptions coming out of `util.py`.

eode: @dimaryaz @akarve Ready for re-review.

compiler/quilt/tools/const.py
```diff
@@ -70,3 +70,21 @@ class TargetType(Enum):
...
+    the module raising it, it should use a `QuiltException` or subclass.
+
+    `QuiltException` is the base for all quilt exceptions, such as
+    `CommandException`, `BuildException`, etc.
+
+    All `QuiltException` instances should have a user friendly message,
+    even if not expected to be seen by a user.
+    """
+    def __init__(self, message, *args):
+        super(QuiltException, self).__init__(message, *args)
```
dimaryaz: Looks like you don't need to call `super`; examples in Python docs don't show it, and you can't really do anything useful with the parent class. (`Exception` has a `message` attribute, but it's deprecated; not sure what happens to `*args` - might as well get rid of those.)

compiler/quilt/tools/build.py
```diff
@@ -80,14 +81,8 @@ def _is_valid_group(group):
     return isinstance(group, dict) or group is None
 
 def _pythonize_name(name):
```
dimaryaz: Might as well remove it now.

compiler/quilt/tools/const.py
eode: There's a whole small topic around this that I won't go into, that relates to language consistency vs convenience, and Python3's choices regarding usage of parameters by a function they weren't passed to -- however, again, I'll skip all the cruft about features that may or may not stay.  I am, however, happy to have that conversation if you want, it's kinda interesting.  ..but, noise.

What relates directly to us is this:
```python
# Without super call, python2:
str(QuiltException('Error message')) -> ""
# With super call, python2:
str(QuiltException('Error message')) -> "Error message"

# With or without super call, python3:
str(QuiltException('Error message')) -> "Error message"
```

We use `"Some Prefix: " + str(error)` fairly frequently.

There's also a similar issue that this has caused me to notice, though, too -- if you rely on Python3's 'convenience' of (effectively -- I don't know what the C code does, haven't looked at it) calling super() for you when you subclass an exception, it always uses all arguments, which may not be (Read:  Is not, in our case) ideal, because passing more than one argument to `QuiltException` would result in a string repr of the tuple of arguments (instead of a string of the message).

..so:
I'll change that `super` call to only include the 'message' argument.

dimaryaz: Haha, ok, sounds good.

PR #538, opened 2018-04-03
Use full-text search to match keywords in package names

dimaryaz:
Currently, we're looking for substrings in package names - however, that won't be possible anymore once we fix search so different keywords can match different things (package name, node names, README).
To fix that, let's find dictionary words in package names and add them to full-text search.


akarve:
Dictionary is the approach I requested and it is the only way that we can split package names like `dogscats` into meaningful substrings.

PR #553, opened 2018-04-12
Move fragment downloading into a helper function

dimaryaz:
This will become useful very soon.

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -1023,17 +1023,34 @@ def install(package, hash=None, version=None, tag=None, force=False):
             return
 
     dataset = response.json(object_hook=decode_node)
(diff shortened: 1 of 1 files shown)


dimaryaz:
What's a good place? `util.py`? I'm guessing it's too high-level for `store.py` - especially with all the print statements.

I guess I should move uploading, too.

kevinemoore:
I think the bulk of it makes sense to include in store.py, but you're right, not the print statements. The store is the right place to check if fragments already exist at least. Or, should we put a file or class together that handles all communication with the registry?

kevinemoore:
On second thought, the actual mechanics of the download, temporary file creation and unzipping aren't really related to the PackageStore and could go in util.py or another helper location.

PR #555, opened 2018-04-12
Move fragment downloading and uploading into a new file

dimaryaz:
Makes the code quite a bit cleaner.

### DIFF
--- a/compiler/quilt/tools/data_transfer.py
+++ b/compiler/quilt/tools/data_transfer.py
@@ -1,0 +1,282 @@
+"""
+High-level helper functions for uploading and downloading fragments.
+"""
+
+from __future__ import print_function
+import gzip
+import os
+import re
+from shutil import copyfileobj, move
+import tempfile
+from threading import Thread, Lock
+
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+from six import iteritems, itervalues
+from tqdm import tqdm
+
+from .hashing import digest_file
+from .util import FileWithReadProgress
+
+
+PARALLEL_UPLOADS = 20
+PARALLEL_DOWNLOADS = 20
+
+S3_CONNECT_TIMEOUT = 30
+S3_READ_TIMEOUT = 30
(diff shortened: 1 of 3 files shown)


compiler/quilt/tools/data_transfer.py
```diff
@@ -0,0 +1,282 @@
...
+    retries = Retry(total=3,
+                    backoff_factor=.5,
+                    status_forcelist=[500, 502, 503, 504])
+    sess.mount('https://', HTTPAdapter(max_retries=retries))
+    return sess
+
+def download_fragments(store, obj_urls, obj_sizes):
+    assert len(obj_urls) == len(obj_sizes)
+
+    if not obj_urls:
```
kevinemoore: This check or at least printing the message should probably move back to command.py (download_fragments doesn't really know why obj_urls is empty if it is).

compiler/quilt/tools/data_transfer.py
```diff
@@ -0,0 +1,282 @@
...
+                                if response.status_code == requests.codes.RANGE_NOT_SATISFIABLE:
+                                    with lock:
+                                        progress.update(original_size)
+                                else:
+                                    if not response.ok:
+                                        message = "Download failed for %s:\nURL: %s\nStatus code: %s\nResponse: %r\n" % (
+                                            obj_hash, response.request.url, response.status_code, response.text
+                                        )
+                                        with lock:
+                                            tqdm.write(message)
```
kevinemoore: Probably for another PR, but maybe it'd be cleaner to factor out these prints in case the function is called from somewhere other than the CLI? That might be hard to do with the threading though.

compiler/quilt/tools/data_transfer.py
dimaryaz: Sure. (I put it there because it's gonna come up again when I add the "materialize" function... but whatever, not a big deal.)

compiler/quilt/tools/data_transfer.py
dimaryaz: Yeah, this one is going to be really difficult to factor out...

PR #556, opened 2018-04-13
WIP: Add a "materialize" command

dimaryaz:
Takes either a package name or a node.

TODO: Is this the API we want? Alternative: a `_materialize()` method on a node.

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -940,6 +940,53 @@ def install(package, hash=None, version=None, tag=None, force=False, meta_only=F
 
     pkgobj.save_contents()
 
+def materialize(package):
+    if isinstance(package, string_types):
+        node, pkgobj, info = _load(package)
+    elif isinstance(package, nodes.Node):
+        node = package
+    else:
+        raise ValueError("Expected a Node or a package name, but got %r" % package)
+
+    store = PackageStore()
+
+    hashes = set()
+
+    stack = [node]
+    while stack:
+        obj = stack.pop()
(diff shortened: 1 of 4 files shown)


compiler/quilt/tools/command.py
```diff
@@ -940,6 +940,53 @@ def install(package, hash=None, version=None, tag=None, force=False, meta_only=F
...
+        node = package
+    else:
+        raise ValueError("Expected a Node or a package name, but got %r" % package)
+
+    store = PackageStore()
+
+    hashes = set()
+
+    stack = [node]
+    while stack:
```
kevinemoore: This is reimplementing tree traversal (we now have a helper, right?) and is duplicating the fragment download logic from install. Maybe it'll look cleaner after merging your data_transfer changes?

dimaryaz: The helper is for core nodes (which we don't have until we "build" the package... yeah, I know, unnecessarily confusing.)

As for duplicating: where do you see that? My data_transfer changes _are_ merged.

kevinemoore:
Both install and materialize start with a list of hashes, then strip away the ones already installed. Install gets that list via the database, but that's basically a shortcut around downloading and parsing the package manifest. 

It seems strange to me now that install automatically upgrades the package version. So if a user does:
quilt install foo/bar/subpath  (waits while a new version is pushed, then)
quilt install foo/bar
They'll get prompted to either overwrite (and upgrade) or cancel all together. Materialize seems like a more logical behavior there.

dimaryaz:
Ok sure. I'll try that.

PR #562, opened 2018-04-17
Represent node paths as lists instead of dot or slash separated strings

dimaryaz:
It's silly to join node names with slashes, then replace them with dots, then split on dots. Regardless of how we expose it to the user, internally, we should just use lists.
Also, rename parameters like "path" to make it clear what they actually do. Fixes one bug, too.

### DIFF
--- a/compiler/quilt/tools/package.py
+++ b/compiler/quilt/tools/package.py
@@ -76,7 +76,7 @@ class Package(object):
 
         Usage:
             p['item']
-            p['path/item]
+            p['path/item']
 
         :param item: Node name or path, as in "node" or "node/subnode".
         """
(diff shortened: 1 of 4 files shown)


compiler/quilt/tools/package.py
```diff
@@ -243,22 +236,21 @@ def save_df(self, dataframe, name, path, ext, target):
                 objhash = digest_file(path)
                 move(path, self._store.object_path(objhash))
                 hashes.append(objhash)
-            self._add_to_contents(buildfile, hashes, ext, path, target)
+            self._add_to_contents(node_path, hashes, ext, source_path, target)
```
dimaryaz: This fixes a bug: `path` gets overwritten with something unrelated in the `for` loop above.

PR #569, opened 2018-04-21
Check the available disk space before downloading fragments

dimaryaz:


compiler/quilt/tools/data_transfer.py
```diff
@@ -64,6 +64,13 @@ def download_fragments(store, obj_urls, obj_sizes):
     # Some objects might be missing a size; ignore those for now.
     total_bytes = sum(size or 0 for size in itervalues(obj_sizes))
 
+    # Check if we have enough disk space. There's no way to check reliably because we also need
+    # space for the temporary gzip'ed files, but that's better than nothing.
+    free_space = 5 # get_free_space(store.object_path('.'))
```
akarve: i don't understand why `get_free_space` is commented out.

akarve: does this respect `QUILT_PRIMARY_PACKAGE_DIR`?

dimaryaz: Oh crap, was testing out of space errors. The fact that it didn't break any unit tests is kinda bad...

Yeah, it checks the directory where the objects will actually get saved,

akarve:
Will tests now fail with a hardcoded value?

compiler/quilt/tools/data_transfer.py
```diff
@@ -64,6 +64,13 @@ def download_fragments(store, obj_urls, obj_sizes):
...
+    if total_bytes > free_space:
+        print("Not enough disk space! Required: %d, available: %d" % (total_bytes, free_space))
```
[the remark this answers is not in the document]
dimaryaz: Wanted to clarify: what's the full message? This?
```
Error: Insufficient space for install. Required: %d, available: %d
```

dimaryaz:
Yep, they'll fail. The problem was, test fragments didn't have a size set (cause we didn't even have sizes saved in the DB until recently), so the required space was 0.

compiler/quilt/tools/data_transfer.py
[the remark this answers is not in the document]
akarve: Yeah, that works

PR #570, opened 2018-04-22
Implement getting/setting metadata on nodes

dimaryaz:
- Split `DataNode` into two subclasses: dataframes that are backed by the store, and in-memory ones.
- Stop using core nodes when modifying packages: those should only be used for serialization, and aren't relevant there.
- Store metadata in the `Node` class, to make it easily accessible by the user. (Right now, the user can change any metadata, not just "custom" - though changes will be ignored.)
- Save custom metadata when building a new package.

### DIFF
--- a/compiler/quilt/imports.py
+++ b/compiler/quilt/imports.py
@@ -15,7 +15,7 @@ import sys
 
 from six import iteritems
 
-from .nodes import DataNode, GroupNode, PackageNode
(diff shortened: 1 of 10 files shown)


dimaryaz:
(Just rebased this and made small fixes)

dimaryaz:
Will be sending out a different version of this

PR #571, opened 2018-04-22
WIP Package table

nl0:
Use a sortable table to display packages on profile page.

Depends on #587 and #588.

### TODO

- [x] complete table design
- [ ] wire up the data
- [ ] test it

### Required API changes

To be able to sort the packages we need to add the following data for each package to the packages endpoint:

* size
* update timestamp
* views (total and last week)
* installs (total and last week)

### DIFF
--- a/catalog/app/components/PackageTable/index.js
+++ b/catalog/app/components/PackageTable/index.js
@@ -1,0 +1,297 @@
+import orderBy from 'lodash/orderBy';
+import omit from 'lodash/fp/omit';
+import { List, ListItem } from 'material-ui/List';
+import {
+  Table,
+  TableBody,
+  TableHeader,
+  TableHeaderColumn,
+  TableRow,
+  TableRowColumn,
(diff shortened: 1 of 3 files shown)


nl0:
@akarve I've implemented a sortable package table. looks like this:
catalog_3000_profile laptop with mdpi screen

Next I need the API adjustments to make it work.

akarve:
@nl0 can you spec the API changes that you need so I can make them happen? also, in general are you comfortable with flask?

nl0:
@akarve i've added the required api changes description in the first post. re:flask: i wrote some flask code and i was using python(2) extensively in general back then, but it was quite some time ago (5 years or so), so i can write something looking at examples and docs, but it may be slow

akarve:
hmm can we get sparklines into this table?

nl0:
@akarve i'll see what i can do in terms of UI

nl0:
doesnt seem relevant anymore

PR #573, opened 2018-04-27
Export tweaks and progress bar

eode:
### Summary
* Add progress bar using tqdm
* Rename and export `XLSX/XLS` files as `.csv`
* Fix export filename collision detection when dealing with TableNode/Dataframe data
* For round-trip csv export/build, ensure floats are close, with tolerance of 1e-15 instead of identical
* Fix `csv` test files that had lower precision than other test data

### DIFF
--- a/compiler/quilt/test/test_command.py
+++ b/compiler/quilt/test/test_command.py
@@ -1374,8 +1374,8 @@ class CommandTest(QuiltTestCase):
         expected_exports = sorted([
             Path('data/README.md'),
             Path('data/100Rows13Cols.csv'),
-            Path('data/100Rows13Cols.tsv'),
-            Path('data/100Rows13Cols.xlsx.csv'),  # originally xlsx
+            Path('data/100Rows13Cols_tsv.csv'),   # originally tsv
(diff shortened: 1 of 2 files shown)


kevinemoore:
Seems strange to me that we'd convert table nodes that used to be Excel into csv, but keep table nodes from TSV's as TSV's. To me, it would be cleaner and more convenient to users to use a consistent format (e.g., CSV) for all Table Node exports.

akarve:
@kevinemoore that is per my request and eliminates a funky dependency (pyopenxl) and slow serialization to XLS when all we have is a table. The issue is that XLS=>DF (our default) is lossy and there's no reason to export an Excel file at that point. (Though I am now wondering @eode what happens to multiple Excel sheets when they are auto compiled to a DF?). In any case transform: id Excel files will remain Excel files.

kevinemoore:
@akarve all of the table transformations, including CSV, are potentially lossy because we don't include all the parsing options in the manifest. IMHO, it would be a cleaner model not to pretend to restore the source file, but simply to export table/DataFrame data into a consistent, widely read format.

PR #576, opened 2018-05-02
Migrate to a two-level object directory

dimaryaz:
An object hash like `abcdef...` will be stored as `ab/abcdef...`, to prevent too many files in a single directory.


dimaryaz:
Oops, will fix Python2.

As for the whole filename: sure. (I was just copying Git, which removes the two chars.) Just need to rename 9M files now :)

PR #577, opened 2018-05-02
Export symlinks

eode:
### Summary
Allows use of symlinks when exporting (Linux, OSX, Vista+) where `node._node` is a `FileNode`:

```bash
quilt export some/package some_folder --symlinks
```
```python
import quilt
quilt.export('some/package', 'some_folder', symlinks=True)
```
Requires `admin` privileges on Windows.


JacksonMaxfield:
Tested symlinks and regular export, no issues to report. Great job!

JacksonMaxfield:
This does not work on windows. The folder is created but no symlinks are.

akarve:
@JacksonMaxfield Thanks for testing on Windows. Would you provide a repro including OS details and package?

JacksonMaxfield:
@akarve 

Windows 10, Command Prompt
NOTE: Must be signed in as administrator for this to even run.

Steps to reproduce:

Basic Input:
```bash
pip install git+https://github.com/quiltdata/quilt.git@export_symlinks#subdirectory=compiler
quilt install aics/cell_line_samples

quilt export aics/cell_line_samples aics_cls --symlinks
```

Output:
```bash
Warning: Exporting using symlinks to the package store.
	This is an advanced feature.
	The package store must not be written to directly, or it may be corrupted.
	Managing permissions and usage are your responsibility.

Are you sure you want to continue? (yes/No)
```

Ensure command:
```bash
yes
```

Terminal Output:
```bash
Exporting file 10 of 10 [00:00]: README.md
```

Directory Output:
Directory `aics_cls` properly created but when opened, there were no links to files.

I didn't however check in command prompt if there were files that Windows Explorer hides by default.
I will do that tonight...

eode:
Windows changed the return value of the symlinking function to an int (from a bool) in Windows 10, so there's a decent chance that there's an error that's not being displayed because it's an even number.

I'll post a fix for that, and we'll troubleshoot it from there.

eode:
So, the incorrect return value *may* be a Windows bug -- although there's precious little information about it, save a conversation on StackOverflow.

In any case, I've done a couple things:
* Don't check for admin access, as it's possible someone may have the specific permission which allows creating symlinks
* On Windows, even if a success is reported, check to ensure the symlink exists, and fail explicitly if it doesn't.  This doesn't really fix the issue, but it makes the failure obvious.

My best recommendation for the Windows symlink issue is:
* Ensure Windows has the latest updates
* Ensure that you're an administrator, or have the `SeCreateSymbolicLink` permission (see [this superuser article](https://superuser.com/a/105381/1361461) 
* If you're an administrator *at all*, ensure you have elevated privileges, *even if you have the `Create Symbolic Links` privilege*.

eode:
btw, current Appveyor error isn't related to code -- it was a failure when Appveyor attempted to connect to PyPi.

eode:
Updated Appveyor tests to include Server 2016 per convo with Aneesh and above issue by JacksonMaxfield (which we're unable to replicate).

@JacksonMaxfield : We've made some changes that *might* address your issue, and updated the API docs and troubleshooting docs with some potential means of resolving the issue.  I'm unable to replicate it on Windows 8, Server 2012, or Server 2016.  If the problem persists, will you open an issue regarding it, and we'll deal with it separately from this PR?

PR #578, opened 2018-05-03
Move dataframe saving/loading code from Package to PackageStore

dimaryaz:
The logic belongs in PackageStore rather than Package because objects are owned by the store, and might not even have a package associated with them. (This will be the case once filtering is implemented.)


dimaryaz:
I feel like stores should be self-contained... But hard to say without some actual use cases.

PR #579, opened 2018-05-03
Add `__slots__` to core nodes and remove `format`

dimaryaz:
Lets us save a bit of memory by getting rid of `__dict__` in each node.
The `format` removal only affects the in-memory representation; JSON stays the same for now, just in case.


akarve:
Linux/Python3 is failing and branch has fallen behind master

PR #584, opened 2018-05-09
Clean up group nodes. Fix `GroupNode._data()` as much as possible.

dimaryaz:
- Make `GroupNode` a subclass of `Node` rather than `DataNode`
- In `_data()`, iterate over the actual children of the group, rather than whatever they used to be when the package was loaded
- Return errors in corner cases like data nodes from different stores or in-memory dataframes

### DIFF
--- a/compiler/quilt/nodes.py
+++ b/compiler/quilt/nodes.py
@@ -26,11 +26,12 @@ class Node(object):
     def __repr__(self):
         return self._class_repr()
 
-    def __setattr__(self, name, value):
-        if name.startswith('_') or isinstance(value, Node):
-            super(Node, self).__setattr__(name, value)
-        else:
-            raise AttributeError("{val} is not a valid package node".format(val=value))
+    def __call__(self):
+        return self._data()
+
+    def _data(self):
+        raise NotImplementedError
+
 
 class DataNode(Node):
     """
@@ -41,9 +42,6 @@ class DataNode(Node):
         self._package = package
         self._node = node
         self.__cached_data = data
-
-    def __call__(self):
-        return self._data()
 
     def _data(self):
         """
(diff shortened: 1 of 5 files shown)


compiler/quilt/nodes.py
```diff
@@ -57,17 +55,24 @@ def _data(self):
             elif isinstance(self._node, core.FileNode):
                 self.__cached_data = store.get_file(self._node.hashes)
             else:
-                # XXX: This is wrong.
-                hash_list = list(core.find_object_hashes(self._node, sort=True))
-                self.__cached_data = store.load_dataframe(hash_list)
+                assert False
```
akarve: is this caught and rolled up into an error message? if not we should at least have a descriptive string here

dimaryaz: No - but this can't happen. I.e., it's a bug in the code, not a user error, if it does happen.

compiler/quilt/nodes.py
```diff
@@ -108,13 +113,43 @@ def _keys(self):
...
+        store = None
+        hash_list = []
+        stack = [self]
+        while stack:
+            node = stack.pop()
+            if isinstance(node, GroupNode):
+                stack.extend(child for _, child in sorted(node._items(), reverse=True))
+            else:
+                if not isinstance(node._node, core.TableNode) or not node._node.hashes:
+                    raise NotImplementedError("Can only merge on-disk dataframes")
```
akarve: `Can only merge built dataframes. `.build` this package and try again.`

compiler/quilt/tools/core.py
```diff
@@ -59,20 +59,15 @@ def __json__(self):
...
         """
         stack = [self]
         while stack:
             obj = stack.pop()
             yield obj
-            if sort:
-                stack.extend(child for name, child in sorted(iteritems(obj.get_children()), reverse=True))
-            else:
-                stack.extend(itervalues(obj.get_children()))
+            stack.extend(itervalues(obj.get_children()))
```
akarve: so what guarantees order now that there's no call to `sorted`?

akarve: i mean `preorder` implies RLV and user might get LRV etc with this method?

akarve: it's fine if we want to loosen the contract but we should be explicit about it.

dimaryaz: There's only one place we use this method now, and it doesn't care about the order. In fact, we really don't need a preorder method.

PR #588, opened 2018-05-10
components/PackageHandle: refactoring and tweaks

nl0:
Refactor and tweak `PackageHandle` component a little:

* use `display: inline-flex` to avoid vertical alignment inconsistencies
  and deprecate now useless "drop" prop

* make `VisibilityIcon` component local to `PackageHandle`, because it's not
  used anywhere else (and not supposed to)


nl0:
@akarve another extracted PR, please take a look.

PR #590, opened 2018-05-10
utils/sagaTools: waitTil

nl0:
Add `utils/sagaTools` module with a `waitTil` helper.


nl0:
@akarve another extracted PR. please take a look

PR #591, opened 2018-05-10
containers/Notifications: factor out notification display and make it configurable

nl0:
`containers/Notifications`: factor out notification display and make it configurable


nl0:
@akarve another one. take a look please

PR #594, opened 2018-05-10
Enhance package_preview's install timeseries, and add total installs

meffij:


registry/quilt_server/views.py
```diff
@@ -964,37 +964,77 @@ def _iterate_data_nodes(node):
...
     )
 
+    total = (
+        db.session.query(
+            Event
+        )
+        .filter(Event.package_owner == owner)
+        .filter(Event.package_name == package_name)
+        .filter(Event.type == event_type)
+        .count()
```
dimaryaz: Use `sa.func.count`, just in case. This one does weird stuff with subqueries.

PR #596, opened 2018-05-11
components/Error: accept nodes

nl0:
Allows `Error` component to accept react nodes, not just strings, as arguments.


nl0:
@akarve another micro PR. take a look pls.

PR #518, opened 2018-03-26
WIP Factor-out auth-related logic

nl0:
Factor-out auth-related logic and adjust it a little. Log-out on auth failure.

Depends on:
  - [x] #589 
  - [x] #590 
  - [x] #591 
  - [x] #596 
  - [x] #597 
  - [x] #599 

### TODO

- [x] Unit Tests
- [x] rebase
- [ ] more manual testing
- [x] figure out 500 vs 401 behaviour
- [x] styling
- [ ] rewrite tests

PR #597, opened 2018-05-11
app/testing: helpers for writing tests

nl0:
Tools for convenient test writing.


nl0:
@akarve these are testing tools extracted from the #518, take a look pls

PR #600, opened 2018-05-11
Implement package filtering

dimaryaz:


compiler/quilt/nodes.py
```diff
@@ -139,6 +140,38 @@ def _data(self):
 
         return store.load_dataframe(hash_list)
 
+def _create_filter_func(filter_dict):
+    filter_name = filter_dict.pop('name', None)
+    if filter_name is not None and not isinstance(filter_name, string_types):
+        raise ValueError("Invalid 'name'")
```
akarve: probably worth printing the invalid thing too

akarve: And do we want to say `Invalid 'name' in filtering dict`. I am frankly unclear as to why a name key would always be defined btw.

compiler/quilt/nodes.py
```diff
@@ -139,6 +140,38 @@ def _data(self):
...
+        raise ValueError("Invalid 'meta'")
+
+    if filter_dict:
+        raise ValueError("Unexpected data in the filter: %r" % filter_dict)
+
+    def helper(value, expected):
+        if isinstance(expected, dict):
+            if isinstance(value, dict):
+                for expected_key, expected_value in iteritems(expected):
+                    if not helper(value.get(expected_key), expected_value):
```
akarve: recursive is most readable; wondering at what package depth we run out of call stack space? most packages should be pretty shallow so not worth doing this iteratively.

compiler/quilt/nodes.py
```diff
@@ -139,6 +140,38 @@ def _data(self):
...
+    filter_name = filter_dict.pop('name', None)
+    if filter_name is not None and not isinstance(filter_name, string_types):
+        raise ValueError("Invalid 'name'")
+
+    filter_meta = filter_dict.pop('meta', None)
+    if filter_meta is not None and not isinstance(filter_meta, dict):
+        raise ValueError("Invalid 'meta'")
+
+    if filter_dict:
+        raise ValueError("Unexpected data in the filter: %r" % filter_dict)
```
akarve: I did `foo._filter({"color":"orange"})` and this message confused me.
1. we probably shouldn't require the "meta" key on top? (i know, the spec shows it.) we just make it clear in the filter documentation/contract that it filters meta only.
1. the error message should be something like this: `Unexpected key {} passed to _filter: {}` And maybe we won't need that message anymore?

akarve:
Oh and we should add a doc entry on `._filter`

compiler/quilt/nodes.py
dimaryaz: Ok - but not sure what you mean by "why a name key would always be defined"?

compiler/quilt/nodes.py
dimaryaz: Yeah... I'd wait till it becomes a problem.

compiler/quilt/nodes.py
dimaryaz: If we don't require "meta", how do we handle the node name? (It can't be a part of the metadata -  nodes can't know their own names. If we add a hack just for the filter, then it becomes a yet another reserved word...)

akarve: oh i see. i had forgotten about that. i thought the node name was a second positional argument. would it work in that case?

btw - i know understand what `.pop('name',...) does above.

compiler/quilt/nodes.py
```diff
@@ -139,6 +140,38 @@ def _data(self):
...
+    filter_name = filter_dict.pop('name', None)
+    if filter_name is not None and not isinstance(filter_name, string_types):
+        raise ValueError("Invalid 'name': %r" % filter_name)
+
+    filter_meta = filter_dict.pop('meta', None)
+    if filter_meta is not None and not isinstance(filter_meta, dict):
+        raise ValueError("Invalid 'meta': %r" % filter_meta)
+
+    if filter_dict:
+        raise ValueError("Unexpected data in the filter: %r" % filter_dict)
```
kevinemoore: I saw this error and didn't know what to do. Maybe include the top-level keys in the error message?

PR #515, opened 2018-03-25
Shortify group nodes representation

kurlov:
### Summary

Shortify nodes printing in case it exceeds limit.

Related to #493

PR #601, opened 2018-05-12
Shortify group nodes reworked

kurlov:
### Summary
Shortify nodes printing in case it exceeds limit.

 This is reworked #515


compiler/quilt/nodes.py
```diff
@@ -79,11 +79,21 @@ def __setattr__(self, name, value):
 
     def __repr__(self):
         pinfo = super(GroupNode, self).__repr__()
-        group_info = '\n'.join(name + '/' for name in sorted(self._group_keys()))
-        if group_info:
-            group_info += '\n'
-        data_info = '\n'.join(sorted(self._data_keys()))
-        return '%s\n%s%s' % (pinfo, group_info, data_info)
+        items = [name + '/' for name in sorted(self._group_keys())]
+        if not items:
```
kevinemoore: I think you'll want to remove this "if not items". With that in place, it skips the data nodes for top-level packages that contain only data nodes (not any group nodes).

PR #612, opened 2018-05-16
String improvements

akarve:


akarve:
Merging because unit tests not present for JS and since #616 fixes CI failure.

PR #614, opened 2018-05-17
attempt to fix appveyor errors

akarve:


appveyor.yml
```diff
@@ -50,8 +50,8 @@ install:
       }
       else
       {
-        pip install --upgrade pip
-        pip install -U pip
+        c:\python35-x64\python.exe -m pip install --upgrade pip
+        c:\python35-x64\python.exe -m pip install -U pip
```
dimaryaz: That can't be right - especially if you're using `pip` on the line below...

akarve: i kind of agree but... it worked and i am following precisely the error messages that appveyor gave us :( (see old failures)

maybe it only pertains to self-modifying pip commands

dimaryaz: Hmm. Ok, but can you confirm that the `pip` below belongs to the same python as here? (Otherwise, your change would be equivalent to just dropping the pip upgrade.)

Actually... `-U` is literally a short name for `--upgrade` - so this already makes no sense.

akarve:
See #616

PR #616, opened 2018-05-17
Appveyor failures: can we live without pip -U pip?

akarve:


akarve:
Should fix CI failures. Only remaining failure is fixed by #613.

PR #622, opened 2018-05-21
Add asa parameter to _data

kevinemoore:
Adding an argument to the _data method to allow generation of data by a
custom deserializer.


compiler/quilt/nodes.py
```diff
@@ -29,10 +29,10 @@ def _class_repr(self):
     def __repr__(self):
         return self._class_repr()
 
-    def __call__(self):
-        return self._data()
+    def __call__(self, asa=None):
+        raise NotImplementedError
```
dimaryaz: Why not always forward it to `_data`?

compiler/quilt/nodes.py
```diff
@@ -47,20 +47,30 @@ def __init__(self, package, node, data, meta):
         self._node = node
         self.__cached_data = data
 
-    def _data(self):
+    def _data(self, asa=None):
         """
-        Returns the contents of the node: a dataframe or a file path.
+        Returns the contents of the node: a dataframe or a file path, or passes
+        the node and is contents to a callable.
```
dimaryaz: its

compiler/quilt/nodes.py
```diff
@@ -47,20 +47,30 @@ def __init__(self, package, node, data, meta):
...
-            if isinstance(self._node, core.TableNode):
-                self.__cached_data = store.load_dataframe(self._node.hashes)
-            elif isinstance(self._node, core.FileNode):
-                self.__cached_data = store.get_file(self._node.hashes)
-            else:
-                assert False
-        return self.__cached_data
+            return asa(self, [store.object_path(obj) for obj in self._node.hashes])
+        elif asa is not None:
+            raise AttributeError("{asa} is not callable".format(asa=asa))
```
dimaryaz: Should be a `TypeError`

compiler/quilt/nodes.py
```diff
@@ -47,20 +47,30 @@ def __init__(self, package, node, data, meta):
...
+    def _data(self, asa=None):
         """
-        Returns the contents of the node: a dataframe or a file path.
+        Returns the contents of the node: a dataframe or a file path, or passes
+        the node and is contents to a callable.        
         """
-        if self.__cached_data is None:
-            # TODO(dima): Temporary code.
+        if callable(asa):
             store = self._package.get_store()
```
dimaryaz: What do we do if the node is created using `_set(...)`, and doesn't have a hash?

compiler/quilt/nodes.py
```diff
@@ -47,20 +47,30 @@ def __init__(self, package, node, data, meta):
...
-            elif isinstance(self._node, core.FileNode):
-                self.__cached_data = store.get_file(self._node.hashes)
-            else:
-                assert False
-        return self.__cached_data
+            return asa(self, [store.object_path(obj) for obj in self._node.hashes])
+        elif asa is not None:
+            raise AttributeError("{asa} is not callable".format(asa=asa))
+        else:
+            if self.__cached_data is None:
```
dimaryaz: I wonder if the whole `else` clause should be a function used as a default value for `asa`... Though can probably wait till later.

compiler/quilt/nodes.py
```diff
@@ -112,13 +122,17 @@ def _add_group(self, groupname):
         child = GroupNode({})
         setattr(self, groupname, child)
 
-    def _data(self):
+    def _data(self, asa=None):
         """
         Merges all child dataframes. Only works for dataframes stored on disk - not in memory.
         """
+        if asa is not None and not callable(asa):
+            raise AttributeError("{asa} is not callable".format(asa=asa))
```
dimaryaz: You should probably use `%r` (or its equivalent for `format`)

compiler/quilt/nodes.py
```diff
@@ -112,13 +122,17 @@ def _add_group(self, groupname):
...
+        
         store = None
         hash_list = []
         stack = [self]
+        store = None
```
dimaryaz: It's already initialized above?

compiler/quilt/test/test_import.py
```diff
@@ -430,3 +430,39 @@ def test_team_set_non_node_attr(self):
         df = pd.DataFrame(dict(a=[1, 2, 3]))
         with self.assertRaises(AttributeError):
             package4.newdf = df
+
+    def test_datanode_asa(self):
+
+        def test_lambda(node, hashes):
```
dimaryaz: Might be good to actually return something from the lambda, and check the `.csv(...)` return value.

compiler/quilt/test/test_import.py
```diff
@@ -430,3 +430,39 @@ def test_team_set_non_node_attr(self):
...
+            for path in hashes:
+                assert os.path.exists(path)
+        
+        mydir = os.path.dirname(__file__)
+        build_path = os.path.join(mydir, './build.yml')
+        command.build('foo/package', build_path)
+        
+        pkg = command.load('foo/package')
+
+        pkg.dataframes._meta['foo'] = 'bar'
```
dimaryaz: Not really useful?

compiler/quilt/nodes.py
dimaryaz: Actually, Python's default error message for calling a non-callable object is pretty reasonable, so you could just do a check for `is not None` and skip `callable(...)`

compiler/quilt/nodes.py
kevinemoore: The only things that make sense to me are: (1) return cached data, (2) pass cached data to asa, or (3) assert. What's your preference? We could also add a parameter to the asa signature to accepted cached data.

dimaryaz: Well... I think 1) would be incorrect because that's equivalent to just ignoring `asa`; 2) would likely break because the `asa` function normally accepts file paths, not dataframes. I guess 3) then? Though a `ValueError` or something similar is better than `assert` since it's not a programming error.

compiler/quilt/nodes.py
```diff
@@ -127,12 +122,11 @@ def _data(self, asa=None):
         Merges all child dataframes. Only works for dataframes stored on disk - not in memory.
         """
         if asa is not None and not callable(asa):
```
dimaryaz: Same as in `DataNode`, you can drop this check and the assert below, and let Python do its thing.

compiler/quilt/nodes.py
```diff
@@ -47,20 +47,25 @@ def __init__(self, package, node, data, meta):
...
+    def _data(self, asa=None):
         """
-        Returns the contents of the node: a dataframe or a file path.
+        Returns the contents of the node: a dataframe or a file path, or passes
+        the node and its contents to a callable.        
         """
-        if self.__cached_data is None:
-            # TODO(dima): Temporary code.
+        if asa is not None:
             store = self._package.get_store()
```
dimaryaz: Still need to do something here if it's an in-memory dataframe?

docs/api-python.md
```diff
@@ -132,6 +132,7 @@ Packages contain three types of nodes:
 * Retrieve the contents of a `DataNode` with `_data()`, or simply `()`: `PACKAGE.NODE.ANOTHER_NODE()`
   * Columnar data (`XLS`, `CSV`, `TSV`, etc.) returns as a `pandas.DataFrame`
   * All other data types return a string to the path of the object in the package store
+  * Provide a custom deserialzer by passing a fuction to `data(asa=FUNCTION)` with the signature `function(node, list_of_hashes)`.
```
akarve: much better, two things to make it super nice:
1. explain why a single DataNode may return multiple hashes
1. explain what happens with GroupNodes

PR #626, opened 2018-05-23
Allow asa calls on non-table GroupNodes

kevinemoore:
Previously, calling _data on GroupNodes was only allowed if all the
node’s children were TableNodes with matching schemas. This change
allows _data to be called with an asa function to return all child-node
paths regardless of their type.

### DIFF
--- a/compiler/quilt/nodes.py
+++ b/compiler/quilt/nodes.py
@@ -50,7 +50,7 @@ class DataNode(Node):
     def _data(self, asa=None):
         """
         Returns the contents of the node: a dataframe or a file path, or passes
-        the node and its contents to a callable.        
+        the node and its contents to a callable.
         """
         if asa is not None:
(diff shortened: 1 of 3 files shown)


compiler/quilt/nodes.py
```diff
@@ -128,13 +128,14 @@ def _data(self, asa=None):
...
+        alldfs = True
         while stack:
             node = stack.pop()
             if isinstance(node, GroupNode):
                 stack.extend(child for _, child in sorted(node._items(), reverse=True))
             else:
                 if not isinstance(node._node, core.TableNode):
-                    raise ValueError("Group contains non-dataframe nodes")
+                    alldfs = False
                 if not node._node.hashes:
```
dimaryaz: You probably need to check for `node._node is None` now, too (same as in the previous PR)

PR #638, opened 2018-05-30
Support downloads of empty fragments

dimaryaz:
They are possible now that compression is optional.


dimaryaz:
Ha, yeah, the nginx setup is janky: not only does it add the "gzip" header to all fragments - compressed or not - but it adds it to error messages, too. So the real failure was probably a 404 or similar - but quilt failed to print out the error because of the incorrect header.

How did you reproduce this?

Also, in `no_upload_compression`, I'm removing the header - but then you'll need to uncompress all of the stored fragments. I'll add a script for that.

PR #647, opened 2018-06-07
Mixpanel tracking + config generation script

nl0:
## Description

Every history action is now being sent to mixpanel if configured (otherwise just printed to the console).

Also, there's now a script to generate a config file for production: `npm run config $PROD_ENV_FILE` (result is written to `config.js`)

### DIFF
--- a/catalog/app/utils/tracking.js
+++ b/catalog/app/utils/tracking.js
@@ -1,0 +1,45 @@
+import { takeEvery, select } from 'redux-saga/effects';
+
+const loadMixpanel = (token) =>
+  import('mixpanel-browser')
+    .then((mp) => {
+      mp.init(token);
+      return mp;
+    });
+
+const consoleTracker = Promise.resolve({
+  // eslint-disable-next-line no-console
+  track: (evt, opts) => console.log(`track: ${evt}`, opts),
+});
+
+export const mkTracker = (token) => {
+  const tracker = token ? loadMixpanel(token) : consoleTracker;
+
+  return {
+    nav: (loc, user) => tracker.then((inst) =>
+      inst.track('WEB', {
+        type: 'navigation',
+        origin: window.location.origin,
+        location: `${loc.pathname}${loc.search}${loc.hash}`,
+        user,
+      })),
+  };
+};
+
+export default function* tracking({
+  selectUsername,
+  locationChangeAction,
+  routerStartAction,
+  token,
+}) {
+  const tracker = mkTracker(token);
+  let started = false;
+  yield takeEvery([locationChangeAction, routerStartAction], function* onLocationChange({ type, payload: location }) {
+    if (!started) {
+      if (type !== routerStartAction) return;
+      started = true;
+    }
+    const user = yield select(selectUsername);
+    tracker.nav(location, user);
+  });
+}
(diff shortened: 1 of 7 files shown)


nl0:
@akarve pls review

catalog/package.json
```diff
@@ -41,7 +41,8 @@
     "test:clean": "rimraf ./coverage",
     "test": "cross-env NODE_ENV=test jest --coverage",
     "test:watch": "cross-env NODE_ENV=test jest --watchAll",
-    "coveralls": "cat ./coverage/lcov.info | coveralls"
+    "coveralls": "cat ./coverage/lcov.info | coveralls",
+    "config": "sh -c 'env $(grep -v '^#' $0 | xargs) envsubst < config.js.tmpl > config.js'"
```
akarve: do we really need to use `grep` to filter out comments? bash will ignore them anyway if we need it the regex for comments is more like `'^\s*#'`.

akarve: this line does not work as written (I'm on Mac); when fixed please document in catalog/README.md

catalog/app/utils/tracking.js
```diff
@@ -0,0 +1,23 @@
...
+  // eslint-disable-next-line no-console
+  track: (evt, opts) => console.log(`track: ${evt}`, opts),
+});
+
+export default (token) => {
+  const tracker = token ? loadMixpanel(token) : consoleTracker;
+
+  return {
+    nav: (loc, action = 'LOAD') => tracker.then((inst) =>
+      inst.track('Navigation', {
```
akarve: let's reorganize and add a few fields:
```
inst.track('WEB', {
  location: `${loc.pathname}${loc.search}${loc.hash}`,
  origin: loc.origin,
  type: 'navigation',
  user: $USERNAME,
})
```

nl0: done

catalog/package.json
nl0: ok, now it's using bash parsing (via `source`).

> this line does not work as written (I'm on Mac)

@akarve could you please provide your input / output? i'm on mac also, but it works for me

catalog/app/utils/tracking.js
```diff
@@ -0,0 +1,45 @@
...
+  selectUsername,
+  locationChangeAction,
+  routerStartAction,
+  token,
+}) {
+  const tracker = mkTracker(token);
+
+  let started = false;
+
+  while (true) {
```
akarve: for some reason, this makes me nervous. isn't it better to use `yield take(LOCATION_CHANGE|ROUTER_START, nav)`...?

akarve: ^ We could do that in App/sagas.js

nl0: > We could do that in App/sagas.js

@akarve `App/sagas.js` is reworked heavily in #518, so i was avoiding doing anything there for now.

> for some reason, this makes me nervous.

what exactly?

>  isn't it better to use yield take(LOCATION_CHANGE|ROUTER_START, nav)...?

you mean `takeEvery`? like, forking on every action instead of using a blocking loop?

nl0:
@akarve under the hood `takeEvery` uses similar approach, listening for actions and forking sagas (this loop is in a generator function, so it's nothing dangerous), but i hear you -- i'll tweak the code to use `takeEvery`.

nl0:
@akarve done

PR #648, opened 2018-06-07
Load by package hash

kevinemoore:
Extend the load command to allow find package instances by their hash.

## Description
This PR implemented loading local package instances by hash (assigning local tags and versions isn't supported yet).

## TODO
Add support for load by tag and version (separate PR after local tags and versions are implemented)

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -1281,16 +1281,19 @@ def reset_password(team, username):
 
 def _load(package):
     info = parse_package_extended(package)
-    team, user, name = info.team, info.user, info.name
-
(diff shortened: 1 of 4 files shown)


compiler/quilt/tools/command.py
```diff
@@ -1281,16 +1281,19 @@ def reset_password(team, username):
 
 def _load(package):
     info = parse_package_extended(package)
-    team, user, name = info.team, info.user, info.name
-
-    pkgobj = PackageStore.find_package(team, user, name)
+    pkgobj = PackageStore.find_package(info.team,
```
dimaryaz: Can you also make it fail if a tag or a version is specified?

PR #652, opened 2018-06-08
Let's not use ubuntu:latest in our Dockerfiles

dimaryaz:
That causes accidental OS upgrades. Hardcode 16.04, which is what we've been using.


akarve:
Does this interfere with father's fargate's ability to manage patch level?

dimaryaz:
No, that's for our own Docker image. We don't control how Fargate itself patches things.

PR #656, opened 2018-06-13
Make fragment store read-only

dimaryaz:
The user should not modify the store fragments since that would change their hashes - so let's make them read-only.


akarve:
Windows CI is unhappy on what look like permission errors in the unit tests:
https://ci.appveyor.com/project/akarve/quilt/build/1.0.402/job/gv7ftu3vhmbhkae1

PR #667, opened 2018-06-28
registry: add env_config.py

nl0:
## Description

Add `env_config.py`: config file that takes the values from the environment

### DIFF
--- a/registry/quilt_server/env_config.py
+++ b/registry/quilt_server/env_config.py
@@ -1,0 +1,40 @@
+"""
+Config file that takes configuration from the environment.
+Overrides values in config.py.
+"""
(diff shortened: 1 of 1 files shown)


nl0:
@dimaryaz @akarve @meffij I was continuously annoyed by the necessity of editing the config files directly and inability to set the required config values via env vars, so I've created this config, which is more convenient when running the registry in docker-compose or similar docker-based environment

nl0:
actually, I would propose to keep only the base config file going forward (with some adjustments, of course) and get rid of the other `*_config.py` files to make things simpler (see https://12factor.net/config for more info)

PR #677, opened 2018-07-01
utils/reactTools: JSDoc comments and little tweaks

nl0:
rtm


nl0:
> This merits a mention in the developer README. i.e. if you plan to use HOCs as the main way to render components then there should be a note on how and why.

ok, i'll add some notes on this going forward

PR #678, opened 2018-07-01
testing tools + mocks

nl0:
* `feature` interface:
   * add step back support
   * add support for running code in `afterEach` block
* mocks for some components
* `fetch-mocks` for request testing
* various reusable step definitions
* `testing/util`: refactor, new helpers


nl0:
@akarve rtm

catalog/app/testing/util.js
```diff
@@ -1,17 +1,143 @@
...
+
+  Object.assign(Mock.prototype, methods);
+
+  return Mock;
+};
+
+/**
+ * Create a selector (suitable for enzyme or cheerio wrappers) that selects
+ * a mock component created by `mockComponent[Class]` or its nested element / prop.
+ *
```
akarve: do we have to have an empty line between `@param` calls? it's not a big deal but we should be consistent across the project.

catalog/app/testing/RequestTracker.js
```diff
@@ -1,70 +0,0 @@
-import invariant from 'invariant';
```
akarve: What became of the logic in this file?

nl0:
@akarve 
> looks reasonable but (maybe this happened in a separate PR?) need docs on how tools works and basics of writing new tests.

yep, it's in my backlog, I'll write it up soon

catalog/app/testing/RequestTracker.js
nl0: `fetch-mock` package provides similar functionality

catalog/app/testing/util.js
nl0: I usually put an empty line if params are not related and/or they have multiline description. If params are related (example: object props) and have short descriptions, I usually don't put an empty line.

catalog/app/testing/feature.js
```diff
@@ -99,15 +138,23 @@ const renderSteps = (steps, getCtx, next) => {
...
+    renderSteps(rest, getCtx, next);
+    return;
+  }
+
   if (steps[0].type === Then) {
-    const isNotWhen = ({ type }) => type !== When;
-    const thens = takeWhile(steps, isNotWhen);
-    const rest = dropWhile(steps, isNotWhen);
+    const isThen = ({ type }) => type !== When && type !== Branch;
+    const thens = takeWhile(steps, isThen);
```
meffij: [partition](https://lodash.com/docs/4.17.10#partition) is a function that expresses your intent here better I think

nl0:
> I agree with Aneesh that I'd like to see some documentation of how this testing strategy works, maybe some rationale given how heavy it is.

@akarve @meffij I wanted to do it in a subsequent PR if you don't mind

catalog/app/testing/feature.js
nl0: @meffij `partition` does something slightly different: it puts all the elements satisfying the predicate into the first group, and all the others into the second, but i need to split the array at the first occasion of the element satisfying the predicate

akarve:
@nl0 OK to doc in separate PR but from this point forward all PRs should contain their own docs. makes review easier. too easy to forget docs otherwise.

nl0:
@akarve sure, it totally makes sense

catalog/app/testing/feature.js
meffij: I see, so you aren't guaranteed to take the whole array.

meffij:
For reference, this PR is being merged because it's low-risk, but more effort is needed to make sure we understand the new testing ideas and that we're sure everything works correctly.

PR #682, opened 2018-07-05
travis: change directory back from catalog

nl0:
adjust travis setup according to @dimaryaz's note. rtm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ install:
     fi
   - |
     if [[ "$TEST_CATALOG" = true ]]; then
-      cd catalog && npm ci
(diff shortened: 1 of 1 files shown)


nl0:
@dimaryaz this doesn't seem to work

dimaryaz:
Oh, apparently, `(...)` causes bash to start a child shell - so `cd ..` gets lost when the child shell exists. But, we can then use parentheses to avoid `cd ..` completely. I'll send out a PR.

dimaryaz:
Actually, even that worked, `cd ..` would overwrite the return code from `npm` - so if `npm` failed, Travis would see a successful code.

Anyways, #687 should work.

PR #689, opened 2018-07-07
travis: bundle size monitoring

nl0:
monitor bundle size changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,8 +82,12 @@ script:
     fi
   - |
     if [[ "$TEST_CATALOG" = true ]]; then
-      (cd catalog && npm test)
+      (cd catalog && npm test && npm run build)
     fi
 
 after_success:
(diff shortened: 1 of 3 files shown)


nl0:
doesn't work as intended yet

nl0:
hm, nope, it does, actually.
@akarve rtm

nl0:
@akarve look at the checks section:
 > bundlesize — Total bundle size is 654.54KB/null (-null)

it will show how the PRs affect bundle size

akarve:
fine but do we need to set targets for it to be useful? or does it store past bundle sizes? it's just Xbk/null right now.

nl0:
@akarve i dont see a point for setting targets in our case, just monitor the changes.

> does it store past bundle sizes? it's just Xbk/null right now.

yep, it should show how every PR affects bundle size

PR #691, opened 2018-07-11
Don't skip header row when reading csv/excel files in unit tests

kevinemoore:
## Description
Skipping the header in Pandas read_csv and read_excel can produce DataFrames with funky column names and indexes, the testing of which is outside the scope of the current unit tests that specify the line skips.

In future, we should add Parquet/Pandas round-trip tests that cover all types of Pandas DataFrames, including uncommon indexing and column names.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -165,3 +165,16 @@ Symbolic links on Windows have a few quirks to be aware of.
  
  ### Solution
  `sudo pip install quilt # ¯\_(ツ)_/¯`
+
+## `TypeError: data type "mixed-integer" not understood` when reading a DataFrame from a package
+This error can occur for package nodes built using the Pandas "skiprows" parameter to skip a source file's header row (usually row 0). For example, this build.yml file, skips the header row in source.xlsx:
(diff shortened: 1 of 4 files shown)


docs/faq.md
```diff
@@ -165,3 +165,16 @@ Symbolic links on Windows have a few quirks to be aware of.
...
+## `TypeError: data type "mixed-integer" not understood` when reading a DataFrame from a package
+This error can occur for package nodes built using the Pandas "skiprows" parameter to skip a source file's header row (usually row 0). For example, this build.yml file, skips the header row in source.xlsx:
+```yaml
+    foo:
+        file: source.xlsx
+        kwargs:
+            skiprows: [0,...]
+            names: ['column0',...]
+```
+To skip the header row and supply a different set of column names, use the "names" parameter:
```
akarve: so `names:...` fixes the issue? it would be very helpful to know that row 0 actually isn't a row, but is the file header; beacause skiprows: [0...] on a CSV skips the first row.

PR #694, opened 2018-07-15
travis: adjust catalog build cfg for bundlesize calculations

nl0:
`bundlesize` couldnt properly compare build size with master because of changing hashes in filenames, so we need to make a build w/o hashes for this to work


nl0:
@akarve rtm. when this fix is in master, we should (supposedly) see proper bundle size diffs in PRs

PR #696, opened 2018-07-17
Allow insecure catalog connection

kevinemoore:
## Description
Adds logic to allow overriding the default catalog URL, which assumes HTTPS. Print a warning if the URL isn't HTTPS.

## Depends On
New Auth (so merging into flask-sec)


meffij:
Closing in favor of #708

PR #704, opened 2018-07-21
Make tokens last 90 days by default

meffij:


registry/quilt_server/auth.py
```diff
@@ -272,8 +272,9 @@ def revoke_user_code_tokens(user):
         db.session.delete(code)
     revoke_tokens(user)
 
-def get_exp(mins=30):
-    return datetime.utcnow() + timedelta(minutes=mins)
+def get_exp(mins=None):
+    delta = timedelta(minutes=mins) if mins is not None else timedelta(days=90)
```
akarve: this is funky. either don't take` minutes` as keyword arg or take both `minutes` and `days` and pass all of them on to `timedelta`. i'm guessing you're aiming for backwards compatibility, but i don't think it's worth it given how confusing this is.

atlernatively, make `mins=60*24*30` the default. and that brings me to another issue, `mins=None` is not a very useful default.

akarve: you could also take `**kwargs` and pass it through to `timedelta`; just put that in the docstring

registry/quilt_server/auth.py
```diff
@@ -12,7 +12,7 @@
 from . import app, db
 from .const import (VALID_EMAIL_RE, VALID_USERNAME_RE, blacklisted_name,
```
akarve: these should be alphabetized

registry/quilt_server/auth.py
```diff
@@ -272,15 +272,17 @@ def revoke_user_code_tokens(user):
         db.session.delete(code)
     revoke_tokens(user)
 
-def get_exp(mins=30):
-    return datetime.utcnow() + timedelta(minutes=mins)
+def get_exp(**kwargs):
```
akarve: semantically, this is not a `get` method. it is really `calc_exp` or something like that.

registry/quilt_server/const.py
```diff
@@ -36,3 +36,4 @@ def blacklisted_name(username):
 PASSWORD_RESET_SALT = 'reset'
```
akarve: should also be alphabetized or at least grouped and alphabetized within group

registry/quilt_server/const.py
```diff
@@ -36,3 +36,4 @@ def blacklisted_name(username):
...
 MAX_LINK_AGE = 60 * 60 * 24 # 24 hours
 CODE_EXP_MINUTES = 10
+TOKEN_EXP_DEFAULT = {'days': 90}
```
akarve: minor: this is not an "expiry" default but a "time to live" or "lifetime" default. expiry varies by token.

registry/quilt_server/const.py
```diff
@@ -22,17 +22,15 @@ class PaymentPlan(Enum):
...
     'anonymous',
     'quilt'
 ])
 
-def blacklisted_name(username):
-    return username in BAD_NAMES
-
 ACTIVATE_SALT = 'activate'
-PASSWORD_RESET_SALT = 'reset'
+CODE_TTL_DEFAULT = {'minutes': 10}
```
dimaryaz: Not really sure if this is better than just using seconds (and we already have `MAX_LINK_AGE` in seconds below)... But I don't have a strong opinion here, so whatever.

registry/quilt_server/const.py
```diff
@@ -22,17 +22,15 @@ class PaymentPlan(Enum):
 
 FTS_LANGUAGE = 'english'
 
-BAD_NAMES = set([
+BAD_NAMES = frozenset([
```
akarve: not a blocker but deserves a more descriptive name like `INVALID_USERNAMES`

PR #707, opened 2018-07-23
Fix logout

meffij:
Fix oversight on format of tokens in headers


registry/quilt_server/views.py
```diff
@@ -269,6 +269,8 @@ def wrapper(*args, **kwargs):
                               is_admin=user.is_admin,
                               is_active=user.is_active)
 
+                g.token_str = token
```
dimaryaz: I'd call it `g.auth_token`, just to make it more clear

PR #708, opened 2018-07-23
Allow insecure catalog connection with extra environment variable

meffij:


registry/quilt_server/prod_config.py
```diff
@@ -10,7 +10,11 @@
 REGISTRY_URL = os.environ['REGISTRY_URL']
 CATALOG_HOST = os.environ['CATALOG_HOST']
 
-CATALOG_URL = 'https://%s' % CATALOG_HOST
+CATALOG_URL = os.environ.get('CATALOG_URL', 'https://%s' % CATALOG_HOST)
```
dimaryaz: I'm fine with `CATALOG_URL` - but we should then drop `CATALOG_HOST`. No point in having two of them

registry/quilt_server/prod_config.py
```diff
@@ -10,7 +10,11 @@
...
+if not CATALOG_URL.startswith("https"):
+    print("WARNING: INSECURE CONNECTION TO CATALOG")
+    # require verbose environment variable to be defined
+    ALLOW_INSECURE_CATALOG_ACCESS = os.environ['ALLOW_INSECURE_CATALOG_ACCESS']
```
dimaryaz: Just make it an assert or something similar? No point in assigning it to a variable if you only need to check that it is defined.

registry/quilt_server/prod_config.py
```diff
@@ -8,9 +8,12 @@
 SQLALCHEMY_DATABASE_URI = os.environ['SQLALCHEMY_DATABASE_URI']
 
 REGISTRY_URL = os.environ['REGISTRY_URL']
-CATALOG_HOST = os.environ['CATALOG_HOST']
 
-CATALOG_URL = 'https://%s' % CATALOG_HOST
+CATALOG_URL = os.environ.get('CATALOG_URL')
```
dimaryaz: This should now use `[]`, so it crashes if the var is missing

PR #722, opened 2018-07-26
Implement subpackage push in the registry

dimaryaz:


registry/quilt_server/views.py
```diff
@@ -661,9 +661,14 @@ def download_object_preview(owner, obj_hash):
         )
 
 @app.route('/api/package/<owner>/<package_name>/<package_hash>', methods=['PUT'])
+@app.route('/api/package_update/<owner>/<package_name>/<path:package_path>', methods=['POST'])
```
dimaryaz: (I don't actually know what the path should be. Also, I'm guessing it should to be a `POST`, not a `PUT`, since it's not creating a new entity?)

registry/quilt_server/views.py
```diff
@@ -778,6 +787,47 @@ def package_put(owner, package_name, package_hash):
                     dict(team=app.config['TEAM_ID'], user=owner, pkg=package_name)
                 )
 
+    if package_path is not None:
```
kevinemoore: Probably cleaner to implement this conditional in a separate function.

registry/quilt_server/views.py
```diff
@@ -778,6 +787,47 @@ def package_put(owner, package_name, package_hash):
...
+                raise ApiException(requests.codes.not_found, "Invalid source subpath: %r" % component)
+            base_subnode_parent = base_subnode
+            try:
+                base_subnode = base_subnode.children.get(component)
+            except AttributeError:
+                raise ApiException(requests.codes.not_found, "Target subpath is not a group node: %r" % component)
+            if base_subnode is None:
+                break
+        try:
+            base_subnode_parent.children[component] = subnode
```
kevinemoore: I can live with this limitation, but it's not ideal. It would be nice to support e.g., changing a single data node into group node with the original data in the folder.

akarve: agree

registry/quilt_server/views.py
dimaryaz: Moving everything into a new function becomes kind of awkward: we need the `tag` variable later, so the function would need to return `(contents, tag)`. Also, locking the DB table as a side effect is weird. I'll move the node traversal logic for now.

kevinemoore: That's enough refactoring for this PR.

PR #723, opened 2018-07-26
Minimize time patch is in effect

meffij:
Could reduce flakiness of code tests


registry/tests/auth_test.py
```diff
@@ -455,15 +455,16 @@ def api_root(token_request):
     def testCodeExpires(self):
         self.code_immediate_expire_mock = mock.patch('quilt_server.auth.CODE_TTL_DEFAULT',
```
dimaryaz: You should just use `with patch(...):`, so it unpatches it automatically. Also, much simpler.

PR #725, opened 2018-07-27
Fix comment email templates

meffij:


registry/tests/mail_test.py
```diff
@@ -47,3 +49,10 @@ def testLinkWorksCorrectly(self, send_email):
         body = call['body']
         assert html.find(expected_test_link) != -1
         assert body.find(expected_test_link) != -1
+
+        expected_test_link = 'http://localhost:3000/test_user/pkg/comments'
```
akarve: we should check for the presence of `commenting_user` as well

meffij: done

PR #730, opened 2018-07-27
Implement subpackage builds

dimaryaz:


compiler/quilt/test/test_build.py
```diff
@@ -527,8 +527,42 @@ def test_parquet_source_file(self):
...
+
+        df = DataFrame(dict(a=[1, 2, 3]))
+        arr = np.array([4, 5, 6])
+        path = str(mydir / 'build_simple.yml')
+
+        command.build('test/foo/empty')
+        command.build('test/foo/df', df)
+        command.build('test/foo/arr', arr)
+        command.build('test/foo/file', path)  # Adds as a plain file
+        command.build('test/foo/stuff', path, build_file=True)  # Builds a subpackage
```
akarve: Nice-to-have, but not required: instead do `data_file=True` if the user wants to "add as a plain file" so that the signature of `quilt.build` is consistent across both packages and subpackages.

akarve:
TODO (in future PR?):
- [ ] why does Codecov say this will decrease coverage by so much
- [ ] docs

dimaryaz:
Codecov: no idea - doesn't make any sense
Delete a node: not directly; can delete /a/b by replacing the whole /a subtree, but can't delete /a.
Replace a sub-package tree: yes
Globbing: yes. Globbing is processed separately, so not affected by subpackages. It's the same thing as just explicitly listing all of the matches.

PR #731, opened 2018-07-27
Remove package class

kevinemoore:
## Description
Removes the Package class. This simplifies the codebase by reducing the number of conceptual representations of a package (from 3 to 2 by my count). The functionality was moved mostly to PackageStore and some to core.py (Node).

## TODO
- [x] Conflics
- [x] Pylint


compiler/quilt/tools/core.py
```diff
@@ -74,13 +79,102 @@ def __json__(self):
...
         return val
-
+   
 
 class RootNode(GroupNode):
     __slots__ = ()
 
     json_type = 'ROOT'
 
+    def __getitem__(self, item):
```
dimaryaz: I'd rather just get rid of this. I don't know why it was added to Package in the first place, it was never a public API, and supporting the "/" syntax just feels weird to me.

compiler/quilt/nodes.py
```diff
@@ -258,11 +256,12 @@ class PackageNode(GroupNode):
     """
     def __init__(self, package, meta):
         super(PackageNode, self).__init__(meta)
-        self._package = package
+        self._package = None
+        self._node = package if package is not None else None
```
dimaryaz: We really shouldn't store the `_node`. We're now back to keeping two package trees in memory.

kevinemoore: I agree, but this is just the first step. We used to store a reference to package object, which contained a reference to this root node. This just cuts out the Package object.

compiler/quilt/tools/core.py
kevinemoore: I agree, but it's used in build. Can we do that in a separate PR?

compiler/quilt/nodes.py
kevinemoore: I looked into removing it and it won't be easy until we merge core nodes and import nodes. The main issue is that the core node is the only way to get the package hash. I propose stopping here with this PR, then merging core and import nodes in a separate PR. Once we do that, we'll be down to a single tree.

compiler/quilt/tools/core.py
kevinemoore: Nevermind. I removed those and all changes to core.py.

compiler/quilt/tools/util.py
```diff
@@ -350,3 +350,33 @@ def fs_link(path, linkpath, linktype='soft'):
...
+
+def find_in_package(corenode, item):
+    """Find a (core) node in a package tree.
+    :param item: Node name or path, as in "node" or "node/subnode".
+    """
+    if not isinstance(corenode, core.Node):
+        raise QuiltException("Corenode must be a valid Quilt package node. %r" % corenode)
+
+    node = corenode
+    path = pathlib.PurePosixPath(item)
```
dimaryaz: I wouldn't use `pathlib` here, since this isn't an actual file path. Just split at `/`?

compiler/quilt/tools/store.py
```diff
@@ -96,6 +109,9 @@ def __init__(self, location=None):
             )
             raise StoreException(msg.format(self._path))
 
+    def __eq__(self, rhs):
```
dimaryaz: If you define it, you also need to define `__ne__` and `__hash__`.

(`__ne__` can be just `not (self == rhs)`, and `__hash__` can be `hash(self._path)`)

PR #736, opened 2018-07-31
Fix subpackage push

dimaryaz:
It got broken by a merge conflict.
Add a unit test.

### DIFF
--- a/compiler/quilt/test/test_push.py
+++ b/compiler/quilt/test/test_push.py
@@ -18,7 +18,7 @@ from .utils import QuiltTestCase
 def _decode_body(body):
     with gzip.GzipFile(fileobj=BytesIO(body)) as fd:
         ungzipped_body = fd.read()
-    return json.loads(ungzipped_body, object_hook=decode_node)
+    return json.loads(ungzipped_body.decode('utf-8'), object_hook=decode_node)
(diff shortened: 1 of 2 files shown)


compiler/quilt/test/test_push.py
```diff
@@ -59,11 +67,70 @@ def test_push(self):
...
+        for blob_hash in all_hashes:
+            urls = upload_urls[blob_hash]
+
+            self.requests_mock.add(responses.HEAD, urls['head'], status=404)
+            self.requests_mock.add(responses.PUT, urls['put'])
+
+        self._mock_update_package('foo/bar', 'foo', contents, upload_urls)
+
+        # Push a subpackage.
+        command.push('foo/bar/foo')
```
akarve: this _should_ fail without `is_pubilc=True`?

akarve: at least for a mocked up "free" user.

dimaryaz: Well, there's no server... It's all mocked by the `callback1` and `callback2` methods below. I could make them fail without `is_public=True`, but that wouldn't test anything other than "does the client pass the is_public parameter to the server?", so really trivial.

PR #737, opened 2018-07-31
Suppress unhelpful warning for python2.7 users

meffij:


compiler/quilt/__init__.py
```diff
@@ -7,6 +7,10 @@
 # None: CLI params have not yet been parsed to determine mode.
 _DEV_MODE = None
 
+# Suppress numpy warnings for Python 2.7
+import warnings
```
akarve: `pylint` wants to know why this `import` isn't at the top of the file; and it's kinda right

PR #744, opened 2018-08-07
[WIP] Add capability to build directly from `DataNode` objects

AffineParameter:
## Description
This PR allows a user to build a package component with a `DataNode` directly (for "on the fly" build workflows). This additional functionality allows users to explicitly craft the new nodes data and metadata simultaneously, representing a ~2x built-time improvement over it's existing workaround on my use case (35s vs 67s).

Workaround:
```python

for dense_data, sparse_data from data_to_build:
    # dense_data = np.random.uniform(0, 1, size=(int(1.5e5), 25, 4))
    # sparse_data = metadata_dict

    # build dense data `numpy` node
    quilt.build(data_node_path, dense_data)

    # re-load the pkg and assign it
    pkg = quilt.load(q_pkg_name)
    
    # add the sparse data to the `_meta` attr
    pkg[group['name']][l_id]._meta.update(sparse_data)
    
    # rebuild w/ additional metadata
    quilt.build(q_pkg_name, pkg)
```

Proposal:
```python

for dense_data, sparse_data from data_to_build:
    # dense_data = np.random.uniform(0, 1, size=(int(1.5e5), 25, 4))
    # sparse_data = metadata_dict

    # build dense data `numpy` node
    quilt.build(
        data_node_path, 
        quilt.nodes.DataNode(None, None, dense_data, sparse_data)
    )
```

## Depends On
As far as I know, there are no dependencies

## TODO
Unit tests run for me in 2.7.13 & 3.6, though I wasn't able to identify where to add a new one or amend an existing one. Also, not sure what documentation would be useful for this.

- [ ] Unit Tests
- [ ] Documentation

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -532,7 +532,7 @@ def _build_internal(package, path, dry_run, env, build_file):
                     rmtree(tmpdir)
         else:
             build_from_path(package, path, dry_run=dry_run, env=env)
-    elif isinstance(path, nodes.GroupNode):
+    elif isinstance(path, (nodes.GroupNode, nodes.DataNode)):
         assert not dry_run  # TODO?
         build_from_node(package, path)
     elif isinstance(path, string_types + (pd.DataFrame, np.ndarray)):
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -532,7 +532,7 @@ def _build_internal(package, path, dry_run, env, build_file):
                     rmtree(tmpdir)
         else:
             build_from_path(package, path, dry_run=dry_run, env=env)
-    elif isinstance(path, (nodes.GroupNode, nodes.DataNode)):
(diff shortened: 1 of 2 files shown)


compiler/quilt/tools/command.py
```diff
@@ -532,7 +532,7 @@ def _build_internal(package, path, dry_run, env, build_file):
                     rmtree(tmpdir)
         else:
             build_from_path(package, path, dry_run=dry_run, env=env)
-    elif isinstance(path, nodes.GroupNode):
+    elif isinstance(path, (nodes.GroupNode, nodes.DataNode)):
```
dimaryaz: Actually, you can just check for `nodes.Node` -  it's their parent class, and there are no other subclasses.

AffineParameter: Will do!

AffineParameter:
Jumping on it, and thanks for the review!

PR #746, opened 2018-08-07
Fix building of group metadata

dimaryaz:
It wasn't actually getting saved.


dimaryaz:
@AffineParameter: this fixes the bug you found

PR #747, opened 2018-08-07
Workaround pyarrow bugs

kevinemoore:
## Description
Retry table->pandas conversion without the pandas metadata in case we got buggy pandas metadata from an older version of pyarrow.

### DIFF
--- a/compiler/quilt/tools/store.py
+++ b/compiler/quilt/tools/store.py
@@ -376,7 +376,16 @@ class PackageStore(object):
         objfiles = [self.object_path(h) for h in hash_list]
         dataset = ParquetDataset(objfiles)
         table = dataset.read(nthreads=4)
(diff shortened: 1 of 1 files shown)


compiler/quilt/tools/store.py
```diff
@@ -376,7 +376,16 @@ def _read_parquet_arrow(self, hash_list):
         objfiles = [self.object_path(h) for h in hash_list]
         dataset = ParquetDataset(objfiles)
         table = dataset.read(nthreads=4)
-        dataframe = table.to_pandas()
+        try:
+            dataframe = table.to_pandas()
+        except:
```
dimaryaz: Catch a more specific exception - `AssertionError`, `Exception`, etc..; you don't want to catch things like `KeyboardInterrupt` or `SystemExit`.

PR #752, opened 2018-08-15
Add save() function

diegoscarabelli:
## Description
Add function to build a data node with associated metadata and push it to the registry.

### DIFF
--- a/compiler/quilt/__init__.py
+++ b/compiler/quilt/__init__.py
@@ -120,4 +120,5 @@ from .tools.command import (
     version_list,
 )
 
-from .tools.util import save
+from .util import save
+
--- a/compiler/quilt/__init__.py
+++ b/compiler/quilt/__init__.py
@@ -119,3 +119,5 @@ from .tools.command import (
     version_add,
     version_list,
 )
+
+from .tools.util import save
--- a/compiler/quilt/tools/util.py
+++ b/compiler/quilt/tools/util.py
@@ -12,7 +12,6 @@ import sys
 from appdirs import user_config_dir, user_data_dir
 from collections import namedtuple
 from six import BytesIO, string_types, Iterator
-from typing import Union
 
 from .const import QuiltException
 from .compat import pathlib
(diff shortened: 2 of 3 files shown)


compiler/quilt/tools/util.py
```diff
@@ -377,3 +378,25 @@ def find_in_package(corenode, path):
...
+         params: dict):
+    """Build and push data to Quilt registry at user/package/data_node,
+    associating params as metadata for the data node.
+    :param user: user name
+    :param package: package name
+    :param data_node: data node name
+    :param data: data to save
+    :param params: metadata
+    """
+    quilt.build(f'{user}/{package}/{data_node}', data)
```
akarve: These are not compatible with Python 2.7; and I think that's why CI is blowing up. I _think_ `"{}/{}/{}".format(...)` will work

compiler/quilt/__init__.py
```diff
@@ -119,3 +119,5 @@ def handle_interrupt(signum, stack):
     version_add,
     version_list,
 )
+
+from .tools.util import save
```
akarve: A blank line at the very end of each file should eliminate the "No EOF" warning we see above

compiler/quilt/tools/util.py
```diff
@@ -377,3 +378,25 @@ def find_in_package(corenode, path):
...
+    pkg = quilt.load(f'{user}/{package}')
+    for key, value in params.items():
+        if isinstance(value, np.ndarray):
+            value = value.astype(float)
+            params[key] = value.tolist()
+    pkg[data_node]._meta.update(params)
+    quilt.build(f'{user}/{package}', pkg)
+    quilt.push(f'{user}/{package}/{data_node}', is_public=True)
```
akarve: the `is_public` kwarg should be passed through from the top of the function (i.e. not everyone will want to push the packages `publicly`; similar strategy for other build, push args would be helpful. it is not uncommon to pass `kwargs` through to subroutines, but you may need to filter out inapplicable keywords.

akarve:
A few inline comments. Please bump up to `quilt.util` (instead of `quilt.tools.util`).

akarve:
TODO:
- [ ] investigate numpy imports failing in existing tests

compiler/quilt/tools/util.py
kevinemoore: Typing also isn't supported in Python2.

diegoscarabelli:
@kevinemoore @akarve I should have implemented the changes that you suggested. It seems that the tests have passed and there are no conflicts. Let me know if there are any further tweaks ti apply.

PR #754, opened 2018-09-05
Send welcome email for normal users on activate

meffij:


registry/quilt_server/auth.py
```diff
@@ -79,7 +79,9 @@ def get_admins():
 def activate_response(link):
     payload = verify_activation_link(link)
     if payload:
-        _activate_user(User.query.filter_by(id=payload['id']).with_for_update().one_or_none())
+        user = User.query.filter_by(id=payload['id']).with_for_update().one_or_none()
```
dimaryaz: Should really be `one()`, not `one_or_none()`

PR #757, opened 2018-09-27
Yaml security fix

eode:
## Description
We have a minor security issue that allows arbitrary code execution.  While arbitrary execution is typically a major issue, this one is fairly minor because:

* quilt is not typically executed with admin permissions
* attacker must get the user to use a compromised build (or other YAML) file in order to perform the attack

Since we're not loading data from YAML files, just loading configs, this is a trivial issue to fix -- all existing `load()` calls have been replaced by `safe_load()` calls.

## Depends On
Nothing

## TODO
- [x] Bug Fix
- [x] Unit Tests

### DIFF
--- a/compiler/pytest.ini
+++ b/compiler/pytest.ini
@@ -1,7 +1,9 @@
 [pytest]
 filterwarnings =
     default
+    ; Warning in cookies module triggered by responses lib.
     ignore:::cookies
+    ; Pandas triggers a bug in the current version of Cython.
+    ; Unfortunately, we lose monitoring of any potential ImportWarnings,
     ignore::ImportWarning
-    ignore:::IPython.lib.pretty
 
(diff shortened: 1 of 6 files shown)


compiler/pytest.ini
```diff
@@ -0,0 +1,9 @@
+[pytest]
+filterwarnings =
+    default
+    ; Warning in cookies module triggered by responses lib.
+    ignore:::cookies
+    ; Pandas triggers a bug in the current version of Cython.
+    ; Unfortunately, we lose monitoring of any potential ImportWarnings,
+    ignore::ImportWarning
+
```
eode: Was getting extraneous warnings from pytest.  The 'ignore::ImportWarning' is broad, but can be removed once cython bugfix is released.

PR #758, opened 2018-09-27
Update env

kevinemoore:
## Description
Removing the dependence on docker_config.py in docker-compose-dev and docker-compose-local-auth. The docker-compose files now include required variables for new-auth.


registry/docker-compose-local-auth.yml
```diff
@@ -33,10 +35,14 @@ services:
...
       - AWS_SECRET_ACCESS_KEY=abcd
-      - QUILT_SERVER_CONFIG=docker_config.py
-      - AUTH_PROVIDER=quilt
+      - QUILT_SERVER_CONFIG=dev_config.py
+      - SQLALCHEMY_DATABASE_URI=postgresql://postgres:testing@db/packages
       - DEPLOYMENT_ID=dev
       - REGISTRY_URL=http://localhost:5000
+      - QUILT_SECRET_KEY=aaaaaaaaaa
+      - S3_ENDPOINT = http://s3:5001
+      - DISABLE_SIGNUP
```
dimaryaz: Does this do anything? If it sets it to an empty string, then I think that's already the default

registry/docker-compose-local-auth.yml
```diff
@@ -33,10 +35,14 @@ services:
...
+      - ENABLE_USER_ENDPOINTS
```
dimaryaz: It's hard-coded to `True` in `dev_config.py`

PR #759, opened 2018-09-28
Fix zero-byte file downloads

eode:
## Description
Zero-byte files weren't downloading due to HTTP `Range` only supporting files of at least one byte.  For package consistency, we want to actually *know* that that file exists and is zero bytes, even though we already know what the contents *should* be.

This uses a different code path for zero-byte files, and verifies that the file exists and downloads correctly.  IMO this is the most readable way of handling the situation, and it has no performance penalty.

## TODO

- [ ] Unit Tests -- should I make a test for this?

### DIFF
--- a/compiler/quilt/tools/data_transfer.py
+++ b/compiler/quilt/tools/data_transfer.py
@@ -95,6 +95,33 @@ def download_fragments(store, obj_urls, obj_sizes):
                             try:
                                 existing_file_size = output_file.tell()
 
+                                # For zero-byte downloads, we don't need to resume, and range download
+                                # must be at least 1 anyways. 
+                                if total_bytes == 0:
+                                    response = s3_session.get(
+                                        url,
+                                        timeout=(S3_CONNECT_TIMEOUT, S3_READ_TIMEOUT)
+                                    )
+                                    if not response.ok:
+                                        message = (
+                                            "Download failed for {obj_hash}:\n"
+                                            "URL: {response.request.url}\n"
+                                            "Status code: {response.status_code}\n"
+                                            "Response: {response.text!r}\n"
+                                        ).format(**locals())  # Splat **kwargs operators ftw
+                                        with lock:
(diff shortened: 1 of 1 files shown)


compiler/quilt/tools/data_transfer.py
```diff
@@ -95,6 +95,33 @@ def _worker_thread():
                             try:
                                 existing_file_size = output_file.tell()
 
+                                # For zero-byte downloads, we don't need to resume, and range download
+                                # must be at least 1 anyways. 
+                                if total_bytes == 0:
```
dimaryaz: Actually, `total_bytes` is the total size of the whole package, not an individual file. That would be `original_size`.

(Also, technically, size might not be known cause some really old quilt clients don't set it... But it shouldn't happen in practice.)

eode: Hrm.. Thanks!

eode: Should we handle that case?  We could probably find a way to work around that.  Adds complexity, but if it's needed, it's needed.

dimaryaz: I'd punt on that. But, you should change the `or 0` code above to just raise an error.

eode: Are you familiar enough to know whether or not I can raise an error that recommends re-uploading the package?  I.e., would '--reupload' fix the issue?

dimaryaz: Familiar enough, and it's not worth worrying about. In prod, we have a total of 5 such files, created by us while testing stuff. And the clients that didn't have size support are so old that no one uses them anymore.

eode: Sweet.

eode: Now raises `QuiltException("Malformed package: No size for object " + obj_hash")`

PR #762, opened 2018-10-02
Fix corepy deviation between registry & compiler

eode:
## Description
Fix trivial deviations between compiler and registry core.py.


eode:
This is a pretty trivial change, but still.  Deviations can add up.

PR #769, opened 2018-10-16
Docs and changes re: handling other S3-compatible providers

eode:
#### Code Changes
* fix testing to return 206 instead of 200 for range downloads
* fix download code to handle case where the range request is ignored
* Added `registry\quilt_server\configs\example.env`
  * should have *all* server-side quilt config values, including those for the catalog
* Change `env_config.py` to include every config var
* Change `views.py` to handle GCP S3 Interop endpoint
* Created `docker-compose-env.py` for starting a server session from config file
* Created simple nginx dockerfile that just contains our nginx config template and startup command.

#### Doc Changes
* We now can (from what I can tell) support Google Cloud Platform Storage and Azure Blob Storage
* This is largely a matter of documentation, but in setting it up and testing, the existing docs for a non-development setup were outdated or partial / missing information
* That resulted in a larger documentation update (in this readme)
* Server modifies signed URL to work with google
  * Client and server must be upgraded to include these changes work properly with GCP.
* Minio setup added to docs
* Docker section moved to bottom, with a link at the top, as these docs are primarily about Quilt, and docker setup is secondary (though necessary).

## TODO
* [x] Join the env files now shown in the AWS section, accounting for any naming collisions that may exist
* [x] Clearly comment the resultant unified env file -- what aspects are critical, what aspects are optional, etc.
* [x] Do a simple docker-nginx image including the nginx config file template
* [x] Publish that nginx as quiltdata/nginx
* [x] Make the 'AWS' docs entry be something like *setting up a public-facing server*, and generalize it slightly.
* [x] Make a `production` option for `docker-compose` that depends on the merged env file.  Setting up a server would be:  Check out github repo, edit env file, and do `docker-compose start production`.
* [x] Redo AWS section to include bare minimum of all information to include alternate S3 sources
* [x] Fold my small section on choosing an alternate S3 Provider into the *Setting up a public-facing server* section.
* [x] clean up any misc 

#### Probably a different PR
* ~[ ] Fold 'env_config.py' into 'prod_config.py', with sensible conservative defaults~ out-of-scope
* ~[ ] Place comments in relevant templates that use the env files, warning about potential for name conflicts~


eode:
It seems that Minio doesn't handle the URLs correctly.  I can access Azure Blob via boto, but when quilt requests a pre-signed URL, it comes back as an AWS link, and I think that's what's provided by minio, as Quilt seems to handle Google S3 just fine.  I haven't tried S3Proxy, but it didn't look quite as solid as Minio.  Maybe I can open a bug with Minio.

registry/README.md
```diff
@@ -118,66 +118,109 @@ Note: ```quilt login``` is its own test that the server is working.  You can now
...
-* an S3 bucket
-* an EC2 instance to run the registry, authentication service and catalog. Add the following policies to the security group: Postgres, HTTP/HTTPS, SSH and a custom TCP rule to enable port 5000.
-* an Elastic Load Balancer (ELB) to terminate SSL connections to the registry and catalog
-
-Once the resources have been created, ssh into the EC2 instance and configure the environment and run the services via Docker. You can make it easier to connect to your registry by creating a new DNS record to point to the EC2 instance's external IP address (e.g., quilt.yourdomain.com).
-
+## Systems configuration
+Running a Quilt Registry services:
+* a Postgres database. [Create a Postres Database in Amazon RDS](https://aws.amazon.com/rds/postgresql/)
+* an S3 bucket and access keys
```
kevinemoore: Looks like you're dropping AWS specific language at the top, but keeping or maybe adding specific mentions of S3. If it's not meant to be AWS specific, it would be clearer to use more general terms for the storage and compute resources.

eode:
Questions:
* Is the section on Github Auth still valid?
* Is the section on headless installation still needed?

eode:
I'm going to break this PR up into bite-sized chunks.

eode:
I'm going to break some of this into smaller PRs to reduce complexity.

PR #771, opened 2018-10-23
Make compiler and registry compatible with GCP

eode:
## Description
Make compiler and registry compatible with Google Cloud Platform Storage in S3 Interop mode.

Making #769 smaller.

### DIFF
--- a/compiler/quilt/tools/data_transfer.py
+++ b/compiler/quilt/tools/data_transfer.py
@@ -152,14 +152,23 @@ def download_fragments(store, obj_urls, obj_sizes):
                                 encoding = response.raw.headers.pop('Content-Encoding', None)
 
                                 # Make sure we're getting the expected range.
-                                content_range = response.headers.get('Content-Range', '')
-                                match = CONTENT_RANGE_RE.match(content_range)
-                                if not match or not int(match.group(1)) == range_start:
-                                    with lock:
-                                        tqdm.write("Unexpected Content-Range: %s" % content_range)
-                                    break
+                                if response.status_code == 200:  # server ignored range request
+                                    compressed_size = response.headers.get('Content-Length', '').strip()
+                                    if not compressed_size:
+                                        # HTTP: Content length is specified by 'Content-Length' header, or
+                                        # by closing the connection when done. :-/
+                                        compressed_size = len(response.content)
+                                    else:
+                                        compressed_size = int(compressed_size)
+                                else:
+                                    content_range = response.headers.get('Content-Range', '')
+                                    match = CONTENT_RANGE_RE.match(content_range)
+                                    if not match or not int(match.group(1)) == range_start:
+                                        with lock:
+                                            tqdm.write("Unexpected Content-Range: %s" % content_range)
+                                        break
 
-                                compressed_size = int(match.group(3))
+                                    compressed_size = int(match.group(3))
 
                                 # We may have started with a partially-downloaded file, so update the progress bar.
                                 compressed_read = existing_file_size
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -17,6 +17,7 @@ from functools import wraps
 import gzip
 import json
 import pathlib
+import re
 import time
 
 import boto3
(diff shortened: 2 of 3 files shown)


eode:
Do we do automated Registry testing?

..in any case, registry tests pass for dev environment.

compiler/quilt/tools/data_transfer.py
```diff
@@ -152,14 +152,23 @@ def _worker_thread():
...
-                                if not match or not int(match.group(1)) == range_start:
-                                    with lock:
-                                        tqdm.write("Unexpected Content-Range: %s" % content_range)
-                                    break
+                                if response.status_code == 200:  # server ignored range request
+                                    compressed_size = response.headers.get('Content-Length', '').strip()
+                                    if not compressed_size:
+                                        # HTTP: Content length is specified by 'Content-Length' header, or
+                                        # by closing the connection when done. :-/
+                                        compressed_size = len(response.content)
```
dimaryaz: This forces the whole contents to be downloaded - and loaded into memory - right? So probably not a good idea. Does it actually happen in practice? If not, I'd just fail in this case.

registry/quilt_server/views.py
```diff
@@ -503,14 +504,21 @@ def _mp_track(**kwargs):
...
             Bucket=PACKAGE_BUCKET_NAME,
             Key='%s/%s/%s' % (OBJ_DIR, owner, blob_hash)
         ),
         ExpiresIn=PACKAGE_URL_EXPIRATION
     )
+    # This is similar to what's done in the google documentation for a simple
+    # migration, but for pre-signed URLs (for which no example is given).
+    #    https://cloud.google.com/storage/docs/migrating#keys
+    # Modify URLs for Google Storage
+    if url.lower().startswith('https://storage.googleapis.com/'):
```
dimaryaz: Could you use `urllib.parse` for this? Regex matching does look really hacky.

compiler/quilt/tools/data_transfer.py
eode: Then we can give up on Google Cloud Platform compatibility.

My *suspicion* is that GCP Storage ignores the range request if the file is too small to warrant it, since their stated documentation is that they support the range request.  However, they do not always return a 206 response (and aren't really required to by the HTTP standard, as it's an optional server feature).

We could, however print a warning (via `tqdm`), like "Range request ignored by S3 server, file will be fully loaded into memory: <s3 path>", perhaps.

registry/quilt_server/views.py
eode: Sure, I was just trying to perturb the URL as little as possible and keep as concise as possible.

compiler/quilt/tools/data_transfer.py
dimaryaz: Well, it's fine if it ignores the range request - but does it really not return `Content-Length`?

registry/quilt_server/views.py
```diff
@@ -503,14 +507,26 @@ def _mp_track(**kwargs):
...
         ),
         ExpiresIn=PACKAGE_URL_EXPIRATION
     )
+    # This is similar to what's done in the google documentation for a simple
+    # migration, but for pre-signed URLs (for which no example is given).
+    #    https://cloud.google.com/storage/docs/migrating#keys
+    parsed_url = urlparse(url)
+    if parsed_url.hostname.lower() in GOOGLE_STORAGE_ENDPOINTS:
+        parsed_query = parse_qsl(parsed_url.query)
+        for index, item in enumerate(parsed_query):
```
dimaryaz: Hm, this looks unnecessarily difficult... Would something like this work? https://stackoverflow.com/questions/2506379/add-params-to-given-url-in-python I guess it won't preserve the order, but I can't imagine that making a difference.

compiler/quilt/tools/data_transfer.py
eode: It's not that Google doesn't send `Content-Length`, it's that no server is required to for a regular 200 response.  I mean, they should, who wouldn't, right?  But http doesn't require it.

registry/quilt_server/views.py
eode: *shrug* this retains order, the other doesn't.  The complexity's about the same, though this has more lines.  I can shoot for brevity here.

eode: Fixed.

PR #790, opened 2018-12-20
Allow push package by hash

kevinemoore:
## Description
Adding a hash argument to push to allow pushing package instances other than `latest`.

TODO:
- [ ] add a push-by-hash unit test

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -648,7 +648,7 @@ def log(package):
             str(entry.get('tags', [])), str(entry.get('versions', []))))
     _print_table(table)
 
-def push(package, is_public=False, is_team=False, reupload=False):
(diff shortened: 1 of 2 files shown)


compiler/quilt/tools/command.py
```diff
@@ -648,19 +648,21 @@ def log(package):
             str(entry.get('tags', [])), str(entry.get('versions', []))))
     _print_table(table)
 
-def push(package, is_public=False, is_team=False, reupload=False):
+def push(package, hash=None, is_public=False, is_team=False, reupload=False):
```
dimaryaz: Make it the last parameter, just in case someone uses the API with non-keyword args.

kevinemoore: Sure. I was trying to match install, but I think you're right. Better to make it the last arg to avoid breaking existing code.

PR #793, opened 2019-01-08
Add `quilt config` for Teams registries

kevinemoore:
## Description

Adding a feature to allow using `quilt config` to set a custom url for a Teams Registry--that allows using the team namespace to address packages.

### DIFF
--- a/compiler/quilt/tools/command.py
+++ b/compiler/quilt/tools/command.py
@@ -131,7 +131,12 @@ def _save_auth(cfg):
 
 def get_registry_url(team):
     if team is not None:
-        return "https://%s-registry.team.quiltdata.com" % team
+        # Check config
+        cfg = _load_config()
+        if 'team_name' in cfg and cfg['team_name'] == team:
+            return cfg['registry_url']
+        else:
+            return "https://%s-registry.team.quiltdata.com" % team
 
     global _registry_url
     if _registry_url is not None:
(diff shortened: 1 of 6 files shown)


compiler/quilt/tools/command.py
```diff
@@ -131,7 +131,12 @@ def _save_auth(cfg):
 
 def get_registry_url(team):
     if team is not None:
-        return "https://%s-registry.team.quiltdata.com" % team
+        # Check config
+        cfg = _load_config()
+        if 'team_name' in cfg and cfg['team_name'] == team:
```
dimaryaz: `if cfg.get('team_name') == team:`

compiler/quilt/tools/command.py
```diff
@@ -131,7 +131,12 @@ def _save_auth(cfg):
 
 def get_registry_url(team):
```
dimaryaz: What happens when a team is configured, but the user does something like `quilt install foo/bar` without a team prefix?

I think you may need to modify `url = cfg.get('registry_url', '')` below, and make sure a team URL is not returned if team is `None`.

compiler/quilt/tools/command.py
```diff
@@ -148,21 +153,33 @@ def get_registry_url(team):
...
             raise CommandException("Invalid URL: %s" % answer)
         canonical_url = urlunparse(url)
-    else:
-        # When saving the config, store '' instead of the actual URL in case we ever change it.
-        canonical_url = ''
-
-    cfg = _load_config()
-    cfg['registry_url'] = canonical_url
+        
+        if team:
```
dimaryaz: `if team is not None`, to be consistent with the checks above?

compiler/quilt/tools/command.py
```diff
@@ -148,21 +153,33 @@ def get_registry_url(team):
...
+            _check_team_id(team)
+            cfg['team_registry_url'] = canonical_url
+            cfg['team_id'] = team
+        else:
+            cfg['registry_url'] = canonical_url
+            cfg['team_id'] = ''
```
dimaryaz: Should this be `None`?

compiler/quilt/tools/command.py
```diff
@@ -131,7 +131,12 @@ def _save_auth(cfg):
 
 def get_registry_url(team):
     if team is not None:
-        return "https://%s-registry.team.quiltdata.com" % team
+        # Check config
+        cfg = _load_config()
+        if team and 'team_id' in cfg and cfg.get('team_id') == team:
```
dimaryaz: I meant, `'team_id' in cfg` isn't needed - `cfg.get('team_id') == team` should be sufficient, I think?

compiler/quilt/tools/command.py
dimaryaz: Ok. Doesn't really matter, I guess... Just make sure `config('')` doesn't do something bad. The two different checks, `if team` vs `if team is not None`, look like they could cause problems.

PR #794, opened 2019-01-09
Update landing page

akarve:
Landing page
* "manage data like code"
* Remove Teams from Pricing


akarve:
Merging in spite of unit tests failures since #795 proves and fixes the cause thereof

PR #801, opened 2019-02-05
Role support for Quilt registry

meffij:


eode:
This (or the PR it is split from) is a dependency for the office add-in.

registry/quilt_server/views.py
```diff
@@ -2337,3 +2475,35 @@ def comments_list(owner, package_name):
...
+        AccessKeyId(string): access key ID
+        SecretKey(string): secret key
+        SessionToken(string): session token
+    """
+    role_id = g.user.role_id
+    if not role_id:
+        raise ApiException(requests.codes.bad_request,
+                           "You have no attached role to assume")
+    role = Role.query.filter_by(id=role_id).one_or_none()
+    if not role:
```
dimaryaz: Shouldn't be possible - i.e., this should be enforced by Postgres.

You can also add a `db.relationship` to the `User` class so you can just access `g.user.role` instead of running a query manually. (There are examples in other classes in model.py)

registry/quilt_server/dev_config.py
```diff
@@ -14,8 +14,8 @@
 
 REGISTRY_URL = 'http://localhost:5000'
 
-AWS_ACCESS_KEY_ID = 'fake_id'
-AWS_SECRET_ACCESS_KEY = 'fake_secret'
+AWS_ACCESS_KEY_ID = os.getenv('AWS_ACCESS_KEY_ID', 'fake_id')
```
dimaryaz: I'm a bit nervous about using the default AWS environment names. Currently, we explicitly override access key and secret with bogus values - but with your change, it will pick up the real ones if they're set using env variables rather than `~/.aws`.

I'd rather call them something like `DEV_ACCESS_KEY_ID`, etc.

registry/quilt_server/models.py
```diff
@@ -244,3 +245,8 @@ class Comment(db.Model):
     contents = db.Column(db.String(MAX_COMMENT_LENGTH), nullable=False)
 
     package = db.relationship('Package', back_populates='comments')
+
+class Role(db.Model):
+    id = db.Column(postgresql.UUID, primary_key=True)
+    name = db.Column(db.String(200), unique=True)
+    arn = db.Column(db.String(200))
```
kevinemoore: Does it make sense to make ARN the primary key? It should at least be unique to prevent adding multiple entries in the registry db for the same IAM Role.

registry/quilt_server/views.py
```diff
@@ -88,6 +91,13 @@
     aws_secret_access_key=app.config.get('AWS_SECRET_ACCESS_KEY')
 )
 
+sts_client = boto3.client(
+    'sts',
+    region_name='us-east-1',
```
kevinemoore: We absolutely cannot hard-code the region here. I don't the region is needed at all for this code. @meffij I have a branch `cross-account` that uses `sts` if you want an example.

registry/quilt_server/views.py
```diff
@@ -2307,6 +2317,134 @@ def _comment_dict(comment):
...
+        username(string): username of user to edit
+        role_name(string): name of role to attach to user
+
+    A user can only have one role at a time.
+    To remove a role from a user, set role_name to the empty string.
+    """
+    data = request.get_json()
+    username = data['username']
+    role_name = data['role']
+    user = User.query.filter_by(name=username).one_or_none()
```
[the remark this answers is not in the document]
kevinemoore: You should be able to do this with User.update().where(...).values(...)

registry/quilt_server/views.py
```diff
@@ -2307,6 +2317,134 @@ def _comment_dict(comment):
         contents=comment.contents
     )
 
+@app.route('/api/users/attach_role', methods=['POST'])
```
kevinemoore: This endpoint seems oddly named. If there's an `attach_role`, I'd expect a corresponding `detach_role`. On the other hand if you want a single endpoint to attach and detach, maybe `set_role` would be a better name.

registry/quilt_server/views.py
```diff
@@ -2307,6 +2317,134 @@ def _comment_dict(comment):
...
+        user.role_id = role.id
+
+    db.session.add(user)
+    db.session.commit()
+    return {}
+
+@app.route('/api/roles/edit', methods=['POST'])
+@api(enabled=ENABLE_USER_ENDPOINTS, require_admin=True, schema=ROLE_DETAILS_SCHEMA)
+@as_json
+def edit_role():
```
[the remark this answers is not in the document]
kevinemoore: I second @dimaryaz's suggestion for separate add/edit/delete endpoints.

registry/quilt_server/models.py
```diff
@@ -244,3 +245,8 @@ class Comment(db.Model):
     contents = db.Column(db.String(MAX_COMMENT_LENGTH), nullable=False)
 
     package = db.relationship('Package', back_populates='comments')
+
+class Role(db.Model):
+    id = db.Column(postgresql.UUID, primary_key=True)
+    name = db.Column(db.String(200), unique=True)
```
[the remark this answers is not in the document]
meffij: That seems correct

registry/quilt_server/views.py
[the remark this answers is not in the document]
meffij: Probably not needed, not sure why I added it originally

registry/quilt_server/schemas.py
```diff
@@ -219,6 +219,37 @@
...
+    'type': 'object',
+    'properties': {
+        'username': {
+            'type': 'string'
+        },
+        'role': {
+            'type': 'string'
+        }
+    },
+    'required': ['username'],
```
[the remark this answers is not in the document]
meffij: The endpoint split should make the schemas more tenable

registry/quilt_server/views.py
stevededalus: I switched this to set_roles in latest commit

registry/quilt_server/views.py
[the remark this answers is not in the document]
stevededalus: this has been removed

stevededalus: this has been removed

registry/quilt_server/views.py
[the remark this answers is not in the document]
stevededalus: Done, PTAL

registry/quilt_server/views.py
```diff
@@ -2307,6 +2317,134 @@ def _comment_dict(comment):
...
+    To remove a role from a user, set role_name to the empty string.
+    """
+    data = request.get_json()
+    username = data['username']
+    role_name = data['role']
+    user = User.query.filter_by(name=username).one_or_none()
+    if user is None:
+        raise ApiException(requests.codes.bad_request,
+                           "No user exists by the provided name.")
+    if role_name is '':
```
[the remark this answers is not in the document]
stevededalus: Done

registry/quilt_server/views.py
```diff
@@ -2307,6 +2317,134 @@ def _comment_dict(comment):
...
+    """
+    roles_list = []
+    roles = Role.query.all()
+    for role in roles:
+        role_dict = {
+            'name': role.name,
+            'arn': role.arn
+        }
+        roles_list.append(role_dict)
+    return {'results': roles_list}
```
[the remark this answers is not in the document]
stevededalus: Done.

registry/quilt_server/models.py
stevededalus: I think it shouldn't be. If an admin creates a role and assigns a bunch of users to it and thne wants to change the ARN they should be able to with our changing the friendly_name of the role. Similarly, they may want to have the same ARN for different friendlynames

registry/quilt_server/models.py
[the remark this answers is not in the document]
stevededalus: Done

registry/quilt_server/models.py
```diff
@@ -200,6 +200,7 @@ class User(db.Model):
     is_active = db.Column(db.Boolean, default=False, nullable=False)
     date_joined = db.Column(postgresql.TIMESTAMP(True), server_default=db.func.now(), nullable=False)
     old_id = db.Column(db.BigInteger) # for django ID -- probably not necessary but good to keep around
+    role_id = db.Column(postgresql.UUID, db.ForeignKey('role.id'), nullable=True)
```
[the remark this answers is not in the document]
stevededalus: Done.

stevededalus:
I believe the remaining issues are:
1.) Fix credential env variables
2.) Locking on Update role
3.) Add a db.relationship for Role to the User model

Otherwise, PTAL

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,174 @@ def _comment_dict(comment):
...
+    role_name = data['name']
+    arn = data.get('arn', None)
+    role = Role.query.filter_by(name=role_name).one_or_none()
+    if role is None:
+        if arn is None:
+            raise ApiException(
+                requests.codes.bad_request,
+                "Creating a role requires a role ARN"
+                )
+        if not VALID_NAME_RE.match(role_name):
```
kevinemoore: Minor, but it'd be better to lift this check outside the database transaction.

registry/quilt_server/models.py
```diff
@@ -244,3 +245,8 @@ class Comment(db.Model):
     contents = db.Column(db.String(MAX_COMMENT_LENGTH), nullable=False)
 
     package = db.relationship('Package', back_populates='comments')
+
+class Role(db.Model):
+    id = db.Column(postgresql.UUID, primary_key=True)
+    name = db.Column(db.String(200), unique=True, nullable=False)
+    arn = db.Column(db.String(200), nullable=False)
```
kevinemoore: I think arn should be unique. Otherwise, users can create aliases for the same role. Unless there's a reason to support that?

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,174 @@ def _comment_dict(comment):
...
+@api(enabled=ENABLE_USER_ENDPOINTS, require_admin=True, schema=ROLE_DETAILS_SCHEMA)
+@as_json
+def delete_role(role_id):
+    """
+    Delete the role at role_id and remove it from associated user accounts.
+    """
+    # delete role
+    # must remove role from all users with that role due to foreign key constraint
+    role = Role.query.get(role_id)
+    users = User.query.filter_by(role_id=role.id).all()
```
kevinemoore: Using update will be much more efficient here. Although we don't expect millions of users in a single role, it's faster and cleaner to let the database do the work.

registry/quilt_server/dev_config.py
stevededalus: Is this DEV_ACCESS_KEY or is this an ACCESS key for the service account for the registry?

registry/quilt_server/views.py
stevededalus: Done

registry/quilt_server/views.py
stevededalus: Done

registry/quilt_server/models.py
meffij: I don't see any reason to make a rule that they can't alias -- we could warn when creating an aliasing role but I don't think it's a correctness issue either way and it's simpler to allow them to alias

registry/quilt_server/dev_config.py
stevededalus: Done.

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+
+def _role_dict(role):
+    role_dict = {
+        'id': role.id,
+        'name': role.name,
+        'arn': role.arn
+    }
+    return role_dict
+
+@app.route('/api/roles', methods=['POST'])
```
dimaryaz: Should have a trailing slash - since "roles" is a group, and we're creating a new item

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+
+        role_dict = {
+            'name': role.name,
+            'arn': role.arn
+        }
+    """
+    role = Role.query.get(role_id)
+    return _role_dict(role)
+
+@app.route('/api/roles', methods=['GET'])
```
dimaryaz: Should have a trailing slash here, too

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+@api(enabled=ENABLE_USER_ENDPOINTS, require_admin=True)
+@as_json
+def list_roles():
+    """
+    Lists the roles the registry knows about.
+
+    Returns a JSON object with the top-level key 'results', with a value
+        that is a list of dicts of the form {'name': role_name, 'arn': role_arn}
+    """
+    roles_list = []
```
dimaryaz: Unused?

registry/quilt_server/views.py
dimaryaz: Still not possible?

registry/migrations/versions/93eca89056df_.py
```diff
@@ -0,0 +1,38 @@
...
+down_revision = 'cab5dbaad226'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('role',
+    sa.Column('id', postgresql.UUID(), nullable=False),
+    sa.Column('name', sa.String(length=200), nullable=True),
```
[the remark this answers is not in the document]
stevededalus: what's the command to regenerate it?

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+@app.route('/api/roles/<role_id>', methods=['DELETE'])
+@api(enabled=ENABLE_USER_ENDPOINTS, require_admin=True, schema=ROLE_DETAILS_SCHEMA)
+@as_json
+def delete_role(role_id):
+    """
+    Delete the role at role_id and remove it from associated user accounts.
+    """
+    # delete role
+    # must remove role from all users with that role due to foreign key constraint
+    role = Role.query.get(role_id)
```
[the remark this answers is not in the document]
stevededalus: What's the issue with it not existin? will `db.session.delete(None)` fail?

registry/quilt_server/views.py
stevededalus: removed

registry/quilt_server/views.py
stevededalus: removed

registry/quilt_server/views.py
stevededalus: I don't think that is correct for rest semantics:
https://hackernoon.com/restful-api-designing-guidelines-the-best-practices-60e1d954e7c9
https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+def get_role(role_id):
+    """
+    Gets a role by id.
+
+        role_dict = {
+            'name': role.name,
+            'arn': role.arn
+        }
+    """
+    role = Role.query.get(role_id)
```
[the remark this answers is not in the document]
stevededalus: I don't think that is correct for rest semantics:
https://hackernoon.com/restful-api-designing-guidelines-the-best-practices-60e1d954e7c9
https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+    arn = data.get('arn', None)
+    role_name = data['name']
+    if not VALID_NAME_RE.match(role_name):
+        raise ApiException(
+            requests.codes.bad_request,
+            "Invalid name for role"
+            )
+    role = Role.query.filter_by(name=role_name).one_or_none()
+    if role is None:
+        if arn is None:
```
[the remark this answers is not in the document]
stevededalus: removed

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,171 @@ def _comment_dict(comment):
...
+    There is one required parameter:
+        name(string): name of the role to operate on
+
+    There is an optional paramters:
+        arn(string): ARN of the IAM role associated with the Quilt role.
+
+    To create a role, you must provide an unused name and an arn.
+    """
+    data = request.get_json()
+    arn = data.get('arn', None)
```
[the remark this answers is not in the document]
stevededalus: Done

registry/quilt_server/views.py
stevededalus: I don't think that is correct for rest semantics:
https://hackernoon.com/restful-api-designing-guidelines-the-best-practices-60e1d954e7c9
https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design

registry/quilt_server/views.py
```diff
@@ -2307,6 +2316,178 @@ def _comment_dict(comment):
...
+    """
+    # delete role
+    # must remove role from all users with that role due to foreign key constraint
+    role = Role.query.get(role_id)
+    if not role:
+        raise ApiException(
+            requests.codes.not_found,
+            "Role not found"
+            )
+    User.query.filter(role_id==role_id).update({"role_id": None})
```
dimaryaz: Hm... Does this work? I have a feeling that `role_id==role_id` is the same as `True`. Also, unit test?

`filter_by(role_id=role_id)` might work, though

registry/tests/auth_test.py
```diff
@@ -470,3 +479,370 @@ def testCodeExpires(self):
             data={'refresh_token': code}
         )
         assert token_request.status_code == 401
+
+    def testRoles(self):
+        with mock.patch('quilt_server.views.sts_client') as client:
+            def f(**params):
```
dimaryaz: You can just do:
```
client.assume_role.return_value = {'Credentials': ...}
```

registry/tests/auth_test.py
```diff
@@ -470,3 +479,370 @@ def testCodeExpires(self):
...
+            # simulate a bad role ARN
+            def f(*params, **kwargs):
+                raise ClientError(
+                        error_response={
+                            'Error': {}
+                        },
+                        operation_name='assumeRole'
+                    )
+
+            client.assume_role = f
```
dimaryaz:
```
client.assume_role.side_effect = ClientError(...)
```

registry/quilt_server/views.py
stevededalus: Done, added a test

registry/tests/auth_test.py
stevededalus: done

registry/tests/auth_test.py
stevededalus: done

PR #802, opened 2019-02-05
Dockerfile changes from 797

meffij:


registry/Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 MAINTAINER Quilt Data, Inc. contact@quiltdata.io
 
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-RUN apt-get update -y
-RUN apt-get install -y python3 python3-dev python3-pip build-essential
+RUN apt update -y
```
dimaryaz: Why not `apt-get`? I don't really know what `apt` is...

registry/Dockerfile
```diff
@@ -1,11 +1,12 @@
...
+RUN apt-get install -y python3.6
```
dimaryaz: Why split it into a separate command? This makes it more fragile - if the first line is cached and only the second one runs, you can end up with different versions of python and python-dev.

PR #809, opened 2019-02-15
Fix schemas for new endpoints

meffij:
Some new endpoints are broken due to schemas

Before this change, `delete_role` and `get_role` require JSON that is not necessary or relevant to their function (both should be fine with just the role_id in the URL but the schemas require extra JSON data).

Also, the 'id' property of the schema is redundant now and never used. Removing it.

### DIFF
--- a/registry/quilt_server/schemas.py
+++ b/registry/quilt_server/schemas.py
@@ -241,9 +241,6 @@ ROLE_DETAILS_SCHEMA = {
         },
         'arn': {
             'type': 'string'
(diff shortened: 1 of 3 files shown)


stevededalus:
LGTM, I think the unit tests need to be updated too.

i.e. auth_test.py line 679 passes a body to delete etc. I guess since there's no schema enforcement you can pass it anything and it doesn't fail. But updating the tests will make it clearer for people reading the code.

PR #813, opened 2019-02-21
Allow team-wide push to team packages

kevinemoore:
## Description

Teams users have asked to allow non-owners to push updates to shared packages. This allows a non-owner to push to a package if the package already exists and is shared to the team.

### DIFF
--- a/registry/quilt_server/views.py
+++ b/registry/quilt_server/views.py
@@ -681,8 +681,6 @@ def package_put(owner, package_name, package_hash=None, package_path=None):
     if not VALID_NAME_RE.match(package_name):
         raise ApiException(requests.codes.bad_request, "Invalid package name")
 
-    print("%s: %s/%s" % (g.auth.user, owner, package_name))
-
     # TODO: Description.
     data = json.loads(request.data.decode('utf-8'), object_hook=decode_node)
     dry_run = data.get('dry_run', False)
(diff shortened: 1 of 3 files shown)


kevinemoore:
The fragments are owned by the package owner (i.e. the package creator). True, that is a bit weird and definitely requires trust amongst the team. It would allow a malicious team member to spam the hash space of the package creator, but I can't think of any other issues. Can you?

registry/quilt_server/views.py
```diff
@@ -762,6 +761,18 @@ def package_put(owner, package_name, package_hash=None, package_path=None):
...
     else:
+        team_access = (
+                Access.query
+                .filter(sa.and_(
+                    Access.package == package,
+                    Access.user == TEAM
+                ))
+                .one_or_none()
+            )
+        if not team_access and g.auth.user != owner:
```
dimaryaz: Kind of a corner case, but: if a package has been shared with your team, you can push it. But it's been shared with just you, then you cannot. Should this be a normal access check instead?

registry/quilt_server/views.py
```diff
@@ -1620,6 +1623,24 @@ def tag_put(owner, package_name, package_tag):
...
+        # Check if the package is shared team-wide
+        team_access = (
+                Access.query
+                .filter(sa.and_(
+                    Access.package == instance.package,
+                    Access.user == TEAM
+                ))
+                .one_or_none()
+            )
+        if not team_access:
```
dimaryaz: Same here

kevinemoore: This is admittedly a half-way solution (in place of adding a two-level shared-accesss model), but in most cases (so we hear) Teams packages are usually shared to the whole team--and teams are frustrated that only one user can push--or completely private so this covers the missing common case.

PR #829, opened 2019-05-28
Rename quilt to quilt3

kevinemoore:
## Description
Renaming the module from `quilt` to `quilt3` to avoid confusion with the linux package manager and reduce errors due to API changes.


api/python/setup.py
```diff
@@ -31,10 +31,10 @@ def run(self):
             sys.exit(info)
 
 setup(
-    name="quilt",
+    name="quilt3",
     version=VERSION,
     packages=find_packages(),
-    description='Quilt: where data comes together',
+    description='Quilt3: where data comes together',
```
ResidentMario: Probably fine to just call it "Quilt" still?

PR #835, opened 2019-05-30
Catalog: only show analytics when there's `analyticsBucket` in config

nl0:


dimaryaz:
Just saw @akarve's comment. Completely agree - though if this PR fixes exceptions when the analytics bucket isn't set at all, it's still good to have.

PR #837, opened 2019-05-30
Clean up `quilt config`

dimaryaz:
- Make `quilt config` with no arguments print the current catalog URL
- Make it possible to unset the catalog URL by passing an empty value
- Remove the weird `('http://', 'https:/')` logic, and just give an error if the user didn't enter a proper URL


ResidentMario:
Changes look good! Cleaning up how config works and what it looks like has been a task in the backlog for months and it's good to see some incremental progress towards that goal.

PR #839, opened 2019-05-30
Set the default catalog to None and provide a helpful authentication error

ResidentMario:
This PR changes the default catalog to `None` and adds an error message explaining the need to authenticate using `t4.config` before logging in using `t4.login` (it also fixes a docstring).


ResidentMario:
Hmm, looks like there are a lot of tests that rely on the previous default behavior that I will now need to fix...

PR #843, opened 2019-06-03
Flatten install; remove local magic word; tighten API package code paths

ResidentMario:
This PR implements a number of changes to the API discussed at the board on Monday. My summary from Tuesday sync:

> Dima brought up the fact that he was confused as to how it works, largely on investigation because every method can kind of do everything. Our documentation does a good job of having users follow specific known good code paths, and it's time we were more restrictive and opinionated in our inputs at the top level as well. There are also some simplifying improvements we came up with, and tweaks that can be made to move us away from having the "local" magic string. All in all, I think the raft of changes we came up with are exactly what we need to move to code clarity in our codebase.


ResidentMario:
TODO: a lot.

api/python/quilt3/packages.py
```diff
@@ -333,19 +333,36 @@ def install(cls, name, registry=None, top_hash=None, dest=None, dest_registry=No
...
+                raise QuiltException(
+                    f"Invalid package destination path {dest!r}. 'dest', if set, must point at "
+                    f"the local filesystem. To copy a package to a remote registry use 'push' or "
+                    f"'build' instead."
+                )
 
-        return pkg.push(name=name, dest=dest, registry=dest_registry)
+        pkg = cls.browse(name=name, registry=registry, top_hash=top_hash)
+        dest = fix_url(dest)
+        message = pkg._meta.get('message', None)  # propogate the package message
```
dimaryaz: propagate

api/python/quilt3/packages.py
```diff
@@ -956,45 +974,57 @@ def push(self, name, dest=None, registry=None, message=None):
...
-                                        "registry with quilt.config")
-
-                dest = registry
+            registry = get_from_config('default_remote_registry')
+            if not registry:
+                raise QuiltException("No registry specified and no default remote "
+                                     "registry configured. Please specify a "
+                                     "registry or configure a default remote "
+                                     "registry with quilt3.config")
+            registry_parsed = urlparse(fix_url(registry))
```
dimaryaz: Theoretically, the `default_local_registry` could be set to a `file://` URL, an S3 URL with a path, etc.

I'd move the checks below out of the `else` clause, and run them unconditionally.

api/python/quilt3/packages.py
```diff
@@ -956,45 +974,57 @@ def push(self, name, dest=None, registry=None, message=None):
...
-            if dest is None:
-                # Specifying registry but not a dest doesn't make sense. Throw.
-                raise ValueError("Registry specified but dest is undefined.")
-            else:
-                # If both dest and registry are specified, no further work needed.
-                pass
+            dest_parsed = urlparse(fix_url(dest))
+            if dest_parsed.scheme != registry_parsed.scheme:
+                raise QuiltException(
+                    f"Invalid package destination path {dest!r}. 'dest', if set, must path "
```
dimaryaz: "must path" - typo?

ResidentMario:
> Are we ok breaking the API - flipping the "registry" and "dest" parameters? (I'm guessing, yes - just need to make it clear when releasing the new API.)

My opinion is that we should release this as part of a major version increment (e.g. `3.0.4` -> `3.1.0`) in order to signal to users that there are breaking changes. We also need to broadcast the changes we made on Quilt Users and in conversation with a handful of our mainline users. Jackson is probably impacted by this change, but most users probably are not.

> Should we worry about the existing packages in the "data" directory? I'm guessing no, since the T4->Quilt3 rename already broke everything.

True, also those packages will continue working, that folder just won't be written to in the future.

> Should ".quilt" and the packages live in a dedicated directory that's not a sibling of "auth.json", "config.yml", etc.? It doesn't look like we can have a conflict, but I think it's nice to keep user data separate. (Though adding yet another level of directories is kind of annoying, too.)

I suggest moving these files under a folder in `.quilt` (perhaps `config` or something) in a follow-up PR.

api/python/quilt3/packages.py
ResidentMario: Good eye. But, in my opinion instead of doing this we should disallow setting `default_local_registry` to a non-conformant remote URL in the `quilt3.config` code, and assume that it is valid here. E.g. we should restrict `default_local_registry` to a local folder at set time.

akarve:
> Are we ok breaking the API - flipping the "registry" and "dest" parameters? (I'm guessing, yes - just need to make it clear when releasing the new API.)
yes; release notes and docs should address this

> Should we worry about the existing packages in the "data" directory?
no; release notes and docs should address this

> Should ".quilt" and the packages live in a dedicated directory that's not a sibling of "auth.json", "config.yml", etc.?
i'm cool either way; @ResidentMario you call it

api/python/quilt3/packages.py
dimaryaz: Ok, that's fine with me

dimaryaz:
So I think we really should move the local registry into a dedicated subdir of `BASE_DIR`. Don't really care what we call it - `registry`, `local_registry`, `local`, `data`, etc.

Re-using the same directory might not break anything right now, but it feels dirty:
- We always have to be careful with files we create in `BASE_DIR` and in the registry dir to make sure there's no overlap
- The user might want to move or delete the local registry; difficult to do if it contains unrelated files

ResidentMario:
The `data` folder we were writing to originally did that. We can go back to doing that I guess.f

ResidentMario:
I have isolated package data to a `package` folder that is a subdirectory of the top-level registry returned by `quilt3.util.get_package_registry`. In a follow-up PR I'd like to get rid of `get_package_registry` entirely but this PR is already looking really big so I don't want to do that here. Note that whilst adapting to the `package` folder I also did a couple of code cleanups and removed a non-feature: support for anonymous builds, e.g. `quilt3.Package.build` without a package name, which we never really supported.

PR #850, opened 2019-06-05
Tone down QuiltConfig repr

akarve:
Absent this change, `quilt3.config()` volunteers a confusing amount of information.


ResidentMario:
(this was previously backlogged, discussed this and other issues with config with Brian but there hasn't been much action on this front)

PR #860, opened 2019-06-07
Only allow S3 URLs in the preview endpoint

dimaryaz:
We really shouldn't be making requests to anything other than S3.

### DIFF
--- a/lambdas/preview/index.py
+++ b/lambdas/preview/index.py
@@ -64,10 +64,12 @@ def lambda_handler(request):
     compression = request.args.get('compression')
 
     parsed_url = urlparse(url, allow_fragments=False)
-    if not (parsed_url.scheme == 'https' and parsed_url.netloc.endswith(S3_DOMAIN_SUFFIX) and
-            parsed_url.username is None and parsed_url.password is None):
+    if not (parsed_url.scheme == 'https' and
+            parsed_url.netloc.endswith(S3_DOMAIN_SUFFIX) and
+            parsed_url.username is None and
+            parsed_url.password is None):
         return make_json_response(400, {
-            'title': 'Invalid S3 URL'
+            'title': 'Invalid url=. Expected S3 virtual-host URL.'
         })
 
     try:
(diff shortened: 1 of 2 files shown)


lambdas/preview/index.py
```diff
@@ -23,6 +24,8 @@
 MAX_LINES = 512 # must be positive int
 MIN_VCF_COLS = 8 # per 4.2 spec on header and data lines
 
+S3_DOMAIN_SUFFIX = '.s3.amazonaws.com'
```
akarve: are we sure we can't get path-style requests?

lambdas/preview/index.py
```diff
@@ -59,6 +62,14 @@ def lambda_handler(request):
     url = request.args['url']
     input_type = request.args.get('input')
     compression = request.args.get('compression')
+
+    parsed_url = urlparse(url, allow_fragments=False)
+    if not (parsed_url.scheme == 'https' and parsed_url.netloc.endswith(S3_DOMAIN_SUFFIX) and
```
akarve: one condition per line would read better

lambdas/preview/index.py
```diff
@@ -59,6 +62,14 @@ def lambda_handler(request):
...
+            parsed_url.username is None and parsed_url.password is None):
+        return make_json_response(400, {
+            'title': 'Invalid S3 URL'
```
akarve: Suggestion: `'Invalid url=. Expected S3 virtual-host URL.`

lambdas/preview/index.py
dimaryaz: Pretty sure... I was actually trying to get JS to make path-style requests cause it would make the S3 proxy simpler, but couldn't find a way.

lambdas/preview/index.py
dimaryaz: Just curious, why `url=`? Is that some kind of convention? (That doesn't even match the request syntax...)

akarve: did i read [the code](https://github.com/quiltdata/quilt/blob/56a172ff299406d94cf3dd9b4da59c2d9c5b6850/lambdas/preview/index.py#L62) wrong? i thought the caller does `foo/preview?url=`. anyway we should reiterate the calling syntax in the message so the caller can diagnose.

dimaryaz: Oops, you're right, I was looking at the wrong thing.

PR #873, opened 2019-06-10
Add id to elastic put

stevededalus:
## Description

Use the key + version to explicitly define an id in elastic to allow idempotent updates

Tested here:
https://github.com/quiltdata/deployment/pull/296

## TODO

- [ ] Determine valid id types 
- [x] Test keys with slashes

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -135,7 +135,7 @@ def post_to_es(event_type, size, text, key, meta, version_id=''):
             connection_class=RequestsHttpConnection
         )
 
-        res = es.index(index=ES_INDEX, doc_type='_doc', body=data)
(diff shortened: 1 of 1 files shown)


lambdas/es/indexer/index.py
```diff
@@ -135,7 +135,7 @@ def post_to_es(event_type, size, text, key, meta, version_id=''):
             connection_class=RequestsHttpConnection
         )
 
-        res = es.index(index=ES_INDEX, doc_type='_doc', body=data)
+        res = es.index(index=ES_INDEX, doc_type='_doc', body=data, id=f'{key}_{version_id}')
```
dimaryaz: As far as I can tell, `version_id` will be `None` if the object has no versioning, but `''` if the object is being deleted (versioned or not). That's probably not right?

akarve: I'll make sure we don't have that bug in #939. but i think we can close this PR as #939 handles it.

PR #874, opened 2019-06-10
Implement Google Sign-In

dimaryaz:
Not the prettiest code, but I don't know enough React / Redux / etc. to fix it.

Small bug: if Google Sign-In fails - e.g., because the user doesn't have an account - the form complains about username and password being blank.


nl0:
continued in #888

PR #877, opened 2019-06-14
Support proxy service in preview lambda

akarve:


akarve:
@dimaryaz k but then we need to stop signing URLs right? i should have kept this as a draft PR to get the open stack working. assuming this will be superseded by your changest.

PR #888, opened 2019-06-18
SSO UI

nl0:
This PR introduces SSO UI and slightly alters `config.json` schema to accommodate this.

Removed config key: `disableSignUp`

Added config keys:

`passwordAuth` (optional, defaults to `false`):

* `true`: allows signing in and signing up using username/password pair

* `'SIGN_IN_ONLY'`: allows signing in, but disallows signing up using username/password (similar to removed `disableSignUp` option)

* `false`: disallows signing in and signing up using username/password

`ssoAuth`: same as password auth, but for SSO

`ssoProviders`: array of enabled sso providers (only `"google"` supported for now)

`googleClientId`: string, google client id, required if google sso is enabled

## TODO

- [x] when `passwordAuth === true`, display the password input (required) on the sso signup form
- [x] hide pass reset hints if `passwordAuth === false`
- [x] only show the spinner on sso button when processing quilt request (grey out while waiting for the sso popup)


catalog/app/containers/Auth/messages.js
```diff
@@ -30,6 +30,36 @@ export default defineMessages({
     defaultMessage: 'Authenticating...',
   },
 
+  // SSO
```
nl0: @akarve pls have a look at the wording and see if it's okay

nl0:
description updated

catalog/app/containers/Auth/messages.js
```diff
@@ -30,6 +30,36 @@ export default defineMessages({
...
+  ssoGoogleUse: {
+    id: `${scope}.SSO.Google.use`,
+    defaultMessage: 'Use Google Account',
```
dimaryaz: Should probably be something like "Sign in with Google" -though I'll let @akarve decide.

Also, it appears in uppercase cause of our button styles - but, in my opinion, it doesn't look good on the Google button.

catalog/app/containers/Auth/saga.js
```diff
@@ -71,6 +71,10 @@ function* signUp(credentials) {
       if (e.status === 409 && e.json && e.json.message === 'Email already taken.') {
         throw new errors.EmailTaken({ originalError: e })
       }
+      if (HTTPError.is(e, 401, /domain \S+ is not allowed/i)) {
```
dimaryaz: I could just return an error code from the server, the auth domain, etc.

catalog/app/utils/Config.js
```diff
@@ -18,32 +18,37 @@ const parseJSON = (msg = 'invalid JSON') =>
...
-  mixpanelToken: isNullable(String),
+  sentryDSN: isNullable(R.is(String)),
+  intercomAppId: isNullable(R.is(String)),
+  mixpanelToken: isNullable(R.is(String)),
   apiGatewayEndpoint: R.is(String),
   defaultBucket: R.is(String),
   signInRedirect: R.is(String),
   signOutRedirect: R.is(String),
-  disableSignUp: isNullable(Boolean),
+  passwordAuth: isNullable(oneOf([true, false, 'SIGN_IN_ONLY'])),
```
dimaryaz: Reminds me of [this](https://thedailywtf.com/articles/What_Is_Truth_0x3f_) :)

catalog/app/containers/Auth/SSOGoogle.js
```diff
@@ -0,0 +1,107 @@
...
+      <GoogleLogin
+        clientId={cfg.googleClientId}
+        onSuccess={handleSuccess}
+        onFailure={handleFailure}
+        cookiePolicy="single_host_origin"
+        disabled={!!mutex.current}
+        render={({ onClick, disabled }) => (
+          <M.Button variant="outlined" onClick={handleClick(onClick)} disabled={disabled}>
+            {mutex.current === MUTEX_ID ? (
+              <M.CircularProgress size={18} />
```
dimaryaz: Right now, the spinner shows up as soon as I click the Google button - but I don't think that's right. It's spinning while the Google dialog box is open, even when there's nothing happening.

I we should let Google do its thing, and only show the spinner in `handleSuccess`, while we're sending the request to our own server.

(It's fine to just disable or grey out the button while the popup is open, though.)

catalog/app/containers/Auth/SSOGoogle.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+        mutex.release(MUTEX_ID)
+      }
+    },
+    [dispatch, mutex.release, sentry, notify],
+  )
+
+  const handleFailure = React.useCallback(
+    (code, details) => {
+      if (code !== 'popup_closed_by_user') {
```
dimaryaz: It's `code.error`

catalog/app/containers/Auth/SSOSignUp.js
```diff
@@ -0,0 +1,148 @@
...
+        await result.promise
+      } catch (e) {
+        if (e instanceof errors.UsernameTaken) {
+          throw new SubmissionError({ username: 'taken' })
+        }
+        if (e instanceof errors.InvalidUsername) {
+          throw new SubmissionError({ username: 'invalid' })
+        }
+        if (e instanceof errors.EmailDomainNotAllowed) {
+          throw new SubmissionError({ _error: 'emailDomain' })
```
dimaryaz: We display the error message in the form that's asking the user to enter a username - but, it's too late at that point; no username will make the error go away. @akarve: I'm guessing it's not a big deal, but just wanted to point it out.

catalog/app/containers/Auth/SSOSignUp.js
```diff
@@ -0,0 +1,148 @@
...
+  if (authenticated) {
+    return <Redirect to={next || cfg.signInRedirect} />
+  }
+
+  return (
+    <Container>
+      <Form onSubmit={onSubmit}>
+        {({ submitting, submitFailed, invalid, error }) => (
+          <>
+            <Field
```
dimaryaz: Not required right now, but nice to have in the future: when `passwordAuth === true`, the user needs to enter both a username and a password in order to sign up.

(I.e., if we allow passwords at all, we shouddn't allow the user to sign up without setting a password.)

catalog/app/utils/Config.js
nl0: :) i can make this a little more descriptive like `isNullable(oneOf(['SIGN_UP', 'SIGN_IN_ONLY']))` (with `null` meaning disabled) or maybe `[false, 'SIGN_UP', 'SIGN_IN_ONLY']`

catalog/app/containers/Auth/SSOGoogle.js
nl0: the spinner shows that the ui is busy / locked, but i can change it to be displayed only when "our" request takes place if you think it's better

catalog/app/containers/Auth/SSOGoogle.js
nl0: yep, it should be destructured, i've pushed a fix for this

catalog/app/containers/Auth/SSOSignUp.js
nl0: @dimaryaz you can add this check to the sign-in request handler, so i can catch the error before redirecting to this form

catalog/app/containers/Auth/SSOSignUp.js
nl0: makes sense, i'll add this

catalog/app/containers/Auth/messages.js
nl0: re:wording: when someone uses sso, they may be either signed-in or signed-up, so i preferred more neutral wording (asana uses this btw).
as for button style: it's consistently used throughout the app and i think it looks okayish ([official branding guide](https://developers.google.com/identity/branding-guidelines) says to use regular casing tho)

catalog/app/containers/Auth/saga.js
nl0: yep, you can, not a big deal tho

catalog/app/containers/Auth/SSOSignUp.js
nl0: @dimaryaz i've changed the check to be `!!passwordAuth` instead, bc it makes sense to require password even if password sign-up is disabled (but sign-in is enabled)

nl0:
@dimaryaz i've addressed your comments, take a look when you have a minute pls

catalog/app/containers/Auth/SSOSignUp.js
dimaryaz: I could do that... but, right now, if the user exists, we don't do any more checks. (The admin could've added the user manually, regardless of domain restrictions, or whether signups are allowed at all.)

For now, let's not worry about this.

catalog/app/containers/Auth/messages.js
akarve: OK as Use Google Account

catalog/app/containers/Auth/messages.js
```diff
@@ -30,14 +30,44 @@ export default defineMessages({
...
+  },
+  ssoGoogleError: {
+    id: `${scope}.SSO.Google.error`,
+    defaultMessage: "Couldn't sign in with your Google account",
```
akarve: `Unable to sign in with Google. <way to fix or pobable cause>`. All error messages should be `PROBLEM. SOLUTION.` what could cause this? admin not creating domain for this?

catalog/app/containers/Auth/messages.js
```diff
@@ -30,14 +30,44 @@ export default defineMessages({
...
+  },
+  ssoGoogleNotFound: {
+    id: `${scope}.SSO.Google.notFound`,
+    defaultMessage: "There's no user linked to your Google account",
```
akarve: `No Quilt user linked to this Google account. Notify your Quilt administrator.`

catalog/app/containers/Auth/messages.js
```diff
@@ -30,14 +30,44 @@ export default defineMessages({
...
+  },
+  ssoSignUpHeading: {
+    id: `${scope}.SSO.SignUp.heading`,
+    defaultMessage: 'Complete sign-up',
```
akarve: not sure where this appears and what user is supposed to do? Does this mean they need to finish sign up or...?

nl0: this is a heading for the sso sign-up form (displayed when a person uses sso, but doesn't have quilt account yet), it has username and (when enabled) password prompts

catalog/app/containers/Auth/SSOSignUp.js
nl0: when the user exists, it's not required. it's required when user doesn't exist, like, you can check if the email has valid domain and add some extra message / prop when throwing `'User does not exist'` api exception

catalog/app/containers/Auth/messages.js
nl0: notification renders the details string from google after this message

nl0:
@akarve strings adjusted

catalog/app/utils/Config.js
dimaryaz: @akarve: we won't be able to do JSON validation once we add any of the new parameters here - or at least, I can't think of a way to do it.

The new `config.json`, before CFT substitutes whatever parameters, will need to be something like this:
```
  ...
  "passwordAuth": ${PasswordAuth},
  "ssoAuth": ${SsoAuth},
  "ssoProviders": ${SsoProviders},
   ...
```
But, that's no longer valid JSON. We _could_ switch from booleans and arrays to strings... but that's kind of ugly.

catalog/app/containers/Auth/messages.js
akarve: OK it should be `Complete sign up`

catalog/app/utils/Config.js
akarve: I think we should add json-schema overrides in `/deployment` (and validate after the overrides are applied to the json-schema) so that some keys can be validated as "Sub expressions" at template time. But the catalog should evaluate the final, JSON form of the strong. We probably will have to turn some booleans into strings, though, since "Sub expressions" can't emit JSON. I think that's a good compromise.

PR #892, opened 2019-06-21
Per-bucket search

kevinemoore:
## Description

This PR supports separate indexing for each bucket in the elastic cluster. Bucket::search is now scoped to a single bucket and there is a new top-level API function to search across all buckets indexed by the search cluster.

### DIFF
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -9,9 +9,11 @@ import humanize
 from .data_transfer import copy_file, get_bytes, put_bytes, delete_object, list_objects
 from .formats import FormatRegistry
 from .packages import Package
+from .search_util import search as util_search
 from .util import (QuiltConfig, QuiltException, CONFIG_PATH,
-                   CONFIG_TEMPLATE, fix_url, parse_file_url, parse_s3_url, read_yaml, validate_url,
-                   write_yaml, validate_package_name, get_from_config, get_package_registry)
+                   CONFIG_TEMPLATE, find_bucket_config, fix_url, get_from_config,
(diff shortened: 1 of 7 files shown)


api/python/tests/test_bucket.py
```diff
@@ -314,7 +314,11 @@ def test_search_bucket(self, config_mock, search_mock):
         b.search('blah', limit=1)
 
         config_mock.assert_called_once_with('navigator_url')
-        search_mock.assert_called_once_with('blah', 'https://es-fake.endpoint', limit=1, aws_region='us-meow')
+        search_mock.assert_called_once_with('blah',
+                                            'https://es-fake.endpoint',
+                                            limit=1,
+                                            aws_region='us-meow',
```
stevededalus: maybe drop the supertroopers reference, this is a public repo.

api/python/quilt3/api.py
```diff
@@ -475,3 +477,43 @@ def config(*catalog_url, **config_values):
...
+            "size": <size of object in bytes>,
+            "text": <indexed text of object>,
+            "source": <source document for object (what is actually stored in ElasticSeach)>,
+            "time": <timestamp for operation>,
+        }...]
+        ```
+    """
+    default_bucket = get_from_config('defaultBucket')
+    navigator_url = get_from_config('navigator_url')
+    config_url = navigator_url + '/config.json'
```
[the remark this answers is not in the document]
stevededalus: looks like they're required arguments to find_bucket_config

api/python/tests/test_bucket.py
kevinemoore: Ha! I didn't add that or even notice the Super Troopers connection. :-)

PR #893, opened 2019-06-21
Nicer looking HTML previews for CSV

akarve:
Per user request


lambdas/preview/index.py
```diff
@@ -112,6 +114,30 @@ def lambda_handler(request):
...
+        html - html version of *first sheet only* in workbook
+        info - metadata
+    """
+    # doing this locally because it might be slow
+    import pandas
+    import re
+    # this shouldn't balloon memory because head is limited in size by _from_stream
+    data = pandas.read_csv(io.StringIO('\n'.join(head)))
+    html = data._repr_html_() # pylint: disable=protected-access
+    html = re.sub(r'</table>\n<p>\d+ rows × \d+ columns</p>\n</div>$', '', html)
```
dimaryaz: Wouldn't this leave unclosed `<table>` and `<div>` tags??

Also, I wonder if it's possible to use this thing to customize the HTML: https://pandas.pydata.org/pandas-docs/version/0.23/generated/pandas.io.formats.style.Styler.html

lambdas/preview/index.py
```diff
@@ -112,6 +114,30 @@ def lambda_handler(request):
 
     return make_json_response(200, ret_val)
 
+def extract_csv(head):
+    """
+    csv file => data frame => html
+    Args:
+        file_ - file-like object opened in binary mode, pointing to XLS or XLSX
```
kevinemoore: These comments could use a quick update to include CSV files.

lambdas/preview/index.py
akarve: sorry. fixed with capture groups and added some tests to sanity check HTML.

PR #894, opened 2019-06-21
Preview html files from `defaultBucket` in iframes

nl0:


catalog/app/components/Preview/loaders/Html.js
```diff
@@ -0,0 +1,35 @@
...
+
+import { PreviewData } from '../types'
+
+import * as Text from './Text'
+import * as utils from './utils'
+
+export const detect = utils.extIn(['.htm', '.html'])
+
+const IFrameLoader = ({ handle, children }) => {
+  const sign = AWS.Signer.useS3Signer()
```
akarve: we should note, if it's true, that unlike markdown and vega, we don't sign any URLs internal to the HTML. is that right?

nl0: that's right

catalog/app/components/Preview/renderers/IFrame.js
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+import { styled } from '@material-ui/styles'
+
+const IFrame = styled((props) => (
```
akarve: Please add note on Content-Type `text/html`

nl0: where should i put that?

PR #895, opened 2019-06-22
Improve upload progress bars

dimaryaz:


api/python/tests/test_data_transfer.py
```diff
@@ -394,21 +394,20 @@ def test_multipart_upload(self):
...
-                    },
-                    expected_params={
-                        'Bucket': 'example',
-                        'Key': 'large_file.npy',
-                        'UploadId': '123',
-                        'Body': fd.read(2048),
-                        'PartNumber': part_num
-                    }
-                )
+        for part_num in range(1, 6):
```
akarve: out of curiosity, what's special about `6`?

dimaryaz: Just the file size divided by the chunk size.

(Could be written as an expression instead of hard-coding 6... but then again, it's good to verify it's a reasonable number - not 1 and not 6000.)

PR #903, opened 2019-06-25
Change signer for S3 to https://open-proxy.quiltdata.com

stevededalus:
Should we land and merge this? Does it need to be in `config.json`?


nl0:
catalog changes from this pr are superseded by #904

PR #906, opened 2019-06-27
Search: clip notebook previews

nl0:


akarve:
Actually, have a look at #915. If you pass `exclude_output:true` to the preview endpoint, there will be no output cells. That results in a 10X smaller response in my unit tests. #915 also lets you, for certain file types, set an upper limit on the number of bytes returned.

PR #907, opened 2019-06-27
Preview csv files using preview endpoint

nl0:


akarve:
Something went wrong in this branch and the preview endpoint is getting `type=csv` but it should be `input=csv`.

nl0:
> Something went wrong in this branch and the preview endpoint is getting `type=csv` but it should be `input=csv`.

fixed

PR #910, opened 2019-06-27
only sign es requests for defaultBucket + dont proxy analyticsBucket

nl0:


akarve:
There is one other thing we need to correct: and that is, we should sign _any_ bucket referenced in config.json (i.e. we should also sign analyticsBucket requests). I didn't explain that well before.

nl0:
@akarve the error should be fixed now

nl0:
> There is one other thing we need to correct: and that is, we should sign _any_ bucket referenced in config.json (i.e. we should also sign analyticsBucket requests). I didn't explain that well before.

ok, i'll add a check for `analyticsBucket`

PR #911, opened 2019-06-27
Auth: disable retries when refreshing tokens

nl0:


catalog/app/containers/Auth/messages.js
```diff
@@ -7,10 +7,6 @@ export default defineMessages({
     id: `${scope}.notificationAuthLost`,
     defaultMessage: 'Authentication lost. Sign in again.',
   },
-  notificationAuthError: {
```
akarve: even if we don't retry, don't we still want to show an error?

nl0: the one above will be shown (`'Authentication lost. Sign in again.'`)

PR #913, opened 2019-06-27
Fix analytics auth

dimaryaz:


akarve:
@nl0 I requested this change in #910 but we can take it here instead

PR #926, opened 2019-07-03
remove-ipywidgets-dep

JacksonMaxfield:
After conversations with @akarve and my frustrations with how, even though Quilt3 has `ipywidgets` as a dependency, progress bars aren't being rendered in JupyterLab (which I would say is the now-a-days, standard workbench). Additionally, having it as a dependency, leads to a fair bit of bloat as it causes `jupyter-core`, `ipykernals`, and etc, to be installed when you install `quilt3` which to me shouldn't really be the case imo.

This is what progress bars look like on current master / `quilt3` version `3.0.4` while using JupyterLab:
image

This is what they will look like after this PR:
image

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -15,12 +15,8 @@ import boto3
 from boto3.s3.transfer import TransferConfig
 from s3transfer.utils import OSUtils, signal_transferring, signal_not_transferring
 
-import warnings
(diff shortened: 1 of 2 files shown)


akarve:
this will mercifully decrease `quilt3` package size

JacksonMaxfield:
Bump. @dimaryaz

PR #927, opened 2019-07-05
Validate config.json and federations against the schemas

nl0:
### TODO

- [x] remove schemas from `lambdas/navigator_config` and adjust python code to use `catalog/*-schema.json` @nl0 Done in #941

### DIFF
--- a/catalog/app/utils/Config.js
+++ b/catalog/app/utils/Config.js
@@ -1,3 +1,4 @@
+import Ajv from 'ajv'
 import PT from 'prop-types'
 import * as R from 'ramda'
 import * as React from 'react'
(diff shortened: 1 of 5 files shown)


nl0:
@dimaryaz @akarve should we move federation schema to the catalog as well and use it for validating federations? i'm asking bc i remember talking about ditching federation jsons or smth.

also @dimaryaz could you please adjust navigator_config code to use `catalog/config-schema.json` and remove the original file?

nl0:
i've implemented validating federations

akarve:
@dimaryaz  - this PR settles where json schemas should live. I think we can now eliminate `lambdas/navigator_config/index.py.validate_configs`, update `navigator_config/setup.py` (which references the json schemas), and finally update `deployment/.../env_schema.py` to point to the new locations on GitHub

catalog/app/utils/Config.js
```diff
@@ -93,10 +60,19 @@ const fetchConfig = async (path) => {
   }
 }
 
+const AUTH_MAP = {
+  ENABLED: true,
+  DISABLED: false,
+  SIGN_IN_ONLY: 'SIGN_IN_ONLY',
+}
+
 const transformConfig = (cfg) => ({
```
dimaryaz: Haha, I didn't realize this function existed when I added `fixConfig`.

PR #928, opened 2019-07-05
Fix package listing when there's > 1000 revisions across packages or 'latest' pointer is missing

nl0:


catalog/app/containers/Bucket/PackageList.js
```diff
@@ -103,7 +103,7 @@ export default ({
                         <Link to={urls.bucketPackageDetail(bucket, name)}>{name}</Link>
                       </Typography>
                       <Typography variant="body1">
-                        Updated on {modified.toLocaleString()}
+                        Updated on {modified ? modified.toLocaleString() : '???'}
```
akarve: Can you make this `"[unknown: missing latest in .quilt]"` or something like that?

PR #930, opened 2019-07-05
Bucket/Summary: fix image detection logic (ignore case)

nl0:


catalog/app/containers/Bucket/Summary.js
```diff
@@ -30,7 +30,7 @@ import * as requests from './requests'
 
 const README_RE = /^readme\.md$/i
 const SUMMARIZE_RE = /^quilt_summarize\.json$/i
-const IMAGE_EXTS = ['.jpg', '.jpeg', '.png', '.gif']
+const IMAGE_EXTS = ['.jpg', '.jpeg', '.png', '.gif', '.webp', '.bmp']
```
akarve: Thanks can you add .tif and .tiff?

nl0: @akarve afaik tif isnt supported by browsers (and by our thumbnailing endpoint as well)

akarve: oh yeah, let's take this PR and then we'll modify the image thumbnailer to return JPG or PNG when the input is TIF/F

PR #939, opened 2019-07-08
Elastic performance improvements

akarve:
* Code refactor for readability and maintainability
* Use bulk indexing calls to improve throughput
* [x] modify user agent to screen Elastic requests from object analytics


lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,183 @@
...
-        '.md',
-        '.rmd',
+    "to_index": [
+        ".ipynb",
+        ".md",
+        ".rmd",
     ]
 }
-
+DOC_SIZE_LIMIT_BYTES = 10_000_000# 10MB
```
kevinemoore: Suggestion: add a parameter to flush the queue and send to elastic when the queue hits a certain size. That way the lambda will be able to process large batches without running out of memory.

lambdas/es/indexer/test/data/event_records.py
```diff
@@ -0,0 +1,16 @@
+[{
```
stevededalus: Looks like this isn't used, drop it.

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,220 @@
...
+
+    def send_all(self):
+        """flush self.queue in a bulk call"""
+        if self.is_empty():
+            return
+        elastic_host = os.environ["ES_HOST"]
+        try:
+            awsauth = AWSRequestsAuth(
+                # These environment variables are automatically set by Lambda
+                aws_access_key=os.environ["AWS_ACCESS_KEY_ID"],
```
dimaryaz: I guess this is from the existing code, but would be nice to clean up. You're already creating a `boto3.session.Session()` below, so might as well get the credentials from it - `Session().get_credentials()`

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,220 @@
...
+                aws_service="es"
+            )
+
+            time_remaining = get_time_remaining(self.context)
+
+            elastic = Elasticsearch(
+                hosts=[{"host": elastic_host, "port": 443}],
+                http_auth=awsauth,
+                max_backoff=time_remaining,
+                # Give ES time to repsond when under laod
```
dimaryaz: "respond", "load"

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,220 @@
...
+                    print(error)
+                    # can be dict or string *sigh*
+                    inner = error.get("index", {})
+                    error_info = inner.get("error")
+                    doc_id = inner.get("_id")
+
+                    if isinstance(error_info, dict):
+                        error_type = error_info.get("type", "")
+                        if 'mapper_parsing_exception' in error_type:
+                            replay = next(doc for doc in self.queue if doc["_id"] == doc_id)
```
dimaryaz: This is kinda weird... also O(N^2). A better way would be: iterate over `errors` and collect all `doc_id`s, then iterate over `self.queue` and select the docs to resend.

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,220 @@
...
+                            replay['user_meta'] = replay['system'] = {}
+                            error_queue.append_document(replay)
+                # recursive but never goes more than one level deep
```
dimaryaz: Still, the recursion bothers me a bit. What about moving everything but the error logic into a helper function, and then:
```
_, errors = self._send_all_helper(self.queue)
if errors:
  error_queue = ...
  self._send_all_helper(error_queue)
```
?

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,220 @@
...
+                error_queue.send_all()
+
+        except Exception as ex:# pylint: disable=broad-except
+            print("Fatal, unexpected Exception in send_all", ex)
```
dimaryaz: What happens if you don't catch it?

lambdas/es/indexer/index.py
```diff
@@ -94,189 +240,278 @@ def extract_text(notebook_str):
...
-        'text': text,
-        'key': key,
-        'updated': datetime.utcnow().isoformat(),
-        'version_id': version_id
-    }
-    data = {**data, **transform_meta(meta)}
-    data['meta_text'] = ' '.join([data['meta_text'], key])
+    for cell in formatted.get("cells", []):
+        if "source" in cell and "cell_type" in cell:
+            if cell["cell_type"] == "code" or cell["cell_type"] == "markdown":
```
dimaryaz: You can do it all in a single `if`:
```
if 'source' in cell and cell.get('cell_type') in ('code', 'markdown'):
```

lambdas/es/indexer/index.py
```diff
@@ -94,189 +240,278 @@ def extract_text(notebook_str):
...
 
-def handler(event, _):
-    """fetch the S3 object from event, extract relevant data and metadata,
-    dispatch post_to_es
+    return {
+        "system_meta": helium,
+        "user_meta": user_meta,
+        "comment": comment,
+        "target": target,
+        "meta_text": " ".join(meta_text_parts)
```
dimaryaz: I never understood this logic... How is this better than just `json.dumps(helium)`? Isn't `meta_text_parts` the same thing, just slightly flatter?

lambdas/es/indexer/index.py
```diff
@@ -94,189 +240,278 @@ def extract_text(notebook_str):
...
+        *,
+        etag,
+        version_id
+):
+    """retry head or get operation to S3 with; stop before we run out of time.
+    retry is necessary since, due to eventual consistency, we may not
+    always get the required version of the object.
+    """
+    if operation not in ["get", "head"]:
+        raise ValueError(f"unexpected operation: {operation}")
```
dimaryaz: Just make it:
```
if operation == 'head':
  ...
elif operation == 'get':
  ...
else:
  raise ...
```

lambdas/es/indexer/index.py
```diff
@@ -94,189 +240,278 @@ def extract_text(notebook_str):
...
+    # use a local function so that we can parameterize to time_remaining
+    @retry(
+        # debug
+        stop=(stop_after_delay(time_remaining) | stop_after_attempt(MAX_RETRY)),
+        wait=wait_exponential(multiplier=2, min=4, max=30)
+    )
+    def call():
+        # we can't use Range= if size == 0 because byte 0 doesn't exist
+        # and we'll get an exception
+        if size == 0:
```
dimaryaz: I think a dict of params would be better - then you can handle `size==0` and `version_id` independently, instead of having 4 different cases.

lambdas/es/indexer/index.py
akarve: will improve. tecnnically the queue has a fixed size so it's O(1) but the constant factor is unnecessarily large, i'll give you that.

lambdas/es/indexer/index.py
akarve: i think it has do with the fact that the JSON characters would inhibit plain text indexing of comment and target

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,218 @@
...
+#  lambda in order to display accurate analytics in the Quilt catalog
+#  a custom user agent enables said filtration
+USER_AGENT = "quilt3-python"
 
-S3_CLIENT = boto3.client("s3")
+def bulk_send(elastic, list_):
+    """make a bulk() call to elastic"""
+    return bulk(
+        elastic,
+        iter(list_),
```
dimaryaz: Can you see if it works without `iter`?

The docs say "iterator"... however, that function then calls `streaming_bulk`, for which the docs say "iterable". One of them is clearly wrong.

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,218 @@
...
+                inner = error["index"]
+                info = inner.get("error")
+                if isinstance(info, dict):
+                    if "mapper_parsing_exception" in info.get("type", ""):
+                        doc = id_to_doc[inner["id_"]]
+                        # zero out structured metadata and try again
+                        doc["user_meta"] = doc["system"] = {}
+                        send_again.append(doc)
+        # we won't retry after this (elasticsearch might retry 429s tho)
+        bulk_send(elastic, send_again)
```
dimaryaz: I think this will make a call to elastic even if `send_again` is empty? Probably good to wrap this (and the code above) in `if errors:`.

lambdas/es/indexer/index.py
```diff
@@ -1,74 +1,218 @@
...
+
+    def append_document(self, doc):
+        """append well-formed documents (used for retry or by append())"""
+        if doc["text"]:
+            # document text dominates memory footprint; OK to neglect the
+            # small fixed size for the JSON metadata
+            self.size += min(doc["size"], DOC_LIMIT_BYTES)
+        self.queue.append(doc)
+
+    def is_empty(self):
```
dimaryaz: I'd just get rid of it. `self.queue` is already public...

PR #943, opened 2019-07-09
Allow for `ssoProvider` to be empty

akarve:
(and fix an indent)

### DIFF
--- a/catalog/config-schema.json
+++ b/catalog/config-schema.json
@@ -95,7 +95,7 @@
     },
     "AuthMethodConfig": {
       "type": "string",
-       "description": "must work for actual values AND template substitution",
+      "description": "must work for actual values AND template substitution",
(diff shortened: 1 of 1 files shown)


catalog/config-schema.json
```diff
@@ -64,7 +64,7 @@
     },
     "ssoProviders": {
       "type": "string",
-      "pattern": "^(google)|(\\$\\{\\w+\\})$",
+      "pattern": "^(google)?|(\\$\\{\\w+\\})$",
```
dimaryaz: Actually, the parens are wrong (both in the old and the new version): it's gonna be "^..." or "...$". Needs to be:
`^((google)?|\\$\\{\\w+\\})$`
or just
`^(|google|\\$\\{\\w+\\})$`

PR #944, opened 2019-07-09
Catalog: preview TIFFs

nl0:


nl0:
the ones from the aics bucket are rendered as black boxes tho 🤷‍♂ 
cc @dimaryaz

PR #945, opened 2019-07-09
Tweak renovate config

nl0:
Because of rate limiting (2 PRs / hour AFAIR) and limited schedule (early Mondays), renovate only does ~8 PRs per week, so we're falling behind on updates and bug fixes.

We should probably figure out a better solution, but let's catch up first.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,6 @@
 {
   "extends": [
-    "config:base",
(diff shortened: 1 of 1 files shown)


nl0:
@akarve i'm gonna merge this and catch up on updates in the following days

PR #1028, opened 2019-07-22
Use the S3 Select proxy when not authenticated

dimaryaz:


catalog/app/utils/AWS/S3.js
```diff
@@ -43,10 +43,22 @@ export const useRequest = (extra) => {
...
+  const s3SelectClient = useS3({
+    endpoint: `${cfg.apiGatewayEndpoint}/s3select/`,
+    s3ForcePathStyle: true,
+    ...extra,
+  })
   const authenticated = reduxHook.useMappedState(Auth.selectors.authenticated)
   return React.useMemo(
     () => ({ bucket, operation, params }) => {
-      const client = cfg.shouldProxy(bucket) ? proxyingClient : regularClient
+      let client
```
dimaryaz: (I'm sure there's a more functional way of doing this, but I couldn't figure it out.)

nl0: it's fine, actually

PR #1033, opened 2019-07-24
Raise exceptions on potential S3 and Elastic calls from Indexer lambda

stevededalus:
This will allow SQS/lambda to properly handle message failures.

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -185,7 +185,7 @@ class DocumentQueue:
                     # that, but if there's an error without an id we need to assume it applies to
                     # the batch.
                     send_again = self.queue
-                    print("unhandled indexer error:", error)
+                    print("unhandled indexer error (missing index field):", error)
 
             # we won't retry after this (elasticsearch might retry 429s tho)
             if send_again:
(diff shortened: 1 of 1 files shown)


lambdas/es/indexer/index.py
```diff
@@ -163,25 +163,35 @@ def send_all(self):
...
-                            handled = True
-                            send_again.append(doc)
-                if not handled:
-                    print("unhandled indexer error:", error)
+                        else:
+                            print("unhandled indexer error:", error)
+                    # Always retry, regardless of whether we know to handle and clean the request
+                    # or not. This can catch temporary 403 on index write blocks and other
+                    # transcient issues.
+                    send_again.append(doc)
```
akarve: If index is not in error we won't retry (but should). I _think_ in this context that index is the operation that failed but it could also fail on a delete. There were also some "_index" keys if I recall. Seeing all the errors in the logs will help; or if there's API docs on this (I couldn't find them in the Python wrapper, but maybe the REST API has them).

stevededalus: this will retry, its handled by the else in line 183 right below

PR #1034, opened 2019-07-25
Handle null helium metadata

stevededalus:
Also only log helium errors when the helium header is present. Right now this produced ~55 million unnecessary log events

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -427,7 +427,7 @@ def handler(event, context):
                 if meta and "helium" in meta:
                     try:
                         decoded_helium = json.loads(meta["helium"])
-                        meta["helium"] = decoded_helium if decoded_helium else {}
+                        meta["helium"] = decoded_helium or {}
                     except (KeyError, json.JSONDecodeError):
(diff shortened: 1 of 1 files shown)


lambdas/es/indexer/index.py
```diff
@@ -424,11 +424,12 @@ def handler(event, context):
...
                 # decode Quilt-specific metadata
-                try:
-                    if "helium" in meta:
-                        meta["helium"] = json.loads(meta["helium"])
-                except (KeyError, json.JSONDecodeError):
-                    print("Unable to parse Quilt 'helium' metadata", meta)
+                if meta and "helium" in meta:
+                    try:
+                        decoded_helium = json.loads(meta["helium"])
+                        meta["helium"] = decoded_helium if decoded_helium else {}
```
dimaryaz: `... = decoded_helium or {}`

dimaryaz: Also, might be outside the scope of this PR, but would be nice to clean it up: instead of sticking decoded JSON back into `meta["helium"]`, just pass it as a separate variable. (`transform_meta` doesn't even touch any metadata besided `helium`...)

PR #1039, opened 2019-07-26
Full text search against text, file name

akarve:


akarve:
This is better handled with copy_to and an appropriate analyzer; see #1048 and #1050

PR #1046, opened 2019-07-29
Don't retry on some HTTP client errors

stevededalus:
- Don't retry on 402, 403, and 404 (Payment required, access forbidden, and not found) errors: they're unlikely to be transient.

- Drop retry count significantly: the purpose of the queue is to handle the majority of the retry logic

- Drop timeout based on context: if the lambda execution environment kills a process that's fine

- Reraise exception in tenacity decorators: right now they're raising a retry exception instead of the underlying exception.

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -197,7 +197,7 @@ class DocumentQueue:
         self.size = 0
         self.queue = []
 
-def get_contents(context, bucket, key, ext, *, etag, version_id, s3_client, size):
+def get_contents(bucket, key, ext, *, etag, version_id, s3_client, size):
     """get the byte contents of a file"""
     content = ""
     if ext in CONTENT_INDEX_EXTS:
@@ -206,8 +206,7 @@ def get_contents(context, bucket, key, ext, *, etag, version_id, s3_client, size
                 # we have no choice but to fetch the entire notebook, because we
                 # are going to parse it
                 # warning: huge notebooks could spike memory here
-                get_notebook_cells(
-                    context,
+                get_notebook_cells(                    
                     bucket,
                     key,
                     size,
@@ -219,7 +218,6 @@ def get_contents(context, bucket, key, ext, *, etag, version_id, s3_client, size
             content = trim_to_bytes(content)
         else:
             content = get_plain_text(
-                context,
                 bucket,
                 key,
                 size,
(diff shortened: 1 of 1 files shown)


lambdas/es/indexer/index.py
```diff
@@ -487,13 +487,19 @@ def retry_s3(
     else:
         arguments['IfMatch'] = etag
 
-    time_remaining = get_time_remaining(context)
+    def not_known_exception(exception):
+        error_code = exception.response.get(['Error'], {}).get(['Code'], 218)
+        return error_code not in ["402", "403", "404"]
```
dimaryaz: What error codes do we expect to see and retry? In general, no 4xx codes should be re-tried...

akarve: That seems a bit extreme. e.g. we want to retry 429s (too many requests).

akarve: Seems more correct to enumerate the 400s we do want to retry (it's true most aren't worth trying) there are only a few to retry; see Client Errors https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

lambdas/es/indexer/index.py
```diff
@@ -487,13 +487,19 @@ def retry_s3(
     else:
         arguments['IfMatch'] = etag
 
-    time_remaining = get_time_remaining(context)
```
akarve: should also remove `context` argument to this function

lambdas/es/indexer/index.py
```diff
@@ -34,7 +34,7 @@
 CHUNK_LIMIT_BYTES = 20_000_000
 DOC_LIMIT_BYTES = 2_000
 ELASTIC_TIMEOUT = 30
-MAX_RETRY = 10 # prevent long-running lambdas due to malformed calls
+MAX_RETRY = 4 # prevent long-running lambdas due to malformed calls
```
akarve: You can drop `RETRY_429` as well, for the same reasons. Even `RETRY_429 = MAX_RETRY`.

lambdas/es/indexer/index.py
akarve: I think 408, 409, and 429 are the only ones worth retrying.

lambdas/es/indexer/index.py
stevededalus: Done.

stevededalus:
I'm going to keep this PR smaller in scope, can revisit playing the 429 retries and more restrictive error code lists in the future

PR #1047, opened 2019-07-30
Catalog: package list UI revamp

nl0:
## TODO

- [x] filter
- [x] sort
- [x] pagination
- [x] optimize analytics request (use one query for all the packages)

To be merged only after #857


akarve:
Can we make search work without <enter> or is that bad?

PR #1049, opened 2019-07-31
Index key parts so we can fulltext search for file names

akarve:
* [ ] requires mapping change in `enterprise`


akarve:
to be handled in mappings by copy_to and analyzer

PR #1050, opened 2019-07-31
Explicitly indicate the schema that goes to Elastic

akarve:
* Eliminate confusing dictionary merge with transform_text
* Avoid hack of adding key to meta_text field (better to use multi-field search instead)


lambdas/es/indexer/index.py
```diff
@@ -99,6 +99,7 @@ def append(
             version_id
     ):
         """format event as a document and then queue the document"""
+        more_meta = transform_meta(meta or {})
```
stevededalus: can you rename this to something clearer than `more_meta`? `processed_meta` or something like that? even `obj_meta` makes a little more sense.

lambdas/es/indexer/index.py
```diff
@@ -103,21 +112,23 @@ def append(
...
             # mapper_parsing_exception that below code won't handle
+            "comment": more_meta["comment"],
+            "content": text,# field for full-text search
             "etag": etag,
             "ext": ext,
             "event": event_type,
-            "size": size,
-            "text": text,
             "key": key,
+            "key_text": key_to_parts(key),
```
akarve: Eliminate this line `key_text` is handled by the mappings now

lambdas/es/indexer/index.py
```diff
@@ -68,6 +68,14 @@ def bulk_send(elastic, list_):
         raise_on_exception=False
     )
 
+def key_to_parts(key):
```
akarve: Eliminate this function; handled by mappings and analyzer

PR #1053, opened 2019-08-01
Fix previews for very long lines

dimaryaz:
`iter_lines()` doesn't work so well if a line is very long.
Instead, handle lines manually. Also, make the gzip'ed and non-gzip'ed code paths the same.


lambdas/preview/index.py
```diff
@@ -319,36 +318,44 @@ def extract_txt(head):
...
+        # TODO why are we hitting dec.eof after ~65kb of large gz files?
+        # not >= since we might get lucky and complete a line if we wait
+        if size > max_bytes or line_count > max_lines or (dec and dec.eof):
+            break
+
+    lines = b''.join(buffer).splitlines()
+
+    # If we stopped because of max_bytes, then drop the last, possibly incomplete line -
+    # as long as we have more than one line.
+    if size > max_bytes and len(lines) > 1:
```
akarve: What about the case where there's only one line but we've exceeded max_bytes (at worse by CHUNK but still)?

dimaryaz: Do you mean, we should truncate it to `max_bytes`? (I was just assuming that `max_bytes + CHUNK` is fine)

dimaryaz: (Truncating gets messy if we exceed both lines and bytes, etc.)

akarve: I agree max_bytes + CHUNK is fine actually

akarve: But we should comment on this because cranking up CHUNK has a weird side effect of violating max_bytes more and more.

dimaryaz: So I actually have a fix for `max_bytes + CHUNK` - but guess what, it breaks a couple unittests that depend on the current behavior :-/ I'll fix that later.

PR #1055, opened 2019-08-02
Use aicsimageio to format n-dim images prior to thumbnail creation

JacksonMaxfield:
Allows the handling and formatting of n-dimensional image formats while retaining the current functionality in image rendering.

Changes:
* Supported formats is now most formats supported by [`imageio`](https://imageio.readthedocs.io/en/stable/formats.html). (It _should_ be all of them, but very only tested a few.
* All images now get cast to PNG image format regardless of what the input file format was.
* For n-dimensional imaging formats, the use of [`aicsimageio`](https://github.com/AllenCellModeling/aicsimageio) means that strong assumptions as to the dimension ordering. If the n-dimensional image is not greyscale, RGB, or RGBA encoded, it assumes that the dimensions are `Scene-Timepoint-Channel-SpacialZ-SpacialY-SpacialX` ordered.
* Cleaned up and added more tests a bit by parameterizing a single test instead of having many single file tests.
* Changed tests from comparing bytes to comparing `np.array_equal` because bytes may be different but values are what matter (yay `uint8` / `uint16` / `float64`, etc.)


akarve:
@JacksonMaxfield thanks for moving this forward. a few things:
* [ ] can you down-res `sat_rgb*.tiff` so they are sub 1MB (keeps repo manageable in size)?
* [ ] @dimaryaz do we want to keep `SUPPORTED_BROWSER_FORMATS` and return in the original format, if possible?

lambdas/thumbnail/index.py
```diff
@@ -62,28 +188,24 @@ def lambda_handler(request):
 
     resp = requests.get(url)
     if resp.ok:
-        image_bytes = BytesIO(resp.content)
-        with Image.open(image_bytes) as image:
-            orig_format = image.format
-            orig_size = image.size
+        img = AICSImage(resp.content)
```
akarve: @dimaryaz we're gonna have some huge responses here (300MB and up); not sure what we can do besides cranking RAM

lambdas/thumbnail/index.py
```diff
@@ -47,7 +53,127 @@
     'additionalProperties': False
 }
 
-SUPPORTED_BROWSER_FORMATS = {'JPEG', 'GIF', 'PNG'}
+
+def generate_factor_pairs(n: int) -> List[Tuple[int]]:
```
akarve: Can we move these to a separate file?

JacksonMaxfield: The helper functions? Sure.

akarve: Actually they are fine here for now; it's still a short file

JacksonMaxfield:
@dimaryaz I just removed the full copies of `aicsimageio` and `tifffile` from the lambda thumbnail directory and merged current master branch. Should be good to go for review 👍

lambdas/thumbnail/requirements.txt
```diff
@@ -1,10 +1,14 @@
+#aicsimageio==3.0.0
```
dimaryaz: Uncomment it?

Also, please include all dependencies - i.e., dependencies of `aicsimageio`, etc. - not just the direct ones. `pip freeze` is probably the easiest way - just make sure you don't have `pytest`-related stuff there, and remove the lambda itself.

lambdas/thumbnail/index.py
```diff
@@ -47,7 +54,146 @@
     'additionalProperties': False
 }
 
-SUPPORTED_BROWSER_FORMATS = {'JPEG', 'GIF', 'PNG'}
```
dimaryaz: Need to bring this back: we want to keep the original image format, but only as long as it's well-supported.

(The reasoning is, converting PNG to JPEG can add artifacts; converting JPEG to PNG can make it unnecessarily large, etc.)

lambdas/thumbnail/index.py
```diff
@@ -47,7 +54,146 @@
...
+
+    return img
+
+
+def _format_n_dim_ndarray(img: AICSImage) -> np.ndarray:
+    # Even though the reader was n-dim, check if the actual data is simply greyscale and return
+    if len(img.reader.data.shape) == 2:
+        return img.reader.data
+
+    # Even the the reader was n-dim,
```
dimaryaz: "though"?

lambdas/thumbnail/index.py
```diff
@@ -47,7 +54,146 @@
...
+        # Make rows of images
+        rows = []
+        for y_i in range(min_grid_shape[0]):
+            row = []
+            for x_i in range(min_grid_shape[1]):
+                row.append(projections.pop(0))
+
+            rows.append(row)
+
+        # Concatinate each row then concatinate all rows together into a single 2D image
```
dimaryaz: "Concatenate"

lambdas/thumbnail/index.py
```diff
@@ -47,7 +54,146 @@
...
+        # Get min grid shape
+        # For 6 channels this returns (2, 3)
+        min_grid_shape = choose_min_grid(len(projections))
+
+        # Make rows of images
+        rows = []
+        for y_i in range(min_grid_shape[0]):
+            row = []
+            for x_i in range(min_grid_shape[1]):
+                row.append(projections.pop(0))
```
dimaryaz: `.pop(0)` is O(N), so potentially expensive... Do something like this instead? https://stackoverflow.com/a/312464

lambdas/thumbnail/index.py
```diff
@@ -47,7 +54,146 @@
...
+
+            rows.append(row)
+
+        # Concatinate each row then concatinate all rows together into a single 2D image
+        merged = []
+        for row in rows:
```
dimaryaz: This is mostly a matter of preference... But I'd either move this into the previous loop, or just do a list comprehension:
```
merged = [np.concatenate(...) for row in rows]
```

lambdas/thumbnail/setup.py
```diff
@@ -4,4 +4,5 @@
     name='t4_lambda_thumbnail',
     version='0.0.1',
     py_modules=['index'],
+    packages=['tifffile', 'aicsimageio', 'aicsimageio.readers', 'aicsimageio.writers', 'aicsimageio.vendor']
```
dimaryaz: Shouldn't have this anymore

PR #1065, opened 2019-08-08
Catalog: fix elastic requests

nl0:


catalog/app/containers/Bucket/requests.js
```diff
@@ -377,8 +376,10 @@ const mergeHits = R.pipe(
 
 export const search = async ({ es, query }) => {
   try {
-    const result = await es(query, {
-      _source: 'key,size,type,updated,user_meta,version_id',
+    const result = await es({
+      query,
+      fields: ['content', 'comment', 'key_text', 'meta_text'],
+      _source: ['key', 'version_id', 'updated', 'size', 'user_meta'],
```
akarve: i believe that we can remove this line

catalog/app/utils/AWS/ES.js
```diff
@@ -28,17 +28,24 @@ export const useES = ({ endpoint: ep, bucket }) => {
   const region = React.useMemo(() => getRegion(endpoint), [ep])
 
   const search = React.useCallback(
-    (query, { _source, size = DEFAULT_SEARCH_SIZE }) => {
+    ({ query, fields, _source, size = DEFAULT_SEARCH_SIZE }) => {
```
akarve: same comment here and below; we shouldn't need _source

catalog/app/containers/Bucket/requests.js
nl0: this reduces bandwidth usage significantly by including only specific fields (the ones we need on the front-end) into the response (https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-source-filtering.html)

PR #1067, opened 2019-08-09
Get number of bytes to index from environment

akarve:


lambdas/es/indexer/index.py
```diff
@@ -32,7 +32,7 @@
 ]
 # 10 MB, see https://amzn.to/2xJpngN
 CHUNK_LIMIT_BYTES = 20_000_000
-DOC_LIMIT_BYTES = 2_000
+DOC_LIMIT_BYTES = os.getenv('DOC_LIMIT_BYTES') or 2_000
```
stevededalus: I think you `os.getenv('DOC_LIMIT_BYTES', 2_000)`

akarve: It is safer as written because it is robust to mistakes in the environment such as `DOC_LIMIT_BYTES = ''` or `DOC_LIMITE_BYTES = 0`.

dimaryaz: Oh wait... Won't `getenv` give you a string?

dimaryaz: Oh wait... Won't getenv give you a string?

PR #1068, opened 2019-08-10
Better search across multiple fields

akarve:
See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-multi-match-query.html


catalog/app/containers/Bucket/requests.js
```diff
@@ -379,6 +379,7 @@ export const search = async ({ es, query }) => {
     const result = await es({
       query,
       fields: ['content', 'comment', 'key_text', 'meta_text'],
+      type: 'best_fields',
```
nl0: @akarve docs state that `type` defaults to `best_fields`, so seems like this line does nothing.

also probably we should omit `fields` to use all the fields for querying :

> `multi_match` query defaults to the `index.query.default_field` index settings, which in turn defaults to `*`

akarve:
```suggestion
      type: 'cross_fields',
```

```suggestion
      type: 'best_fields',
```
there are some fields we know we don't want to search, so `fields` is correct.
i'm not satisfied with the results as returned (especially around file names, so before boosting `'key_text^2'` i'm going to try `cross_fields`)

catalog/app/containers/Bucket/requests.js
```diff
@@ -379,6 +379,7 @@ export const search = async ({ es, query }) => {
     const result = await es({
       query,
       fields: ['content', 'comment', 'key_text', 'meta_text'],
+      type: 'cross_fields',
```
nl0: @akarve `es` function doesn't support `type` argument currently. i'll tweak it to do that in a sec

PR #1076, opened 2019-08-13
Dockerize

kevinemoore:
## Description
Updating the Dockerfile for catalog.


catalog/nginx-web.conf
```diff
@@ -25,4 +25,9 @@ server {
 		add_header X-Frame-Options DENY;
 		try_files $uri /index.html =404;
 	}
+
+    location = /healthcheck {
```
dimaryaz: Just use `/` for healthchecking in the CFN template?

catalog/Dockerfile
```diff
@@ -25,7 +26,15 @@ RUN npm install
 
 # Install the app
 COPY . .
-COPY config.json.example static/config.json
-COPY federation.json.example static/federation.json
-RUN npm run build:dll
+RUN npm run build
 CMD npm start
```
dimaryaz: You're running both `npm` and `nginx`??

catalog/validate_config.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+
+import json
+from jsonschema import Draft4Validator
+
+schema = None
+schema_path = "./config-schema.json"
+with open(schema_path, "r") as schema_file:
+    schema = json.loads(schema_file.read())
```
dimaryaz: `schema = json.load(schema_file)`

catalog/validate_config.py
```diff
@@ -0,0 +1,18 @@
...
+
+config_path = "./config.json.tmpl"
+with open(config_path, "r") as config_file:
+    catalog_config = json.loads(config_file.read())
```
dimaryaz: Same here

catalog/nginx-web.conf
kevinemoore: I tried that and it didn't work. I'm not sure exactly why.

catalog/Dockerfile
kevinemoore: Good catch!

catalog/Dockerfile
```diff
@@ -18,14 +19,16 @@ EXPOSE 3000
 RUN mkdir /opt/app
```
dimaryaz: Remove these, too - and all of the install code above besides nginx and gettext.

dimaryaz: Might be a good idea to switch to a newer ubuntu, too - 16.04 is pretty ancient at this point. (Or use the nginx image?)

PR #1078, opened 2019-08-16
Add a proper ES indexer unittest. Fix requirements.

dimaryaz:


lambdas/es/indexer/test/test_index.py
```diff
@@ -0,0 +1,174 @@
...
+            "Records": [{
+                "body": json.dumps({
+                    "Message": json.dumps({
+                        "Event": "s3:TestEvent"
+                    })
+                })
+            }]
+        }
+
+        index.handler(event, None)
```
stevededalus: can you add a comment about what this is doing?

stevededalus: (one sentence doc string for this and test_index)

stevededalus: it's a little strange to see tests with no asserts

PR #1079, opened 2019-08-16
Move the text preview logic into lambas/shared

dimaryaz:
Truncating byte, lines, and ungzipping will be shared by the preview endpoint and the indexer.

TODO: The preview code has a byte limit of 1MB (though it rarely matters, since the line limit is 500) - while the indexer had a 2000 byte limit. Is this going to cause problems?


akarve:
Re: indexer vs preview byte limits, we should just parameterize the S3 calls to take different byte limits from the caller, right? Pending a PR in deployment, the number of bytes that are consumed by the indexer will be dependent on an environment variable, because we want to set aggressively low limits for open indexing, and probably much more generous limits for customer and example buckets.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -0,0 +1,55 @@
...
+def get_preview_lines(chunk_iterator, compression, max_lines, max_bytes):
+    """
+    Read a (possibly compressed) text file, and return up to `max_lines` lines and `max_bytes` bytes.
+    """
+    buffer = []
+    size = 0
+    line_count = 0
+
+    if compression:
+        assert compression == 'gz', 'only gzip compression is supported'
```
akarve: we don't need this because of https://github.com/quiltdata/quilt/pull/1079/files#diff-4ea57d41f3e2cb15d690feb1230f3d8dL63

lambdas/es/indexer/index.py
```diff
@@ -300,11 +302,11 @@ def get_plain_text(bucket, key, size, *, etag, s3_client, version_id):
             size,
             etag=etag,
             s3_client=s3_client,
-            limit=DOC_LIMIT_BYTES,
+            limit=MAX_BYTES,
             version_id=version_id
         )
-        # ignore because limit might break a long character midstream
-        text = obj["Body"].read().decode("utf-8", "ignore")
+        lines = get_preview_lines(obj["Body"], None, MAX_LINES, MAX_BYTES)
```
akarve: shouldn't we also handle the case of when we have, say, `file.json.gz`? i think we need to check for `.gz` and try accordingly.

lambdas/es/indexer/index.py
```diff
@@ -18,6 +18,9 @@
 import nbformat
 from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponential
 
+from t4_lambda_shared.preview import get_preview_lines, MAX_BYTES, MAX_LINES
+
+
 CONTENT_INDEX_EXTS = [
```
akarve: i think we want to make this one shared list between preview and parquet? anything we can show in plaintext on a webpage also belongs in elastic... for example after this refactor `.parquet` should just be searchable in elastic. i'm not sure if that means we index the HTML preview or, probably smarter, something more like the pandas `head` of the file. the thing to do might be to parameterize get_preview_lines to return either HTML or text. But as written I think we are still missing the property that "anything we can preview, we can and will index". we're not far though. i think we want to wrap get_preview_lines with something that also handles the binary cases (like Parquet; or also handles IPYNB in the smart way that we do (only code and markdown cells)).

dimaryaz: The preview endpoint doesn't know file extensions, though - that's done by the catalog. In any case, I'd rather do one thing at a time.

lambdas/es/indexer/index.py
dimaryaz: Yeah - but want to do one thing at a time.

lambdas/shared/t4_lambda_shared/preview.py
dimaryaz: But, it's now used by the indexer, too - that's the whole point of this change.

PR #1081, opened 2019-08-18
Add ~/.quilt/config.yml support

remilapeyre:
This makes it possible for the user to define personal configuration in
$HOME/.quilt/config.yml that won't get overwritten by quilt3.config() and
makes it possible to save only a subset of the configuration options
with those not defined defaulting to their values in CONFIG_PATH or the
config template.

### DIFF
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -10,7 +10,7 @@ from .data_transfer import copy_file, get_bytes, put_bytes, delete_object, list_
 from .formats import FormatRegistry
 from .packages import Package
 from .search_util import search as util_search
-from .util import (QuiltConfig, QuiltException, CONFIG_PATH,
+from .util import (QuiltConfig, QuiltException, CONFIG_PATH, load_config,
                    CONFIG_TEMPLATE, find_bucket_config, fix_url, get_from_config,
                    get_package_registry, parse_file_url, parse_s3_url, read_yaml,
                    validate_url, validate_package_name, write_yaml)
@@ -465,10 +465,7 @@ def config(*catalog_url, **config_values):
         return QuiltConfig(CONFIG_PATH, config_template)
 
     # Use local configuration (or defaults)
-    if CONFIG_PATH.exists():
-        local_config = read_yaml(CONFIG_PATH)
-    else:
-        local_config = read_yaml(CONFIG_TEMPLATE)
+    local_config = load_config()
 
     # Write to config if needed
     if config_values:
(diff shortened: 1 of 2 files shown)


akarve:
@remilapeyre Thanks for submitting. We are reworking some of the config code now. Can you explain your workflow and which variables you would rather not be over-written? Is this to make it easier to swtich between stacks? I _think_ where we want to go is for `.config()` to be a _union_ operation (client keeps adding knowledge of stacks) rather than a singleton (as it is now).

remilapeyre:
Hi, we are currently in the process of testing quilt to see if it fits our needs. Especially we don't need the catalog and we don't plan to deploy it so the only config option we need to set is `default_remote_registry`. It's easier for us to set just this one option rather than having to copy-paste the whole file. This is also more future proof as any addition of a new parameter would require us to update all configuration files.

> I think where we want to go is for .config() to be a union operation (client keeps adding knowledge of stacks) rather than a singleton (as it is now).

This was my thought too and is what the change at https://github.com/quiltdata/quilt/pull/1081/files#diff-efa4288976a32b38e0a87974484f8c7eL306-R312 does.

The last part of this change is to let the user define the configuration in `~/.quilt/config.yml` which seems more usual and convenient than `~/Library/Application Support/Quilt/config.yml` for the user. `~/Library/Application Support/Quilt/config.yml` also has the problem that it will get overwritten by quilt which forbid the user to override some settings if he wants to for whatever reason.

akarve:
@remilapeyre I'm so sorry for the long hiatus. I will make sure that we turn around future PRs faster.

Does `quilt3.config(default_remote_registry=WHATEVER)` not meet your requirements? After discussion, the reason to avoid ~/.quilt is that the recommended way of storing state for all applications is via appdirs so that we use the platform-specific home for user preferences instead of an arbitrary location. If we had a specific, e.g. `.quiltrc` file in `.` that would be different, but no such file at the moment.

remilapeyre:
I think this make sense. I would still change https://github.com/quiltdata/quilt/pull/1081/files#diff-efa4288976a32b38e0a87974484f8c7eL306-L309 to

```python
    local_config = read_yaml(CONFIG_TEMPLATE)
    if CONFIG_PATH.exists():
        local_config.update(read_yaml(CONFIG_PATH))
```

so that the user can override `default_remote_registry` without having to redefine all parameters.

Other than that, I think we should close this PR.

akarve:
Agreed. Would you like to open a clean PR for that or should I backlog as an Issue?

> I think this make sense. I would still change https://github.com/quiltdata/quilt/pull/1081/files#diff-efa4288976a32b38e0a87974484f8c7eL306-L309 to
> 
> ```python
>     local_config = read_yaml(CONFIG_TEMPLATE)
>     if CONFIG_PATH.exists():
>         local_config.update(read_yaml(CONFIG_PATH))
> ```
> 
> so that the user can override `default_remote_registry` without having to redefine all parameters.
> 
> Other than that, I think we should close this PR.

remilapeyre:
I opened a PR at https://github.com/quiltdata/quilt/pull/1521.

PR #1105, opened 2019-09-05
Retry 403s when null, otherwise don't retry 403s

stevededalus:
Not super excited about this, but would like a way to update the lambdas so we can flush the elevation tiles from the RODA queue. It's starting to backup a lot from the `log/` file 403s.

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -397,14 +397,31 @@ def handler(event, context):
 
                 ext = pathlib.PurePosixPath(key).suffix.lower()
 
-                head = retry_s3(
-                    "head",
-                    bucket,
-                    key,
-                    s3_client=s3_client,
-                    version_id=version_id,
-                    etag=etag
-                )
+                try: 
+                    head = retry_s3(
+                        "head",
+                        bucket,
+                        key,
+                        s3_client=s3_client,
+                        version_id=version_id,
+                        etag=etag
+                    )
+                except botocore.exceptions.ClientError as exception: 
+                    # "null" version sometimes results in 403s for buckets
+                    # that have changed versioning, retry without it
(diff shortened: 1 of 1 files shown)


lambdas/es/indexer/index.py
```diff
@@ -48,6 +48,9 @@
 #  a custom user agent enables said filtration
 USER_AGENT_EXTRA = " quilt3-lambdas-es-indexer"
 
+# Key prefixes that can be ignored if 403 responses are returned when accessing
+FORBIDDEN_PREFIXES = tuple(filter(None, os.environ.get('FORBIDDEN_PREFIXES', "").split(" ")))
```
akarve: Prefer list comprehension over `filter`. The `or` is a bit safer than `.get(X, DEFAULT)` because if we set `FORBIDDEN_PREFIXES=""` it will forbid everything... probably dangerous.

```suggestion
FORBIDDEN_PREFIXES = [int(x) for x in (os.environ.get("FORBIDDEN_PREFIXES") or "402 403 404").split(" ")]  
```

lambdas/es/indexer/index.py
```diff
@@ -448,6 +451,16 @@ def handler(event, context):
                     size=size,
                     text=text
                 )
+            except botocore.exceptions.ClientError as boto_exc:# pylint: disable=broad-except
+                if key.startswith(FORBIDDEN_PREFIXES) and \
```
akarve: Any reason not to unify with logic in https://github.com/quiltdata/quilt/blob/74dda6ce5205562c013b42e87e02a7044aad8351/lambdas/es/indexer/index.py#L506?

stevededalus: that makes sense, done

PR #1106, opened 2019-09-05
Skip s3 'directories' in `set_dir`

### LINKED ISSUE #1100: Package.list_object_versions can return keys ending in /, breaking root.set
In packages.py, set_dir, around line 621, setting a directory for a package calls list_object_versions to get all the objects under the specified dir to add.  It is possible to have some of the returned objects end in /.  For example, we have sequencer data that is copied into S3 through a storage gateway.  Calling list_object_versions on the root directory of the S3 contents returns results like:

```
from datetime import date, datetime
from quilt3.packages import list_object_versions
import json

objects, _ = list_object_versions('celsius-sequencing', '190828_NB552139_0023_AHKCYJBGXB/')
def json_serial(obj):
    if isinstance(obj, (datetime, date)):
        return obj.isoformat()
for i in range(3):
    print(f'{json.dumps(objects[i], default=json_serial, indent=2)}')
```
Results in 
```
{
  "ETag": "\"d41d8cd98f00b204e9800998ecf8427e\"",
  "Size": 0,
  "StorageClass": "STANDARD",
  "Key": "190828_NB552139_0023_AHKCYJBGXB/",
  "VersionId": "L7UoB6tk.T5bH8XHzNWx63ZjgG_KvCBW",
  "IsLatest": true,
  "LastModified": "2019-08-28T20:12:19+00:00",
  "Owner": {
    "DisplayName": "aws",
    "ID": "5f378d7af9023313f9eb8f0ea138443d2d7629af0efa4c66572dfdb5360dd5c1"
  }
}
{
  "ETag": "\"d41d8cd98f00b204e9800998ecf8427e\"",
  "Size": 0,
  "StorageClass": "STANDARD",
  "Key": "190828_NB552139_0023_AHKCYJBGXB/Config/",
  "VersionId": "7TLUsbBstpl.ei8TeSbe1GZY_mudfTXI",
  "IsLatest": true,
  "LastModified": "2019-08-28T20:12:35+00:00",
  "Owner": {
    "DisplayName": "aws",
    "ID": "5f378d7af9023313f9eb8f0ea138443d2d7629af0efa4c66572dfdb5360dd5c1"
  }
}
{
  "ETag": "\"8048e95a2c72097c274ccbdce9115ebb\"",
  "Size": 264379,
  "StorageClass": "STANDARD",
  "Key": "190828_NB552139_0023_AHKCYJBGXB/Config/Effective.cfg",
  "VersionId": "3vuUg9PqKDWlCyWfk8jbKyEqw2iF4Qk5",
  "IsLatest": true,
  "LastModified": "2019-08-28T20:13:07+00:00",
  "Owner": {
    "DisplayName": "aws",
    "ID": "5f378d7af9023313f9eb8f0ea138443d2d7629af0efa4c66572dfdb5360dd5c1"
  }
}
```

when root.set is called with the second item, it raises an exception around:
```
        if not logical_key or logical_key.endswith('/'):
            raise QuiltException(
                f"Invalid logical key {logical_key!r}. "
                f"A package entry logical key cannot be a directory."
            )
```

We need to be able to add these files.  Currently, I have a patch in place that just ignores the QuiltException.  Obviously not ideal.

kevinemoore:
## Description
InPackage::set_dir, skip empty files in S3 that have keys that end in /. Resolves #1100.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -623,9 +623,10 @@ class Package(object):
             for obj in objects:
                 if not obj['IsLatest']:
                     continue
-                # Skip S3 pseduo directory files
-                if obj['Key'].endswith('/') and obj['Size'] == 0:
-                    warnings.warn(f'Logical keys cannot end in "/", skipping: {obj["Key"]}')
+                # Skip S3 pseduo directory files and Keys that end in /
+                if obj['Key'].endswith('/'):
+                    if obj['Size'] != 0:
+                        warnings.warn(f'Logical keys cannot end in "/", skipping: {obj["Key"]}')
                     continue
                 obj_url = make_s3_url(src_bucket, obj['Key'], obj.get('VersionId'))
                 entry = PackageEntry([obj_url], obj['Size'], None, None)
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1,3 +1,4 @@
+from collections import deque
 import copy
 import hashlib
 import io
@@ -6,7 +7,7 @@ import pathlib
 import os
 import time
 from urllib.parse import quote, urlparse, unquote
-from collections import deque
+import warnings
 
 import jsonlines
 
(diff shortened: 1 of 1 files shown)


api/python/quilt3/packages.py
```diff
@@ -622,6 +623,10 @@ def set_dir(self, lkey, path=None, meta=None):
             for obj in objects:
                 if not obj['IsLatest']:
                     continue
+                # Skip S3 pseduo directory files
+                if obj['Key'].endswith('/') and obj['Size'] == 0:
+                    warnings.warn(f'Logical keys cannot end in "/", skipping: {obj["Key"]}')
```
dimaryaz: When does this actually happen? If regular nested directories in s3 show up here, we should skip them silently; a warning doesn't really help the user.

PR #1111, opened 2019-09-06
Split document_queue into new module for importing elsewhere

stevededalus:


lambdas/es/indexer/index.py
```diff
@@ -4,44 +4,22 @@
...
-from aws_requests_auth.aws_auth import AWSRequestsAuth
 import boto3
 import botocore
-from elasticsearch import Elasticsearch, RequestsHttpConnection
-from elasticsearch.helpers import bulk
 import nbformat
 from tenacity import retry, retry_if_exception, stop_after_attempt, wait_exponential
 
 from t4_lambda_shared.preview import get_bytes, get_preview_lines, MAX_LINES
+from .document_queue import (DocumentQueue, CHUNK_LIMIT_BYTES, CONTENT_INDEX_EXTS, DOC_LIMIT_BYTES,
```
dimaryaz: Are all of these constants still used? I think they're mostly internal to `document_queue.py` now?

stevededalus: good catch, cleaned them up

PR #1121, opened 2019-09-11
Global search

nl0:
Global search available under `/search` url.
Requires top-level `"globalSearch": true` config prop.

## TODO

- [x] polish the ui, make it mobile-friendly
- [x] querystring-based pagination
- [x] handle null versions
- [x] make breadcrumbs copyable
- [x] make sure the old bucket search layout isnt broken
- [x] global search box in navbar when outside of bucket
- [x] test with the new search endpoint


nl0:
@akarve should be good to go

nl0:
@akarve fyi i've merged `master` here with Dima's search endpoint changes, so the api gateway requires the search endpoint to be compatible with these catalog changes

PR #1125, opened 2019-09-13
Use the new search endpoint.

dimaryaz:
TODO: Completely remove searchEndpoint from bucket config.


nl0:
so, is it just a proxy-like thing with the same interface as vanilla es?

PR #1129, opened 2019-09-13
String updates

akarve:


catalog/app/website/pages/About/About.js
```diff
@@ -85,10 +85,19 @@ const Team = (props) => (
     <M.Container maxWidth="lg" style={{ position: 'relative' }}>
       <Heading>Team</Heading>
       <M.Box maxWidth={480} mt={3} mx="auto">
-        <M.Typography variant="body1" color="textSecondary" align="center">
-          Quilt was founded by Kevin Moore and Aneesh Karve. They have been fast friends
-          ever since they met in 2005 as graduate students in Computer Science at
-          UW-Madison.
+        <M.Typography variant="body1" color="textSecondary">
```
nl0: @akarve `Typography` usually renders as `p`, so it's better to use two subsequent `Typography` components instaed of nesting `p`s

akarve: OK

PR #1149, opened 2019-09-16
Bump request size to 99MB for larger containers

akarve:
Increased document throughput for content-heavy indexing


lambdas/es/indexer/document_queue.py
```diff
@@ -20,11 +20,11 @@
     ".parquet",
     ".rmd",
     ".tsv",
-    ".txt"
+    ".txt
```
stevededalus: dropped a quote somehow

PR #1155, opened 2019-09-18
Continue indexing if content extraction fails

akarve:
* [x] ~Make `content` nullable or add a new field to schema so that we can determine via simple queries which files failed content extraction~ For now we can query for files with eligible extensions, non-zero size, and `content==""`.


lambdas/es/indexer/index.py
```diff
@@ -53,43 +53,47 @@ def get_contents(bucket, key, ext, *, etag, version_id, s3_client, size):
...
-                key,
-                size,
-                compression,
-                etag=etag,
-                s3_client=s3_client,
-                version_id=version_id
-            )
+                )
+        # It's OK to catch "everything here"
+        except Exception as exc:#pylint: disable=broad-except
```
stevededalus: No, this isn't correct.

What you need to do is catch an exception and then load an that object into elastic based on HEAD with an empty content. Then raise the exception up after processing. We need to see the failures in the DLQ and dashboard for monitoring and debugging. We also need to load something into elastic for stats. This just silently eats the error AND fails to put it into elastic.

akarve: To clarify, as written, the document will still end up in Elastic because that happens after https://github.com/quiltdata/quilt/blob/b270147f9c271fcbed0b7eae16a133f24a3e419b/lambdas/es/indexer/index.py#L269, which should always succeed.

I can re-raise but don't understand the implications yet. If I raise, the whole batch fails, and all get retried until they end up in the DLQ? Mostly I'm wondering how some good docs will ever reach ES if we fail the entire batch, especially as the queue comes to be dominated by failing jobs, towards the end.

stevededalus: As we talked about f2f, if you don't reraise you lose all the retry and metrics/observability that you get from the SQS queue.

My recommendation here is to let the exception pass through. On line 269 `text = get_contents( `, you should catch the exception there, it makes the behavior of falling back to `head` level info explicit since its tied to the `head` and `insert` calls in the function.

Then on catch, if you want you can store the exception in a variable `recordException`. Add the empty doc as normal and continue through the loop. Outside the loop, if the variable is set, the you can raise it. This will let you process all documents in the batch without sacrificing the control flow and observability of the system.

```
batch_processor.send_all()
raise recordException
```

akarve: Fixed.

lambdas/es/indexer/index.py
```diff
@@ -262,15 +264,20 @@ def handler(event, context):
...
-                    bucket,
-                    key,
-                    ext,
-                    etag=etag,
-                    version_id=version_id,
-                    s3_client=s3_client,
-                    size=size
-                )
+                try:
+                    text = get_contents(
```
kevinemoore: Looks to me that if `get_contents` throws an exception (which will be caught), text will keep is previous value and we'll end up adding it twice. I think you want to `continue` here or maybe cleaner, move more of the logic into the try block.

akarve: Per discussion: we won't add the document twice. We actually want files for which `get_contents()` fails to have an entry in ES. At the same time, the same files will end up in the DLQ when we re-raise at the very end of `handler()`.

I moved `text = ""` to the `except` clause so that it's easier to read.

PR #1156, opened 2019-09-18
New search API

dimaryaz:
Search and stats queries are now hard-coded in the search lambda.

But, `bucketSummary` needs to be updated to use `overview_url` + `'summary.json'`

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -20,6 +20,40 @@ def lambda_handler(request):
     """
     Proxy the request to the elastic search.
     """
+    action = request.args.get('action')
+    index = request.args.get('index')
+
+    if action == 'search':
+        query = request.args.get('query', '')
+        body = {
+            "query": {
+                "simple_query_string" : {
+                    "query": query,
+                    "fields": ['content', 'comment', 'key_text', 'meta_text'],
+                    "default_operator": "or"
+                }
+            }
+        }
+        # TODO: should be user settable; we should proably forbid `content` (can be huge)
+        _source = ['key', 'version_id', 'updated', 'last_modified', 'size', 'user_meta']
+        size = 1000
+    elif action == 'stats':
+        body = {
+            "query": { "match_all": {} },
+            "aggs": {
+                "totalBytes": { "sum": { "field": 'size' } },
+                "exts": {
+                    "terms": { "field": 'ext' },
+                    "aggs": { "size": { "sum": { "field": 'size' } } },
+                },
+                "updated": { "max": { "field": 'updated' } },
+            }
+        }
+        size = 0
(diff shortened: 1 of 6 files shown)


lambdas/search/index.py
```diff
@@ -20,6 +20,40 @@ def lambda_handler(request):
...
+        body = {
+            "query": {
+                "simple_query_string" : {
+                    "query": query,
+                    "fields": ['content', 'comment', 'key_text', 'meta_text'],
+                    "default_operator": "or"
+                }
+            }
+        }
+        # TODO: should be user settable; we should proably forbid `content` (can be huge)
```
kevinemoore: Are these the fields that Elastic will search to find matches? Or, the fields that are returned in the result? Either way, don't we want the document contents?

dimaryaz: Fields that get returned, I think... This was a comment from @akarve's example

lambdas/search/index.py
```diff
@@ -20,6 +20,40 @@ def lambda_handler(request):
...
+    index = request.args.get('index')
+
+    if action == 'search':
+        query = request.args.get('query', '')
+        body = {
+            "query": {
+                "simple_query_string" : {
+                    "query": query,
+                    "fields": ['content', 'comment', 'key_text', 'meta_text'],
+                    "default_operator": "or"
```
akarve: eliminate - it's the default

lambdas/search/index.py
akarve: Yes, this prevents Elastic from sending back unnecessary huge stuff, e.g. `content`

lambdas/search/index.py
```diff
@@ -37,11 +71,6 @@ def lambda_handler(request):
         connection_class=RequestsHttpConnection
     )
 
-    index = request.pathParameters['proxy']
-    body = request.args.get('source')
-    _source = request.args.get('_source')
-    size = request.args.get('size', '1000')
-
     result = es_client.search(index, body, _source=_source, size=size, timeout=MAX_QUERY_DURATION)
```
akarve: Oh, for this we should probably have a global override-able by the environment so that we can skip indexes that are under load or incomplete. So:
```
INDEX_OVERRIDES = os.environ.get("INDEX_OVERRIDES") or "-elevation-tiles-prod*"
...
indexes = index + "," + INDEX_OVERRIDES
```
Keep in mind that "index" can be comma-separated list; ditto for INDEX_OVERRIDES.

akarve: Note the `-` in front of `the index name.

dimaryaz: Ok; I'll do that in a separate PR

PR #1169, opened 2019-09-20
Patch urllib3 for security

akarve:


akarve:
@stevededalus I ended up doing this in a dedicated environment - it gets hairy otherwise. should be OK since lambdas are all self-contained environments.

PR #1197, opened 2019-09-24
Make Athena CloudTrail queries more efficient

dimaryaz:
Only scan the last 7 days of CloudTrail logs, and keep a persistent table of processed data in S3.

### DIFF
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -3,7 +3,7 @@ Lambda function that runs Athena queries over CloudTrail logs and .quilt/named_p
 and creates summaries of object and package access events.
 """
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 import os
 import textwrap
 import time
@@ -252,6 +252,7 @@ def run_multiple_queries(query_list):
     results = [None] * len(query_list)
 
     remaining_queries = list(enumerate(query_list))
+    remaining_queries.reverse()  # Just to make unit tests more sane: we use pop() later, so keep the order the same.
     pending_execution_ids = set()
 
     while remaining_queries or pending_execution_ids:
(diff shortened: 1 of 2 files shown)


lambdas/access_counts/index.py
```diff
@@ -190,20 +267,65 @@ def delete_temp_dir():
         errors = delete_response.get('Errors')
         if errors:
             print(errors)
-            raise Exception("Failed to delete the temporary directory")
+            raise Exception(f"Failed to delete dir: bucket={bucket!r}, prefix={prefix!r}")
 
 
 def handler(event, context):
```
akarve: nice-to-have: unit test for this function; even one will do

lambdas/access_counts/index.py
```diff
@@ -190,20 +267,65 @@ def delete_temp_dir():
...
         wait_for_query(query_id)
 
-    for query_id in [run_query(CREATE_OBJECT_ACCESS_LOG), run_query(CREATE_PACKAGE_HASHES)]:
+    # Create new Athena tables.
+    for query_id in [run_query(CREATE_CLOUDTRAIL), run_query(CREATE_OBJECT_ACCESS_LOG), run_query(CREATE_PACKAGE_HASHES)]:
         wait_for_query(query_id)
 
+    # Create CloudTrail partitions, after the CloudTrail table is created.
+    for query_id in [run_query(q) for q in partition_queries]:
+        wait_for_query(query_id)
```
akarve: Because these queries are issued sequentially (right?) we shouldn't hit the service limits of a max of 20 DDL simultaneous DDL queries? In any case do these rate limits deserves a note somewhere https://docs.aws.amazon.com/athena/latest/ug/service-limits.html?

lambdas/access_counts/index.py
```diff
@@ -18,33 +21,107 @@
 ACCESS_COUNTS_OUTPUT_DIR = os.environ['ACCESS_COUNTS_OUTPUT_DIR']
 
 
+LAST_UPDATE_KEY = 'ObjectAccessLog.last_updated_ts'
```
akarve: We could hit consistency issues here since we reuse the key name each rune... I guess the one hour cron spacing prevents that.

lambdas/access_counts/index.py
```diff
@@ -190,20 +267,65 @@ def delete_temp_dir():
...
+    # Create CloudTrail partitions, after the CloudTrail table is created.
+    for query_id in [run_query(q) for q in partition_queries]:
+        wait_for_query(query_id)
+
+    # Scan CloudTrail and insert new data into "object_access_log".
+    insert_query = INSERT_INTO_OBJECT_ACCESS_LOG.format(start_ts=start_ts.timestamp(), end_ts=end_ts.timestamp())
+    wait_for_query(run_query(insert_query))
+
+    # Save the end timestamp.
+    s3.put_object(Bucket=QUERY_RESULT_BUCKET, Key=LAST_UPDATE_KEY, Body=str(end_ts.timestamp()), ContentType='text/plain')
```
akarve: Just to be clear: we don't have object versioning on in `QUERY_RESULT_BUCKET`? And we don't expect to hit any consistency issues?

lambdas/access_counts/index.py
dimaryaz: Actually, they're issued in parallel (hence the weird `for` loop with a list comprehension, rather than `wait_for_query(run_query(q))`). I tested it on 365 days, where it would've run 365 queries in parallel, but nothing failed... But you're right, we shouldn't do that.

lambdas/access_counts/index.py
dimaryaz: No versioning. Shouldn't hit any issues as long as there are no overlapping lambdas. It runs once an hour, so should be safe?

lambdas/access_counts/index.py
akarve: Any reason not to use .txt as the extension? That way it will show in catalog, people can open it, etc

akarve: Any reason not to use .txt as the extension? That way it will show in catalog, people can open it, etc.

lambdas/access_counts/index.py
akarve: Since we use Glue now (true?) I think [these are the right docs](https://docs.aws.amazon.com/general/latest/gr/aws_service_limits.html#limits_glue). Same numbers

My reading on "Number of DDL queries you can submit at the same time" is not how many queries are running, but how many you can submit in a single request. I could be totally wrong. Maybe the safe thing to do is: only set 20 partitions in per `run_query` and space out the calls.

lambdas/access_counts/tests/test_access_counts.py
```diff
@@ -0,0 +1,197 @@
...
+                'CommonPrefixes': [{
+                    'Prefix': 'AWSLogs/123456/CloudTrail/ng-north-1/'
+                }]
+            }
+        )
+
+        for idx, table in enumerate(['cloudtrail', 'object_access_log', 'package_hashes']):
+            self._start_query(f'DROP TABLE IF EXISTS {table}', str(idx))
+
+        for _ in range(3):
```
akarve: Less fragile if len(some_list), here and elsewhere

PR #1201, opened 2019-09-27
See if pyarrow 0.14.1 works in CI

akarve:


akarve:
Optimistically leaving the dependency open on top. CI will start failing if 0.15 is bad.

PR #1221, opened 2019-09-30
package.set('KEY', object)

armandmcqueen:
## Description

Add an in-memory object to a package


armandmcqueen:
There is an alternative approach that serializes during `package.set()`. This make `set` a non-instant operation, but it guarantees that data will not be lost between `package.set()` and `package.build()` if say the Jupyter notebook kernal is restarted

armandmcqueen:
> There is an alternative approach that serializes during `package.set()`. This make `set` a non-instant operation, but it guarantees that data will not be lost between `package.set()` and `package.build()` if say the Jupyter notebook kernel is restarted

Closing this option following f2f discussion of what packages are.

Conceptually, a package is an immutable datastructure. It should have a complete manifest, a hash for each entry and a top-hash for the package. 

`package.set()` creates a new, intermediate datastructure that is different than a package. This intermediate datastructure is mutable, it has incomplete entries (objects still in memory, no hash or size computed, etc) and it may contain information that the end user doesn't want in the final package (performance optimization related information such as what has changed, etc). This intermediate datastructure is converted to an immutable package during the `build` step. While the code does not currently do a good job of distinguishing between these two datastructures, that is a refactor we should do in the future.

api/python/quilt3/packages.py
```diff
@@ -782,6 +919,10 @@ def build(self, name=None, registry=None, message=None):
         Returns:
             The top hash as a string.
         """
+        for _, entry in self.walk():
```
armandmcqueen: After discussion about what a package is, I am changing this so that we support serialization on build()

api/python/quilt3/packages.py
```diff
@@ -769,7 +979,7 @@ def _set_commit_message(self, msg):
 
         self._meta.update({'message': msg})
 
-    def build(self, name=None, registry=None, message=None):
+    def build(self, name=None, registry=None, message=None, serialization_dir=None):
```
kevinemoore: Here's another example where complexity creeps in. When objects are passed in (e.g., set, users will be ready to specify a path for serialization, but not during build).

armandmcqueen:
Updated PR following discussion with @kevinemoore 

Serialization now happens at `package.set()` time. This has substantial benefits in terms of code simplicity and consistent behavior from the `Package` object, since we don't need to differentiate between `Packages` that have `LazySerializedPackageEntries` and `Packages` that only have `PackageEntries`. 

Given primary use below, performance will be unchanged compared to lazy serialization implementation:
```
pkg.set('KEY', df, serialization_location="./df.parquet")
new_pkg = pkg.push("quilt/test")
```
This is also valid: 
```
pkg.set('KEY', df) # Will write to a location determined by tempfile
new_pkg = pkg.push("quilt/test")
```
Important to note that in this model, the user is responsible for deleting intermediate files. This could probably be improved by searching through local packages to see which physical keys are orphaned, but this user experience is okay as long as we provide a convenience method that can delete a file:// URL:
```
Package.delete_local_file(pkg.get('KEY'))
```

Requesting additional review from @kevinemoore and @dimaryaz

api/python/quilt3/packages.py
```diff
@@ -880,10 +914,49 @@ def set(self, logical_key, entry=None, meta=None):
             entry = PackageEntry([url], size, None, orig_meta)
         elif isinstance(entry, PackageEntry):
             entry = entry._clone()
+
+        elif object_is_serializable(entry):
+            # Use file extension from serialization_location, fall back to file extension from logical_key
```
kevinemoore: Using the file extension here seems a bit fragile, especially since there are potentially two (for both the logical and physical keys). Since this is really a convenience mechanism, I'm not sure how important it is to support multiple serialization methods. Users can manually serialize then add the file if they want to control serialization.

kevinemoore: We haven't previously required that logical keys have extensions. If we make that change here, we should be sure to do it intentionally and document it.

api/python/quilt3/packages.py
```diff
@@ -880,10 +914,49 @@ def set(self, logical_key, entry=None, meta=None):
...
+                serialization_path.parent.mkdir(exist_ok=True, parents=True)
+            else:
+                suffix = f'.{ext}' if ext else None
+                tmpfile = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
+                serialization_path = tmpfile.name
+
+            write_url = fix_url(f'file://{serialization_path}')
+            put_bytes(serialized_object_bytes, write_url, meta=new_meta)
+            size, _, _ = get_size_and_meta(write_url)
+            entry = PackageEntry([write_url], size, hash_obj=None, meta=new_meta)
```
kevinemoore: Why not hash the file?

api/python/quilt3/packages.py
```diff
@@ -880,10 +914,49 @@ def set(self, logical_key, entry=None, meta=None):
...
+            if serialization_format_opts is None:
+                serialization_format_opts = {}
+            serialized_object_bytes, new_meta = format_handlers[0].serialize(entry, meta=None, ext=ext,
+                                                                             **serialization_format_opts)
+            if serialization_location:
+                serialization_path = pathlib.Path(serialization_location).expanduser().absolute()
+                serialization_path.parent.mkdir(exist_ok=True, parents=True)
+            else:
+                suffix = f'.{ext}' if ext else None
+                tmpfile = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
```
kevinemoore: Suggestion (and feel free to implement in a separate PR):
Serialized object files should go somewhere in the local package registry (i.e. under APP_DIR). That way all the quilt3 state will be in one place and a lot easier to track down. Also, temp files can get deleted even if the user saves the package manifest.

api/python/tests/integration/test_packages.py
```diff
@@ -542,6 +560,52 @@ def test_set_package_entry(self):
...
+        Package.delete_local_file(pkg.get("mydataframe1.parquet"))
+        Package.delete_local_file(pkg.get("mydataframe2.csv"))
+        Package.delete_local_file(pkg.get("mydataframe3.tsv"))
+        Package.delete_local_file(pkg.get("mydataframe4.parquet"))
+        Package.delete_local_file(pkg.get("mydataframe5.csv"))
+        Package.delete_local_file(pkg.get("mydataframe6.tsv"))
+
+        for lk, entry in pkg.walk():
+            file_path = parse_file_url(urlparse(entry.physical_keys[0]))
+            assert not (pathlib.Path(file_path)).exists(), "The serialization files should have been deleted exist"
```
armandmcqueen: Assertion typo

api/python/quilt3/packages.py
armandmcqueen: The key thing I want to accomplish with this piece of code is preventing `pkg.set('mydataframe.csv', df)` from doing something stupid like serializing the dataframe as parquet. We aren't requiring the user to give us a file extension, we are just respecting it if is is specified, since it looks like it is not uncommon for logical keys to have extensions.

I added an exception if the user gives conflicting file extensions.

api/python/quilt3/packages.py
armandmcqueen: It looks like we leverage parallelism to calculate hashes quickly - https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/data_transfer.py#L703.

If we hash on set, we are removing that opportunity for parallelism.

api/python/quilt3/packages.py
armandmcqueen: > Serialized object files should go somewhere in the local package registry (i.e. under APP_DIR)

Are you thinking in the case that the user doesn't specify a serialization location or in general?

> Also, temp files can get deleted even if the user saves the package manifest.

It's always possible that someone removes the underlying files on disk so won't this be true no matter where we write to?

kevinemoore: TemporaryFiles are deleted automatically (at least by default I think). Storing serialization files in the local Quilt registry seems a lot safer to me.

akarve: In this case the danger of bloating local disk is worse. I think TempFiles are appropriate.

api/python/quilt3/packages.py
akarve: Yes and the set contract is "return right away"

kevinemoore: Hashing memory-size files isn't a noticeable delay. We could even calculate the hash as we write the bytes to disk, but that's probably over-optimizing.

api/python/quilt3/packages.py
akarve: As written in this PR (selecting the serializer from the file extension) is correct and true to the user's intention. We are actually moving away from a worse problem that we had before: we would _always_ select parquet for dataframes, irrespective of the user's choice of file extension. Proposed logic is strictly better.

akarve:
We should be aware that we have started to break the intention of the `set()` method which was to be fast, and defer work to build or push. The reason for this is to predictably defer slow work to an isolated step in the package lifecycle. As a rule, compute and I/O (outside of say, file or object existence checks) belongs in build or push.

Large dataframes can take multiple seconds to serialize, for instance. Until now `set` has never had the effect of being slow.

api/python/quilt3/packages.py
armandmcqueen: Adding it to my future todo. These tempfiles aren't actually temporary as far as I can tell (if delete=False, the file isn't cleaned up on file close and it writes to `/var/folder/...` which as far as I know isn't ever automatically cleaned up.

PR #1222, opened 2019-10-01
[WIP] Alternative package.set('KEY', object) approach with early serialization

armandmcqueen:
## Description

Alternative implementation where serialization happens immediately. Trades off slower `package.set()` for a guarantee that data will not be lost between `package.set()` and `package.build()`/`package.push()`

Any discussion should take place in #1221


armandmcqueen:
Closing as I do not like this solution. See explanation in #1221

PR #1228, opened 2019-10-03
Change `list_packages` to return just the package names

dimaryaz:
The current implementation downloads all of the package manifests;
this is not going to scale. For now, make it simple.


akarve:
@dimaryaz can you updated the docs as well?

PR #1231, opened 2019-10-04
Improve package.set('KEY', object) tempfile behavior

armandmcqueen:
## Description

Improved serialized file behavior for `package.set('KEY', obj)`.

- Files are now written to Quilt's appdir if the user doesn't specify a location
- Files are automatically deleted on pkg.push() if the user didn't specify a `serialization_location`

Enables:
```
pkg.set('KEY', df)
pkg.push("my/repo")
```
without leaving any intermediary files on disk


api/python/quilt3/packages.py
```diff
@@ -869,16 +878,16 @@ def set(self, logical_key, entry=None, meta=None, serialization_location=None, s
             entry(PackageEntry OR string OR object): new entry to place at logical_key in the package.
                 If entry is a string, it is treated as a URL, and an entry is created based on it.
                 If entry is None, the logical key string will be substituted as the entry value.
-                If entry is an object and we know how to serialize it, it will immediately be serialized and written to
-                disk, either to serialization_location or to a location determined by the tempfile library
+                If entry is an object and quilt knows how to serialize it, it will immediately be serialized and written
```
akarve: Can you note where users can find which file formats are known to Quilt?

api/python/quilt3/packages.py
```diff
@@ -869,16 +878,16 @@ def set(self, logical_key, entry=None, meta=None, serialization_location=None, s
...
-                characters. Will be passed as kwargs to the FormatHandler.serialize() function. See docstrings for
-                individual FormatHandlers too for full list of options -
+            serialization_format_opts(dict): [only used when entry is an object]. Options to help Quilt understand how
+                the object should be serialized. Useful for underspecified file formats like csv when content contains
+                confusing characters. Will be passed as kwargs to the FormatHandler.serialize() function. See docstrings
+                for individual FormatHandlers too for full list of options -
                 https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/formats.py
-            serialization_location(string): [when entry is an object]. Where the serialized object should be written,
-                e.g. "./mydataframe.parquet"
+            serialization_location(string): [only used when entry is an object]. Where the serialized object should be
```
akarve: We should note that `serilization_format` and `_location` are optional.

api/python/tests/integration/test_packages.py
```diff
@@ -581,29 +581,40 @@ def test_set_package_entry_as_object(self):
         pkg.set("mydataframe6.tsv", df, meta={'user_meta': 'blah6'})
 
         for lk, entry in pkg.walk():
-            file_path = parse_file_url(urlparse(entry.physical_keys[0]))
-            assert (pathlib.Path(file_path)).exists(), "The serialization files should exist"
+            file_path = parse_file_url(urlparse(entry.get()))
+            assert pathlib.Path(file_path).exists(), "The serialization files should exist"
 
-            self.file_sweeper_path_list.append(file_path)
+            self.file_sweeper_path_list.append(file_path)  # Make sure files get deleted even if test fails
```
akarve: the test set up / tear down should ensure deletion, right @dimaryaz?

api/python/quilt3/packages.py
armandmcqueen: That's harder than it sounds. I added a function to the `FormatRegistry` that prints out what formats Quilt supports - https://github.com/quiltdata/quilt/pull/1231/files#diff-4c173192a4cd94c4125490ccf3298e37R315. 

It's a bit hacky, but it works - let me know if you think I should beautify it some more.

api/python/quilt3/formats.py
```diff
@@ -312,6 +312,43 @@ def for_meta(cls, meta):
         name = cls._get_name_from_meta(meta)
         return cls.for_format(name)
 
+    @classmethod
+    def show_all_supported_formats(cls):
```
kevinemoore: Seems like this would be a very useful function to have return a dict, list or other parseable format (i.e. instead of simply printing).

armandmcqueen: Fixed

PR #1233, opened 2019-10-04
[WIP] Quick and dirty RODA login() fix

armandmcqueen:
## Description

- Don't assume a new role for RODA
- Downside: If you have a current role that restricts s3 buckets, may not be able to access some public buckets. But no fewer permissions that you had before.

### DIFF
--- a/api/python/quilt3/session.py
+++ b/api/python/quilt3/session.py
@@ -182,7 +182,10 @@ def login(force_open_quiltdata_com_login=False):
             f"\"quilt3.config('$URL')\", replacing '$URL' with your catalog homepage."
         )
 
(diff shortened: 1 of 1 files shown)


api/python/quilt3/session.py
```diff
@@ -177,6 +182,11 @@ def login():
             f"\"quilt3.config('$URL')\", replacing '$URL' with your catalog homepage."
         )
 
+    if registry_url ==  OPEN_DATA_URL and not force_open_quiltdata_com_login:
+        print("Skipping login. open.quiltdata.com is a Quilt stack for many public, open datasets. As the buckets "
+              "are public, you do not need to login to use features or read data.")
```
akarve: they do need to log in to search

PR #1251, opened 2019-10-07
Simplify package_delete

dimaryaz:
Don't remove the actual manifest hashes - just the files referencing them

### DIFF
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -1,7 +1,7 @@
 import pathlib
 from urllib.parse import urlparse, unquote
 
-from .data_transfer import copy_file, get_bytes, delete_object, put_bytes, list_objects, list_url
+from .data_transfer import copy_file, get_bytes, delete_url, put_bytes, list_objects, list_url
 from .formats import FormatRegistry
 from .search_util import search_api
 from .util import (QuiltConfig, QuiltException, CONFIG_PATH,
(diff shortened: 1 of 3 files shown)


akarve:
Note: this is a stepping stone to `.quilt` refactor, after which we'll be able to "deep delete" more safely.

api/python/quilt3/data_transfer.py
```diff
@@ -447,14 +447,8 @@ def _calculate_etag(file_path):
...
 
-    if key.endswith('/'):
-        paginator = s3_client.get_paginator('list_objects_v2')
-        for response in paginator.paginate(Bucket=bucket, Prefix=key):
-            for obj in response.get('Contents', []):
-                s3_client.delete_object(Bucket=bucket, Key=obj['Key'])
-    else:
-        s3_client.head_object(Bucket=bucket, Key=key)  # Make sure it exists
-        s3_client.delete_object(Bucket=bucket, Key=key)  # Actually delete it
+    s3_client.head_object(Bucket=bucket, Key=key)  # Make sure it exists
```
armandmcqueen: Why do we need to head instead of just handling the exception if the delete fails?

api/python/quilt3/data_transfer.py
```diff
@@ -512,6 +506,10 @@ def list_objects(bucket, prefix, recursive=True):
         return prefixes, objects
 
 
+def _looks_like_dir(s):
+    return not s or s.endswith('/')
```
armandmcqueen: This won't work on Windows, right? Do we care?

armandmcqueen: This won't work correctly on Windows, right?

api/python/quilt3/data_transfer.py
dimaryaz: Cause delete doesn't fail if the object doesn't exist... it's just a no-op.

api/python/quilt3/data_transfer.py
dimaryaz: It's only used on URLs, so it's going to be `/`

api/python/quilt3/api.py
```diff
@@ -128,58 +128,23 @@ def delete_package(name, registry=None):
...
+    usr, pkg = name.split('/')
 
     registry_base_path = get_package_registry(fix_url(registry) if registry else None)
-    registry_url = urlparse(registry_base_path)
-    pkg_namespace, pkg_subname = name.split("/")
-
-    tophashes_with_packages = _tophashes_with_packages(registry)
-
-    if registry_url.scheme == 'file':
-
```
kevinemoore: Is this removing the delete code for local/file-system package stores. Is that intentional? I wouldn’t think we need to remove that to improve the S3 code performance.

dimaryaz: You mean, the cleanup code? I wanted it to be consistent between local and S3, so it'll do the same thing - delete the package versions, but not the actual manifests. We can add a cleanup helper later - or hopefully, just migrate to a better directory structure.

kevinemoore: OK, that makes sense. I couldn't tell from just the diff if users could still call delete on local packages.

dimaryaz:
I kinda screwed up this PR: I opened it on top of #1228, and then forgot. Going to redo it.

PR #1252, opened 2019-10-08
Remove useless compression logic

dimaryaz:
No point in setting compression for each column individually - that's what pyarrow does already.

Fixes pyarrow 0.15


armandmcqueen:
PR #1231 pinned `pyarrow == 0.14.1` to pass tests. This PR should unpin pyarrow since code now works with 0.15.0

PR #1254, opened 2019-10-09
Overview searches

kevinemoore:
## Description
Adding two new actions to the search lambda: images (returns a list of images from the bucket) and other (returns a list of other pre-viewable files).


lambdas/search/index.py
```diff
@@ -50,6 +73,26 @@ def lambda_handler(request):
...
         _source = []
+    elif action == 'images':
+        body = {
+            'query': {
+                'terms': {
+                    "ext": IMG_EXTS
+                }
+            }
+        }
+        size = 10
```
akarve: I think we should request 100 of these, e.g. https://open.quiltdata.com/b/nasanex. @nl0 can yo confirm they are lazy loaded?

akarve: both `size = ` values should be set by GLOBALS, e.g. `NUM_PREVIEW_IMAGES`, `NUM_PREVIEW_FILES`

lambdas/search/index.py
```diff
@@ -50,6 +73,26 @@ def lambda_handler(request):
...
+        _source = []
+    elif action == 'other':
+        body = {
+            'query': {
+                'terms': {
+                    "ext": OTHER_EXTS
```
akarve: close but not quite what we want. we really should have a bias towards `readme.*` files

akarve: i guess in the general case that implies a full scan; can we at least look for `ROOT/readme.*`?

lambdas/search/index.py
nl0: confirming

lambdas/search/index.py
kevinemoore: A simple search for "README" works pretty well to grab just READMEs. Do we need anything fancier than that?

PR #1262, opened 2019-10-10
Add list_package_versions. Make list_packages not rely on `latest`.

dimaryaz:


api/python/quilt3/api.py
```diff
@@ -162,10 +162,27 @@ def list_packages(registry=None):
...
+        if len(parts) == 3:
+            pkg = f'{parts[0]}/{parts[1]}'
+            # A package can have multiple versions, but we should only return the name once.
+            if pkg != prev_pkg:
+                prev_pkg = pkg
+                yield pkg
+
+
+def list_package_versions(name, registry=None):
+    validate_package_name(name)
```
armandmcqueen: This is a user-facing method, right? We should add doc-string for docs.quiltdata.com

dimaryaz: I realized that the "versions" this method is returning aren't even useful for anything - need to resolve them to hashes first. Sigh.
And that brings up the question: how would we indicate which hash is "latest"?

armandmcqueen: Maybe by yielding a hash, timestamp pair? 
```
> list(quilt3.list_package_versions("quiltdata/example"))
"quiltdata/example"        00817c1cf85ac050e0c8fb9652a7a6     2019-10-03 12:34:56
"quiltdata/example"        9dd6b60e1ebebef962d3411ede8625     2019-10-01 01:23:45
```

api/python/quilt3/api.py
```diff
@@ -162,10 +162,39 @@ def list_packages(registry=None):
...
+
+    registry_base_path = get_package_registry(fix_url(registry) if registry else None)
+
+    package = registry_base_path.rstrip('/') + '/named_packages/' + name + '/'
+    for path, _ in list_url(package):
+        parts = path.split('/')
+        if len(parts) == 1:
+            pkg_hash, _ = get_bytes(package + parts[0])
+            yield parts[0], pkg_hash.decode().strip()
```
armandmcqueen: Current user experience will be:
```
> list(quilt3.list_package_versions("quiltdata/example"))
[
("1552516174", "cf4ffaa6082aef9caa9d5b92a39808db065f894fbb307b8a121637bbd6f46667"),
("1552517308", "7927f4c3cfab5193a9fe383908303c570646af890cabc2a7a34a5ba5251c7be9"),
("latest", "5d4354f610a497d06c2fb53771e0f7277f473c837252c50a1a4b64eed4fdaca2")
]
```

Do you think it will be obvious to users that the version is a unix timestamp? If not maybe we should make that clearer either by returning datetime objects/formatted timestamp strings or by calling it out in the docstring?

dimaryaz: Well... Technically, we can't assume they're timestamps. Or at least, we'd need to deal with cases when they're not.

armandmcqueen: Only the "latest" would not be a timestamp, right? In any case, it's not critical for this PR.

PR #1264, opened 2019-10-10
Fix bug when pushing a package containing temp files twice 

armandmcqueen:
## Description

When a package with tempfiles is pushed, update the package to point to the new location so the original package isn't pointing to a temp file that no longer exists

Fixes:
```
pkg.set('KEY', pd.DataFrame([...])
pkg.push('user/pkg')
pkg.push('user/pkg') # Previously crashed due to file not existing
```

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1101,7 +1101,19 @@ class Package(object):
 
         self._fix_sha256()
         pkg = self._materialize(dest)
+
+        def physical_key_is_temp_file(pk):
+            if not file_is_local(pk):
+                return False
+            return pathlib.Path(parse_file_url(urlparse(pk))).parent == APP_DIR_TEMPFILE_DIR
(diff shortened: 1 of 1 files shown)


api/python/quilt3/packages.py
```diff
@@ -1101,7 +1101,19 @@ def push(self, name, registry=None, dest=None, message=None):
 
         self._fix_sha256()
         pkg = self._materialize(dest)
+
+        def physical_key_is_temp_file(pk):
+            if not file_is_local(pk):
+                return False
+            return pathlib.Path(parse_file_url(urlparse(pk))).parent == APP_DIR_TEMPFILE_DIR
+        temp_file_logical_keys = [lk for lk, entry in self.walk() if physical_key_is_temp_file(entry.physical_keys[0])]
```
dimaryaz: This logic is now duplicated here and in `_delete_temporary_files`... Maybe do it all at once - find temp files, delete them, and fix the logical keys?

armandmcqueen: Good point. Updated it so we only do that work once

PR #1285, opened 2019-10-16
Implement deleting a specific package hash

dimaryaz:


dimaryaz:
It defaults to `default_local_registry` - and that's the same for pretty much all other APIs, I think?

PR #1289, opened 2019-10-17
lambdas/search: adjust overview queries

nl0:


lambdas/search/index.py
```diff
@@ -78,34 +78,44 @@ def lambda_handler(request):
         _source = []
     elif action == 'images':
         body = {
-            'query': {
-                'terms': {
-                    "ext": IMG_EXTS
-                }
-            }
+            'query': {'terms': {'ext': IMG_EXTS}},
+            'sort': [{'last_modified': {'order': 'desc'}}],
```
kevinemoore: Sorts are very expensive--e.g., the query will need to fetch all the files to check their timestamps instead of stopping after finding enough matches. Let's not sort.

nl0: moved to inner_hits

lambdas/search/index.py
```diff
@@ -78,34 +78,43 @@ def lambda_handler(request):
...
+                    'size': 1,
+                    'sort': [{'last_modified': 'desc'}],
+                    '_source': ['key', 'version_id'],
+                },
+            },
         }
         size = NUM_PREVIEW_IMAGES
         _source = []
-    elif action == 'other':
+    #TODO: deprecate 'other'
```
kevinemoore: `other` should only be used by the catalog so go ahead and remove that now?

nl0: removed

PR #1294, opened 2019-10-18
Bucket/Overview: adjust analytics window options

nl0:


catalog/app/containers/Bucket/Overview.js
```diff
@@ -248,8 +248,10 @@ function SparklineSkel({ height, width, lines, window = 30, animate = false, chi
 }
 
 const ANALYTICS_WINDOW_OPTIONS = [
-  { value: 30, label: 'Last 30 days' },
-  { value: 365, label: 'Last 365 days' },
+  { value: 30, label: 'Last 1 month' },
```
akarve: Why not `31`?
```suggestion
  { value: 31, label: 'Last 1 month' },
```

PR #1297, opened 2019-10-18
Make sure we don't write to more than 100 Athena partitions at once.

dimaryaz:


lambdas/access_counts/index.py
```diff
@@ -27,6 +27,9 @@
 # Timestamp for the dir above.
 LAST_UPDATE_KEY = f'{OBJECT_ACCESS_LOG_DIR}.last_updated_ts.txt'
 
+# Athena does not allow us to write more than 100 partitions at once.
+MAX_OPEN_PARTITIONS = 100
```
akarve: Not a requirement for this PR but we should probably use [`boto3` service quotas](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/service-quotas.html#ServiceQuotas.Client.get_service_quota) instead of hardcoding.

lambdas/access_counts/index.py
```diff
@@ -322,6 +325,11 @@ def handler(event, context):
         # We start from scratch, so make sure we don't have any old data.
         delete_dir(QUERY_RESULT_BUCKET, OBJECT_ACCESS_LOG_DIR)
 
+    # We can't write more than 100 days worth of data at a time due to Athena's partitioning limitations.
+    # Moreover, we don't want the lambda to time out, so just process 100 days and let the next invocation handle the rest.
+    if (end_ts - start_ts).days >= MAX_OPEN_PARTITIONS:
```
akarve: why `>=` and not `>`? i would probably do the same thing, just asking :)

dimaryaz: Hm, actually, neither of them is right...

Suppose `start_ts` is day=1, hour=23, and `end_ts` is day=3, hour=1. The difference is 1 day (plus two hours), while the time range spans 3 days.

akarve: Good point. Don't you guarantee elsewhere that end_ts is on a date boundary so that the code never writes incomplete days? In that case I think the worst case is MAX_OPEN_PARTITIONS - 2 days right?

PR #1330, opened 2019-10-30
Fix unnecessary copying when modifying only a part of a package

dimaryaz:
If the destination of an object is the same as the source (minus the version ID), then skip the copy.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1141,22 +1141,28 @@ class Package(object):
         pkg._meta = self._meta
         # Since all that is modified is physical keys, pkg will have the same top hash
         file_list = []
+        entries = []
         for logical_key, entry in self.walk():
             # Copy the datafiles in the package.
             physical_key = _to_singleton(entry.physical_keys)
+            unversioned_physical_key = physical_key.split('?', 1)[0]
             new_physical_key = dest_url + "/" + quote(logical_key)
-            file_list.append((physical_key, new_physical_key, entry.size, entry.meta))
+            new_entry = entry._clone()
+            if unversioned_physical_key == new_physical_key:
+                # No need to copy - re-use the original physical key.
+                pkg.set(logical_key, new_entry)
+            else:
+                entries.append((logical_key, new_entry))
+                file_list.append((physical_key, new_physical_key, entry.size, None))
 
         results = copy_file_list(file_list)
 
-        for (logical_key, entry), versioned_key in zip(self.walk(), results):
+        for (logical_key, new_entry), versioned_key in zip(entries, results):
             # Create a new package entry pointing to the new remote key.
             assert versioned_key is not None
-            new_entry = entry._clone()
             new_entry.physical_keys = [versioned_key]
             pkg.set(logical_key, new_entry)
         return pkg
-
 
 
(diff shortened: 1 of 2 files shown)


api/python/quilt3/packages.py
```diff
@@ -1141,24 +1141,30 @@ def _materialize(self, dest_url):
         pkg._meta = self._meta
         # Since all that is modified is physical keys, pkg will have the same top hash
         file_list = []
+        entries = []
         for logical_key, entry in self.walk():
             # Copy the datafiles in the package.
             physical_key = _to_singleton(entry.physical_keys)
+            unversioned_physical_key = physical_key.split('?', 1)[0]
```
akarve: Object keys may contain `?`
```suggestion
            unversioned_physical_key = physical_key.rsplit('?', 1)[0]
```

armandmcqueen: Can we split on `?versionId=` instead? I don't think we can safely assume that filenames won't have question marks in them.

api/python/tests/integration/test_packages.py
```diff
@@ -319,7 +319,20 @@ def test_load_into_quilt(self):
             expected_params={
                 'Body': ANY,
                 'Bucket': 'my_test_bucket',
-                'Key': 'Quilt/package/foo',
+                'Key': 'Quilt/package/foo1',
+                'Metadata': {'helium': '{}'}
```
akarve: I guess because of https://github.com/quiltdata/quilt/blob/ee5ccbdfcb026fe265ba3fc102465c92e99441aa/api/python/tests/integration/test_packages.py#L40 we cover the case with absolutely no metadata?

api/python/quilt3/packages.py
dimaryaz: This is why we URL-encode paths...

armandmcqueen: The logic we use does not work. Try
```
pkg = Package()
pkg.set("k", f"s3://quilt-ml-data/tst/myfile?.txt")
```
You will get a 404 despite the object existing. 

The core problem is illustrated by this function:
```
def test_question_mark():
    url = fix_url(str("s3://quilt-ml-data/tst/myfile?.txt"))
    parsed_url = urlparse(url)
    b, k, v = parse_s3_url(parsed_url)
    print("Expected:", "quilt-ml-data", "tst/myfile?.txt", None)
    print("Actual:", b, k, v)
```

We can fix it or not - `?` is [not a recommended character to use in an s3 key](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys), but I don't see a good reason to split on `?` instead of `?versionId=`

dimaryaz: Um... https://quilt-example.s3.amazonaws.com/dima/q/foo?.txt doesn't work, either - because it needs to be https://quilt-dima.s3.amazonaws.com/dima/q/foo%3F.txt . That is how URLs work. (Also, what if the file name contains "?versionId="?)

armandmcqueen: I don't think this is worth blocking on, but there's currently no way for the user to know that they need to url encode the key that s3.list_object_v2 returns before calling pkg.set()

PR #1331, opened 2019-10-31
Telemetry decorator proposal

armandmcqueen:
## Description

Prototype decorator class initial implementation. 

@dimaryaz please take a look and give comments


api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,118 @@
+import functools
+from concurrent.futures import ThreadPoolExecutor, wait
+from requests_futures.sessions import FuturesSession
+from threading import Lock
+import uuid
+import atexit
+import time
+
+TELEMETRY_URL = "https://gha7zraiy7.execute-api.us-east-1.amazonaws.com/Stage/metrics/installer"
```
kevinemoore: Quick suggestion: use a DNS name for this (so we don't have to update clients when we redeploy the service).

armandmcqueen: It's a test API for now, but it will go to quiltdata.cloud/metrics when this is finalized

dimaryaz: Definitely need a DNS name for this... But does API Gateway even support it? Or would we make it part of the registry?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,118 @@
...
+TELEMETRY_USER_AGENT = "QuiltCli"
```
dimaryaz: Should probably call it "Quilt3" or something.

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,118 @@
...
+API_CALL_TIMEOUT_SECS = 5
```
dimaryaz: If it's all async, then it probably doesn't matter what the timeout is?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,118 @@
...
+                "telemetry_schema_version": "test",
+            }
+            r = ApiTelemetry.session.post(TELEMETRY_URL,
+                                          timeout=API_CALL_TIMEOUT_SECS,
+                                          json=data,
+                                          headers={'User-Agent': TELEMETRY_USER_AGENT})
+
+            with ApiTelemetry.pending_reqs_lock:
+                ApiTelemetry.pending_reqs.append(r)
+            func(*args, **kwargs)
```
dimaryaz: Good thing none of our APIs have any return values ;)

api/python/quilt3/telemetry.py
armandmcqueen: API Gateway supports custom domain names. The Telemetry API in the prod account is already set up to be called via quiltdata.cloud/metrics.

armandmcqueen:
Batching is an interesting idea - do you think performance will be bad enough that we need it. I think it's pretty complicated to implement well, so if we can avoid it in the near term, that would be preferable (in many cases there won't be a next time to upload leftover files because the code is running on ephemeral infrastructure).

api/python/quilt3/telemetry.py
armandmcqueen: I think we can get rid of the timeout as long as we confirm that we can make a connection from the client to the API.

armandmcqueen:
Current state:

There are a couple open items I'm working on, but for the most part this is done and is ready for review.

- I need to look into whether having the Telemetry Proxy API running only in us-east-1 has any noticeable latency costs (I don't expect much since everything is async in a background thread)
- I need to improve the logic around the `QUILT_DISABLE_USAGE_METRICS` environment variable so that `QUILT_DISABLE_USAGE_METRICS=False` doesn't disable the metrics.
- I am planning to punt on checking whether the client can connect to the API

We have chosen not to batch requests at the moment because it seems unlikely to be a performance problem for most people and is extra implementation effort. We may need to revisit that.

We are keeping USER_AGENT as `QuiltCli` for now because it makes the telemetry proxy API simpler if we don't have a separate user agent for the installer and the `quilt3` python client. We can revisit that after reinvent.

One important thing that is missing is a page in the docs that explains our telemetry policy. Unsure if that needs to be done for reinvent or not.

api/python/quilt3/VERSION
```diff
@@ -0,0 +1 @@
+3.1.5
```
akarve: We should really get this from `setup.py`, in the same directory, so as not to dupilcate state.

armandmcqueen: State is not duplicated. `setup.py` get the version from this file as does `__init__`. This allows users to call `quilt3.__version__`.

api/python/quilt3/api.py
```diff
@@ -151,6 +165,14 @@ def list_package_versions(name, registry=None):
     Returns:
         A sequence of tuples containing the named version and hash.
     """
+    return _list_package_versions(name=name, registry=registry)
+
+
+def _list_package_versions(name, registry=None):
```
akarve: the `_method` = telemetry-free makes sense but, just a thought, is there an easier way to do this that doesn't require method renaming? idk like  global singleton where you can turn off telemetery; i don't love that either just wondering aloud if there's another pattern for this already.

akarve:
Resolving conflicts and catching up with master should fix CI failures.

api/python/quilt3/api.py
armandmcqueen: I was thinking maybe we should move to a design where we have an `InternalPackage` class that hold the internal logic and a `Package` class that is for the end-user. We put telemetry and documentation on the `Package` class methods which ensures we never get more than one event per user action. It would also clarify what APIs we can change freely and what APIs we are promising the user will stay the same. 

But that is a bit more of a substantial change.

api/python/quilt3/packages.py
```diff
@@ -245,6 +246,7 @@ def __init__(self):
         self._children = {}
         self._meta = {'version': 'v0'}
 
+    @ApiTelemetry("package.__repr__")
```
dimaryaz: `Package.__repr__`

api/python/quilt3/packages.py
```diff
@@ -318,6 +320,7 @@ def meta(self):
         return self._meta.get('user_meta', dict())
 
     @classmethod
+    @ApiTelemetry("Package.install")
```
dimaryaz: I feel like this should be outside of `@classmethod`? Though I'm not actually sure how `@classmethod` works.

api/python/quilt3/packages.py
```diff
@@ -459,6 +463,7 @@ def __getitem__(self, logical_key):
             pkg = pkg._children[key_fragment]
         return pkg
 
+    @ApiTelemetry("package.fetch")
```
dimaryaz: `Package.fetch`

api/python/quilt3/telemetry.py
dimaryaz: ?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+    pending_reqs = []
+    pending_reqs_lock = Lock()
+    telemetry_disabled = False
+
+    @staticmethod
+    def create_session():
+        return FuturesSession(executor=ThreadPoolExecutor(max_workers=2))
+
+    def __init__(self, api_name):
+        ApiTelemetry.telemetry_disabled = ApiTelemetry.telemetry_is_disabled()
```
dimaryaz: Why are you setting globals in the constructor?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @staticmethod
+    def check_telemetry_disabled_by_envvar():
+        # TODO: Potential bug where 'QUILT_DISABLE_USAGE_METRICS=False' doesn't work as expected
+        return bool(os.environ.get(DISABLE_USAGE_METRICS_ENVVAR, False))
+
+
+    @staticmethod
+    def telemetry_is_disabled():
+        disabled_via_config = ApiTelemetry.check_telemetry_disabled_in_config()
```
dimaryaz: Why not just...
`return ApiTelemetry.check_telemetry_disabled_in_config() or ApiTelemetry.check_telemetry_disabled_by_envvar() or ...`?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class ApiTelemetry:
+    session = None
+    session_id = str(uuid.uuid4())
+    pending_reqs = []
+    pending_reqs_lock = Lock()
+    telemetry_disabled = False
+
+    @staticmethod
```
dimaryaz: Use `@classmethod` for everything; the existence of `@staticmethod` is kind of a mistake in Python.

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+            'python_version_patch': platform.python_version_tuple()[2]
+        }
+        print(f"Sending data: {data}")
+        r = ApiTelemetry.session.post(TELEMETRY_URL, json=data, headers={'User-Agent': TELEMETRY_USER_AGENT})
+
+        with ApiTelemetry.pending_reqs_lock:
+            ApiTelemetry.pending_reqs.append(r)
+
+
+    def __call__(self, func):
```
dimaryaz: You could actually get the function name from `func` instead of passing a string into the constructor

api/python/quilt3/util.py
```diff
@@ -366,6 +369,19 @@ def get_install_location():
         loc = get_from_config('default_local_registry').rstrip('/')
     return loc
 
+
+
+def set_config_value(key, value):
```
dimaryaz: There's no way we don't already have some function that does this? (I think that's what `config(...)` is supposed to do?)

api/python/quilt3/packages.py
dimaryaz: (I don't really care about lowercase vs uppercase - but we should be consistent.)

api/python/quilt3/packages.py
```diff
@@ -793,6 +806,11 @@ def build(self, name=None, registry=None, message=None):
         Returns:
             The top hash as a string.
         """
+        return self._build(name=name, registry=registry, message=message)
+
+
+    def _build(self, name=None, registry=None, message=None):
+        """   Telemetry-free version of build()   """
```
dimaryaz: It's a bit weird to mention telemetry in every function... I would remove the comments from the private functions.

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @staticmethod
+    def create_session():
+        return FuturesSession(executor=ThreadPoolExecutor(max_workers=2))
+
+    def __init__(self, api_name):
+        ApiTelemetry.telemetry_disabled = ApiTelemetry.telemetry_is_disabled()
+
+        if ApiTelemetry.session is None:
+            ApiTelemetry.session = ApiTelemetry.create_session()
```
dimaryaz: It's not exactly thread-safe... https://github.com/ross/requests-futures/issues/47

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+            'client_type': TELEMETRY_CLIENT_TYPE,
+            'client_version': quilt3_version,
+            'platform': sys.platform,
+            'python_implementation': platform.python_implementation(),
+            'python_version_major': platform.python_version_tuple()[0],
+            'python_version_minor': platform.python_version_tuple()[1],
+            'python_version_patch': platform.python_version_tuple()[2]
+        }
+        print(f"Sending data: {data}")
+        r = ApiTelemetry.session.post(TELEMETRY_URL, json=data, headers={'User-Agent': TELEMETRY_USER_AGENT})
```
dimaryaz: Sessions aren't exactly thread-safe... I'd lock everything.
https://github.com/ross/requests-futures/issues/47

api/python/quilt3/telemetry.py
armandmcqueen: Does it need to be thread-safe?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,146 @@
...
+            return config_value
+        else:
+            set_config_value("telemetry_disabled", False)
+            return False
+
+
+    @staticmethod
+    def check_telemetry_disabled_by_envvar():
+        # TODO: Potential bug where 'QUILT_DISABLE_USAGE_METRICS=False' doesn't work as expected
+        return bool(os.environ.get(DISABLE_USAGE_METRICS_ENVVAR, False))
```
akarve: might be slightly more correct to default to `''` instead of `False` because when it's set `environ.get` will give you a string.

api/python/quilt3/packages.py
armandmcqueen: I believe a custom decorator should always be on the inside of a classmethod decorator - https://stackoverflow.com/questions/1987919/why-can-decorator-not-decorate-a-staticmethod-or-a-classmethod

api/python/quilt3/telemetry.py
armandmcqueen: As opposed to at import time?

api/python/quilt3/telemetry.py
armandmcqueen: The function name isn't quite good enough for analytics

api/python/quilt3/telemetry.py
armandmcqueen: I don't think that's what we want to do. You should not be able to call these methods from an instance of the class.

api/python/quilt3/util.py
armandmcqueen: We have `api.config()`, but that's method that needs telemetry on it and the telemetry code needs to be able to set the config value, so it became a circular dependency.

api/python/quilt3/packages.py
armandmcqueen: Agreed

api/python/quilt3/telemetry.py
dimaryaz: Sure... But `@staticmethod` doesn't stop you from doing that:

```
In [1]: class Foo():
   ...:     @staticmethod
   ...:     def blah():
   ...:         print('blah')
   ...:

In [2]: Foo.blah()
blah

In [3]: x = Foo()

In [4]: x.blah()
blah
```

Basically, it has no redeeming qualities. Also, according to Guido:
"""
They're basically an accident -- back in the Python 2.2 days when I was inventing new-style classes and descriptors, I meant to implement class methods but at first I didn't understand them and accidentally implemented static methods first. Then it was too late to remove them and only provide class methods.
"""

api/python/quilt3/packages.py
```diff
@@ -569,6 +579,7 @@ def load(cls, readable_file):
 
         return pkg
 
+    @ApiTelemetry("package.set_dir")
```
akarve: For now let's not instrument the `set*` methods. We would benefit from visibility into this but until we have some kind of event coalescing it could blow up analytics.

api/python/quilt3/packages.py
```diff
@@ -851,6 +875,8 @@ def manifest(self):
         for logical_key, entry in self.walk():
             yield {'logical_key': logical_key, **entry.as_dict()}
 
+
+    @ApiTelemetry("package.set")
```
akarve: Per above, let's temporarily not instrument this call, as it could be high-volume.

api/python/quilt3/telemetry.py
dimaryaz: Sure. Like, it doesn't make sense for each call to `ApiTelemetry(...)` to modify global state used by other `ApiTelemetry` instances?

dimaryaz: I guess you could do something like, `if ApiTelemetry.telemetry_disabled is None:`

api/python/quilt3/api.py
```diff
@@ -182,6 +212,15 @@ def config(*catalog_url, **config_values):
     # Return current config
     return QuiltConfig(CONFIG_PATH, local_config)
 
+
+@ApiTelemetry("api.disable_telemetry")
+def disable_telemetry():
+    """ Permanently disable sending of anonymous usage metrics """
+    _config(telemetry_disabled=True)
```
dimaryaz: This doesn't disable anything for the current session, though.

armandmcqueen: Good catch! I added a somewhat hacky workaround for that in `telemetry.py` which should solve this for now

api/python/quilt3/telemetry.py
armandmcqueen: Fixed

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+
+
+class ApiTelemetry:
+    session = None
+    session_id = str(uuid.uuid4())
+    pending_reqs = []
+    pending_reqs_lock = Lock()
+    telemetry_disabled = False
```
dimaryaz: Um... this needs to be initialized to `None` - otherwise `if ApiTelemetry.telemetry_disabled is None:` doesn't work?

api/python/quilt3/telemetry.py
```diff
@@ -0,0 +1,159 @@
...
+            "telemetry_schema_version": TELEMETRY_SCHEMA_VERSION,
+            'client_type': TELEMETRY_CLIENT_TYPE,
+            'client_version': quilt3_version,
+            'platform': sys.platform,
+            'python_implementation': platform.python_implementation(),
+            'python_version_major': platform.python_version_tuple()[0],
+            'python_version_minor': platform.python_version_tuple()[1],
+            'python_version_patch': platform.python_version_tuple()[2]
+        }
+        print(f"Sending data: {data}")
```
dimaryaz: ...

PR #1337, opened 2019-11-01
WIP

kevinemoore:
## Description
Comments on the API proposal

### DIFF
--- a/code_samples.md
+++ b/code_samples.md
@@ -11,6 +11,21 @@ API names are rough and will need to be improved.
 ## Usecase: Download
 - I am training a ML model and I want to make a large amount of data accessible to my training model. Currently my code is written to read that data from files in a directory, probably on EFS or FSx. While not an immediate concern to me, I would like to be able to reproduce this model in the future.
 
+>How important is location to this workflow? One motivation for download is that the code expects local files (e.g., in a particular folder). A second motivation is for faster access (especially if files are accessed repeatedly). 
+>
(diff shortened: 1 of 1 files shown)


code_samples.md
```diff
@@ -117,13 +139,17 @@ PackageBuilder(
 Exception: The manifest has not been pushed to a remote repository. Use push_manifest() to first push the manifest to s3 or use allow_local_manifest=True if this is intentional
 ```
 
+>I was confused by this example. I would expect a package to be reproducible regardless of whether or not it lives in S3. More precisely, the package should contain enough information to verify that a dataset matches a defined state. The underlying files might no longer be available, but any consumer of the package should be able to verify if the contents match the expected hashes.
```
armandmcqueen: The thinking was that no one else can reproduce the results if the package is local. I don't have a strong opinion about this, but I do think we need guardrails so someone doesn't think their work can be replicated when it can't

code_samples.md
```diff
@@ -238,14 +264,15 @@ Wall time: 123 s
 > %time pkg.cache("other/huge_file.json")
 CPU times: user 123 s, sys: 253 µs, total: 123 s
 Wall time: 123 s
+_I normally expect caching to transparent_
```
armandmcqueen: The challenge is knowing when to clean up the on-disk cache.

armandmcqueen:
Folding comments into refactor branch

PR #1339, opened 2019-11-01
Use the original URI to prevent Nginx from decoding S3 paths

dimaryaz:


s3-proxy/nginx.conf
```diff
@@ -57,7 +57,10 @@ http {
             }
 
             # Proxy everything else to S3.
-            proxy_pass 'https://$s3_bucket.s3.amazonaws.com/$s3_path$is_args$args';
+            # Use $request_uri rather than $s3_path because it needs to stay encoded.
+            if ($request_uri ~ "^/[^/]+/(.*)") {
```
akarve: so the scheme has already been stripped from `$requiest_uri`? I am curious if we will ever get something other than s3:// paths and if not then why do we need to handle multiple leading `/`?

dimaryaz: `$request_uri` contains just the path, starting with `/`, without a scheme. But, it contains the bucket name first - so `/S3_BUCKET/S3_PATH`

dimaryaz:
@nl0: yeah, that one seems to work.

PR #1350, opened 2019-11-04
Use Ubuntu 18.04 (LTS) for Nginx images

dimaryaz:
- In the catalog, just upgrade 16.04 to 18.04.
- In the s3 proxy, switch from a global config to a site config (so we don't override things like the nginx user which is different in Ubuntu, Amazon Linux, etc.)


dimaryaz:
I have no preference here, so I can switch it back to Ubuntu... Though with `apt-get`, we don't have a whole lot of control over the Nginx version - just whatever version Ubuntu ships.

PR #1351, opened 2019-11-05
Initial feedback

akarve:


code_samples.md
```diff
@@ -153,6 +260,8 @@ Unclear the best representation of meta when displaying a Package/PackageBuilder
 > pkg.set("train2017/1.jpg", "./replacement/1.jpg", meta={})
```
akarve: TODO: what about `.set("KEY", obj)`?

akarve: Suggest that we should maintain this functionality for existing user base

armandmcqueen:
Folding comments into refactor branch

PR #1353, opened 2019-11-06
Clean up handling of delete events, bad variable refs, simplify

akarve:


lambdas/es/indexer/index.py
```diff
@@ -212,15 +214,32 @@ def handler(event, context):
...
+                # Handle delete  first and then continue so that
+                # head_object and get_object (below) don't fail
+                if event_name == OBJECT_DELETE:
+                    batch_processor.append(
+                        event_name,
+                        bucket=bucket,
+                        ext=ext,
+                        etag=etag,
+                        key=key,
+                        last_modified=datetime.datetime.now(),
```
dimaryaz: `tz=datetime.timezone.utc`

PR #1354, opened 2019-11-06
Work around contextual version conflict dateutil:botcore

akarve:


api/python/setup.py
```diff
@@ -55,6 +55,7 @@ def run(self):
         'boto3>=1.8.0',
         'jsonlines==1.2.0',
         'packaging>=16.8',
+        'python-dateutil<=2.8.0',           # 2.8.1 conflicts with botocore
```
dimaryaz: I'd do `<2.8.1` just to be consistent with botocore

PR #1371, opened 2019-11-13
Fix an exception when reading Parquet metadata

dimaryaz:
Don't try to parse the metadata as JSON cause it might not be JSON.
Also don't read the file twice.


lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -63,15 +63,17 @@ def extract_parquet(file_, as_html=True):
...
-    meta = pq.read_metadata(file_)
+    pf = pq.ParquetFile(file_)
+
+    meta = pf.metadata
 
     info = {}
     info['created_by'] = meta.created_by
     info['format_version'] = meta.format_version
     info['metadata'] = {
         # seems silly but sets up a simple json.dumps(info) below
```
akarve: Delete this comment now?

PR #1388, opened 2019-11-18
Use ChunksizeAdjuster to adjust chunk size

dimaryaz:
...because of course it's a thing.

S3 allows a maximum of 10,000 chunks - and with the default chunk size of 8MB,
we would go over that if the file is bigger than 80GB.

Boto adjusts chunk sizes to deal with that, so let's do the same.

Also adjust unit tests to use realistic file sizes (because there's also
a minimum allowed chunk size).

### DIFF
--- a/api/python/tests/test_data_transfer.py
+++ b/api/python/tests/test_data_transfer.py
@@ -443,10 +443,16 @@ class DataTransferTest(QuiltTestCase):
 
 
     def test_multipart_copy(self):
-        size = 20 * 1024 * 1024
+        size = 100 * 1024 * 1024 * 1024
+
+        # size / 8MB would give us 12501 chunks - but the maximum allowed is 10000,
+        # so we should end with 16MB chunks instead.
         chunksize = 8 * 1024 * 1024
+        assert size / chunksize > 10000
+        chunksize *= 2
 
         chunks = -(-size // chunksize)
+        assert chunks <= 10000
 
         self.s3_stubber.add_response(
             method='head_object',
--- a/api/python/tests/test_data_transfer.py
+++ b/api/python/tests/test_data_transfer.py
@@ -371,14 +371,25 @@ class DataTransferTest(QuiltTestCase):
 
 
     def test_multipart_upload(self):
-        path = DATA_DIR / 'large_file.npy'
+        name = 'very_large_file.bin'
+        path = pathlib.Path(name)
+
+        size = 30 * 1024 * 1024
+        chunksize = 8 * 1024 * 1024
+
+        chunks = -(-size // chunksize)
+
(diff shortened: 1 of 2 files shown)


api/python/tests/test_data_transfer.py
```diff
@@ -371,14 +371,25 @@ def test_upload_large_file_etag_match_metadata_mismatch(self):
 
 
     def test_multipart_upload(self):
-        path = DATA_DIR / 'large_file.npy'
+        name = 'very_large_file.bin'
+        path = pathlib.Path(name)
+
+        size = 30 * 1024 * 1024
+        chunksize = 8 * 1024 * 1024
```
akarve: Don't we want to push for the case where we have more than 10,000 parts per `Part number must be an integer between 1 and 10000, inclusive`

akarve: Or at least where `size/chunksize > CHUNK_LIMIT`?

dimaryaz: Sure... I'll add it to the multipart copy test, though. The logic is the same - but a unittest creating and uploading a 100GB file seems a bit extreme. In the copy test, there's no actual IO going on, so it's fine.

dimaryaz: (That is, in the upload test, it wouldn't take up 100GB on disk, and it wouldn't be uploaded - but it would still be read)

api/python/tests/test_data_transfer.py
```diff
@@ -371,14 +371,25 @@ def test_upload_large_file_etag_match_metadata_mismatch(self):
...
+
+        chunks = -(-size // chunksize)
```
JacksonMaxfield: Sorry, why double negative?

api/python/tests/test_data_transfer.py
```diff
@@ -414,27 +425,34 @@ def test_multipart_upload(self):
...
-        file_size = 5000
+        size = 100 * 1024 * 1024 * 1024
+
+        # size / 8MB would give us 12501 chunks - but the maximum allowed is 10000,
+        # so we should end with 16MB chunks instead.
+        chunksize = 8 * 1024 * 1024
+        assert size / chunksize > 10000
+        chunksize *= 2
+
+        chunks = -(-size // chunksize)
```
JacksonMaxfield: Same thing here. Probably just not understanding.

api/python/tests/test_data_transfer.py
dimaryaz: Haha, this is the same as `import math`, ..., `math.ceil(size / chunksize)` - just trying to be clever and save a few bytes. I can change it to `math.ceil`.

PR #1399, opened 2019-11-22
Implement Package.rollback

dimaryaz:


api/python/quilt3/packages.py
```diff
@@ -1100,6 +1100,34 @@ def physical_key_is_temp_file(pk):
...
+        Set the "latest" version to the given hash.
+
+        Args:
+            name(str): Name of package to rollback.
+            top_hash(str): Hash to rollback to.
+            registry(str): Registry where package is located.
+                Defaults to the default local registry.
+        """
+        if registry is None:
+            registry = get_from_config('default_local_registry')
```
dimaryaz: Question: do we default to local or remote?

akarve: pretty disruptive operation so no default at all and make `registry` required?

api/python/quilt3/packages.py
```diff
@@ -1100,6 +1100,34 @@ def physical_key_is_temp_file(pk):
...
+        else:
+            registry = fix_url(registry)
+
+        registry = registry.rstrip('/')
+        validate_package_name(name)
+
+        hash_path = f'{registry}/.quilt/packages/{quote(top_hash)}'
+        latest_path = f'{registry}/.quilt/named_packages/{quote(name)}/latest'
```
akarve: this would be cleaner if it called a shared function that was schema-version aware, but this is OK for now

PR #1401, opened 2019-11-22
Cache remote manifests locally.

dimaryaz:


api/python/quilt3/packages.py
```diff
@@ -446,20 +446,21 @@ def browse(cls, name=None, registry=None, top_hash=None):
...
-            with open(parse_file_url(src_url)) as open_file:
-                pkg = cls.load(open_file)
-        elif src_url.scheme == 's3':
-            body = get_bytes(uri)
-            pkg = cls.load(io.BytesIO(body))
+        if file_is_local(pkg_manifest_uri):
+            local_pkg_manifest = parse_file_url(urlparse(pkg_manifest_uri))
         else:
-            raise NotImplementedError
+            local_pkg_manifest = CACHE_PATH / "manifest" / binascii.hexlify(pkg_manifest_uri.encode()).decode()
```
akarve: Shouldn't there be a universal method, also used in the ObjectCacheClass, that goes from a string to the potential cache location? Because we do this in a few places and it's easy to get wrong (e.g. skip encode, don't decode, etc.)).

api/python/tests/integration/test_packages.py
```diff
@@ -157,52 +157,73 @@ def test_read_manifest(self):
...
+                'Body': BytesIO(b'abcdef'),
+            },
+            expected_params={
+                'Bucket': 'test-bucket',
+                'Key': '.quilt/named_packages/Quilt/test/latest',
+            }
+        )
+
+        self.s3_stubber.add_response(
+            method='head_object',
```
akarve: why do we head manifests at all?

PR #1413, opened 2019-11-25
Added some Package and PackageEntry utility functions

armandmcqueen:


api/python/quilt3/packages.py
```diff
@@ -225,6 +226,41 @@ def get_cached_path(self):
             return ObjectPathCache.get(physical_key)
         return None
 
+    def get_bytes(self, use_cache_if_available=True):
```
akarve: looks good. how about unit tests for a few of these, at least get_bytes?

api/python/quilt3/packages.py
```diff
@@ -225,6 +226,41 @@ def get_cached_path(self):
...
+        """
+        Returns the bytes of the object this entry corresponds to. If 'use_cache_if_available'=True, will first try to
+        retrieve the bytes from cache.
+        """
+        if use_cache_if_available:
+            cached_path = self.get_cached_path()
+            if cached_path is not None:
+                cached_file_url = fix_url(cached_path)
```
dimaryaz: `fix_url` is intended strictly for user input, where we want to handle both URLs and file paths (and rely on heuristics - e.g.,you could technically have a local file named `s3:`, but it's unlikely). Here, you know it's going to be a file path, so just use `pathlib`:
`pathlib.Path(cached_path).as_uri()`
or just read the bytes directly:
`pathlib.Path(cached_path).read_bytes()`

api/python/quilt3/packages.py
```diff
@@ -718,7 +754,31 @@ def get(self, logical_key):
...
 
-    # def get_as_pathlib(self):
+
+
+
+    def readme(self):
+        """
+        Find the README PackageEntry. Returns (readme_logical_key, readme_package_entry)
+
+        The README is the entry with the logical key 'readme.md' (case-insensitive). Will raise a QuiltException if
```
dimaryaz: Do we really want it to be case-insensitive? I wouldn't bother with the extra complexity.

api/python/quilt3/packages.py
```diff
@@ -225,6 +226,40 @@ def get_cached_path(self):
...
+
+
+    def get_as_string(self, use_cache_if_available=True):
+        """
+        Return the object as a string. Assumes that the file is encoded using utf-8.
+
+        If 'use_cache_if_available'=True, will first try to retrieve the object from cache.
+        """
+        obj_bytes = self.get_bytes(use_cache_if_available=use_cache_if_available)
+        return obj_bytes.decode("utf-8")
```
akarve: do we want `.decode("utf-8", "ignore")`? it's a bit more robust

armandmcqueen: No, I think we generally want to fail loudly if the bytes are not valid utf-8

api/python/quilt3/packages.py
```diff
@@ -225,6 +226,40 @@ def get_cached_path(self):
...
+        if use_cache_if_available:
+            cached_path = self.get_cached_path()
+            if cached_path is not None:
+                return get_bytes(pathlib.Path(cached_path).as_uri())
+
+        physical_key = _to_singleton(self.physical_keys)
+        data = get_bytes(physical_key)
+        return data
+
+    def get_as_json(self, use_cache_if_available=True):
```
akarve: technically this is `get_json_as_dict` because it returns `dict`

api/python/quilt3/packages.py
```diff
@@ -225,6 +226,40 @@ def get_cached_path(self):
...
+        """
+        Return the object as JSON. Assumes that the file is encoded using utf-8.
```
```suggestion
        Returns a JSON file as a `dict`. Assumes that the file is encoded using utf-8.
```

api/python/quilt3/packages.py
```diff
@@ -718,7 +753,23 @@ def get(self, logical_key):
...
+    def readme(self):
+        """
+        Returns the README PackageEntry
+
+        The README is the entry with the logical key 'README.md' (case-sensitive). Will raise a QuiltException if
+        no such entry exists.
+        """
+        if "README.md" not in self:
+            ex_msg = f"This Package is missing a README file. A Quilt recognized README file is a  file named " \
+                     f"'README.md' (case-insensitive)"
```
dimaryaz: "case-insensitive"

PR #1416, opened 2019-11-26
Implement Package.verify

dimaryaz:


api/python/quilt3/packages.py
```diff
@@ -1338,3 +1338,37 @@ def _filter(self, f, include_directories=False):
                 p._set(lk, entity)
 
         return p
+
+    def verify(self, src, extra_files=False):
```
armandmcqueen: I would do something like `extra_files_ok=False`

docs/API Reference/Package.md
```diff
@@ -111,25 +111,6 @@ Generator that traverses all entries in the package tree and returns tuples of (
 with keys in alphabetical order.
 
 
-## Package.load(readable\_file)  {#Package.load}
```
armandmcqueen: Why is this being deleted?

dimaryaz: Hm... Looks like it's because you renamed `load` to `_load`. Want to move the docstring back to `load`?

armandmcqueen: Yeah, I'll open a PR for that

api/python/quilt3/packages.py
```diff
@@ -1338,3 +1338,37 @@ def _filter(self, f, include_directories=False):
...
+            if entry.size != src_size:
+                return False
+            entry_url = src + quote(logical_key)
+            url_list.append(entry_url)
+            size_list.append(src_size)
+
+        if src_dict and not extra_files:
+            return False
+
+        hash_list = calculate_sha256(url_list, size_list)
```
akarve: Ideally we'd have "short-circuit" version of `calculate_sha256` but that can be done in a separate PR. i.e. this will do all hashes even if all but one of them are wrong. Expected-case performance would be much better with short circuit.

PR #1419, opened 2019-11-27
Add ability to only push a subset of data during push

armandmcqueen:
This was tested manually. Integration tests should be added as well, but they will probably come via a different PR.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1051,7 +1051,7 @@ class Package(object):
 
 
     @ApiTelemetry("package.push")
-    def push(self, name, registry=None, dest=None, message=None):
+    def push(self, name, registry=None, dest=None, message=None, selector_fn=lambda logical_key, package_entry: True):
         """
         Copies objects to path, then creates a new package that points to those objects.
         Copies each object in this package to path according to logical key structure,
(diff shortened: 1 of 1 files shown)


api/python/quilt3/packages.py
```diff
@@ -1193,14 +1218,20 @@ def _materialize(self, dest_url):
         file_list = []
         entries = []
         for logical_key, entry in self.walk():
+
+            if selector_fn(logical_key, entry) == False:
+                # print(f"Skipping {logical_key}")
```
```suggestion
```

api/python/quilt3/packages.py
```diff
@@ -1193,14 +1218,20 @@ def _materialize(self, dest_url):
         file_list = []
         entries = []
         for logical_key, entry in self.walk():
+
+            if selector_fn(logical_key, entry) == False:
```
akarve: I don't understand why we set if the selector fails? we don't want `if selector_fn(...):`
```suggestion
            if not selector_fn(logical_key, entry):
```

api/python/quilt3/packages.py
```diff
@@ -1193,14 +1218,20 @@ def _materialize(self, dest_url):
         file_list = []
         entries = []
         for logical_key, entry in self.walk():
+
```
```suggestion
```

api/python/quilt3/packages.py
armandmcqueen: At this point the new package is empty (see line 1215). We need to go through the entries in the old package and add them to the new package. If the selector_fn is false, we add it as is. If the selector_fn is true and the data isn't already in s3, we need to copy it to s3 and then add it to the new package (lines 1234-1235)

akarve: updated suggestion

PR #1421, opened 2019-11-27
Add a command line version of Package.install

dimaryaz:


api/python/quilt3/main.py
```diff
@@ -46,6 +46,40 @@ def create_parser():
     )
     config_p.set_defaults(func=cmd_config)
 
+    shorthelp = "Install a package"
+    install_p = subparsers.add_parser("install", description=shorthelp, help=shorthelp)
+    install_p.add_argument(
+        "name",
+        help="Name of package, in the user/pkg format",
```
akarve: In docs we usually use all caps for template strings.
```suggestion
        help="Name of package, in the USER/PKG format",
```

api/python/quilt3/main.py
```diff
@@ -46,6 +46,40 @@ def create_parser():
...
+        type=str,
+    )
+    install_p.add_argument(
+        "--registry",
+        help="Registry where package is located. Defaults to the default remote registry.",
```
```suggestion
        help="Registry where package is located, usually s3://MY-BUCKET). Defaults to the default remote registry.",
```

armandmcqueen:
```suggestion
        help="Registry where package is located, usually s3://MY-BUCKET.",
```

api/python/quilt3/main.py
```diff
@@ -46,6 +46,40 @@ def create_parser():
...
+        type=str,
+        required=False,
+    )
+    install_p.add_argument(
+        "--dest",
+        help="Local path to download files to.",
+        type=str,
+        required=False,
+    )
+    install_p.add_argument(
```
armandmcqueen: Do we really need this argument? Having both destination and destination registry seems confusing to me.

api/python/quilt3/main.py
```diff
@@ -46,6 +46,40 @@ def create_parser():
...
+        "--top-hash",
+        help="Hash of package to install. Defaults to latest.",
+        type=str,
+        required=False,
+    )
+    install_p.add_argument(
+        "--dest",
```
akarve: What we call `--dest` is probably better called `--prefix`, but not change for now

api/python/quilt3/main.py
dimaryaz: True... But we should fix the Python API

PR #1423, opened 2019-11-28
Increase the number of threads to 10 (default for S3)

dimaryaz:


armandmcqueen:
What impact did this have on download/upload performance?

dimaryaz:
@armandmcqueen: haven't had a chance to test it on EC2; will do it tomorrow. But I think S3's default should be a safe choice.

dimaryaz:
Just did some testing on EC2... Pushing `cv/coco2017` from S3 to S3 goes from ~4MB/s to ~11MB/s.

akarve:
it doesn't seem to help install though? i tried both this has and then one before it.

Before
```
%timeit -n 10 p = quilt3.Package.install("akarve/pytorch-intro", registry="s3://quilt-example")
19.6 s ± 448 ms per loop (mean ± std. dev. of 10 runs, 1 loop each)
```

After
```
%timeit -n 10 p = quilt3.Package.install("akarve/pytorch-intro", registry="s3://quilt-example")
21.7 s ± 745 ms per loop (mean ± std. dev. of 10 runs, 1 loop each)
```

PR #1424, opened 2019-11-28
Use sha256 hashes when caching files

dimaryaz:


api/python/quilt3/packages.py
```diff
@@ -68,15 +67,15 @@ def _delete_local_physical_key(pk):
 
 
 def _filesystem_safe_encode(key):
-    """Encodes the key as hex. This ensures there are no slashes, uppercase/lowercase conflicts, etc."""
-    return binascii.hexlify(key.encode()).decode()
+    """Returns the sha256 of the key. This ensures there are no slashes, uppercase/lowercase conflicts, etc."""
+    return hashlib.sha256(key.encode()).hexdigest()
```
akarve: I thought the choice of hash algorithm would matter but it does not appear to at all (at least on my proc):

```
In [14]: %timeit hashlib.sha256(random_string().encode()).hexdigest()           
108 µs ± 982 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [15]: %timeit hashlib.sha1(random_string().encode()).hexdigest()             
106 µs ± 411 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [16]: %timeit hashlib.md5(random_string().encode()).hexdigest()              
107 µs ± 557 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [17]: %timeit hashlib.sha512(random_string().encode()).hexdigest()           
107 µs ± 564 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [18]: %timeit hashlib.sha256(random_string(1024).encode()).hexdigest()       
844 µs ± 4.32 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [19]: %timeit hashlib.sha1(random_string(1024).encode()).hexdigest()         
852 µs ± 9.44 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [20]: %timeit hashlib.md5(random_string(1024).encode()).hexdigest()          
843 µs ± 7.89 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [21]: %timeit hashlib.sha512(random_string(1024).encode()).hexdigest()       
845 µs ± 4.75 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

api/python/quilt3/packages.py
```diff
@@ -68,15 +67,15 @@ def _delete_local_physical_key(pk):
 
 
 def _filesystem_safe_encode(key):
-    """Encodes the key as hex. This ensures there are no slashes, uppercase/lowercase conflicts, etc."""
-    return binascii.hexlify(key.encode()).decode()
+    """Returns the sha256 of the key. This ensures there are no slashes, uppercase/lowercase conflicts, etc."""
```
```suggestion
    """Returns the sha256 of the key. This ensures there are no slashes, uppercase/lowercase conflicts, 
    avoids `OSError: [Errno 36] File name too long:`, etc."""
```

api/python/quilt3/packages.py
dimaryaz: Ha, yeah, I was gonna use md5, but then did the same test

PR #1426, opened 2019-12-01
Fix Telemetry double counting

armandmcqueen:
## Description

Fixed some places where a single user action could lead to multiple telemetry events.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -456,7 +456,7 @@ class Package(object):
                     f"'build' instead."
                 )
 
-        pkg = cls.browse(name=name, registry=registry, top_hash=top_hash)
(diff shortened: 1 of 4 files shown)


api/python/quilt3/packages.py
```diff
@@ -650,10 +655,6 @@ def _walk_dir_meta(self):
     @classmethod
     @ApiTelemetry("package.load")
     def load(cls, readable_file):
```
dimaryaz: I wonder if this should just be a private method, with no telemetry? I don't think it's particularly useful for a regular user.

armandmcqueen: It's used as a workaround for performance issues in `quilt3`, so I think it is important to measure.

PR #1428, opened 2019-12-01
Drop irrelevant config fields when pulling from catalog

armandmcqueen:
## TODO:

Figure out which other fields we should discard. Below are list of fields:

```
apiGatewayEndpoint: 
binaryApiGatewayEndpoint: 
alwaysRequiresAuth: 
defaultBucket: 
s3Proxy: 
intercomAppId: 
signInRedirect: 
signOutRedirect: 
passwordAuth: 
ssoAuth: 
ssoProviders: 
googleClientId: 
sentryDSN: 
mixpanelToken: 
analyticsBucket:
mode: 
```

### DIFF
--- a/api/python/quilt3/util.py
+++ b/api/python/quilt3/util.py
@@ -62,6 +62,25 @@ registryUrl:
 telemetry_disabled: false
 """.format(BASE_PATH.as_uri() + '/packages')
 
+# When pulling the config from the catalog, don't keep fields that are not relevant for the python client
+CONFIG_FIELD_BLACKLIST = [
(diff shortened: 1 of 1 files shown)


dimaryaz:
It should be pretty simple to grep the Python code for those fields? I think almost none of them are used. I'd make it a whitelist rather than a blacklist.

armandmcqueen:
Added additional fields. I'm going to keep it as a blacklist because this is a purely cosmetic change and a whitelist is more likely to be the cause of significant bugs than a blacklist is.

PR #1431, opened 2019-12-03
Use the cache when installing packages

dimaryaz:
Installing into the same directory will be a no-op.
Installing into a different directory will copy files from the original directory.


akarve:
Really minor, the progress bar shows a `?` when no bytes transferred. Might be nicer to say. "Package already installed."
```
In [4]: quilt3.Package.install("akarve/many-revisions", "s3://quilt-example")   
Copying: 0.00B [00:00, ?B/s]
```

PR #1433, opened 2019-12-06
Preview lambda: Handle case of inconsistent quote characters

akarve:
* [ ] catalog changes from @nl0 to show all relevant `info` text under `info.warnings` and `info.note`


lambdas/preview/index.py
```diff
@@ -173,6 +173,7 @@ def extract_csv(head, separator):
     except pandas.errors.ParserError:
         data = pandas.read_csv(
             io.StringIO('\n'.join(head)),
+            error_bad_lines=False,
```
armandmcqueen: This will cause the preview to silently not accurately match the underlying data. This will be misleading and users will consider it to be a bug. 

I don't think this is the right solution

akarve: I can either accumulate the warnings + errors, and send those to the HTML response, or we can run with `quoting=3` in the second pass. Let me investigate.

akarve: Fixed; I'll file for the catalog to pick the errors up

nl0:
putting it here in case anyone's missing asana updates:
it would be far easier to work on this (catalog changes) when the backend changes are deployed somewhere

PR #1439, opened 2019-12-10
Better package list ux pt. 1: larger hit area (whole tile)

nl0:


catalog/app/containers/Bucket/PackageList.js
```diff
@@ -127,9 +122,38 @@ const usePackageStyles = M.makeStyles((t) => ({
...
+  },
+  updated: {
+    ...t.typography.body2,
+    color: t.palette.text.secondary,
+    position: 'relative',
+  },
 }))
 
-const Package = ({ name, modified, revisions, revisionsTruncated, bucket, views }) => {
+function Package({ name, modified, revisions, revisionsTruncated, bucket, views }) {
```
akarve: why not use arrow functions, there is no `this` reference, so just curious

nl0: oldschool named functions show up properly in the debugger / stack traces (not as "Anonymous") -- that's the only reason

PR #1418, opened 2019-11-27
Catalog: LOCAL mode (use registry credentials for anon users if configured)

nl0:
Use AWS credentials (requested from the registry) for anonymous users when configured (set `"mode": "LOCAL"` in `config.json`)

PR #1442, opened 2019-12-11
Change the S3 proxying logic to fix signatures

dimaryaz:
Right now, the AWS SDK uses the proxy hostname when calculating signatures - causing the real S3 to reject them.

Instead, use the original endpoint when signing the request, and after that, change it to the proxy.


dimaryaz:
@nl0: I don't know if this is the right way to refactor the code. You can just take my changes and use them in your branch, if that's easier.

nl0:
@dimaryaz yep, it's okay. i've ported these changes to the #1418

PR #1445, opened 2019-12-11
Faster JSONL loading for Package.browse()

armandmcqueen:


armandmcqueen:
Performance results on m5.4xl, browsing `cv/coco2017` (after manifest has been cached):
Original code:
```
mean: 21.496473999023436
p50:  21.76580500602722
p90:  22.53216381072998
p90:  24.311231207847595
```
After PR:
```
mean: 9.981648049354554
p50:  9.977904319763184
p90:  10.034168267250061
p90:  10.286012094020844
```

api/python/quilt3/packages.py
```diff
@@ -668,28 +671,41 @@ def load(cls, readable_file):
...
+                    raise PackageException("Duplicate logical key while loading package")
+                subpkg._children[key] = PackageEntry(
+                    obj['physical_keys'],
+                    obj['size'],
+                    obj['hash'],
+                    obj['meta']
+                )
+                tqdm_progress.update(1)
+
+        gc.enable()
```
dimaryaz: Needs to be in a `finally:`

api/python/quilt3/packages.py
```diff
@@ -668,28 +671,41 @@ def load(cls, readable_file):
...
-                obj['physical_keys'],
-                obj['size'],
-                obj['hash'],
-                obj['meta']
-            )
 
+        gc.disable()  # Experiments with COCO (650MB manifest) show disabling GC gives us ~2x performance improvement
+
+        line_count = 0
+        for _ in readable_file:
```
dimaryaz: That seems pretty inefficient. Just count bytes instead? See `OSUtils().open_file_chunk_reader(...)` in `data_transfer.py`

dimaryaz:
Also, this worries me a bit:

```
  ./python/py_defines.h:39:10: fatal error: Python.h: No such file or directory
     39 | #include <Python.h>
        |          ^~~~~~~~~~
  compilation terminated.
  error: command 'gcc' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for ujson
 ```

And if you think that Python headers are easy to install, then how about this one:

```
    building 'ujson' extension
    error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": https://visualstudio.microsoft.com/downloads/
```

File: api/python/quilt3/packages.py (hunk 2 above)
Comment (armandmcqueen) [reply api/python/quilt3/packages.py]: I've found that it is a very small portion of the call so I think it's fine (on my laptop, `cv/coco2017` ~0.7 seconds to scan the line count compared to 70 seconds to download the manifest and 10-11 seconds to parse the jsonl) so I think it's probably fine. The `jsonlines` reader doesn't give a way for us to see how many bytes each parsed json is.

File: api/python/quilt3/packages.py [hunk 3]
```diff
@@ -12,6 +13,12 @@
 import warnings
 
 import jsonlines
+from tqdm import tqdm
+
+try:
+    import ujson as json  # If you have ujson, JSONL parsing is slightly faster, but is not a required dependency
+except ImportError:
+    pass
```
dimaryaz: You should move `import json` here, instead of importing it early and overwriting it

api/python/quilt3/packages.py
dimaryaz: Ok. That's fine I guess.

(But you don't need any support from `jsonlines` - you only need to wrap `readable_file` in order to count bytes.)

api/python/quilt3/packages.py
```diff
@@ -649,28 +656,42 @@ def load(cls, readable_file):
...
+        gc.disable()  # Experiments with COCO (650MB manifest) show disabling GC gives us ~2x performance improvement
+
+        try:
+            line_count = 0
+            for _ in readable_file:
+                line_count += 1
+            readable_file.seek(0)
+
+            reader = jsonlines.Reader(readable_file, loads=json.loads)
+            with tqdm(desc="Loading Manifest", total=line_count) as tqdm_progress:
```
dimaryaz: Add `unit="lines"` or `unit="objects"` or whatever

PR #1450, opened 2019-12-15
Fixed various problems with `quilt3 catalog`

armandmcqueen:
- Fixed bug where new users couldn't use `quilt3 catalog` because they do not have config fields such as `s3Proxy` that are relied on by catalog command.
- Added info about where the catalog will run
- Hid misleading flask output that says " Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)"

### DIFF
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -43,17 +43,14 @@ def cmd_catalog():
     command += ["-p", "3000:80", "quiltdata/catalog"]
     subprocess.Popen(command)
     print("Running a local version of the catalog at http://localhost:3000")
-    orig_stdout = sys.stdout
+
+    # app.run() generates misleading info to stderr ("Running on http://127.0.0.1:5000/") so disable stderr
     orig_stderr = sys.stderr
     try:
         f = open(os.devnull, 'w')
-        sys.stdout = f
         sys.stderr = f
         app.run()
-    except Exception:
-        pass
     finally:
-        sys.stdout = orig_stdout
         sys.stderr = orig_stderr
 
 def cmd_verify(name, registry, top_hash, dir, extra_files_ok):
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -3,11 +3,12 @@ Parses the command-line arguments and runs a command.
 """
 
 import argparse
+import os
 import subprocess
 import sys
 
 from . import api, session
-from .util import get_from_config, QuiltException
+from .util import get_from_config, QuiltException, read_config_from_url, OPEN_DATA_URL
 from .registry import app
 
 def cmd_config(catalog_url):
@@ -27,7 +28,7 @@ def cmd_catalog():
     """
     Run the Quilt catalog locally
     """
-    open_config = api.config()
+    open_config = read_config_from_url(OPEN_DATA_URL)
     command = ["docker", "run", "--rm"]
     env = dict(REGISTRY_URL="http://localhost:5000",
                S3_PROXY_URL=open_config["s3Proxy"],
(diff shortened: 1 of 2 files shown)


armandmcqueen:
It shouldn't work for new users right now:

Get config - https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/main.py#L30
Calls `api.config()` - https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/api.py

It does not have a `catalog_url` and the config YAML doesn't exist so it follows this branch - https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/api.py#L205, which reads the config from the template string - https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/util.py#L31

@kevinemoore maybe you didn't blow away the right config file?

armandmcqueen:
To let people use their own infrastructure, we can use `utils.load_config()` instead of `read_config_from_url`

api/python/quilt3/main.py
```diff
@@ -41,7 +42,15 @@ def cmd_catalog():
         command += ["-e", var]
     command += ["-p", "3000:80", "quiltdata/catalog"]
     subprocess.Popen(command)
-    app.run()
+    print("Running a local version of the catalog at http://localhost:3000")
+
+    # app.run() generates misleading info to stderr ("Running on http://127.0.0.1:5000/") so disable stderr
```
kevinemoore: This seems like overkill. If we hide stderr, we'll hide exceptions, errors, etc. Let's add a line that clarifies it's the credential service running on 5000 then open the catalog URL so users can easily find the right URL.

api/python/quilt3/util.py
```diff
@@ -343,6 +343,20 @@ def get_package_registry(path=None):
...
+    for key, value in new_config.items():
+        if key in CONFIG_FIELD_BLACKLIST:
+            continue
+        config_template[key] = value
+    write_yaml(config_template, CONFIG_PATH, keep_backup=True)
+    return config_template
+
+
+def read_config_from_url(catalog_url):
+    """ Read configuration settings from a Quilt catalog, but do not change the local config.yaml """
```
kevinemoore: Does this refactor still make sense? It could still make the code cleaner.

api/python/quilt3/main.py
```diff
@@ -27,7 +28,7 @@ def cmd_catalog():
     """
     Run the Quilt catalog locally
     """
-    open_config = api.config()
+    open_config = read_config_from_url(OPEN_DATA_URL)
```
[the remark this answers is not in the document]
armandmcqueen: `api.config()` does not default to open.quiltdata.com AFAICT

api/python/quilt3/main.py
```diff
@@ -41,7 +42,15 @@ def cmd_catalog():
         command += ["-e", var]
     command += ["-p", "3000:80", "quiltdata/catalog"]
     subprocess.Popen(command)
-    app.run()
+    print("Running a local version of the catalog at http://localhost:3000")
```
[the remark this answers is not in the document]
armandmcqueen: Yeah, opening the catalog URL would work just as well

api/python/quilt3/main.py
armandmcqueen: We can get rid of it if we launch the catalog automatically. Otherwise I couldn't find a great solution other than suppressing `stderr` or `stdout`. `app.run()` prints out a decent amount of noise so anything that we print gets lost in the noise while "Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)" is the last line printed.

armandmcqueen:
Fixed in other PRs

PR #1452, opened 2019-12-18
Quilt3 catalog fixes

kevinemoore:
## Description
Several small fixes to `quilt3 catalog`:
* open a browser to the local catalog URL
* don't launch the catalog container if it's already running
* run the s3proxy locally


kevinemoore:
> This doesn't work for me. The docker image isn't available:
> 
> ```
> docker: Error response from daemon: pull access denied for quiltdata/s3proxy, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
> ```
`quiltdata/s3proxy` is now public on dockerhub. It should work now.

api/python/quilt3/main.py
```diff
@@ -41,6 +54,31 @@ def cmd_catalog():
...
+    local_catalog_url = "http://localhost:3000"
+    local_s3proxy_url = "http://localhost:5002"
+
+    if not _test_url(local_catalog_url):
+        _launch_local_catalog()
+
+    if not _test_url(local_s3proxy_url):
+        _launch_local_s3proxy()
+
+    open_url(local_catalog_url)
```
armandmcqueen: [Optional]

I would add a check before `open_url` to see if both `_test_url(local_catalog_url)` and `_test_url(local_s3proxy_url)` pass. I had a situation where my docker daemon wasn't running so `quilt3 catalog` opened a "cannot connect" webpage. It was easy to go back to my terminal and see the cause, but IMO it would have been better had the webpage not opened.

kevinemoore: Because the catalog and s3 proxy processes run in the background, we don't know exactly when the processes stabilize or fail. I'd rather avoid adding complex retry logic here, at least for the first version.

PR #1456, opened 2019-12-18
Use a single physical key in PackageEntry

dimaryaz:
Calling `_to_singleton(self.physical_keys)` all over the place is not helping anyone. Just make it a single value.

The manifest is still using a list, so the format is staying the same.


armandmcqueen:
This could break existing user code, right? Maybe we should add a `PackageEntry.physical_keys` property that guarantees backwards compatibility.

PR #1458, opened 2019-12-18
Remove config on install

kevinemoore:
## Description
Running `config` at install time doesn't work with wheel installs so this PR implements auto-configuration to the default stack on the first call to `api.config`.

### DIFF
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -4,9 +4,10 @@ from .data_transfer import copy_file, get_bytes, delete_url, list_url
 from .packages import Package
 from .search_util import search_api
 from .util import (QuiltConfig, QuiltException, CONFIG_PATH,
-                   CONFIG_TEMPLATE, configure_from_url, fix_url,
-                   get_package_registry, load_config, read_yaml,
-                   validate_package_name, write_yaml)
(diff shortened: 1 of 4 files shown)


api/python/quilt3/api.py
```diff
@@ -196,17 +197,17 @@ def _config(*catalog_url, **config_values):
...
+    # Create a custom config with the passed-in values only
+    elif config_values:
+        local_config = load_config()
         config_values = QuiltConfig('', config_values)  # Does some validation/scrubbing
         for key, value in config_values.items():
             local_config[key] = value
         write_yaml(local_config, CONFIG_PATH)
+    # Install the default configuration
+    else:
+        local_config = configure_from_default()
```
armandmcqueen: This line of code is going to be hard to understand long-term - to me, `configure_from_default` does not intuitively suggest that it will read the existing config.yaml if it exists. 

I would either rename `configure_from_default()` to clarify that it will read the current config.yaml if it exists, or expand the comment on line 208.

kevinemoore: Good point. I moved the check for an existing config into api._config. Now `configure_from_default` is equivalent to `configure_from_url`. I would have removed `configure_from_default` all together except for the failure case when we're unable to connect to the default stack.

PR #1461, opened 2019-12-19
Install now shows tophash in output

armandmcqueen:
```
$ quilt3 install test/tiny-pkg --registry="s3://quilt-ml-data" --dest=./tiny-pkg
Copying: 0.00B [00:00, ?B/s]
Successfully installed package 'test/tiny-pkg', tophash=b19dd2 from s3://quilt-ml-data
```

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -471,23 +471,31 @@ class Package(object):
         if len(hash_prefix) == 64:
             top_hash = hash_prefix
         elif 6 <= len(hash_prefix) < 64:
-            all_hashes = list_url(f'{registry}/.quilt/packages/')
-            matches = [h for h, _ in all_hashes if h.startswith(hash_prefix)]
-            if not matches:
+            matching_hashes = [h for h, _
+                               in list_url(f'{registry}/.quilt/packages/')
+                               if h.startswith(hash_prefix)]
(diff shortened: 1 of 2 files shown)


api/python/quilt3/packages.py
```diff
@@ -479,6 +483,12 @@ def resolve_hash(cls, registry, hash_prefix):
             raise QuiltException("Invalid hash: %r" % hash_prefix)
         return top_hash
 
+    @classmethod
+    def shorten_tophash(cls, package_name, registry, top_hash):
+        # TODO: Eventually we should check that we aren't giving a short hash that has a conflict. Not easy to do with
+        #       current .quilt layout
+        return top_hash[:6]
```
akarve: might be safer
```suggestion
        return top_hash[:8]
```

api/python/quilt3/packages.py
```diff
@@ -479,6 +483,12 @@ def resolve_hash(cls, registry, hash_prefix):
             raise QuiltException("Invalid hash: %r" % hash_prefix)
         return top_hash
 
+    @classmethod
+    def shorten_tophash(cls, package_name, registry, top_hash):
```
akarve: did dima not add a method for this when he introduced short hashes?

PR #1462, opened 2019-12-19
Better package list ux pt. 2: new package routes + some more

nl0:
- render tree for the latest revision under the root package route
- show "ROOT /" in the breadcrumbs (maybe smth like "PACKAGE /" would be better? @akarve)
- show revision list under the new `$pkg/revisions` route
- link to the revision list from the revision label (is it obvious enough or do we need a proper "Revisions" link / button somewhere? @akarve)
- fix a crash while rendering VCF with null header
- show `latest` revision

pkg-landing

### DIFF
--- a/catalog/app/components/Preview/renderers/Vcf.js
+++ b/catalog/app/components/Preview/renderers/Vcf.js
@@ -73,7 +73,7 @@ const Vcf = ({ meta, header, data, variants }) => {
               // eslint-disable-next-line react/no-array-index-key
               <TableRow key={`meta:${i}`} className={classes.row}>
                 <TableCell
-                  colSpan={header[0].length}
+                  colSpan={header ? header[0].length : undefined}
                   className={cx(classes.cell, classes.meta)}
                 >
                   {l}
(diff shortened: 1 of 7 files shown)


nl0:
The dropdown:
pkg-revision-dropdown

@akarve 
> i think we can do with ROOT completely the tab sets the context as a package

i've adjusted ui to only show "ROOT" when inside a folder -- it's required to return to the root of the package tree under the *current* revision, whereas the package name links to the package landing page, i.e. root of the tree under the *latest* revision

> seems from a routing perspective that revision instances should hang off of /revisions? e.g. /revisions/tree/1defffsf?

i've tried to mimic github as much as makes sense, so the revision instances are hang off the same `$pkg/tree/$revision/$path` route, whereas `$pkg/revisions` just lists the revisions.

also, there's some confusion with revision "id" (aka timestamp) vs hash -- currently the "id" is used as a primary lookup key for revisions, not the hash, bc it's easier to do this (less requests) given the current `.quilt` layout

PR #1467, opened 2019-12-21
Make bucket unit tests more useful

dimaryaz:
Add a real test for `fetch`.
Don't use full URLs in paths within the bucket.

### DIFF
--- a/api/python/tests/test_bucket.py
+++ b/api/python/tests/test_bucket.py
@@ -1,14 +1,12 @@
-import json
 from unittest.mock import patch
+from io import BytesIO
 import pathlib
 from urllib.parse import urlparse
 
-from botocore.stub import Stubber
 import pandas as pd
 import pytest
-import responses
 
-from quilt3 import Bucket, data_transfer, config
+from quilt3 import Bucket
 from quilt3.util import QuiltException
 
 from .utils import QuiltTestCase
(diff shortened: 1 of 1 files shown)


api/python/tests/test_bucket.py
```diff
@@ -19,16 +17,98 @@ def test_bucket_construct(self):
...
+        with patch('quilt3.data_transfer.s3_transfer_config.max_request_concurrency', 1):
+            bucket.fetch('dir/', './')
+
+        assert pathlib.Path('a').read_bytes() == a_contents
+        assert pathlib.Path('foo/b').read_bytes() == b_contents
+
+        # Fetch a single file.
+
+        self.s3_stubber.add_response(
+            method='head_object',
```
akarve: where does this response actually get consumed?

dimaryaz: Before downloading the file, we "head" it to get the size and decide whether to do a multipart download. That doesn't happen in the directory download because listing the directory gives us sizes.

(That's also what AWS SDK does.)

akarve: that's what i figured, but then above i saw a `.fetch` call with no `head_object` on https://github.com/quiltdata/quilt/pull/1467/files#diff-e2e375e7182ff0261068d685af014694R65

PR #1470, opened 2019-12-26
Quilt3 catalog url

kevinemoore:
## Description
Adds an argument to `quilt3 catalog` to specify an S3 URL to browse to in the local catalog.

### DIFF
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -83,11 +83,8 @@ def cmd_catalog(s3_url=None):
     if not _test_url(local_s3proxy_url):
         _launch_local_s3proxy()
 
-    launch_url = local_catalog_url
-    if s3_url:
-        launch_url += catalog_s3_url(s3_url)
-
-    open_url(launch_url)
+    # open a browser to the local catalog
+    open_url(catalog_s3_url(local_catalog_url, s3_url))
     app.run()
 
 def cmd_verify(name, registry, top_hash, dir, extra_files_ok):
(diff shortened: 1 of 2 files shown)


api/python/quilt3/util.py
```diff
@@ -433,3 +433,11 @@ def validate_key(key):
                 f"Invalid key {key!r}. "
                 f"A package entry key cannot contain a file or folder named '.' or '..' in its path."
             )
+
+def catalog_s3_url(s3_url):
+    bucket, path, version_id = parse_s3_url(urlparse(s3_url))
+    catalog_url = f"/b/{bucket}/tree/{path}"
```
dimaryaz: Need to `quote` the params - `bucket` and `path`

api/python/quilt3/main.py
```diff
@@ -3,15 +3,15 @@
 """
 
 import argparse
-import dns.resolver
 import subprocess
 import sys
 
+import dns.resolver
```
dimaryaz: Why did you move it? It's a built-it package, right?

api/python/quilt3/util.py
```diff
@@ -433,3 +433,11 @@ def validate_key(key):
                 f"Invalid key {key!r}. "
                 f"A package entry key cannot contain a file or folder named '.' or '..' in its path."
             )
+
+def catalog_s3_url(s3_url):
```
dimaryaz: Just a thought - but might be cleaner to pass in the catalog URL, and have this function return the full URL, not just the path and query params.

api/python/quilt3/main.py
kevinemoore: No, I think it's an extra lib, though one we already depended on.

armandmcqueen: We need to add it to `setup.py`

PR #1472, opened 2019-12-27
Let pytest clean up temporary files

dimaryaz:
Instead of maintaining a list of temporary files we create and deleting them after the test is over, we could... let pytest do it for us.

### DIFF
--- a/api/python/tests/integration/test_packages.py
+++ b/api/python/tests/integration/test_packages.py
@@ -19,9 +19,10 @@ from ..utils import QuiltTestCase
 
 
 DATA_DIR = Path(__file__).parent / 'data'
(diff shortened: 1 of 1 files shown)


armandmcqueen:
I don't understand. How does pytest known to clean up the temporary files in the Quilt APP_DIR?

armandmcqueen:
Wait, do our tests completely ignore the `config.yaml`?

dimaryaz:
@armandmcqueen: yeah, all configs are mocked in tests (since tests should run in the same environment, regardless of your actual config): https://github.com/quiltdata/quilt/blob/master/api/python/tests/utils.py#L28

APP_DIR is mocked to be a subdirectory of the test directory: https://github.com/quiltdata/quilt/blob/master/api/python/tests/conftest.py#L31

PR #1474, opened 2019-12-28
Expose linked data (JSON-LD) for packages

nl0:
To be merged after #1462 

### TODO
- [x] package detail
- [x] package list
- [x] package revision list

### Notes

Currently structured data is being looked up under the `['user_meta']['json-ld']` path (as I discovered in some existing packages) in the manifest header (first line). The props used are:
- `name`
- `description`
- `sameAs`
- `identifier`
- `keywords`
- `creator`
- `license`

Moving forward, I think we should optimize and standardize structured data lookup for packages: make `name` (human-readable) and `description` top-level keys, bc they seem to be usable outside of the SEO / JSON-LD cases (btw `description` is required by Google). The rest of the props may live somewhere under the `structured_data` or `linked_data` top-level key, for example (no need for the schema.org boilerplate like `@context` and `@type`, just the actual data is required).

### DIFF
--- a/catalog/app/containers/Bucket/Overview.js
+++ b/catalog/app/containers/Bucket/Overview.js
@@ -1165,7 +1165,7 @@ export default function Overview({
       {AsyncResult.case({
         Ok: () => (
           <M.Box pb={{ xs: 0, sm: 4 }} mx={{ xs: -2, sm: 0 }}>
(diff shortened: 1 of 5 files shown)


nl0:
@akarve @dimaryaz @armandmcqueen @kevinemoore your input on the structured data lookup in the manifests would be really helpful

PR #1477, opened 2019-12-29
Preview: show note & warnings if present

nl0:
Looks like this:
warnings

Tho there was no testing using the real warnings returned by the backend, bc I couldn't find a deployed preview endpoint running the new code and proper broken files causing warnings, so I wasn't sure what format to expect.


akarve:
@nl0 please put all text in a single, yellow block (no separate blue block):
```
Data truncated due to parsing errors:
Warning 1
Warning 2
Warning 3
```
(and we should vertically scroll after 3-5 warnings).

I think Kevin deployed this for you so you should be able to test?

nl0:
warnings2

PR #1478, opened 2019-12-30
Use a PhysicalKey class instead of strings with URLs

dimaryaz:
Parse the URLs only once, when we get them from the user or from the manifest, and store the parsed values.

All user-visible APIs continue to use URLs as input and output.

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -1,11 +1,10 @@
 from codecs import iterdecode
 from concurrent.futures import ThreadPoolExecutor
 import hashlib
-import json
 import pathlib
 import shutil
 from threading import Lock
-from urllib.parse import quote, unquote, urlparse
+from typing import List, Tuple
 import warnings
 
 from botocore import UNSIGNED
@@ -19,7 +18,7 @@ import jsonlines
 from tqdm import tqdm
 
 from .session import create_botocore_session
-from .util import QuiltException, make_s3_url, parse_file_url, parse_s3_url
+from .util import PhysicalKey, QuiltException
 
 
 def create_s3_client():
@@ -62,7 +61,7 @@ def _copy_local_file(ctx, size, src_path, dest_path):
     ctx.progress(size)
     shutil.copymode(src_path, dest_path)
 
-    ctx.done(pathlib.Path(dest_path).as_uri())
+    ctx.done(PhysicalKey(None, dest_path, None))
 
 
 def _upload_file(ctx, size, src_path, dest_bucket, dest_key):
@@ -77,7 +76,7 @@ def _upload_file(ctx, size, src_path, dest_bucket, dest_key):
             )
 
         version_id = resp.get('VersionId')  # Absent in unversioned buckets.
-        ctx.done(make_s3_url(dest_bucket, dest_key, version_id))
+        ctx.done(PhysicalKey(dest_bucket, dest_key, version_id))
     else:
         resp = s3_client.create_multipart_upload(
             Bucket=dest_bucket,
@@ -118,7 +117,7 @@ def _upload_file(ctx, size, src_path, dest_bucket, dest_key):
                     MultipartUpload={"Parts": parts}
                 )
                 version_id = resp.get('VersionId')  # Absent in unversioned buckets.
-                ctx.done(make_s3_url(dest_bucket, dest_key, version_id))
+                ctx.done(PhysicalKey(dest_bucket, dest_key, version_id))
 
         for i, start in enumerate(chunk_offsets):
             end = min(start + chunksize, size)
@@ -148,7 +147,7 @@ def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
             fd.write(chunk)
             ctx.progress(len(chunk))
 
-    ctx.done(pathlib.Path(dest_path).as_uri())
+    ctx.done(PhysicalKey(None, dest_path, None))
 
 
 def _copy_remote_file(ctx, size, src_bucket, src_key, src_version,
@@ -177,7 +176,7 @@ def _copy_remote_file(ctx, size, src_bucket, src_key, src_version,
         resp = s3_client.copy_object(**params)
         ctx.progress(size)
         version_id = resp.get('VersionId')  # Absent in unversioned buckets.
-        ctx.done(make_s3_url(dest_bucket, dest_key, version_id))
+        ctx.done(PhysicalKey(dest_bucket, dest_key, version_id))
     else:
         resp = s3_client.create_multipart_upload(
             Bucket=dest_bucket,
@@ -220,7 +219,7 @@ def _copy_remote_file(ctx, size, src_bucket, src_key, src_version,
                     MultipartUpload={"Parts": parts}
                 )
                 version_id = resp.get('VersionId')  # Absent in unversioned buckets.
-                ctx.done(make_s3_url(dest_bucket, dest_key, version_id))
+                ctx.done(PhysicalKey(dest_bucket, dest_key, version_id))
 
         for i, start in enumerate(chunk_offsets):
             end = min(start + chunksize, size)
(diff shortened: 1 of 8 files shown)


api/python/quilt3/data_transfer.py
```diff
@@ -62,7 +61,7 @@ def _copy_local_file(ctx, size, src_path, dest_path):
     ctx.progress(size)
     shutil.copymode(src_path, dest_path)
 
-    ctx.done(pathlib.Path(dest_path).as_uri())
+    ctx.done(PhysicalKey(None, dest_path, None))
```
armandmcqueen: It might be clearer to use either a constructor method like `fromPath()` or named arguments + default. Having two positional arguments as None is not very intuitive.

api/python/quilt3/data_transfer.py
```diff
@@ -712,7 +653,7 @@ def select(url, query, meta=None, raw=False, **kwargs):
         transparently supported.
 
     Args:
-        url(str):  S3 URL of the object to query
+        src(PhysicalKey):  S3 URL of the object to query
```
armandmcqueen: No longer a URL

api/python/quilt3/packages.py
```diff
@@ -174,7 +174,7 @@ def set(self, path=None, meta=None):
             self
         """
         if path is not None:
-            self.physical_key = fix_url(path)
+            self.physical_key = PhysicalKey.from_url(fix_url(path))
```
armandmcqueen: Should `from_url()` not automatically do `fix_url`?

api/python/quilt3/packages.py
```diff
@@ -852,7 +850,7 @@ def _set_commit_message(self, msg):
 
 
     @ApiTelemetry("package.build")
-    def build(self, name=None, registry=None, message=None):
```
armandmcqueen: This could break user code, right?

api/python/quilt3/packages.py
```diff
@@ -865,35 +863,35 @@ def build(self, name=None, registry=None, message=None):
         Returns:
             The top hash as a string.
         """
-        return self._build(name=name, registry=registry, message=message)
-
-
-    def _build(self, name=None, registry=None, message=None):
-        self._set_commit_message(message)
+        validate_package_name(name)
```
armandmcqueen: This should be a part of `_build`, not a part of `build`

api/python/quilt3/packages.py
```diff
@@ -865,35 +863,35 @@ def build(self, name=None, registry=None, message=None):
...
 
         if registry is None:
             registry = get_from_config('default_local_registry')
         else:
             registry = fix_url(registry)
 
-        registry = registry.rstrip('/')
-        validate_package_name(name)
+        registry_parsed = PhysicalKey.from_url(registry)
```
armandmcqueen: Similarly this should be in `_build` so that `build` and `build` take in the same argument types.

api/python/quilt3/packages.py
```diff
@@ -976,21 +974,18 @@ def _set(self, logical_key, entry=None, meta=None, serialization_location=None,
         validate_key(logical_key)
 
         if entry is None:
-            current_working_dir = pathlib.Path.cwd()
-            logical_key_abs_path = pathlib.Path(logical_key).absolute()
-            entry = logical_key_abs_path.relative_to(current_working_dir)
+            entry = pathlib.Path(logical_key).resolve().as_uri()
+            print('entry1: %r' % entry)
```
armandmcqueen: Should this still be here?

api/python/quilt3/packages.py
```diff
@@ -976,21 +974,18 @@ def _set(self, logical_key, entry=None, meta=None, serialization_location=None,
...
-            parsed_url = urlparse(url)
-            if parsed_url.scheme == 's3':
-                bucket, key, current_version = parse_s3_url(parsed_url)
-                if not current_version and version:
-                    url = make_s3_url(bucket, key, version)
-            entry = PackageEntry(url, size, None, None)
+            if not src.is_local() and src.version_id is None and version_id is not None:
+                src.version_id = version_id
+            entry = PackageEntry(src, size, None, None)
+            print('entry2: %r' % entry)
```
armandmcqueen: Same with this print statement

api/python/quilt3/util.py
```diff
@@ -79,6 +79,116 @@ def __init__(self, message, **kwargs):
...
 
 
+class PhysicalKey(object):
+    __slots__ = ['bucket', 'path', 'version_id']
+
+    def __init__(self, bucket, path, version_id):
+        """
+        """
+        assert bucket is None or isinstance(bucket, str)
+        assert isinstance(path, str)
```
armandmcqueen: We don't accept Path objects?

api/python/quilt3/packages.py
```diff
@@ -466,9 +470,9 @@ def resolve_hash(cls, registry, hash_prefix):
         return top_hash
 
     @classmethod
-    def shorten_tophash(cls, package_name, registry, top_hash):
+    def _shorten_tophash(cls, package_name, registry, top_hash):
```
armandmcqueen: Why is this internal?

api/python/quilt3/data_transfer.py
dimaryaz: Yep. I only added `from_path` later, and missed some of the constructor calls.

I wish Python had a way to make a private constructor (without jumping through a bunch of hoops.)

api/python/quilt3/packages.py
dimaryaz: Not really: in the old code, `name=None` would just cause an exception, so it was already broken.

api/python/quilt3/packages.py
dimaryaz: I'm kind of on the fence here: I feel like user-facing APIs should take user-friendly arguments and do validation - while internal APIs should take whatever is convenient for us (e.g., `PhysicalKey` rather than a `str`) and not bother with validation cause it's unnecessary.

`validate_package_name` is pretty harmless - but passing the registry as a string and parsing it in the caller and in the callee is kind of silly.

In case of `_build` in particular, it doesn't really matter since it's only called by `build`. I'm just wondering what the convention should be in general. Maybe using different names for parameters or type annotations would help.

api/python/quilt3/packages.py
dimaryaz: Oops

api/python/quilt3/packages.py
dimaryaz: Oops2

api/python/quilt3/util.py
dimaryaz: The constructor is meant to be private, and other classes should only call `from_url` or `from_path`. But I don't know of a good way to enforce this.

api/python/quilt3/packages.py
dimaryaz: The problem with `fix_url` is, it's meant to handle ambiguous user input and guess what the user's intention is.

E.g., you can totally have a directory named `file:` or `s3:`, so _technically_, `s3://foo/bar` could mean, `./s3:/foo/bar`. But it's probably not what the user wants.
Also, _technically_, `c:` could be a URL scheme, but `c:/user/blah/` is probably just a path on Windows. And who knows what it should mean on Unix?

Anyways, I wanted to use `fix_url` only at the user-facing API boundaries, and never in the internal code.

armandmcqueen: Okay, makes sense

api/python/quilt3/packages.py
armandmcqueen: Right now `build` should just be `_build` + telemetry. It's not great having two methods that do the same thing, but I think we can reduce complexity if we always know that `_method` does the exact same thing as `method` but without the telemetry. 

Longer term it would really be nice to have that distinction between user-facing APIs and internal APIs. and split up the logic accordingly, but I think we probably want to restructure the code into multiple classes for that (to make it very clear to users what APIs are public and reliable and what APIs are private and prone to breaking changes).

api/python/quilt3/util.py
armandmcqueen: Ah ok. Maybe for now just add a docstring comment indicating that the constructor is supposed to be private and we can try to enforce it via code reviews.

api/python/quilt3/packages.py
dimaryaz: I just wanted it to take the PhysicalKey, instead of doing all the config and URL parsing all over again.

Also, I'm not convinced that shorten_tophash is a good API to have: if collisions could realistically happen and there's a need to check for them, then how do you know a future hash won't cause a collision for the short hash you generate now?

PR #1479, opened 2019-12-30
Fix IPv6 addresses in nginx config

dimaryaz:
nginx gets confused by IPv6 addresses without brackets.


dimaryaz:
Should be fixed by #1482 - but I'd still like to merge this change since it fixes a bug (bad syntax in the nginx config).

PR #1484, opened 2020-01-07
Stripped-down landing page for the local mode

nl0:
local-landing


kevinemoore:
I tested this in local mode and it works fine. I miss the clear calls to action to contribute to Quilt though. Maybe we can add some of that back, e.g., clearer links to Slack and GitHub?

PR #1490, opened 2020-01-09
unpushed amendment for the #1462

nl0:


nl0:
ima merge this trivial pr -- it's an amendment to the last commit from #1462 that i have forgotten to push

PR #1493, opened 2020-01-10
Improve docs: Add CLI and PackageEntry documentation

armandmcqueen:
## Description

* Change the `quilt3 catalog` CLI command to add `--detailed_help` argument, where we keep additional information about the command. This allows us to pipe this info into the markdown documentation file. Don't love this approach, but I couldn't think of a better way to make that info available to the autogenerated documentation while also putting it somewhere where it might stay in sync with the actual behavior of the code.

* Automatically generate CLI documentation from `quilt3 <CMD> --help` as part of documentation building (our CLI has pretty poor start up performance so this takes longer than I would like).

* Added documentation of `PackageEntry` class to `Package.md`. It was not previously documentation anywhere (this is where you will find documentation for things like `pkg["key"].get_bytes()`)

* Regenerated documentation

The CLI docs are not a great long-term solution, but I haven't found a good way to automatically generate documentation from an `argparse` CLI.

### DIFF
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -101,24 +101,24 @@ def cmd_verify(name, registry, top_hash, dir, extra_files_ok):
         return 1
 
 def create_parser():
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(allow_abbrev=False)
 
     subparsers = parser.add_subparsers(metavar="<command>")
     subparsers.required = True
 
     # login
     shorthelp = "Log in to configured Quilt server"
-    login_p = subparsers.add_parser("login", description=shorthelp, help=shorthelp)
+    login_p = subparsers.add_parser("login", description=shorthelp, help=shorthelp, allow_abbrev=False)
     login_p.set_defaults(func=session.login)
 
     # logout
     shorthelp = "Log out of current Quilt server"
-    logout_p = subparsers.add_parser("logout", description=shorthelp, help=shorthelp)
+    logout_p = subparsers.add_parser("logout", description=shorthelp, help=shorthelp, allow_abbrev=False)
     logout_p.set_defaults(func=session.logout)
 
     # config
     shorthelp = "Configure Quilt"
-    config_p = subparsers.add_parser("config", description=shorthelp, help=shorthelp)
+    config_p = subparsers.add_parser("config", description=shorthelp, help=shorthelp, allow_abbrev=False)
     config_p.add_argument(
         "catalog_url",
         help="URL of catalog to config with, or empty string to reset the config",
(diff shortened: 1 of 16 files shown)


api/python/quilt3/main.py
```diff
@@ -74,13 +74,45 @@ def _launch_local_s3proxy():
...
+using AWS credentials in `boto3`, see the AWS documentation: 
+https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html
+
+#### Previewing files in S3
+The Quilt catalog allows users to preview files in S3 without
+downloading. It relies on a API Gateway and AWS Lambda to generate
+certain previews in the cloud. The catalog launched by `quilt3
+catalog` sends preview requests to https://open.quiltdata.com. Preview
+requests contain short-lived signed URLs generated using your AWS
+credentials. It is not recommended to use `quilt3 catalog` to browse
```
```suggestion
credentials. Data are encrypted in transit and no data is retained by Quilt.
Nevertheless, it is recommended that you use `quilt3 catalog` only for public data.
```

api/python/quilt3/main.py
```diff
@@ -74,13 +74,45 @@ def _launch_local_s3proxy():
...
+highly sensitive files. We strongly encourage users with highly
```
```suggestion
We strongly encourage users with
```

api/python/quilt3/main.py
```diff
@@ -74,13 +74,45 @@ def _launch_local_s3proxy():
...
+sensitive information in S3 to run a private Quilt deployment. See
```
```suggestion
sensitive data in S3 to run a private Quilt deployment. See
```

api/python/quilt3/main.py
```diff
@@ -129,14 +161,20 @@ def create_parser():
...
-        nargs="?"
+    catalog_p = subparsers.add_parser("catalog", description=shorthelp, help=shorthelp, allow_abbrev=False)
+    catalog_p.add_argument(
+            "s3_url",
+            help="S3 URL to browse in local catalog",
+            type=str,
+            nargs="?"
+    )
+    catalog_p.add_argument(
+            "--detailed_help",
```
akarve: Just an idea; not religious. Or `--man-page` idk.
```suggestion
            "--verbose",
```

docs/API Reference/Package.md
```diff
@@ -254,6 +273,28 @@ __Arguments__
 * __dest__:  where to copy the objects in the package
 * __registry__:  registry where to create the new package
 * __message__:  the commit message for the new package
+* __selector_fn__:  A filter function that determines which package entries should be pushed. The function takes
```
akarve:
```suggestion
* __selector_fn__:  An optional lambda function that determines which package entries should be copied to S3,
and which should not. The function takes
```

PR #1496, opened 2020-01-10
Added CLI command `quilt3 disable-telemetry`

armandmcqueen:


api/python/quilt3/main.py
```diff
@@ -176,6 +181,10 @@ def create_parser():
     )
     catalog_p.set_defaults(func=cmd_catalog)
 
+    # disable-telemetry
+    shorthelp = "Configure quilt to not send anonymous usage metrics"
```
```suggestion
    shorthelp = "Disable anonymous usage metrics"
```

PR #1498, opened 2020-01-11
Check that hash belongs to the correct package. Use nanosecond timestamps

dimaryaz:
When resolving hash, find all valid hashes for the given package. This obviously adds latency. Check "latest" first, as a shortcut.
Use nanoseconds instead of seconds when naming versions - otherwise, we end up with collisions. (The old version gets overwritten - and that breaks the new hash resolver.)


armandmcqueen:
I think this adds too much latency for the benefit it gives. I wouldn't merge it

PR #1501, opened 2020-01-14
Ignore errors when looking up objects in the cache

dimaryaz:
Cache file can get corrupted if the `set` call is interrupted, etc.


armandmcqueen:
Should we clean up the cache file in that case?

LGTM

dimaryaz:
Probably... but then cleanup could cause an exception, so need to handle that. Probably safer to implement cleanup elsewhere.

PR #1503, opened 2020-01-15
Fully support public buckets

armandmcqueen:
When downloading data, try both a normal s3_client and an unsigned s3_client to determine which should be used.

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -29,7 +29,6 @@ from .util import PhysicalKey, QuiltException
 class S3Api(Enum):
     GET_OBJECT = "GET_OBJECT"
     HEAD_OBJECT = "HEAD_OBJECT"
-    DELETE_OBJECT = "DELETE_OBJECT"
     LIST_OBJECT_VERSIONS = "LIST_OBJECT_VERSIONS"
     LIST_OBJECTS_V2 = "LIST_OBJECTS_V2"
 
@@ -94,10 +93,17 @@ class S3ClientProvider:
 
 
 
-    def find_correct_client(self, api_type, bucket, param_dict, check_fn):
+    def find_correct_client(self, api_type, bucket, param_dict):
         if self.client_type_known(api_type, bucket):
             return self.get_correct_client(api_type, bucket)
         else:
+            check_fn_mapper = {
+                S3Api.GET_OBJECT: check_get_object_works_for_client,
+                S3Api.HEAD_OBJECT: check_head_object_works_for_client,
+                S3Api.LIST_OBJECTS_V2: check_list_objects_v2_works_for_client,
+                S3Api.LIST_OBJECT_VERSIONS: check_list_object_versions_works_for_client
+            }
+            check_fn = check_fn_mapper[api_type]
             if check_fn(self.standard_client, param_dict):
                 self.set_cache(api_type, bucket, use_unsigned=False)
                 assert self.client_type_known(api_type, bucket), f"{self.key(api_type, bucket)}: {self._use_unsigned_client}"
@@ -165,7 +171,6 @@ def check_head_object_works_for_client(s3_client, params):
 
 
 
-# s3_client_provider = S3ClientProvider()
 s3_transfer_config = TransferConfig()
 
 # When uploading files at least this size, compare the ETags first and skip the upload if they're equal;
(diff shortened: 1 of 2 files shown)


api/python/quilt3/data_transfer.py
```diff
@@ -22,29 +23,145 @@
...
+            raise RuntimeError("get_correct_client was called, but the correct client type is not known. Only call "
+                               "get_correct_client() after checking if client_type_known()")
+
+        if self.should_use_unsigned_client(action, bucket):
+            return self.unsigned_client
+        else:
+            return self.standard_client
+
+    def key(self, action: S3Api, bucket: str):
+        return f"{action}/{bucket}"
```
dimaryaz: Why not just use a tuple `(action, bucket)`?

api/python/quilt3/data_transfer.py
```diff
@@ -277,6 +400,8 @@ def _copy_file_list_internal(s3_client, file_list, callback):
 
     stopped = False
 
+    s3_client_provider = S3ClientProvider()  # Share provider accross threads to reduce redundant public bucket checks
```
dimaryaz: Then it needs to be thread-safe?

api/python/quilt3/data_transfer.py
```diff
@@ -22,29 +23,145 @@
...
-        event_name, signal_not_transferring,
-        unique_id='datatransfer-not-transferring')
-    s3_client.meta.events.register_last(
-        event_name, signal_transferring,
-        unique_id='datatransfer-transferring')
 
-    return s3_client
+class S3Api(Enum):
+    GET_OBJECT = "GET_OBJECT"
+    HEAD_OBJECT = "HEAD_OBJECT"
```
dimaryaz: There's no point making a distinction between `HEAD` and `GET` - it's a single S3 permission.

But even `GET` vs `LIST` is tricky - see the 403/404 comment.

api/python/quilt3/data_transfer.py
armandmcqueen: It is threadsafe for this usage AFAICT. Logically it is a cache representation of a static dataset, so the concurrency requirements are very simple

api/python/quilt3/data_transfer.py
```diff
@@ -22,29 +23,145 @@
...
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "AccessDenied":
+            return False
+
+def check_get_object_works_for_client(s3_client, params):
+    try:
+        s3_client.get_object(**params, Range='bytes=0-0')  # Only get the first byte so it is fast
+        return True
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "AccessDenied":
```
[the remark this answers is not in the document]
armandmcqueen: I think this doesn't matter? In that case, the check_fn with the normal_credentials will fail, the check_fn with the unsigned_credentials will fail and the user will get an S3 AccessDenied error as they should

api/python/quilt3/data_transfer.py
```diff
@@ -22,29 +23,145 @@
...
+    try:
+        s3_client.list_objects_v2(**params, MaxKeys=1)  # Make this as fast as possible
+        return True
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "AccessDenied":
+            return False
+
+def check_get_object_works_for_client(s3_client, params):
+    try:
+        s3_client.get_object(**params, Range='bytes=0-0')  # Only get the first byte so it is fast
```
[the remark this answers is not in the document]
armandmcqueen: Fixed, thanks

akarve:
@armandmcqueen encountering a crash:
```
(pytorch_p36) sh-4.2$ quilt3 install quilt/altair \
>     --registry s3://quilt-example \
>     --dest ./my/directory/
  File "/home/ec2-user/anaconda3/envs/pytorch_p36/bin/quilt3", line 11, in <module>
    load_entry_point('quilt3', 'console_scripts', 'quilt3')()
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/main.py", line 312, in main
    return func(**kwargs)
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/telemetry.py", line 138, in decorated
    results = func(*args, **kwargs)
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/packages.py", line 439, in install
    pkg = cls._browse(name=name, registry=registry, top_hash=top_hash)
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/packages.py", line 514, in _browse
    top_hash = get_bytes(top_hash_file).decode('utf-8').strip()
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/data_transfer.py", line 705, in get_bytes
    s3_client = S3ClientProvider().find_correct_client(S3Api.GET_OBJECT, src.bucket, params)
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/data_transfer.py", line 104, in find_correct_client
    if check_fn(self.standard_client, param_dict):
  File "/home/ec2-user/SageMaker/quilt/api/python/quilt3/data_transfer.py", line 158, in check_get_object_works_for_client
    s3_client.head_object(head_args)  # HEAD/GET share perms, but HEAD always fast
  File "/home/ec2-user/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/botocore/client.py", line 355, in _api_call
    "%s() only accepts keyword arguments." % py_operation_name)
TypeError: head_object() only accepts keyword arguments.
```

PR #1504, opened 2020-01-15
Use hosted s3proxy for local mode catalog

kevinemoore:
## Description

Apparently due a bug in Docker on Mac, `quilt3 catalog` had been crashing on calls through the local s3 proxy. This change points local catalogs to the hosted s3proxy in users' config. The default is the s3proxy in the open.quiltdata.com stack.

### DIFF
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -69,7 +69,9 @@ def _launch_local_s3proxy():
     # Workaround for a Docker-for-Mac bug in which the container
     # ends up with a different DNS server than the host.
     # Workaround #2: use only IPv4 addresses.
-    # TODO: reintroduce this once Docker-for-Mac DNS works reliably
+    # Note: leaving this code in though it isn't called so that it
+    # can be reintroduced once Docker-for-Mac DNS works reliably.
+    # TODO: switch back to this local s3proxy or remove this function
     if sys.platform == 'darwin':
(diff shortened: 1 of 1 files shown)


api/python/quilt3/main.py
```diff
@@ -69,6 +69,7 @@ def _launch_local_s3proxy():
     # Workaround for a Docker-for-Mac bug in which the container
     # ends up with a different DNS server than the host.
     # Workaround #2: use only IPv4 addresses.
+    # TODO: reintroduce this once Docker-for-Mac DNS works reliably
```
armandmcqueen: I don't understand this TODO

api/python/quilt3/main.py
```diff
@@ -147,23 +142,19 @@ def cmd_catalog(navigation_target=None, detailed_help=False):
     while True:
         if time.time() - start_time > failure_timeout_secs:
             catalog_failed = _test_url(local_catalog_url)
-            s3proxy_failed = _test_url(local_s3proxy_url)
```
armandmcqueen: If we only have the one catalog running, this logic can be simplified:
```
while True:
    if _test_url(local_catalog_url):
        break
    else:
        if time.time() - start_time > failure_timeout_secs:
            raise QuiltException("The local catalog container did not successfully launch")
        time.sleep(pool_interval_secs)
```

PR #1506, opened 2020-01-15
Make progress bar messages more specific

dimaryaz:
Display "Downloading Manifest" instead of "Copying", which is very misleading.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -526,7 +526,7 @@ class Package(object):
                 # Copy to a temporary file first, to make sure we don't cache a truncated file
                 # if the download gets interrupted.
                 tmp_path = local_pkg_manifest.with_suffix('.tmp')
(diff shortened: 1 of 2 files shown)


api/python/quilt3/packages.py
```diff
@@ -526,7 +526,7 @@ def _browse(cls, name, registry, top_hash):
                 # Copy to a temporary file first, to make sure we don't cache a truncated file
                 # if the download gets interrupted.
                 tmp_path = local_pkg_manifest.with_suffix('.tmp')
-                copy_file(pkg_manifest, PhysicalKey.from_path(tmp_path))
+                copy_file(pkg_manifest, PhysicalKey.from_path(tmp_path), message="Downloading Manifest")
```
akarve: "Downloading manifest" will give us the same capitalization for all messages (only capitalize first word)

dimaryaz: Lol... I was just copying https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/packages.py#L686
(But I do like lower-case manifest better.)

PR #1508, opened 2020-01-16
WIP feature/dask-support

JacksonMaxfield:
## Description
Dask clients and concurrent futures executors have the same API of submit and map (Specifically designed to be that way). As more and more people are starting to use dask in data engineering and modeling workloads, the problem that arises is it is common for single dask workers to only have a few cores / threads. Instead of utilizing a single workers available resources we can call out to the scheduler that we want to distribute our jobs to the available cluster.

I wasn't able to fully get it working but would love eyes to see if I am missing something.

## TODO

Use this section for work-in-progress pull requests. If you're using this section,
please tag your PR "WIP". 

- [ ] Unit Tests
- [x] Documentation
- [ ] Fix conflict with #2

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -1,29 +1,26 @@
-from collections import deque
-from codecs import iterdecode
-from concurrent.futures import ThreadPoolExecutor
-from enum import Enum
 import hashlib
 import pathlib
 import shutil
-from threading import Lock
-from typing import List, Tuple
 import warnings
+from codecs import iterdecode
+from collections import deque
+from concurrent.futures import ThreadPoolExecutor
+from enum import Enum
+from threading import Lock
+from typing import List
 
+import boto3
+import jsonlines
+from boto3.s3.transfer import TransferConfig
 from botocore import UNSIGNED
 from botocore.client import Config
 from botocore.exceptions import ClientError
-import boto3
-from boto3.s3.transfer import TransferConfig
-from s3transfer.utils import ChunksizeAdjuster, OSUtils, signal_transferring, signal_not_transferring
-
-import jsonlines
+from s3transfer.utils import (ChunksizeAdjuster, OSUtils,
+                              signal_not_transferring, signal_transferring)
 from tqdm import tqdm
 
 from .session import create_botocore_session
 from .util import PhysicalKey, QuiltException
-
-
-
 
 
 class S3Api(Enum):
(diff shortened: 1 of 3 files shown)


api/python/quilt3/data_transfer.py
```diff
@@ -1,31 +1,28 @@
...
 from botocore import UNSIGNED
 from botocore.client import Config
 from botocore.exceptions import ClientError
-import boto3
-from boto3.s3.transfer import TransferConfig
-from s3transfer.utils import ChunksizeAdjuster, OSUtils, signal_transferring, signal_not_transferring
-
-import jsonlines
+from s3transfer.utils import (ChunksizeAdjuster, OSUtils,
+                              signal_not_transferring, signal_transferring)
```
JacksonMaxfield: My editor auto-sorted the imports 🤷‍♀

api/python/setup.py
```diff
@@ -1,13 +1,13 @@
 import os
```
JacksonMaxfield: I noticed in the contributing documentation that you had an invalid pip install command so I rearranged how the deps are setup and changed the docs to be more accurate. (At least from what I can tell)

api/python/quilt3/data_transfer.py
```diff
@@ -444,8 +473,10 @@ def _copy_file_list_internal(file_list, message, callback):
 
     s3_client_provider = S3ClientProvider()  # Share provider across threads to reduce redundant public bucket checks
 
-    with tqdm(desc=message, total=total_size, unit='B', unit_scale=True) as progress, \
-         ThreadPoolExecutor(s3_transfer_config.max_request_concurrency) as executor:
+    # Get the appropriate executor
+    executor = _get_executor(s3_transfer_config.max_request_concurrency)
```
dimaryaz: You should still use `with` here, to ensure the executor gets closed. You can probably make it work using `@contextmanager`.

api/python/quilt3/data_transfer.py
```diff
@@ -512,6 +543,9 @@ def done_callback(value):
 
     assert all(results)
 
+    # Close the executor
+    _close_executor(executor)
```
dimaryaz: Should be done before the `assert`. (Though bringing back `with` will make this go away, anyway.)

api/python/quilt3/data_transfer.py
```diff
@@ -811,8 +845,9 @@ def _process_url(src, size):
                         progress.update(len(chunk))
             return hash_obj.hexdigest()
 
-        with ThreadPoolExecutor() as executor:
-            results = executor.map(_process_url, src_list, sizes)
+        executor = _get_executor()
```
dimaryaz: Same here

api/python/setup.py
```diff
@@ -70,24 +91,12 @@ def run(self):
...
-            'responses',
-            'tox',
-            'detox',
-            'tox-pytest-summary',
-        ],
+        'pyarrow': pyarrow_requires,
+        'tests': test_requires,
+        'all': [
+            *pyarrow_requires,
+            *test_requires,
```
dimaryaz: Technically, `tests` includes all of `pyarrow` dependencies, so this isn't really needed... But if you think it's cleaner this way, then sure.

api/python/quilt3/data_transfer.py
```diff
@@ -429,6 +426,38 @@ def __init__(self, s3_client_provider, progress, done, run):
...
+    # (Specifically designed to be that way)
+    # As more and more people are starting to use dask in data engineering and modeling
+    # workloads, the problem that arises is it is common for single dask workers to
+    # only have a few cores / threads.
+    # Instead of utilizing a single workers available resources we can call out to
+    # the scheduler that we want to distribute our jobs to the available cluster.
+    # https://docs.dask.org/en/latest/futures.html#start-dask-client
+
+    # Check if this function is running in a dask worker
+    try:
```
dimaryaz: Not sure if picking up dask automatically is a good idea. Probably safer to use a config option?

Also, what about unit tests?

api/python/quilt3/data_transfer.py
JacksonMaxfield: Good idea!

api/python/quilt3/data_transfer.py
JacksonMaxfield: I couldn't get the unit tests working with this change which is why I marked it as a WIP PR. as soon as I get them working for the base case of threadpoolexecutor I will add tests for dask worker case.

What about an env variable?
`QUILT_USE_DASK=True`?

JacksonMaxfield:
Closing PR for now. Will try to get this working again at a later time point as this is still incredibly valuable.

PR #1543, opened 2020-02-21
Index parquet metadata (incl schema)

akarve:


lambdas/es/indexer/index.py
```diff
@@ -87,7 +87,8 @@ def get_contents(bucket, key, ext, *, etag, version_id, s3_client, size):
                 s3_client=s3_client,
                 version_id=version_id
             )
-            content = extract_parquet(get_bytes(obj["Body"], compression), as_html=False)[0]
+            body, info = extract_parquet(get_bytes(obj["Body"], compression), as_html=False)
+            content = trim_to_bytes(f"{str(info)}\n{body}", ELASTIC_LIMIT_BYTES)
```
kevinemoore: Ideally, we'd fold the schema into an expanded system_meta, but this is a good first step.

PR #1549, opened 2020-02-27
Install sub package

sir-sigurd:
## Description

This allows to install only the part of the package. Example:

We have package `foo/bar`:
```
(remote Package)
 └─requirements.txt
 └─scripts/
   └─build.py
   └─demolish.py
```

```python
In [17]: import os, tempfile, quilt3                 

In [18]: with tempfile.TemporaryDirectory() as dest: 
    ...:     quilt3.Package.install('foo/bar/scripts', dest=dest) 
    ...:     print([os.path.relpath(os.path.join(top, f), dest) for top, dirs, files in os.walk(dest) for f in files]) 

['demolish.py', 'build.py', 'more_scripts/test.py']
```

```python
In [20]: with tempfile.TemporaryDirectory() as dest: 
    ...:     quilt3.Package.install('foo/bar/scripts/build.py', dest=dest) 
    ...:     print([os.path.relpath(os.path.join(top, f), dest) for top, dirs, files in os.walk(dest) for f in files]) 

['build.py']
```

### DIFF
--- a/api/python/quilt3/util.py
+++ b/api/python/quilt3/util.py
@@ -25,6 +25,7 @@ CONFIG_PATH = BASE_PATH / 'config.yml'
 OPEN_DATA_URL = "https://open.quiltdata.com"
 
 PACKAGE_NAME_FORMAT = r"[\w-]+/[\w-]+$"
+SUBPACKAGE_NAME_FORMAT = r"([\w-]+/[\w-]+)(?:/(.+))?$"
 
 ## CONFIG_TEMPLATE
 # Must contain every permitted config key, as well as their default values (which can be 'null'/None).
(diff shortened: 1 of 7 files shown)


api/python/quilt3/util.py
```diff
@@ -25,6 +25,7 @@
 OPEN_DATA_URL = "https://open.quiltdata.com"
 
 PACKAGE_NAME_FORMAT = r"[\w-]+/[\w-]+$"
+SUBPACKAGE_NAME_FORMAT = r"([\w-]+/[\w-]+)(?:/(.+))?$"
```
kevinemoore: Minor suggestion, but wouldn't it be cleaner to simply replace PACKAGE_NAME_FORMAT to all the optional path, then check that the path is empty in validate_package_name? We might also want a helper function to pull out the package name and sub-package path.

PR #1561, opened 2020-03-15
case studies: superimpose logos

nl0:
I'm not quite happy with placement, but I couldn't come up with anything better

case-studies-w-logos-2
case-studies-w-logos


nl0:
> Thanks for doing this. Can you try a different approach? Instead of embossing the logo, put it in front of the
> 
> #### text. So
> ```
> <logo/><h4>The future of data collaboration in S3</h4>
> 
> OR
> 
> <logo/>Read More
> 
> I realize the logo will be smaller but it will be more natural. In the first example maybe even a column layout will look good (logo in first column) and that will handle wrapping as well.
> ```

ok, i'll try this approach. i understood "superimpose" literally, so i did the watermark-ish thing

nl0:
@akarve wdyt about this?
case-studies-lg
case-studies-sm

akarve:
Ship it :)

Issue #1559, opened 2020-03-11
Please provide option to not see progress bars on download/hash/etc.


gdesmarais-ctx:
tqdm is nice on a console, but messes up log files, especially CloudWatch.  An option to turn it off would be helpful, or even an automatic detection when running in AWS.  I happen to use it in Batch, so I look for `os.environ.get("AWS_BATCH_JOB_ID")`.

akarve:
How would you like that option to maniefst, as an environment variable (e.g. `QUILT_USE_TQDM=''`)? I'm a bit leary of automagic detection of batch, though we'll consider it.

gdesmarais-ctx:
Sure, an env var is reasonable, and I totally understand the batch issue. The only other place I can think of is the api, which would get ugly and is probably not nice oo.

akarve:
Fixed in the upcoming PyPI release.

akarve:
Docs here https://docs.quiltdata.com/references/faq#can-i-turn-off-progress-bars-for-log-files

PR #1564, opened 2020-03-17
Disable or enable tqdm with env

cosmic-byte:
## Description
Disable or Enable tqdm with an environment variable.
eg:
```
QUILT_USE_TQDM=False
QUILT_USE_TQDM=True
```
#### Related issue
#1559

### DIFF
--- a/api/python/pytest.ini
+++ b/api/python/pytest.ini
@@ -7,4 +7,3 @@ filterwarnings =
 
 env =
     QUILT_DISABLE_USAGE_METRICS=True
-    QUILT_MINIMIZE_STDOUT=False
--- a/api/python/pytest.ini
+++ b/api/python/pytest.ini
@@ -7,4 +7,4 @@ filterwarnings =
 
 env =
     QUILT_DISABLE_USAGE_METRICS=True
-    QUILT_USE_TQDM=False
+    QUILT_MINIMIZE_STDOUT=False
--- a/api/python/pytest.ini
+++ b/api/python/pytest.ini
@@ -7,3 +7,4 @@ filterwarnings =
 
 env =
     QUILT_DISABLE_USAGE_METRICS=True
+    QUILT_USE_TQDM=True
--- a/api/python/pytest.ini
+++ b/api/python/pytest.ini
@@ -7,4 +7,4 @@ filterwarnings =
 
 env =
     QUILT_DISABLE_USAGE_METRICS=True
-    QUILT_USE_TQDM=True
+    QUILT_USE_TQDM=False
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -1,4 +1,3 @@
-import os
 from collections import defaultdict, deque
 from codecs import iterdecode
 from concurrent.futures import ThreadPoolExecutor
@@ -23,11 +22,9 @@ from tenacity import retry, stop_after_attempt, wait_exponential
 from tqdm import tqdm
 
 from .session import create_botocore_session
-from .util import PhysicalKey, QuiltException
-
+from .util import PhysicalKey, QuiltException, DISABLE_TQDM
 
 MAX_COPY_FILE_LIST_RETRIES = 3
-USE_TQDM = os.getenv('QUILT_USE_TQDM', '').lower() == 'true'
 
 
 class S3Api(Enum):
@@ -458,13 +455,14 @@ def _copy_file_list_internal(file_list, results, message, callback):
 
     s3_client_provider = S3ClientProvider()  # Share provider across threads to reduce redundant public bucket checks
 
-    def copy_internal(stp, tqdm_progress=None):
+    with tqdm(desc=message, total=total_size, unit='B', unit_scale=True, disable=DISABLE_TQDM) as progress, \
+         ThreadPoolExecutor(s3_transfer_config.max_request_concurrency) as executor:
+
         def progress_callback(bytes_transferred):
             if stopped:
                 raise Exception("Interrupted")
-            if tqdm_progress:
-                with lock:
-                    tqdm_progress.update(bytes_transferred)
+            with lock:
+                progress.update(bytes_transferred)
 
         def run_task(idx, func, *args):
             future = executor.submit(func, *args)
@@ -474,7 +472,7 @@ def _copy_file_list_internal(file_list, results, message, callback):
                 idx_to_futures[idx].append(future)
 
         def worker(idx, src, dest, size):
-            if stp:
+            if stopped:
                 raise Exception("Interrupted")
 
             def done_callback(value):
@@ -535,15 +533,7 @@ def _copy_file_list_internal(file_list, results, message, callback):
                         futures_to_cancel.clear()
         finally:
             # Make sure all tasks exit quickly if the main thread exits before they're done.
-            stp = True
-
-    if USE_TQDM:
-        with tqdm(desc=message, total=total_size, unit='B', unit_scale=True) as progress, \
-             ThreadPoolExecutor(s3_transfer_config.max_request_concurrency) as executor:
-            copy_internal(stopped, progress)
-    else:
-        with ThreadPoolExecutor(s3_transfer_config.max_request_concurrency) as executor:
-            copy_internal(stopped)
+            stopped = True
 
     if not all(results):
         raise QuiltException("Unable to copy some files.")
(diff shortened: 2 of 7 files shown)


api/python/quilt3/data_transfer.py
```diff
@@ -26,6 +27,7 @@
 
 
 MAX_COPY_FILE_LIST_RETRIES = 3
+use_tqdm = os.getenv('QUILT_USE_TQDM').lower() == 'true'
```
akarve: 1. Use all caps for constants
2. `.lower()` will fail if `None` is returned :(
3. `== 'true'` is super fragile; just use the truthiness of the string
```suggestion
USE_TQDM = os.getenv('QUILT_USE_TQDM', '').lower()
```

api/python/quilt3/data_transfer.py
```diff
@@ -473,7 +474,7 @@ def run_task(idx, func, *args):
                 idx_to_futures[idx].append(future)
 
         def worker(idx, src, dest, size):
-            if stopped:
```
akarve: Why change the name of this variable?

api/python/quilt3/packages.py
```diff
@@ -311,6 +311,8 @@ def physical_keys(self):
 class Package(object):
     """ In-memory representation of a package """
 
+    use_tqdm = os.getenv('QUILT_USE_TQDM').lower() == 'true'
```
akarve: Why not just import this from `data_transfer`? As a rule copying the same code twice is not a good idea. Also: please run `pylint` on all files in this PR.

api/python/quilt3/data_transfer.py
cosmic-byte: > 3. `== 'true'` is super fragile; just use the truthiness of the string

In this case, the truthiness of the string might not work since the string is passed as an env variable and can be any value. I had to explicitly compare to "true" to mitigate that.

Screenshot 2020-03-20 at 8 03 22 AM

cosmic-byte: [Here's](https://github.com/quiltdata/quilt/blob/22c0c18fc86ce1f7b002c63fdf373db76fa0a4ce/api/python/quilt3/telemetry.py#L66) a similar usage for `QUILT_DISABLE_USAGE_METRICS`

api/python/quilt3/data_transfer.py
cosmic-byte: shadows the outer variable `stopped` since I moved the operation inside an internal function `copy_internal` with `stopped` as a parameter.

api/python/quilt3/data_transfer.py
```diff
@@ -534,7 +535,15 @@ def done_callback(value):
                         futures_to_cancel.clear()
         finally:
             # Make sure all tasks exit quickly if the main thread exits before they're done.
-            stopped = True
+            stp = True
+
+    if use_tqdm:
+        with tqdm(desc=message, total=total_size, unit='B', unit_scale=True) as progress, \
+             ThreadPoolExecutor(s3_transfer_config.max_request_concurrency) as executor:
+            copy_internal(stopped, progress)
```
cosmic-byte: @akarve this is where I'm passing the original outer `stopped` variable.

api/python/quilt3/data_transfer.py
akarve: You can still check `USE_TQDM == "whatever"` but you have to provide a default value or else that Code will crash for undefined varaibles.

akarve: So for instance your code `os.getenv('QUILT_USE_TQDM').lower()` will crash if the environment variable is undefined.

cosmic-byte: Yes, sure. I will update that change request.

api/python/tests/integration/test_packages.py
```diff
@@ -532,7 +534,12 @@ def test_load_into_quilt(self):
 
         with patch('time.time', return_value=1234567891), \
              patch('quilt3.data_transfer.s3_transfer_config.max_request_concurrency', 1):
-            remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            stdout = io.StringIO()
+
+            # temporarily redirect stderr to capture warnings (usually errors)
+            with redirect_stderr(stdout):
+                remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            assert '100%|##' not in stdout.getvalue()
```
akarve: this is a bit fragile. what if it gets to 99% and fails. check a bit more carefully for the absence of tqdm data.

api/python/quilt3/data_transfer.py
```diff
@@ -26,6 +27,7 @@
 
 
 MAX_COPY_FILE_LIST_RETRIES = 3
+USE_TQDM = os.getenv('QUILT_USE_TQDM', '').lower() == 'true'
```
sir-sigurd: `util.py` might be a better place for this.

api/python/quilt3/util.py
```diff
@@ -25,6 +25,7 @@
 OPEN_DATA_URL = "https://open.quiltdata.com"
 
 PACKAGE_NAME_FORMAT = r"([\w-]+/[\w-]+)(?:/(.+))?$"
+DISABLE_TQDM = os.getenv('QUILT_USE_TQDM', '').lower() != 'true'
```
sir-sigurd: @akarve, this disables `tqdm` by default, is it intended?
Also name `QUILT_USE_TQDM` might be too specific, IMHO `QUILT_INTERACTIVE` or `QUILT_PROGRESS_BARS` or something might be better.

api/python/tests/test_data_transfer.py
```diff
@@ -431,9 +433,17 @@ def test_multipart_copy(self):
...
+
+            # temporarily redirect stderr to capture warnings (usually errors)
+            with redirect_stderr(stdout):
+                data_transfer.copy_file_list([
+                    (PhysicalKey.from_url('s3://example1/large_file1.npy'), PhysicalKey.from_url('s3://example2/large_file2.npy'), size),
+                ])
+            assert not re.search('#+', stdout.getvalue())
+            assert not re.search('[1-100]', stdout.getvalue())
+
+
```
sir-sigurd: These blank lines are redundant.

api/python/quilt3/util.py
akarve: No, disabling `TQDM` by default is not intended. @cosmic-byte please make default "on".
And let's go with `QUILT_MINIMIZE_STDOUT` and not in the docs that this setting is useful for logging. @cosmic-byte please document this here somewhere https://docs.quiltdata.com/api-reference/api, there should be a section on "Environment variables"

docs/FAQ.md
```diff
@@ -31,3 +31,10 @@ $ export QUILT_DISABLE_USAGE_METRICS=true
 
 Or call `quilt3.disable_telemetry()`
 to persistently disable anonymous usage statistics.
+
+## How do I minimize stdout?
```
akarve:
```suggestion
## Can I turn off progress bars for log files?
```

docs/FAQ.md
```diff
@@ -31,3 +31,10 @@ $ export QUILT_DISABLE_USAGE_METRICS=true
...
+
+You can minimize stdout with an environment variable:
```
akarve:
```suggestion
Yes:
```

api/python/tests/test_data_transfer.py
```diff
@@ -431,9 +433,15 @@ def test_multipart_copy(self):
...
-            ])
+            stdout = io.StringIO()
+
+            # temporarily redirect stderr to capture warnings (usually errors)
+            with redirect_stderr(stdout), mock.patch('quilt3.data_transfer.DISABLE_TQDM', False):
+                data_transfer.copy_file_list([
+                    (PhysicalKey.from_url('s3://example1/large_file1.npy'), PhysicalKey.from_url('s3://example2/large_file2.npy'), size),
+                ])
+            assert re.search('#+', stdout.getvalue())
+            assert re.search('[1-100]', stdout.getvalue())
```
sir-sigurd: This regex is equivalent to `'[01]'`, probably that's not something you actually want. Maybe we should simply check that the buffer contents is empty?

api/python/tests/integration/test_packages.py
```diff
@@ -531,8 +535,15 @@ def test_load_into_quilt(self):
         remote_pkg.set('foo2', test_file3)
 
         with patch('time.time', return_value=1234567891), \
+             patch('quilt3.packages.DISABLE_TQDM', True), patch('quilt3.data_transfer.DISABLE_TQDM', True), \
              patch('quilt3.data_transfer.s3_transfer_config.max_request_concurrency', 1):
-            remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            stdout = io.StringIO()
```
akarve: Why is this called `stdout` if we are redirecting `stderr`?

api/python/tests/integration/test_packages.py
```diff
@@ -531,8 +535,15 @@ def test_load_into_quilt(self):
...
+
+            # temporarily redirect stderr to capture warnings (usually errors)
+            with redirect_stderr(stdout), patch('quilt3.packages.DISABLE_TQDM', True):
+                remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            assert not re.search('#+', stdout.getvalue())
+            assert not re.search('[1-100]', stdout.getvalue())
```
akarve: Better but should really look for `%` sign after the number?

api/python/tests/test_data_transfer.py
```diff
@@ -431,9 +433,15 @@ def test_multipart_copy(self):
         )
 
         with mock.patch('quilt3.data_transfer.s3_transfer_config.max_request_concurrency', 1):
-            data_transfer.copy_file_list([
-                (PhysicalKey.from_url('s3://example1/large_file1.npy'), PhysicalKey.from_url('s3://example2/large_file2.npy'), size),
-            ])
+            stdout = io.StringIO()
+
+            # temporarily redirect stderr to capture warnings (usually errors)
```
sir-sigurd: Not quite sure how this comment is relevant, probably it should explain that by default `tqdm` writes to `stderr`.

api/python/tests/integration/test_packages.py
```diff
@@ -531,8 +535,15 @@ def test_load_into_quilt(self):
...
         with patch('time.time', return_value=1234567891), \
+             patch('quilt3.packages.DISABLE_TQDM', True), patch('quilt3.data_transfer.DISABLE_TQDM', True), \
              patch('quilt3.data_transfer.s3_transfer_config.max_request_concurrency', 1):
-            remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            stdout = io.StringIO()
+
+            # temporarily redirect stderr to capture warnings (usually errors)
+            with redirect_stderr(stdout), patch('quilt3.packages.DISABLE_TQDM', True):
+                remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            assert not re.search('#+', stdout.getvalue())
```
akarve: Why not check for the ASCII block character? (I *think* it's █.) i.e. let's look for things more likely unique to tqdm.

api/python/tests/test_data_transfer.py
cosmic-byte: >Maybe we should simply check that the buffer contents is empty?

How do you propose we check that?

api/python/tests/integration/test_packages.py
cosmic-byte: There isn't much output from `tqdm` except for the `0 - 100%`, `|`, and sequence of `#`.
>Why not check for the ASCII block character? (I think it's █.)

I don't understand why we need this?

api/python/tests/test_data_transfer.py
sir-sigurd: `assert not stdout.getvalue()`?

cosmic-byte: Anything can be captured by `stderr` don't you think it's better to check for `tqdm` specific outputs?

sir-sigurd: 1. if something is written to `stderr` during `copy_file_list()` call this might be the problem that we want to be aware of
2. I don't think that `#` and digits are very specific indicators of `tqdm` output

cosmic-byte: Alright cool

api/python/tests/integration/test_packages.py
```diff
@@ -472,6 +474,7 @@ def test_load_into_quilt(self):
         new_pkg.set('foo2', test_file1)
 
         with patch('time.time', return_value=1234567890), \
+             patch('quilt3.packages.DISABLE_TQDM', True), patch('quilt3.data_transfer.DISABLE_TQDM', True),\
```
sir-sigurd, line 477: Why patching is needed here?

api/python/tests/integration/test_packages.py
```diff
@@ -531,8 +534,13 @@ def test_load_into_quilt(self):
...
 
         with patch('time.time', return_value=1234567891), \
+             patch('quilt3.packages.DISABLE_TQDM', True), patch('quilt3.data_transfer.DISABLE_TQDM', True), \
              patch('quilt3.data_transfer.s3_transfer_config.max_request_concurrency', 1):
-            remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            stderr = io.StringIO()
+
+            with redirect_stderr(stderr), patch('quilt3.packages.DISABLE_TQDM', True):
+                remote_pkg.push('Quilt/package', 's3://my_test_bucket/')
+            assert not stderr.getvalue()
```
sir-sigurd, line 543: Why are you patching `quilt3.packages.DISABLE_TQDM` in two places?

cosmic-byte: That is to test different behaviors when tqdm is disabled or enabled

api/python/tests/integration/test_packages.py
cosmic-byte: The reason for patching is to test for different behaviors or tqdm when disabled or enabled. In test_data_transfer, it was enabled while it is disabled here.

api/python/tests/integration/test_packages.py
cosmic-byte: The two places called `.push` which uses tqdm

api/python/tests/integration/test_packages.py
sir-sigurd: Why patching is needed on this line, if you don't check outcomes of the next `push()` call? Do tests fail if you remove this line?

cosmic-byte: I don't think it would. I'll remove it then.

api/python/pytest.ini
```diff
@@ -7,3 +7,4 @@ filterwarnings =
 
 env =
     QUILT_DISABLE_USAGE_METRICS=True
+    QUILT_MINIMIZE_STDOUT=False
```
sir-sigurd, line 10: Remove this line if tests don't fail without it.

api/python/tests/integration/test_packages.py
```diff
@@ -531,8 +533,13 @@ def test_load_into_quilt(self):
         remote_pkg.set('foo2', test_file3)
 
         with patch('time.time', return_value=1234567891), \
+             patch('quilt3.packages.DISABLE_TQDM', True), patch('quilt3.data_transfer.DISABLE_TQDM', True), \
```
sir-sigurd, line 536: Remove this line if tests don't fail without it.

PR #1565, opened 2020-03-18
Fix config read+support tests on Windows

NathanDeMaria:
## Description

`quilt3` fails to load config on Windows as of `3.1.12`, giving me this error on `import quilt3`:

```
AttributeError: 'WindowsPath' object has no attribute 'read'
```

from inside the `yaml` code. Tracing around, I believe the root is that [this check](https://github.com/NathanDeMaria/quilt/blob/2fe625d9717a2098ae40c087086d28bebd7ea7e5/api/python/quilt3/util.py#L238) should accept all `pathlib.Path`s, not just `pathlib.PosixPath`. Without that check, `quilt3` passes a `pathlib` object instead of a stream into `yaml.safe_load`, which triggers the above error.

Changing this worked for me locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,14 @@
-version: 2
+version: 2.1
+
+orbs:
+  win: circleci/windows@2.4.0
 
 workflows:
   version: 2
   test:
     jobs:
-      - test-3.6
+      - test-36
+      - test-windows
       - test-lambdas-indexer
       - test-lambdas-preview
       - test-lambdas-s3select
@@ -25,7 +29,7 @@ workflows:
             branches:
               ignore: /.*/
 jobs:
-  test-3.6: &test-template
+  test-36: &test-template
     docker:
       - image: circleci/python:3.6-jessie
     environment:
@@ -43,6 +47,28 @@ jobs:
           command: |
             . venv/bin/activate
             cd api/python && pytest --cov=./
+            codecov
+
+  test-windows:
+    executor:
+      name: win/default
+      size: "medium"
+    environment:
+      QUILT_DISABLE_USAGE_METRICS: true
+    steps:
+      - checkout
+      - run:
+          name: Setup
+          command: |
+            pip install virtualenv
+            virtualenv venv
+            venv\Scripts\activate.bat
+            pip install -e api\python[tests]
+      - run:
+          name: Run Pytest
+          command: |
+            venv\Scripts\activate.bat
+            cd api\python; pytest --cov=.
             codecov
 
   test-lambdas-indexer:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,13 +62,14 @@ jobs:
           command: |
             pip install virtualenv
             virtualenv venv
-            venv/Scripts/activate.bat
+            source venv/Scripts/activate
             pip install -e api/python[tests]
           shell: bash.exe
       - run:
           name: Run Pytest
           command: |
-            venv/Scripts/activate.bat
+            set -e
+            source venv/Scripts/activate
             cd api/python; pytest --cov=.
             codecov
           shell: bash.exe
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,51 +28,44 @@ workflows:
               only: /[0-9]+(\.[0-9]+)*([abrc]+[0-9]+)?$/
             branches:
               ignore: /.*/
+
+test-base: &test-base
+  environment:
+    QUILT_DISABLE_USAGE_METRICS: true
+  steps:
+    - run: git config --global core.autocrlf false # for Windows
+    - checkout
+    - run:
+        name: Setup venv
+        command: |
+          pip install virtualenv
+          virtualenv venv
+          if [ -e venv/bin/activate ]; then
+            echo ". venv/bin/activate" >> $BASH_ENV
+          else
+            echo ". venv/Scripts/activate" >> $BASH_ENV # for Windows
+          fi
+    - run:
+        name: Install dependencies
+        command: |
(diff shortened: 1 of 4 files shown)


akarve:
@NathanDeMaria Nice catch. Thank you. Do you feel comfortable modifying the .circleci/config.yml file to test windows in this PR? That would be a great help and we could extend official support.

NathanDeMaria:
@akarve does that look reasonable? I'm not very familiar with CircleCI, but this seems to be working

akarve:
@NathanDeMaria Good progress. I took a quick look at the failures and we *might* be down to actual platform differences at this point (e.g. around `\r`). If you bump into any tricky issues, just ask and we'll help.

In theory the unit tests should run locally on a Windows box as well.

NathanDeMaria:
I have a question on what the intended behavior should be for the [`from_path`](https://github.com/quiltdata/quilt/blob/9333971b6054ba7f660fe7439d11ca1eeb7a8eb3/api/python/quilt3/util.py#L139-L150) function - on Windows, that makes the `PhysicalKey` have `/`'s instead of `\\`, but tests like [this one](https://github.com/quiltdata/quilt/blob/9333971b6054ba7f660fe7439d11ca1eeb7a8eb3/api/python/tests/integration/test_packages.py#L1379-L1383) expect that key to be the same as the string form of a Windows path (with `\\`s)

From context, it looks like we _do_ want the `PhysicalKey` to have `/`s (because it's an AWS S3 key), so that `test_install_subpackage` test (and a few others like it) should be expecting something like `PhysicalKey.from_path(path).path` (to avoid repeating the `.from_path` logic). Does that fit with what you're looking for?

api/python/tests/test_data_transfer.py
```diff
@@ -144,12 +145,14 @@ def test_list_local_url(self):
...
-            ('x/blah.txt', 6)
+            ('foo.txt', 3 + len(os.linesep)),
+            ('x/blah.txt', 5 + len(os.linesep))
         ])
 
     def test_etag(self):
-        assert data_transfer._calculate_etag(DATA_DIR / 'small_file.csv') == '"0bec5bf6f93c547bc9c6774acaf85e1a"'
+        small_file_tag = data_transfer._calculate_etag(DATA_DIR / 'small_file.csv')
+        # Tag will be different if the file had Windows/unix line endings
+        assert small_file_tag == ('"52a0d820c8fa7bec4f04c1325901406c"' if os.name == 'nt' else '"0bec5bf6f93c547bc9c6774acaf85e1a"')
```
NathanDeMaria: An alternative here could be to use `.gitattributes` to make sure these files are checked out with unix line endings, removing these conditional branches for the difference between how an OS checks out text files.

I went with this because it was faster than figuring out the right way to make `git` happy 😆, but if there's a reason to go with that approach instead, I can give that a try

sir-sigurd: > An alternative here could be to use .gitattributes to make sure these files are checked out with unix line endings, removing these conditional branches for the difference between how an OS checks out text files.

I think that's the correct approach, because I think that we should treat test data, well, as data, so its newlines shouldn't be autoconverted.

.circleci/config.yml
```diff
@@ -25,7 +29,7 @@ workflows:
             branches:
               ignore: /.*/
 jobs:
-  test-3.6: &test-template
+  test-36: &test-template
```
sir-sigurd, line 32: any reason for this change?

NathanDeMaria: I'm not familiar with circleci, but from what I could tell, version 2.1 here (which I needed to update to for the Windows orb) doesn't allow `.`s in the names of jobs

sir-sigurd: That's OK then.

NathanDeMaria:
These tests pass on a clean checkout on Windows, but fail in the CI builds. My best guess is CircleCI is using an old checked out version of the `.txt` files (still with Windows line endings). I'm not sure how to (or if I have permission to) start the builds from a clean checkout, but I _think_ that's all the work that remains here

sir-sigurd:
> These tests pass on a clean checkout on Windows, but fail in the CI builds. My best guess is CircleCI is using an old checked out version of the `.txt` files (still with Windows line endings). I'm not sure how to (or if I have permission to) start the builds from a clean checkout, but I _think_ that's all the work that remains here

CircleCI runs at first `git clone` and then `git checkout` and as I understand git fixes newlines only when it creates files.
Anyway I decided to use different approach to avoid `.gitattributes` files:

https://github.com/quiltdata/quilt/blob/feb531826ca122d1aa7367a83c5e1b6e42a93193/.circleci/config.yml#L36

Do my changes look OK for you?

NathanDeMaria:
Without the `.gitattributes`, tests will fail locally for new Windows checkouts, unless people have that `git config` set. It'll save people a headache to keep around `.gitattributes` (see tests are failing, realize it's because they need a certain `git config` set, make the change, 🔁 reclone), but I'm not sure how many people plan on running theses tests locally in Windows, so maybe it's not a big deal 😆

sir-sigurd:
> Without the `.gitattributes`, tests will fail locally for new Windows checkouts, unless people have that `git config` set. It'll save people a headache to keep around `.gitattributes` (see tests are failing, realize it's because they need a certain `git config` set, make the change, repeat reclone), but I'm not sure how many people plan on running theses tests locally in Windows, so maybe it's not a big deal laughing

Makes sense. Added `.gitattributes` in https://github.com/quiltdata/quilt/commit/9c4362db0ff6b5bd386453e707bd2d0c46801a9f.

sir-sigurd:
@NathanDeMaria, thanks for your contribution!

PR #1568, opened 2020-03-21
Tree view filtering + load more

nl0:
## TODO

- [x] UI polish

### DIFF
--- a/catalog/app/containers/Bucket/Listing.js
+++ b/catalog/app/containers/Bucket/Listing.js
@@ -1,22 +1,9 @@
-import PT from 'prop-types'
 import * as R from 'ramda'
 import * as React from 'react'
 import { Link } from 'react-router-dom'
-import * as RC from 'recompose'
-import {
-  Box,
-  Card,
-  CardContent,
-  CircularProgress,
-  Divider,
-  Icon,
-  ListItem,
-  Typography,
-} from '@material-ui/core'
-import { withStyles } from '@material-ui/styles'
+import * as M from '@material-ui/core'
 
 import * as Pagination from 'components/Pagination'
-import { composeComponent } from 'utils/reactTools'
 import { readableBytes } from 'utils/string'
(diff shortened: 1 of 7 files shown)


nl0:
@akarve 

> This is only for immediate children of current directory, right?

right

> It does complete substring matching, right?

right

> What happens to folders with thousands of children (e.g. we cannot fetch all candidates in a single request?) - we need to handle this case because some users have gigantic folders and want to cut down to a few files.

discussed in slack -- i'll add a "load more" button

> Filter placeholder text should be descriptive like "Filter children by prefix" or whatever.

"Filter children by substring"? or maybe you can come up with some better wording?

catalog/app/containers/Bucket/Listing.js
```diff
@@ -85,149 +73,265 @@ const computeStats = R.reduce(
...
+                      <M.Icon
+                        style={{
+                          fontSize: 16,
+                          verticalAlign: 'text-bottom',
+                          marginRight: 4,
+                        }}
+                      >
+                        warning
+                      </M.Icon>
+                      Listing truncated to {totalItems} items
```
akarve:
```suggestion
                      Results truncated to {totalItems} items
```

catalog/app/containers/Bucket/Listing.js
```diff
@@ -85,149 +73,265 @@ const computeStats = R.reduce(
...
-    )
+    [filtering.set, inputRef],
+  )
+
+  return (
+    <div className={classes.root}>
+      <M.InputBase
+        {...filtering.input}
+        onKeyDown={handleKeyDown}
+        placeholder="Filter children of current directory"
```
akarve:
```suggestion
        placeholder="Filter current directory by substring"
```

akarve: Wordsmith filtering text

PR #1567, opened 2020-03-20
Doc'd quilt3.login() and quilt3.logout()

sir-sigurd:
## Description

Probably docstrings can be improved, but I think `quilt3.login()` and `quilt3.logout()` should be documented somehow (considering that we're adding `is_logged()`).

PR #1569, opened 2020-03-22
Fix #1514: Add quilt3.logged_in() function

### LINKED ISSUE #1514: [Enhancement proposal] Add `is_logged` method
### Current status
Currently there is not any straightforward way to determine, if a user is logged in a private registry. As I observed, the login persists between sessions. Anyway, it would be nice to have this tiny method, for example to use in automated scripts.

### Proposition
Add `is_logged` method that would simply return a boolean.

sir-sigurd:
## TODO

- [x] regen docs after #1567

### DIFF
--- a/api/python/quilt3/session.py
+++ b/api/python/quilt3/session.py
@@ -239,12 +239,12 @@ def _refresh_credentials():
 
 def logged_in():
     """
-    Return registry URL if Quilt client is authenticated. Otherwise
+    Return catalog URL if Quilt client is authenticated. Otherwise
     return `None`.
     """
     url = get_registry_url()
     if url in _load_auth():
-        return url
+        return get_from_config('navigator_url')
 
 
 class QuiltProvider(CredentialProvider):
--- a/api/python/quilt3/session.py
+++ b/api/python/quilt3/session.py
@@ -237,6 +237,13 @@ def _refresh_credentials():
     return result
 
 
+def logged_in():
+    """
+    Return whether Quilt client is authenticated.
+    """
+    return get_registry_url() in _load_auth()
+
+
 class QuiltProvider(CredentialProvider):
     METHOD = 'quilt-registry'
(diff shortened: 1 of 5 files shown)


akarve:
Sure, return stack URL or `None` if not logged in.

api/python/quilt3/session.py
```diff
@@ -237,6 +237,16 @@ def _refresh_credentials():
...
 
 
+def logged_in():
+    """
+    Return registry URL if Quilt client is authenticated. Otherwise
+    return `None`.
+    """
+    url = get_registry_url()
+    if url in _load_auth():
+        return url
```
kevinemoore, line 247: This seems to return the registry_url. The more meaningful URL is the catalog URL, which is the URL the user specifies in `quilt3.login`. The username might also be as useful here if not more useful.

sir-sigurd: > This seems to return the registry_url. The more meaningful URL is the catalog URL, which is the URL the user specifies in quilt3.login

That's the same URL that is used in `quilt3.login()`:
https://github.com/quiltdata/quilt/blob/43abfd13ffa2411a553550bb7181505c6dd66481/api/python/quilt3/session.py#L166-L191

> The username might also be as useful here if not more useful.

If I don't miss something we can't return username, as we store only refresh/access tokens.

kevinemoore: We should return `navigator_url` from `get_from_config`.

sir-sigurd:
@kevinemoore I updated PR to make `logged_in()` return catalog URL, however I'm not quite sure that returning URL instead of boolean adds any value, since catalog URL can be obtained with `quilt3.config()['navigator_url']`.

kevinemoore:
@sir-sigurd that's true. Maybe the username would be more useful, since that's not in the config. But, it would still be useful, esp. on the command line to know where you're logged in. But, the catalog_url/navigator_url is definitely much clearer than the registry_url.

PR #1571, opened 2020-03-23
Update 401 login message to point to the latest CLI

NathanDeMaria:
## Description

Quick update to this error message to point to the new CLI. I 🤦‍♂'d around with this for too long before I realized I was logging into old quilt


akarve:
Oof

PR #1580, opened 2020-03-27
Rework retrying during hashing

sir-sigurd:
## Description

* retrying moved to calculate_sha256()
* the original exception is raised on last retry

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -12,7 +12,6 @@ import uuid
 import warnings
 
 import jsonlines
-from tenacity import retry, stop_after_attempt, wait_exponential
 from tqdm import tqdm
 
 from .data_transfer import (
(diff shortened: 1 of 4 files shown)


api/python/quilt3/packages.py
```diff
@@ -918,19 +903,6 @@ def _set_commit_message(self, msg):
             )
 
         self._meta.update({'message': msg})
-
```
kevinemoore: @sir-sigurd this dump of the partial (maybe mostly complete) manifest is lost. Was that intentional? @akarve?

akarve: Let's keep that dump.

sir-sigurd: @kevinemoore 
That was intentional.
In this PR no package entries are modified, unless all hashes are calculated. We don't raise our exception (that masks the original one) with path to the partial manifest. If we want to allow users to continue from where it failed, we might need to rethink API.

kevinemoore: @sir-sigurd, how would you suggest restructuring the API? The purpose of this set of changes is to provide a better experience for users pushing millions of files. The ability to restart and continue from where the error occurred is very important.

akarve: I would suggest that we keep the dump *unless we are very certain that incomplete packages will be very, very rare*. Reason is users uploaded say 1M files, and only one failed, but then they had to start from scratch. Open to API redesign but need more color from Sergey.

sir-sigurd:
@akarve, @kevinemoore 
I restored the dump feature, please re-review.

PR #1585, opened 2020-04-02
Listing: support regex filtering, adjust UI

nl0:
default state:
filter-re

no matches in substring mode, query is quoted:
filter-re-substr-no-match

no matches in regex mode, query is rendered as a RegExp:
filter-re-no-match

regexp syntax error:
filter-re-error

results:
filter-re-results


nl0:
@akarve alternatively, we can use a select control (e.g. [filter current directory] "by substring" | "by regex") instead of a checkbox for clearer ux

PR #1586, opened 2020-04-03
Bucket selection autocomplete

nl0:
## TODO

- [x] cleanup and polish

### DIFF
--- a/catalog/app/containers/NavBar/BucketSelect.js
+++ b/catalog/app/containers/NavBar/BucketSelect.js
@@ -1,264 +1,216 @@
 import { push } from 'connected-react-router/esm/immutable'
 import deburr from 'lodash/deburr'
-import PT from 'prop-types'
+import matchSorter from 'match-sorter'
 import * as R from 'ramda'
 import * as React from 'react'
-import * as RC from 'recompose'
+import AutosizeInput from 'react-input-autosize'
 import * as reduxHook from 'redux-react-hook'
 import * as M from '@material-ui/core'
+import Autocomplete from '@material-ui/lab/Autocomplete'
 
 import * as style from 'constants/style'
 import * as BucketConfig from 'utils/BucketConfig'
 import Delay from 'utils/Delay'
 import * as NamedRoutes from 'utils/NamedRoutes'
-import * as RT from 'utils/reactTools'
-
-const NavInput = RT.composeComponent(
-  'NavBar.BucketSelect.NavInput',
-  M.withStyles(({ palette }) => ({
-    underline: {
-      '&:after': {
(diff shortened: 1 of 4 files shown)


nl0:
bucket-autocomplete

nl0:
@akarve 

> 1) can we keep the bucket descriptions?

keep bucket names and add descriptions or replace names with descriptions?

> 2) how does it work? I first expected typing nyc to only leave buckets starting with "nyc" but that's not what happened. Is it substring match against all the things? (name, title, description)

we're using [match-sorter](https://github.com/kentcdodds/match-sorter) here on name, title and description and it does "intelligent" sorting. here's the readme quote:

> This follows a simple and sensible (user friendly) algorithm that makes it easy for you to filter and sort a list of items based on given input. Items are ranked based on sensible criteria that result in a better user experience.
> To explain the ranking system, I'll use countries as an example:
> CASE SENSITIVE EQUALS: Case-sensitive equality trumps all. These will be first. (ex. France would match France, but not france)
> EQUALS: Case-insensitive equality (ex. France would match france)
> STARTS WITH: If the item starts with the given value (ex. Sou would match South Korea or South Africa)
> WORD STARTS WITH: If the item has multiple words, then if one of those words starts with the given value (ex. Repub would match Dominican Republic)
> CASE STARTS WITH: If the item has a defined case (camelCase, PascalCase, snake_case or kebab-case), then if one of the parts starts with the given value (ex. kingdom would match unitedKingdom or united_kingdom)
> CASE ACRONYM If the item's case matches the synonym (ex. uk would match united-kingdom or UnitedKingdom)
> CONTAINS: If the item contains the given value (ex. ham would match Bahamas)
> ACRONYM: If the item's acronym is the given value (ex. us would match United States)
> SIMPLE MATCH: If the item has letters in the same order as the letters of the given value (ex. iw would match Zimbabwe, but not Kuwait because it must be in the same order). Furthermore, if the item is a closer match, it will rank higher (ex. ua matches Uruguay more closely than United States of America, therefore Uruguay will be ordered before United States of America)
> This ranking seems to make sense in people's minds. At least it does in mine. Feedback welcome!

it can be configured to match in a stricter manner if this is what you want

akarve:
@nl0 For 1) ideally we keep the menu looking exactly the same during search BUT we add in parentheses the S3 address near the title. Can that that work.

For 2) if you think it's a good UX keep it

nl0:
@akarve 

> For 1) ideally we keep the menu looking exactly the same during search BUT we add in parentheses the S3 address near the title

done

> For 2) if you think it's a good UX keep it

i've made it a little stricter and also added matching by bucket's tags

bucket-autocomplete-2

PR #1587, opened 2020-04-06
Attempt to optimize GROUP BY in Athena queries

dimaryaz:


lambdas/access_counts/index.py
```diff
@@ -152,14 +152,18 @@ def sql_escape(s):
...
+
 OBJECT_ACCESS_COUNTS = textwrap.dedent("""\
     SELECT
         eventname,
         bucket,
         key,
         CAST(histogram(date) AS JSON) AS counts
     FROM object_access_log
-    GROUP BY eventname, bucket, key
+    GROUP BY 3, 2, 1
```
akarve, line 166: Oh is `bucket` actually higher cardinality than `eventname`?

dimaryaz: Not necessarily, but no way to know. But, there are only 5 or so possible events, while the number of buckets could be pretty much anything. (It's whatever buckets CloudTrail is set up for, so even without multi-bucket setup, it can be more than one.)

PR #1588, opened 2020-04-07
Gzip lambda responses

dimaryaz:
If the lambda response is plain text or JSON, and bigger than 1KB, compress it.

It appears that this only affects the API Gateway -> Lambda communication, not the Browser -> API Gateway, so don't need to switch to the binary gateway.

### DIFF
--- a/lambdas/shared/t4_lambda_shared/decorator.py
+++ b/lambdas/shared/t4_lambda_shared/decorator.py
@@ -4,10 +4,15 @@ Decorators for using lambdas in API Gateway
 
 from base64 import b64decode, b64encode
 from functools import wraps
+import gzip
 import json
 import traceback
 
 from jsonschema import Draft4Validator, ValidationError
+
+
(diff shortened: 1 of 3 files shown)


lambdas/shared/t4_lambda_shared/decorator.py
```diff
@@ -4,12 +4,17 @@
...
 from base64 import b64decode, b64encode
 from functools import wraps
+import gzip
 import json
 import traceback
 
 from jsonschema import Draft4Validator, ValidationError
 
 
+GZIP_MIN_LENGTH = 1024
```
akarve, line 14: We need to keep this in sync with the corresponding API Gateway, right? i.e. worth a comment

dimaryaz: Nah, that's just to avoid gzip'ing tiny responses, where the compressed data will probably be bigger than the original

PR #1591, opened 2020-04-10
Fix thumbnail lambda for imageio rejects and Z Dimension files

JacksonMaxfield:
Somehow, CI got dropped for thumbnail lambda and I was trying to fix / upgrade the thumbnail generator and couldn't run pytest.

Additionally, this fixes a thumbnail generation bug on files such as [this one](https://open.quiltdata.com/b/aics-modeling-packages-test-resources/tree/actk/test_data/data/example_prepared_image_for_feature_extraction_0_2.ome.tiff) where `imageio.get_reader` was failing because it uses a new writer.


akarve:
@JacksonMaxfield we'll need to handle body decoding a bit differently, a la https://github.com/quiltdata/quilt/pull/1588/files#diff-0dc00393af00858bc875f1b2caebc9cdR39

lambdas/shared/t4_lambda_shared/utils.py
```diff
@@ -26,3 +28,12 @@ def make_json_response(status_code, json_object, extra_headers=None):
         headers.update(extra_headers)
 
     return status_code, json.dumps(json_object), headers
+
+
+def read_body(resp):
```
akarve, line 33: perfect can you a) docstring; b) have dima's code in /preview test use this same function?

JacksonMaxfield: Yep! Didn't know if you would want me to pull that in or not in this PR but will do.

lambdas/thumbnail/index.py
```diff
@@ -188,7 +189,7 @@ def _format_n_dim_ndarray(img: AICSImage) -> np.ndarray:
         return np.pad(np.concatenate(merged, axis=0), ((0, 5), (0, 5)), mode="constant")
 
     if "Z" in img.reader.dims:
-        return Image.fromarray(norm_img(img.data[0, 0, 0, :, :, :].max(axis=0)))
+        return norm_img(img.data[0, 0, 0, :, :, :].max(axis=0))
```
akarve, line 192: Can you add a comment as to why we do this?

PR #1595, opened 2020-04-15
Add pylint job to CircleCI

sir-sigurd:


pylintrc
```diff
@@ -0,0 +1,51 @@
+[MESSAGES CONTROL]
+disable=attribute-defined-outside-init,
```
sir-sigurd, line 2: That's a minimal set of disabled checks to make pylint happy with the current codebase. I'm going to enable some of these checks one by one in subsequent PRs.

PR #1597, opened 2020-04-15
Remove doc limit for stats query

kevinemoore:
## Description
Two minor changes to the search lambda:
1. removing the document-per-shard limit only for the stats query, which can take advantage of cached summary statistics insdie elastic.
2. allow the MAX_DOCS_PER_SHARD to be set in the environment.

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -12,8 +12,8 @@ from t4_lambda_shared.decorator import api
 from t4_lambda_shared.utils import get_default_origins, make_json_response
 
 INDEX_OVERRIDES = os.getenv('INDEX_OVERRIDES', '')
+MAX_DOCUMENTS_PER_SHARD = os.getenv('MAX_DOCUMENTS_PER_SHARD')
 MAX_QUERY_DURATION = '15s'
-MAX_DOCUMENTS_PER_SHARD = 10000
 NUM_PREVIEW_IMAGES = 100
 NUM_PREVIEW_FILES = 100
 IMG_EXTS = [
@@ -48,6 +48,7 @@ def lambda_handler(request):
     """
     action = request.args.get('action')
     indexes = request.args.get('index')
+    terminate_after = MAX_DOCUMENTS_PER_SHARD
 
     if action == 'search':
         query = request.args.get('query', '')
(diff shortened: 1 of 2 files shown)


lambdas/search/index.py
```diff
@@ -46,8 +44,10 @@ def lambda_handler(request):
     """
     Proxy the request to the elastic search.
     """
+
     action = request.args.get('action')
     indexes = request.args.get('index')
+    terminate_after = os.getenv('MAX_DOCUMENTS_PER_SHARD')
```
akarve, line 50: This needs a default or else it can return `None`.
If there's no safe default (unlikely) then we should immediately throw an exception.
We should also make sure this number is an integer. The best way to do that is probably:
```
 locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
locale.atoi(terminate_after)
```
Docs are not really clear on kwarg type https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch.search

akarve, line 50: Is `None` really a safe default? We should probably default to an integer?

Also, you need to convert from string (`os.getenv()`) to int (what the API expects). `int("1,000")` will fail because of commas:
```
locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
locale.atoi(terminate_after)
```
Docs are not really clear on kwarg type https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch.search

lambdas/search/index.py
```diff
@@ -75,6 +75,8 @@ def lambda_handler(request):
         }
         size = 0
         _source = []
+        # Consider all documents when computing counts, etc.
+        terminate_after = None
```
akarve, line 79: Are we sure that `terminate_after` is correct for `'search'`? If it terminates the search before going through the entire index, that would be very bad. If it's just a limit, that's different?

lambdas/search/index.py
kevinemoore: None is a safe default. The argument is dropped before sending. See the example url in test_search.py (test_stats).

I didn't convert the string because it isn't necessary since it's being encoded and sent as a string. I also wanted to preserve None (though I could do that manually).

akarve: OK but what if the user sets a string that is not an integer?

kevinemoore: It's set by the stack (or lambda deployment).

PR #1598, opened 2020-04-16
Add pylint git pre-commit hook

cosmic-byte:
## Description
Add to quiltdata/quilt: pylint pre-commit git hook for *.py

If commit hook doesn't work automatically after applying this PR changes,  create a new file `quilt/.git/hooks/pre-commit` and add the following to that file:

```
#!/usr/bin/env bash
git-pylint-commit-hook
```
Save the file and make it executable:

```
chmod +x .git/hooks/pre-commit
```

### DIFF
--- a/api/python/.pylintrc
+++ b/api/python/.pylintrc
@@ -1,2 +1,5 @@
 [TYPECHECK]
 generated-members=requests.codes.*,responses.*
+
+[pre-commit-hook]
+limit=8.0
--- a/api/python/setup.py
+++ b/api/python/setup.py
@@ -71,7 +71,6 @@ setup(
         'tqdm>=4.26.0',
         'urllib3<1.25,>=1.21.1',            # required by requests
         'requests_futures==1.0.0',
-        'git-pylint-commit-hook',
     ],
     extras_require={
         'pyarrow': [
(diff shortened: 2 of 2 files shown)


api/python/setup.py
```diff
@@ -71,6 +71,7 @@ def run(self):
         'tqdm>=4.26.0',
         'urllib3<1.25,>=1.21.1',            # required by requests
         'requests_futures==1.0.0',
+        'git-pylint-commit-hook',
```
akarve, line 74: this belongs in `tests` instead?

api/python/.pylintrc
```diff
@@ -1,2 +1,5 @@
 [TYPECHECK]
 generated-members=requests.codes.*,responses.*
+
+[pre-commit-hook]
+limit=8.0
```
akarve, line 5: Does this automatically only lint *.py that are staged and dirty?

cosmic-byte: >The git-pylint-commit-hook is a tool for checking Python code using pylint before the code is committed to git.
>
>All files that end with .py or that have python in the she bang will be automatically checked before they are committed to the repo.

https://git-pylint-commit-hook.readthedocs.io/en/latest/index.html

api/python/setup.py
akarve: That's fine but shouldn't it be in extras_require?

cosmic-byte:
@akarve I've moved the `git-pylint-commit-hook` to `extras_require`. This implies that for the hook to work, you need to install 
```
pip install .[tests]
```

PR #1606, opened 2020-04-20
parameterise lambda tests

cosmic-byte:
## Description
Reduce cut-and-paste code in .circleci/config.yml lambdas tests by reusing (and parameterizing) a single test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,7 @@ jobs:
       QUILT_DISABLE_USAGE_METRICS: true
     parameters:
       path:
-        description: "lambdas/$path"
+        description: "Relative path to lambda root including setup.py, e.g. 'es/indexer' for lambdas/es/indexer"
         type: string
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,8 @@ jobs:
     docker:
       - image: circleci/python:3.6-jessie
     description: "Test lambdas"
+    environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -100,9 +104,14 @@ jobs:
...
     <<: *test-base
 
-  test-lambdas-indexer:
+  test-lambda:
     docker:
       - image: circleci/python:3.6-jessie
+    description: "Test lambdas"
+    parameters:
+      path:
+        description: "The lambda path"
```
akarve, line 113:
```suggestion
        description: "lambdas/$path"
```

cosmic-byte: I don't think you can reference `parameters.path` inside the description of the parameter definition

cosmic-byte: Oh, you mean just a literal string `"lambdas/$path"`

.circleci/config.yml
```diff
@@ -100,67 +104,16 @@ jobs:
...
-  test-lambdas-search:
+  test-lambda:
     docker:
       - image: circleci/python:3.6-jessie
+    description: "Test lambdas"
+    environment:
+      QUILT_DISABLE_USAGE_METRICS: true
+    parameters:
+      path:
+        description: "lambdas/$path"
```
akarve, line 115: Sorry, here's a better description:
```suggestion
        description: "Relative path to lambda root including setup.py, e.g. 'es/indexer' for lambdas/es/indexer"
```

PR #1608, opened 2020-04-21
add lambda test for thumbnail

cosmic-byte:
## Description
Add lambda test for thumbnail

## Depends On
#1607


akarve:
@cosmic-byte I think you need to pick up some of the test fixes from https://github.com/quiltdata/quilt/pull/1591, have a look

PR #1623, opened 2020-05-07
Python code cleanup: enable all pycodestyle checks

sir-sigurd:


setup.cfg
```diff
@@ -1,2 +1,2 @@
 [pycodestyle]
-select = E3
+max-line-length = 119
```
sir-sigurd, line 2: I chose this number mostly because it's used in Django repo I'm used to and IMHO it's a reasonable number.

PR #1630, opened 2020-05-13
Only consider second file extension if primary ends with '.gz'

akarve:
Customer stacks with lot of *.blah.png, *.boo.png, *.baz.png extensions actually wanted all of those file stats to roll up under .png in the overview. Now we only consider the secondary extension if the primary is .gz.

### DIFF
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -210,7 +210,7 @@ EXTS_ACCESS_COUNTS = textwrap.dedent("""\
             eventname,
             bucket,
             lower(CASE
-                WHEN cardinality(parts) > 2 THEN concat(element_at(parts, -2), '.', element_at(parts, -1))
+                WHEN cardinality(parts) > 2 AND lower(element_at(parts, -1)) = 'gz' THEN concat(element_at(parts, -2), '.', element_at(parts, -1))
                 WHEN cardinality(parts) = 2 THEN element_at(parts, -1)
                 ELSE ''
                 END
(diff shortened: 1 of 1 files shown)


lambdas/access_counts/index.py
```diff
@@ -210,7 +210,7 @@ def sql_escape(s):
             eventname,
             bucket,
             lower(CASE
-                WHEN cardinality(parts) > 2 THEN concat(element_at(parts, -2), '.', element_at(parts, -1))
+                WHEN cardinality(parts) > 2 AND lower(element_at(parts, -1)) = 'gz' THEN concat(element_at(parts, -2), '.', element_at(parts, -1))
                 WHEN cardinality(parts) = 2 THEN element_at(parts, -1)
```
akarve, line 214:
```suggestion
                WHEN cardinality(parts) >= 2 THEN element_at(parts, -1)
```

lambdas/access_counts/index.py
```diff
@@ -210,8 +210,9 @@ def sql_escape(s):
             eventname,
             bucket,
             lower(CASE
-                WHEN cardinality(parts) > 2 THEN concat(element_at(parts, -2), '.', element_at(parts, -1))
-                WHEN cardinality(parts) = 2 THEN element_at(parts, -1)
+                WHEN cardinality(parts) > 2 AND lower(element_at(parts, -1)) = 'gz'
+                    THEN concat(element_at(parts, -2), '.', element_at(parts, -1))
+                WHEN cardinality(parts) >= 2 THEN element_at(parts, -1)
```
dimaryaz, line 215: Why did you change it to `>=` here?

akarve: So that when `bucket/folder/foo.baz.png` flunks the line above, it still turns into  png?

PR #1633, opened 2020-05-13
Refactor some components, fix MUI deprecation warnings

nl0:


nl0:
> Not critical but I'm still seeing this warning in console:
> 
> ```
> Warning: Cannot update a component (`Loader`) while rendering a different component (`Auth.Provider:render`). To locate the bad setState() call inside `Auth.Provider:render`, follow the stack trace as described in https://fb.me/setstate-in-render
> ```

this requires a more in-depth refactoring

PR #1635, opened 2020-05-15
Bucket/File: adjust delete marker display and handle non-existent objects better

nl0:
* render DELETE MARKER in bold in the version list

* dont show download button if selected version is a delete marker

* dont attempt preview if selected version is a delete marker, show
  descriptive message instead

* show "no such package" when navigating to non-existent package
### TODO

- [x] show "Object does not exist" if this is the case
- [x] fix delete marker detection


nl0:
@akarve @kevinemoore this one's ready for review

nl0:
note: i've also update s3proxy config here to expose more aws headers

PR #1643, opened 2020-05-22
always use regional s3 endpoints

nl0:


catalog/app/utils/AWS/S3.js
```diff
@@ -14,6 +14,7 @@ import * as Credentials from './Credentials'
 
 const DEFAULT_OPTS = {
   signatureVersion: 'v4',
+  s3UsEast1RegionalEndpoint: 'regional',
```
kevinemoore, line 17: What is this used for?

nl0: to always use regional endpoints (instead of global `s3.amazonaws.com`), even for the default region (us-east-1)

PR #1647, opened 2020-05-25
Use dir metadata for top hash

sir-sigurd:
## TODO

- [x] Unit Tests

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1175,20 +1175,28 @@ class Package:
         """
         top_hash = hashlib.sha256()
         assert 'top_hash' not in self._meta
-        top_meta = json.dumps(self._meta, sort_keys=True, separators=(',', ':'))
-        top_hash.update(top_meta.encode('utf-8'))
+
+        json_encode = json.JSONEncoder(sort_keys=True, separators=(',', ':')).encode
+        for part in self._get_top_hash_parts():
+            top_hash.update(json_encode(part).encode())
+
+        return top_hash.hexdigest()
+
+    def _get_top_hash_parts(self):
+        yield self._meta
+        for logical_key, meta in self._walk_dir_meta():
+            yield {'logical_key': logical_key, 'meta': meta}
         for logical_key, entry in self.walk():
             if entry.hash is None or entry.size is None:
                 raise QuiltException(
                     "PackageEntry missing hash and/or size: %s" % entry.physical_key
                 )
-            entry_dict = entry.as_dict()
-            entry_dict['logical_key'] = logical_key
-            entry_dict.pop('physical_keys', None)
-            entry_dict_str = json.dumps(entry_dict, sort_keys=True, separators=(',', ':'))
-            top_hash.update(entry_dict_str.encode('utf-8'))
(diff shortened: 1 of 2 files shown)


api/python/quilt3/packages.py
```diff
@@ -1139,20 +1139,28 @@ def top_hash(self):
         """
```
akarve, line 1139: Please add the high-level top_hash algorithm to the docs here (i.e. note that metadata are hashed as of <commit>).

api/python/quilt3/packages.py
```diff
@@ -1139,20 +1139,28 @@ def top_hash(self):
...
                     "PackageEntry missing hash and/or size: %s" % entry.physical_key
                 )
-            entry_dict = entry.as_dict()
-            entry_dict['logical_key'] = logical_key
-            entry_dict.pop('physical_keys', None)
-            entry_dict_str = json.dumps(entry_dict, sort_keys=True, separators=(',', ':'))
-            top_hash.update(entry_dict_str.encode('utf-8'))
-
-        return top_hash.hexdigest()
+            yield {
```
akarve, line 1158: Much cleaner.

api/python/tests/integration/test_packages.py
```diff
@@ -863,6 +863,15 @@ def test_top_hash_stable(self):
         assert pkg.top_hash == top_hash, \
             "Unexpected top_hash for {}/packages/.quilt/packages/{}".format(registry, top_hash)
 
+    def test_top_hash_stable_dir_meta(self):
+        registry = DATA_DIR.as_posix()
+        top_hash = '20de5433549a4db332a11d8d64b934a82bdea8f144b4aecd901e7d4134f8e733'
```
akarve, line 868: this is fine but the unit test will break every time someone modifies `/data`. So I'd use set_dir here, get the `.top_hash` and then assert that the hash changes after calling `.set_meta`.

then maybe patch `get_top_hash_parts` (which doesn't have any test coverage) and return whatever you want so you can check against a stable hash?

api/python/quilt3/packages.py
sir-sigurd: > note that _get_top_hash_parts indicates what from each entry contributes to the hash

This docstring is public, why mention a private method here?

akarve: Agreed to skip mention it on the docstring

api/python/tests/integration/test_packages.py
sir-sigurd: I merged the new test with `test_top_hash_stable()`. We could completely inline the [test manifest](https://github.com/quiltdata/quilt/blob/master/api/python/tests/integration/data/.quilt/packages/20de5433549a4db332a11d8d64b934a82bdea8f144b4aecd901e7d4134f8e733) in the test if it makes sense to you.

akarve:
Per discussion we should merge this as part for quilt3 4.x so as not to merge a potentially breaking change into 3.x.

drernie:
@sir-sigurd This seems outdated and of unclear value.  Closing, unless you want to revive and merge.

PR #1655, opened 2020-06-05
Post-process stats queries to ElasticSearch to aggregate uncompressed file extensions

akarve:


lambdas/search/index.py
```diff
@@ -74,7 +77,7 @@ def lambda_handler(request):
                 },
             }
         }
-        size = 0
```
kevinemoore, line 77: I believe size = 0 is correct. We are only interested in the aggregations and are not fetching/returning any actual documents. See: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#_fast_check_for_any_matching_docs

lambdas/search/tests/test_search.py
```diff
@@ -40,6 +124,110 @@ def _make_event(cls, query):
             'isBase64Encoded': False,
         }
 
+    def test_post_process_stats_all_gz(self):
+        """test stats when all *.gz in bucket"""
+        es_response = ES_STATS_RESPONSES['all_gz']
+        processed = post_process(es_response, 'stats')
+        assert es_response == processed, 'No processing expected'
+        assert set(processed.keys()) == set(es_response.keys()), 'Unexpected top-level key change'
```
akarve, line 132:
```suggestion
```

lambdas/search/index.py
```diff
@@ -74,7 +77,7 @@ def lambda_handler(request):
...
+        size = 0 # We still get all aggregates, just don't need the results
```
akarve, line 80:
```suggestion
        size = 0  # We still get all aggregates, just don't need the results
```

PR #1656, opened 2020-06-05
Don't ask the user to set a password in SIGN_IN_ONLY mode

dimaryaz:
If SSO is enabled with password sign-in but not password sign-up, don't ask the user to set a password.

(This particular combination has never come up before.)

### DIFF
--- a/catalog/app/containers/Auth/SSOSignUp.js
+++ b/catalog/app/containers/Auth/SSOSignUp.js
@@ -113,7 +113,7 @@ export default ({ location: { search } }) => {
                 invalid: <FM {...msg.signUpUsernameInvalid} />,
               }}
             />
-            {!!cfg.passwordAuth && (
+            {cfg.passwordAuth === true && (
               <>
                 <Field
                   component={Layout.Field}


dimaryaz:
That's intentional: in Single Sign-On mode, we don't want the user to be able to sign in using a password, only using the SSO provider. See https://github.com/quiltdata/enterprise/pull/155.

nl0:
> That's intentional: in Single Sign-On mode, we don't want the user to be able to sign in using a password, only using the SSO provider. See [quiltdata/enterprise#155](https://github.com/quiltdata/enterprise/pull/155).

in this case we should set passwordAuth to false, no?

dimaryaz:
Well, we also want to allow the admin to invite non-SSO users who set a password. So password sign-in still needs to be enabled (but SSO users won't have passwords, so won't be able to use it).

This is basically a combination of the two modes we had before: 1) Password only, no signup, only invites, and 2) Google sign-in/sign-up only.

nl0:
> Well, we also want to allow the admin to invite non-SSO users who set a password. So password sign-in still needs to be enabled (but SSO users won't have passwords, so won't be able to use it).
> 
> This is basically a combination of the two modes we had before: 1) Password only, no signup, only invites, and 2) Google sign-in/sign-up only.

well, it seems a little confusing to me, but the change in this PR _shouldnt_ break anything i think

akarve:
> well, it seems a little confusing to me, but the change in this PR _shouldnt_ break anything i think

What's the most confusing part? We have a customer use case where they want SSO on-domain and they need to be able to invite external collabs off-domain/off-Google.

nl0:
> What's the most confusing part?

my mental model was a little different: i assumed that if the user is allowed to sign-in w/ the password, he should be able to set that password, i didnt think about invitation case

PR #1657, opened 2020-06-06
Get index extension overrides from environment

akarve:


lambdas/es/indexer/test/test_index.py
```diff
@@ -534,6 +534,57 @@ def test_multiple_index_events(self):
...
+        """ensure the function that infers overrides from the env works:
+            always returns a valid set(), perhaps empty, lowercases extensions
+        """
+        with patch.dict(os.environ, {'CONTENT_INDEX_EXTS': '.txt'}):
+            assert _get_extension_overrides() == {'.txt'}
+        with patch.dict(os.environ, {'CONTENT_INDEX_EXTS': ' .tXt   '}):
+            assert _get_extension_overrides() == {'.txt'}
+        with patch.dict(os.environ, {'CONTENT_INDEX_EXTS': ' garbage  gar.bage  '}):
+            assert _get_extension_overrides() == set()
+        with patch.dict(os.environ, {'CONTENT_INDEX_EXTS': ' .Parquet, .csv, .tsv'}):
```
akarve, line 585:
```suggestion
        with patch.dict(os.environ, {'CONTENT_INDEX_EXTS': ' .Parquet,.csv, .tsv'}):
```

lambdas/es/indexer/document_queue.py
```diff
@@ -12,15 +12,25 @@
...
 
-CONTENT_INDEX_EXTS = [
+def _get_extension_overrides():
+    """check the environment for index extension overrides; this is a standalone
+    function so that it can be unit tested"""
+    return {
+        t.strip().lower()
+        for t in os.getenv("CONTENT_INDEX_EXTS", "").split(",")
+        if t.strip().startswith(".")
+    }
```
kevinemoore, line 22: This code is correct, but it's a bit confusing to see how (e.g., without the if startswith(".') the or below would break.) I think it will be clearer for the long run if you refactor this just a bit.

CONTENT_INDEX_EXTS (all caps) looks like a constant, but is now being set by the environment. Instead, replace the reference to CONTENT_INDEX_EXTS in index.py with a simple call to a method "get_content_index_exts" (or similar). Then you can write all the logic into one clean function.

akarve: It's still a constant with respect to use within the module for a given lambda (and is set by an environment constant). I'd rather not change that part.
I'll see how much cleaner I can make the comprehension and how it's called.

PR #1665, opened 2020-06-12
Set NO_DOWNLOAD in local catalog

kevinemoore:
## Description
A recent change to config.json (adding the NO_DOWNLOAD flag) requires that the catalog docker environment set `NO_DOWNLOAD` to produce a correct config.json.

### DIFF
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -77,6 +77,7 @@ def _launch_local_catalog():
     env = dict(REGISTRY_URL="http://localhost:5000",
                S3_PROXY_URL=open_config["s3Proxy"],
                ALWAYS_REQUIRE_AUTH="false",
+               NO_DOWNLOAD="false",
                CATALOG_MODE="LOCAL",
                SSO_AUTH="DISABLED",
                PASSWORD_AUTH="ENABLED",


api/python/quilt3/main.py
```diff
@@ -77,6 +77,7 @@ def _launch_local_catalog():
     env = dict(REGISTRY_URL="http://localhost:5000",
                S3_PROXY_URL=open_config["s3Proxy"],
                ALWAYS_REQUIRE_AUTH="false",
+               NO_DOWNLOAD="false",
```
akarve, line 80: Since this is a dict don't you want `False` or do we clean that up elsewhere?

kevinemoore: It's a dict processed by javascript so we've used the string "false" for other boolean variables.

akarve: `json.dumps` takes care of turning `False` into `false`

akarve: You can address this simply but we should be using `json.dumps` and it's not really correct without that because the string "false" is truthy even in JS.

kevinemoore: The goal is to produce command line parameters for docker (e.g., -e NO_DOWNLOAD=false). Using the Python boolean False instead of the string "false" sets the value to "False", which breaks json parsing on config.json. I suppose we could call json.dumps() on each value in the dict. That might be safer in general, but seems like overkill since we're just dealing with command-line text.

akarve: errr fix this in a separate PR with `json.dumps(foo).strip('\"')`?

PR #1666, opened 2020-06-12
Pin open dependencies so CI doesn't fail randomly

akarve:


.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ commands:
       - run:
           name: Setup venv
           command: |
-            pip install virtualenv
+            pip install virtualenv==20.0.23
```
akarve, line 13:
```suggestion
            pip install --user virtualenv==20.0.23
```

.circleci/config.yml
```diff
@@ -223,5 +223,5 @@ jobs:
           command: |
             . venv/bin/activate
             cd api/python
-            pip install twine
+            pip install twine==3.1.1
```
akarve, line 226:
```suggestion
            pip install --user twine==3.1.1
```

.circleci/config.yml
```diff
@@ -223,5 +223,5 @@ jobs:
           command: |
             . venv/bin/activate
             cd api/python
-            pip install twine
+            sudio pip install twine==3.1.1
```
kevinemoore, line 226: Is that a typo? sudio->sudo?

PR #1667, opened 2020-06-13
Replace partitions with projections

akarve:


akarve:
Partially handled by https://github.com/quiltdata/quilt/pull/1708. Future use of projection, instead of using a stride for backfill, should backfill days in an interruptible, iterative loop.

PR #1674, opened 2020-06-19
Optimize the package access counts queries

dimaryaz:
Avoid doing a cross product if a package hash belongs to many packages.
Calculate counts for each (eventname, bucket, hash, date) first, then do the join.

### DIFF
--- a/lambdas/access_counts/index.py
+++ b/lambdas/access_counts/index.py
@@ -166,7 +166,9 @@ OBJECT_ACCESS_COUNTS = textwrap.dedent("""\
     GROUP BY 3, 2, 1
 """)
 
-PACKAGE_ACCESS_COUNTS = textwrap.dedent("""\
+MANIFEST_PREFIX = '.quilt/packages/'
+
+PACKAGE_ACCESS_COUNTS = textwrap.dedent(f"""\
     SELECT
         eventname,
         bucket,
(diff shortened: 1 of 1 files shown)


lambdas/access_counts/index.py
```diff
@@ -169,24 +169,40 @@ def sql_escape(s):
...
+    FROM (
+        SELECT
+            eventname,
+            t.bucket AS bucket,
+            name,
+            date,
+            sum(count) AS count
+        FROM (
+            SELECT eventname, bucket, substr(key, 17) AS hash, date, count(*) AS count from object_access_log
+            WHERE substr(key, 1, 16) = '.quilt/packages/'
```
akarve, line 184: Easier to maintain if you do `MANIFEST_PREFIX = '.quilt/packages/'` above and `len(MANIFEST_PREFIX)`, etc. here.

lambdas/access_counts/index.py
```diff
@@ -169,24 +169,40 @@ def sql_escape(s):
...
+        CAST(map_agg(date, count) AS JSON) AS counts
+    FROM (
+        SELECT
+            eventname,
+            t.bucket AS bucket,
+            name,
+            date,
+            sum(count) AS count
+        FROM (
+            SELECT eventname, bucket, substr(key, 17) AS hash, date, count(*) AS count from object_access_log
```
akarve, line 183: Nice. This is the trick I was missing.

lambdas/access_counts/index.py
```diff
@@ -169,24 +169,40 @@ def sql_escape(s):
...
-        hash,
-        CAST(histogram(date) AS JSON) AS counts
-    FROM object_access_log JOIN package_hashes
-    ON object_access_log.bucket = package_hashes.bucket AND key = concat('.quilt/packages/', hash)
-    GROUP BY 4, 3, 2, 1
+        t.hash AS hash,
+        counts
+    FROM (
+        SELECT eventname, bucket, substr(key, 17) AS hash, CAST(histogram(date) AS JSON) AS counts from object_access_log
+        WHERE substr(key, 1, 16) = '.quilt/packages/'
```
akarve, line 202: Same comments as above, prefer a constant.

lambdas/access_counts/index.py
```diff
@@ -169,24 +169,40 @@ def sql_escape(s):
...
+        GROUP BY 4, 3, 2, 1
+    )
     GROUP BY 3, 2, 1
 """)
 
 PACKAGE_VERSION_ACCESS_COUNTS = textwrap.dedent("""\
     SELECT
         eventname,
-        package_hashes.bucket AS bucket,
+        t.bucket AS bucket,
```
kevinemoore, line 196: Minor nit: replace 't' with something descriptive (e.g., 'access_totals', 'total_counts', ...).

PR #1676, opened 2020-06-22
add get_meta to package

cosmic-byte:
## Description

Add `get_meta` to top level of each package.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -149,6 +149,12 @@ class PackageEntry:
         """
         self._meta['user_meta'] = meta
 
+    def get_meta(self):
+        """
+        Gets the user_meta for this PackageEntry.
+        """
+        return self.meta or None
+
     def _verify_hash(self, read_bytes):
         """
         Verifies hash of bytes
@@ -856,6 +862,12 @@ class Package:
         """
         self._meta['user_meta'] = meta
         return self
+
+    def get_meta(self):
+        """
+        Gets user metadata on this Package.
+        """
+        return self.meta or None
 
     def _fix_sha256(self):
         """
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -149,12 +149,6 @@ class PackageEntry:
         """
         self._meta['user_meta'] = meta
 
-    def get_meta(self):
-        """
-        Gets the user_meta for this PackageEntry.
-        """
-        return self.meta or None
-
     def _verify_hash(self, read_bytes):
         """
         Verifies hash of bytes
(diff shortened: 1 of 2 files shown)


api/python/quilt3/packages.py
```diff
@@ -149,6 +149,12 @@ def set_meta(self, meta):
         """
         self._meta['user_meta'] = meta
 
+    def get_meta(self):
```
akarve, line 152: we already have .meta() so I don't think we want get_meta() for PackageEntry

api/python/quilt3/packages.py
```diff
@@ -857,6 +863,12 @@ def set_meta(self, meta):
         self._meta['user_meta'] = meta
         return self
 
+    def get_meta(self):
+        """
+        Gets user metadata on this Package.
+        """
+        return self.meta or None
```
akarve, line 870: I don't think this is correct because look [what `set_meta` does](https://github.com/quiltdata/quilt/blob/f476bf61b08c83e4541c9d234973ae82d99307eb/api/python/quilt3/packages.py#L853). Also, `or None` doesn't buy us anything? If you want to safely get a value from a `dict` use `.get`.

akarve: Also. `self.meta` is not defined anywhere?

sir-sigurd:
What is the purpose of this change?

cosmic-byte:
> I don't quite understand what this change is for, but it we want parity for getting and setting user meta, we could add setter for `meta` property and deprecate `set_meta()`.
> Edit:
> But that's wouldn't be identical since setter can't return and doesn't allow chaining.

I wasn't quite sure what the purpose of the change is since property `meta` returns the user metadata set by `set_meta`

api/python/quilt3/packages.py
cosmic-byte: Package and PackageEntry classes already have a property [`meta`](https://github.com/quiltdata/quilt/blob/f476bf61b08c83e4541c9d234973ae82d99307eb/api/python/quilt3/packages.py#L381) defined which returns the user metadata or an empty dict.

akarve:
The purpose of this change is to allow the user to get *top level* package metdata. Currently, the user can call `p.set_meta`, or `p.set_dir(".", meta=)` but there is no way to retrieve that data. We should not interfere with set/get on PackageEntry as there is already a, albeit clumsy, way to get metadata.

api/python/quilt3/packages.py
akarve: True but look at how it's used. We want to extract the `user_meta` portion. Also, we shouldn't have to do anything with PackageEntry as I mentioned below.

sir-sigurd:
@akarve 
> The purpose of this change is to allow the user to get top level package metdata.

We have it:
```python
In [3]: Package().set_meta({'key': 'value'}).meta
Out[3]: {'key': 'value'}
```

So I suggest to simply move `Package.meta` right above `Package.set_meta()` :slightly_smiling_face: .

PR #1678, opened 2020-06-22
PackageRegistry v2

sir-sigurd:
## TODO

- [x] deduplicate tests
- [x] more tests for package and package revision removal
- [x] allow users to select registry version
- [x] add comments


api/python/quilt3/backends/local.py
```diff
@@ -20,18 +22,52 @@ def list_packages(self):
...
+        shutil.rmtree(self._pointers_usr_dir(pkg_name).path)
+
+    def delete_package_version(self, pkg_name: str, top_hash: str):
+        super().delete_package_version(pkg_name, top_hash)
+        delete_url(self.pointers_dir(pkg_name))
+        delete_url(self._pointers_usr_dir(pkg_name))
+
+
+class LocalPackageRegistryV2(PackageRegistryV2):
+    _from_path_to_package_name = operator.methodcaller('replace', '@', '/')
```
kevinemoore, line 41: This could be a good opportunity to drop the <user>/<name> requirement for package names.

api/python/quilt3/backends/local.py
```diff
@@ -20,18 +22,52 @@ def list_packages(self):
...
+    def delete_package(self, pkg_name: str):
+        shutil.rmtree(self._pointers_usr_dir(pkg_name).path)
+
+    def delete_package_version(self, pkg_name: str, top_hash: str):
+        super().delete_package_version(pkg_name, top_hash)
+        delete_url(self.pointers_dir(pkg_name))
+        delete_url(self._pointers_usr_dir(pkg_name))
+
+
+class LocalPackageRegistryV2(PackageRegistryV2):
```
kevinemoore, line 40: This could use a class comment describing the layout of the files in a local V2 registry.

api/python/quilt3/backends/s3.py
```diff
@@ -13,8 +26,46 @@ def list_packages(self):
...
         package_dir = self.pointers_dir(pkg_name)
         for path, _ in list_url(package_dir):
             pkg_hash = get_bytes(package_dir.join(path))
             yield path, pkg_hash.decode().strip()
+
+    def delete_package(self, pkg_name: str):
+        delete_url_recursively(self.pointers_dir(pkg_name))
+
+
+class S3PackageRegistryV2(PackageRegistryV2):
```
kevinemoore, line 39: This could use a class comment describing the layout of the files in an S3 V2 registry.

sir-sigurd:
> It might be intended for a later PR, but I don't see how users migrate packages from v1 to v2. Is there going to be a migration command?

Yes, it will be implemented in following PR.

PR #1679, opened 2020-06-25
More robust preview and indexing extraction for Parquet files

akarve:
* No longer depend on row_groups in parquet extraction (there might not be any, e.g. if columns or now rows)
* Require less memory when previewing parquet files (pyarrow supports slices)
* Improve unit test coverage to cover parquet extraction and file type inference
* Send only columns (as opposed to full schema) to ES when indexing Parquet files


lambdas/es/indexer/test/test_index.py
```diff
@@ -336,6 +336,13 @@ def _test_index_events(
                 )
 
             if mock_object:
+                if mock_overrides and 'skip_byte_range' in mock_overrides:
```
akarve, line 339:
```suggestion
                if mock_overrides and mock_overrides.get('skip_byte_range'):
```

PR #1682, opened 2020-06-25
PackageRegistry abstraction

sir-sigurd:
## Description
The first commit adds `LocalPackage` abstraction layer, the second one reworks methods in a more backend specific way to demonstrate why I add `S3PackageRegistryV1` and `LocalPackageRegistryV1`.
In some near future I'm going to move S3 specific stuff under `backends.s3` and add backend specific `PhysicalKey` with helper methods like `get_bytes()` etc.

## TODO
 - [x] - more polishing?

### DIFF
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -1,11 +1,11 @@
-from .data_transfer import copy_file, get_bytes, delete_url, list_url
-from .packages import Package
+from .backends import get_package_registry
+from .data_transfer import copy_file
 from .search_util import search_api
 from .util import (QuiltConfig, QuiltException, CONFIG_PATH,
                    CONFIG_TEMPLATE, configure_from_default, config_exists,
-                   configure_from_url, fix_url, get_package_registry,
+                   configure_from_url, fix_url,
                    load_config, PhysicalKey, read_yaml, validate_package_name,
-                   write_yaml, get_from_config)
+                   write_yaml)
 from .telemetry import ApiTelemetry
 
 
@@ -34,50 +34,7 @@ def delete_package(name, registry=None, top_hash=None):
         top_hash (str): Optional. A package hash to delete, instead of the whole package.
     """
     validate_package_name(name)
-    usr, pkg = name.split('/')
-
-    registry = (
-        get_from_config('default_local_registry')
-        if registry is None else
-        fix_url(registry)
-    )
-    registry_parsed = PhysicalKey.from_url(get_package_registry(registry))
-    named_packages = registry_parsed.join('named_packages')
-    package_path = named_packages.join(name)
-
-    paths = list(list_url(package_path))
-    if not paths:
-        raise QuiltException("No such package exists in the given directory.")
-
-    if top_hash is not None:
-        top_hash = Package.resolve_hash(name, PhysicalKey.from_url(registry), top_hash)
-        deleted = []
-        remaining = []
-        for path, _ in paths:
-            parts = path.split('/')
-            if len(parts) == 1:
-                pkg_hash = get_bytes(package_path.join(parts[0]))
-                if pkg_hash.decode().strip() == top_hash:
-                    deleted.append(parts[0])
-                else:
-                    remaining.append(parts[0])
-        if not deleted:
-            raise QuiltException("No such package version exists in the given directory.")
-        for path in deleted:
-            delete_url(package_path.join(path))
-        if 'latest' in deleted and remaining:
-            # Create a new "latest". Technically, we need to compare numerically,
-            # but string comparisons will be fine till year 2286.
-            new_latest = max(remaining)
-            copy_file(package_path.join(new_latest), package_path.join('latest'))
-    else:
-        for path, _ in paths:
-            delete_url(package_path.join(path))
-
-    # Will ignore non-empty dirs.
-    # TODO: .join('') adds a trailing slash - but need a better way.
-    delete_url(package_path.join(''))
-    delete_url(named_packages.join(usr).join(''))
+    get_package_registry(registry).delete_package(name, top_hash)
 
 
 @ApiTelemetry("api.list_packages")
@@ -93,31 +50,7 @@ def list_packages(registry=None):
     Returns:
         An iterable of strings containing the names of the packages
     """
-    registry_parsed = PhysicalKey.from_url(get_package_registry(fix_url(registry) if registry else None))
-
-    return _list_packages(registry_parsed)
-
-
-def _list_packages(registry):
-    """This differs from list_packages because it does not have
-
-    telemetry on it. If Quilt code needs the functionality to list
-    packages under a different customer-facing API, _list_packages()
-    is the function that should be used to prevent duplicate metrics
-    (each API call that the user makes should generate a single
-    telemetry event).
-    """
-
-    named_packages = registry.join('named_packages')
-    prev_pkg = None
-    for path, _ in list_url(named_packages):
-        parts = path.split('/')
-        if len(parts) == 3:
-            pkg = f'{parts[0]}/{parts[1]}'
-            # A package can have multiple versions, but we should only return the name once.
-            if pkg != prev_pkg:
-                prev_pkg = pkg
-                yield pkg
+    return get_package_registry(registry).list_packages()
 
 
 @ApiTelemetry("api.list_package_versions")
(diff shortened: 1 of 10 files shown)


api/python/quilt3/backends/s3.py
```diff
@@ -0,0 +1,20 @@
...
+
+from .base import PackageRegistryV1
+
+
+class S3PackageRegistryV1(PackageRegistryV1):
+    def list_packages(self):
+        prev_pkg = None
+        for path, _ in list_url(self.pointers_global_dir):
+            pkg = path.rpartition('/')[0]
+            # A package can have multiple versions, but we should only return the name once.
```
kevinemoore, line 11: Does this rely on ordering for de-duplication? Is ordering guaranteed?

api/python/quilt3/backends/base.py
```diff
@@ -0,0 +1,174 @@
...
+    def push_manifest(self, pkg_name: str, top_hash: str, manifest_data: bytes):
+        put_bytes(manifest_data, self.manifest_pk(pkg_name, top_hash))
+        hash_bytes = top_hash.encode()
+        named_path = self.pointers_dir(pkg_name)
+        # TODO: use a float to string formatter instead of double casting
+        timestamp_path = named_path.join(str(int(time.time())))
+        put_bytes(hash_bytes, timestamp_path)
+        put_bytes(hash_bytes, self.pointer_latest_pk(pkg_name))
+
+    def resolve_top_hash(self, pkg_name: str, hash_prefix: str) -> str:
```
kevinemoore, line 145: To support the existing API, don't we need a way to resolve short hashes across all package names?

api/python/quilt3/packages.py
```diff
@@ -486,37 +479,9 @@ def resolve_hash(cls, registry, hash_prefix):
...
-        matches = [h for h, _ in list_url(registry.join('.quilt/packages/'))
-                   if h.startswith(top_hash[:min_shorthash_len])]
-        if len(matches) == 0:
-            raise ValueError(f"Tophash {top_hash} was not found in registry {registry}")
-        for prefix_length in range(min_shorthash_len, 64):
-            potential_shorthash = top_hash[:prefix_length]
-            matches = [h for h in matches if h.startswith(potential_shorthash)]
-            if len(matches) == 1:
-                return potential_shorthash
+        return get_package_registry(registry).resolve_top_hash(None, hash_prefix)
```
kevinemoore, line 484: Does this work?

api/python/quilt3/backends/__init__.py
```diff
@@ -0,0 +1,19 @@
+from quilt3.util import PhysicalKey, get_from_config, fix_url
+
+from .base import PackageRegistry
+from .local import LocalPackageRegistryV1
+from .s3 import S3PackageRegistryV1
+
+
+def get_package_registry(path=None) -> PackageRegistry:
```
kevinemoore, line 8: Let's make have a signature that's consistent with `PhysicalKey.from_path`. Users should also have access to PhysicalKey since that class is part of the API (e.g., `Package.resolve_hash`).

api/python/quilt3/backends/s3.py
sir-sigurd: Yes, [it's guaranteed](https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysUsingAPIs.html):

> List results are always returned in UTF-8 binary order.

api/python/quilt3/backends/__init__.py
sir-sigurd: `Package.resolve_hash()` was added in https://github.com/quiltdata/quilt/commit/8efb0b6a9bac8108b9fde542c0465d487ae6f01f and `registry` parameter was `str`. `PhysicalKey` was added in https://github.com/quiltdata/quilt/pull/1478 and was intended to be private. This PR also made `Package.shorten_tophash()` [private](https://github.com/quiltdata/quilt/pull/1478#discussion_r362641345). So `Package.resolve_hash()` worked for users only in 3.1.7 and 3.1.8. I think we should either make it private or make it take `str` again. Make it private would be easier.

api/python/quilt3/packages.py
sir-sigurd: Well, yes:
https://github.com/quiltdata/quilt/pull/1682/files#diff-60b178a9c2fe184f8e58ee3927efc3ecR105-R106

api/python/quilt3/backends/base.py
sir-sigurd: If we decide to keep `Package.resolve_hash()` public we should add `pkg_name` parameter for it. It would be optional for V1 registry and required for V2 registry.

api/python/quilt3/api.py
```diff
@@ -1,11 +1,11 @@
-from .data_transfer import copy_file, get_bytes, delete_url, list_url
-from .packages import Package
+from .backends import get_package_registry
+from .data_transfer import copy_file
 from .search_util import search_api
 from .util import (QuiltConfig, QuiltException, CONFIG_PATH,
```
akarve, line 4: minor: at this scale, one import per line reads better

api/python/quilt3/backends/s3.py
```diff
@@ -0,0 +1,20 @@
+from quilt3.data_transfer import list_url, get_bytes
+
+from .base import PackageRegistryV1
+
+
+class S3PackageRegistryV1(PackageRegistryV1):
+    def list_packages(self):
+        prev_pkg = None
+        for path, _ in list_url(self.pointers_global_dir):
+            pkg = path.rpartition('/')[0]
```
akarve, line 10: `rpartition` is a strange choice as it will fail if there are multiple slashes in the (sub) package name? or can that never happen?

api/python/quilt3/backends/base.py
```diff
@@ -0,0 +1,174 @@
...
+        package_path = self.pointers_dir(pkg_name)
+        paths = list(list_url(package_path))
+        if not paths:
+            raise QuiltException("No such package exists in the given directory.")
+
+        if top_hash is not None:
+            top_hash = self.resolve_top_hash(pkg_name, top_hash)
+            deleted = []
+            remaining = []
+            for path, pkg_hash in self.list_package_versions(pkg_name):
```
akarve, line 121: this is dangerous if multiple packages have the same path. this can actually happen, e.g. if someone has an empty package with several names. or, better example, a `foo/dev` gets pushed to `foo/prod` and user deletes `foo/prod`. we can punt on deep_delete until v2 where manifests are not shared across handles.

api/python/quilt3/backends/base.py
```diff
@@ -0,0 +1,174 @@
...
+                raise QuiltException("Found multiple matches: %r" % hash_prefix)
+            else:
+                top_hash = matching_hashes[0]
+        else:
+            raise QuiltException("Invalid hash: %r" % hash_prefix)
+        # TODO: verify that name is correct with respect to this top_hash
+        return top_hash
+
+    def shorten_top_hash(self, pkg_name: str, top_hash: str) -> str:
+        min_shorthash_len = 7
```
akarve, line 164: a few lines above the minimum is 6; i guess that's fine if you're supporting legacy short hashes of 6 chars, but maybe a comment and why not a constant for this: `MIN_SHORTHASH_LEN`?

api/python/quilt3/backends/s3.py
sir-sigurd: We allow only a single slash in the package name, so it should work unless some put file `.quilt/named_packages/foo/bar/baz/bla`, but I think we assume that registry is not broken.

api/python/quilt3/backends/base.py
sir-sigurd: This is just a copy of [the current implementation](https://github.com/quiltdata/quilt/blob/ad3b381657f07befc818930d83cae2766300fc2d/api/python/quilt3/packages.py#L480-L519). If we want to change something, we should do it in `master`, but not here.

api/python/quilt3/api.py
sir-sigurd: Let's fix with [isort](https://isort.readthedocs.io/en/stable/) later.

api/python/quilt3/backends/__init__.py
kevinemoore: I don't mind making it private. It does seem useful to be able to read a package from a short hash though.

api/python/quilt3/backends/base.py
akarve: Let's deprecate and make private in V2.

api/python/quilt3/backends/s3.py
akarve: OK

api/python/quilt3/backends/__init__.py
kevinemoore, line 8: Can we keep get_package_registry private (used only internal to the library)? If we make it public, it looks like we're duplicating a lot of API surface (and hiding telemetry in the process).

PR #1684, opened 2020-06-27
Extract_parquet: don't deserialize large tables, trim metadata returned, unit tests

akarve:


lambdas/shared/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pandas==0.25.1
+pyarrow==0.17.1
```
kevinemoore, line 2: You're sure this doesn't add pandas and pyarrow to all lambdas that include t4-lambda-shared?

PR #1689, opened 2020-06-30
Environment variable to skip rows for certain file extensions

akarve:
In some cases (e.g. Parquet data warehouse) minimizes disk footprint and noisy text in ES


lambdas/es/indexer/index.py
```diff
@@ -134,7 +135,11 @@ def get_contents(bucket, key, ext, *, etag, version_id, s3_client, size):
                 s3_client=s3_client,
                 version_id=version_id
             )
-            body, info = extract_parquet(get_bytes(obj["Body"], compression), as_html=False)
+            body, info = extract_parquet(
+                get_bytes(obj["Body"], compression),
+                as_html=False,
+                skip_rows=('.parquet' in SKIP_ROWS_EXTS)
```
kevinemoore, line 141: Hard-coding the file extension here seems a bit fragile, or perhaps just confusing because the previewer (and indexer?) recognizes multiple extensions for parquet data. Seems like it would be more consistent to trigger this switch on the `input` (input_type) instead of the extension. Maybe make a single flag to control indexing and/or previewing all row/col data (parquet, csv, etc.)?

PR #1690, opened 2020-06-30
Deprecate passing sub-package path to Package.install() via package name

sir-sigurd:
## TODO ##

- [x] regen docs once approved

### DIFF
--- a/api/python/quilt3/main.py
+++ b/api/python/quilt3/main.py
@@ -305,7 +305,9 @@ def create_parser():
     install_p = subparsers.add_parser("install", description=shorthelp, help=shorthelp, allow_abbrev=False)
     install_p.add_argument(
         "name",
-        help="Name of package, in the USER/PKG[/PATH] format",
+        help=(
+            "Name of package, in the USER/PKG[/PATH] format ([/PATH] is deprecated, use --path parameter instead)"
+        ),
         type=str,
     )
     install_p.add_argument(
@@ -329,6 +331,12 @@ def create_parser():
     install_p.add_argument(
         "--dest-registry",
         help="Registry to install package to. Defaults to local registry.",
+        type=str,
+        required=False,
+    )
+    install_p.add_argument(
+        "--path",
+        help="If specified, downloads only PATH or its children.",
         type=str,
         required=False,
     )
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -24,7 +24,7 @@ from .telemetry import ApiTelemetry
 from .util import (
     QuiltException, fix_url, get_from_config, get_install_location,
     validate_package_name, quiltignore_filter, validate_key, extract_file_extension,
-    parse_sub_package_name)
+    parse_sub_package_name, RemovedInQuilt4Warning)
 from .util import CACHE_PATH, TEMPFILE_DIR_PATH as APP_DIR_TEMPFILE_DIR, PhysicalKey, \
     user_is_configured_to_custom_stack, catalog_package_url, DISABLE_TQDM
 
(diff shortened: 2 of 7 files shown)


api/python/quilt3/main.py
```diff
@@ -305,7 +305,9 @@ def create_parser():
     install_p = subparsers.add_parser("install", description=shorthelp, help=shorthelp, allow_abbrev=False)
     install_p.add_argument(
         "name",
-        help="Name of package, in the USER/PKG[/PATH] format",
+        help=(
+            "Name of package, in the USER/PKG[/PATH] format (passing PATH is deprecated, use --path parameter instead)"
```
akarve, line 309:
```suggestion
            "Name of package, in the USER/PKG[/PATH] format ([/PATH] is deprecated, use --path parameter instead)"
```

api/python/quilt3/packages.py
```diff
@@ -383,19 +383,22 @@ def meta(self):
...
     @ApiTelemetry("package.install")
-    def install(cls, name, registry=None, top_hash=None, dest=None, dest_registry=None):
+    def install(cls, name, registry=None, top_hash=None, dest=None, dest_registry=None, *, path=None):
         """
         Installs a named package to the local registry and downloads its files.
 
         Args:
-            name(str): Name of package to install. It also can be passed as NAME/PATH,
+            name(str): Name of package to install. It also can be passed as NAME/PATH
+                (deprecated, use 'path' parameter instead),
```
akarve, line 392:
```suggestion
                (/PATH is deprecated, use 'path=' keyword instead),
```

api/python/quilt3/main.py
```diff
@@ -332,6 +334,12 @@ def create_parser():
         type=str,
         required=False,
     )
+    install_p.add_argument(
+        "--path",
+        help="If specified, only the sub-package or the entry specified by PATH will be downloaded.",
```
akarve, line 339:
```suggestion
        help="If specified, downloads only the children of PATH.",
```

api/python/quilt3/packages.py
sir-sigurd: I think `keyword` is not quite accurate. See https://docs.python.org/3/reference/lexical_analysis.html#keywords. I think we should either say `use the 'path' parameter` or `pass it as 'path' keyword argument`. See https://docs.python.org/3/glossary.html#term-parameter and https://docs.python.org/3/glossary.html#term-argument. Which one makes more sense?

api/python/tests/integration/test_packages.py
```diff
@@ -1393,10 +1393,20 @@ def test_install(self):
...
             )))
 
+    def test_install_subpackage_deprecated_and_new(self):
+        pkg_name = 'Quilt/Foo'
+        bucket = 'my-test-bucket'
+        path = 'baz'
+        dest = 'package'
+
+        with pytest.warns(RemovedInQuilt4Warning):
+            with pytest.raises(ValueError):
```
akarve, line 1403: I'm not clear why we also raise a hard ValueError if thi sis a soft deprecation?

sir-sigurd: We raise `ValueError` when the path is passed both via package name and the `path` parameter.

sir-sigurd: i.e. it doesn't make sense to call `install('usr/pkg/path1', path='path2')`.

api/python/quilt3/main.py
sir-sigurd: Is it still clear (was it?) that the path can be a full path to a single file?

akarve: Rewritten.

api/python/quilt3/packages.py
akarve: I tired argument to cover both because its kwarg and positional? See if you like it. Otherwise go with your best estimate.

PR #1693, opened 2020-07-01
Check memory before deserializing large Parquet files during indexing

akarve:


lambdas/es/indexer/index.py
```diff
@@ -77,6 +78,11 @@
 USER_AGENT_EXTRA = " quilt3-lambdas-es-indexer"
 
 
+def get_available_memory():
+    """how much virtual memory is available to us (bytes)?"""
+    return getattr(virtual_memory(), 'available')
```
kevinemoore, line 83: Why not `virtual_memory().available`?

akarve, line 83:
```suggestion
    return virtual_memory().available
```

PR #1694, opened 2020-07-01
Move get_available_memory to shared, add some tests

akarve:
Use memory to decide if we should deserialize parquet

### DIFF
--- a/lambdas/shared/t4_lambda_shared/preview.py
+++ b/lambdas/shared/t4_lambda_shared/preview.py
@@ -10,7 +10,7 @@ import zlib
 from .utils import get_available_memory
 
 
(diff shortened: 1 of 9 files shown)


lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -7,7 +7,10 @@
 import re
 import zlib
 
+from .utils import get_available_memory
 
+
+AVG_PARQUET_CELL_BYTES = 100  # a heuristic to avoid floding memory
```
akarve, line 13:
```suggestion
AVG_PARQUET_CELL_BYTES = 100  # a heuristic to avoid flooding memory
```

PR #1701, opened 2020-07-13
Add update policy to set_dir

cosmic-byte:
## Description
Add `update_policy` to `set_dir` of package


api/python/quilt3/packages.py
```diff
@@ -784,8 +789,14 @@ def set_dir(self, lkey, path=None, meta=None):
             for f in files:
                 if not f.is_file():
                     continue
-                entry = PackageEntry(PhysicalKey.from_path(f), f.stat().st_size, None, None)
+
                 logical_key = f.relative_to(src_path).as_posix()
+
+                # check update policy
+                if update_policy == existing_policy and root.__contains__(logical_key):
```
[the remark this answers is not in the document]
cosmic-byte: I don't think there's any need to explicitly check for `incoming`? since the package entry will still be created if the `logical_key` doesn't exist.
With our current use case, `incoming` update policy and non existing `logical_key` proceeds to create/update an entry.

api/python/quilt3/packages.py
```diff
@@ -784,8 +802,12 @@ def set_dir(self, lkey, path=None, meta=None):
             for f in files:
                 if not f.is_file():
                     continue
-                entry = PackageEntry(PhysicalKey.from_path(f), f.stat().st_size, None, None)
                 logical_key = f.relative_to(src_path).as_posix()
+                # check update policy
+                if update_policy == PackageUpdatePolicy.EXISTING.value and root.__contains__(logical_key):
```
sir-sigurd, line 807:
```suggestion
                if update_policy == PackageUpdatePolicy.EXISTING.value and logical_key in root:
```

api/python/quilt3/packages.py
```diff
@@ -305,6 +306,15 @@ def physical_keys(self):
         return [self.physical_key]
 
 
+class PackageUpdatePolicy(enum.Enum):
+    INCOMING = 'incoming'
+    EXISTING = 'existing'
+
+    @classmethod
+    def has_value(cls, value):
+        return value in cls._value2member_map_
```
sir-sigurd, line 315: I'd rather [omit values](https://docs.python.org/3/library/enum.html#omitting-values) and in `set_dir()` check that `type(update_policy) is PackageUpdatePolicy`.

cosmic-byte: checking `type(update_policy)` would require `update_policy` argument to be strictly an enum of type `PackageUpdatePolicy` which implies that passing string `incoming` or `existing` would fail.

api/python/quilt3/packages.py
```diff
@@ -803,8 +825,13 @@ def set_dir(self, lkey, path=None, meta=None):
                         warnings.warn(f'Logical keys cannot end in "/", skipping: {obj["Key"]}')
                     continue
                 obj_pk = PhysicalKey(src.bucket, obj['Key'], obj.get('VersionId'))
-                entry = PackageEntry(obj_pk, obj['Size'], None, None)
                 logical_key = obj['Key'][len(src_path):]
+
+                # check update policy
+                if update_policy == PackageUpdatePolicy.EXISTING.value and root.__contains__(logical_key):
```
akarve, line 831: `logical_key in`...

api/python/quilt3/packages.py
sir-sigurd: I think this behavior is desirable.
@akarve what do you think?

akarve: I'm emulating pandas kwargs like parser=STRING. Easier for the user to just pass a string. I initially told Greg to use an enum but it's looking clumsy. I think we just do:
```
UPDATE_POLICY = set(
    'incoming',
    'existing',
)
```

And check `if update_policy in UPDATE_POLICIES`.

cosmic-byte: Alright, I will update now.

api/python/quilt3/util.py
```diff
@@ -25,6 +25,7 @@
 
 PACKAGE_NAME_FORMAT = r"([\w-]+/[\w-]+)(?:/(.+))?$"
 DISABLE_TQDM = os.getenv('QUILT_MINIMIZE_STDOUT', '').lower() == 'true'
+PACKAGE_UPDATE_POLICY = {'incoming', 'existing', }
```
sir-sigurd, line 28:
```suggestion
PACKAGE_UPDATE_POLICY = {'incoming', 'existing'}
```

api/python/quilt3/packages.py
```diff
@@ -748,13 +748,21 @@ def set_dir(self, lkey, path=None, meta=None):
...
 
         Returns:
             self
 
         Raises:
             When `path` doesn't exist
+            ValueError: if the 'update_policy' is not a PackageUpdatePolicy enum value.
         """
+        if update_policy not in PACKAGE_UPDATE_POLICY:
+            raise ValueError(f"{update_policy} is not a valid update policy")
```
akarve, line 764: We should tell users how to fix the problem, not just "there was a problem" :)
```suggestion
            raise ValueError(f"Update policy should be one of {str(PACKAGE_UPDATE_POLICY)}, not {update_policy}")
```

api/python/quilt3/packages.py
```diff
@@ -784,8 +792,12 @@ def set_dir(self, lkey, path=None, meta=None):
             for f in files:
                 if not f.is_file():
                     continue
-                entry = PackageEntry(PhysicalKey.from_path(f), f.stat().st_size, None, None)
                 logical_key = f.relative_to(src_path).as_posix()
+                # check update policy
+                if update_policy == 'existing' and logical_key in root:
+                    continue
+
```
akarve, line 799: Let's be consistent with spacing. Remove this newline to match below?

api/python/quilt3/packages.py
```diff
@@ -748,13 +748,21 @@ def set_dir(self, lkey, path=None, meta=None):
             path(string): path to scan for files to add to package.
                 If None, lkey will be substituted in as the path.
             meta(dict): user level metadata dict to attach to lkey directory entry.
+            update_policy(str): can be either 'incoming' (default) or 'existing'.
+                If 'incoming', whenever logical keys match, always take the new entry from set_dir.
+                If 'existing', whenever logical keys match, retain existing entries
+                and ignore new entries from set_dir.
```
akarve, line 754: New docstrings so you'll want to gendocs and add the .md changes in this PR

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
 
             list_object_versions_mock.assert_called_with('bucket', 'foo/')
 
+    def test_set_dir_update_policy(self):
+        """Verify building a package with update policy. """
+
+        # Test for local src dir
+        foodir = pathlib.Path("foo_dir")
```
akarve, line 673: Cleaner and more biotic to use the `/data` directory

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
...
+        bazdir = pathlib.Path("baz_dir")
+        foodir.mkdir()
+        bazdir.mkdir()
+        with open(foodir / 'bar', 'w') as fd:
+            fd.write(fd.name)
+        with open(bazdir / 'bar', 'w') as fd:
+            fd.write(''.join([str(i) for i in range(100)]))
+
+        pkg = Package()
+        pkg = pkg.set_dir("/", ".", meta={'name': 'test_meta'})
```
akarve, line 683: If we're going to set the meta, we should read it out in the context of the test. e.g. check that metadata are preserved after a `set_dir`.

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
...
+        assert 'foo_dir' in pkg.keys()
+        existing_bar_size = pkg['foo_dir/bar'].size
+
+        # existing update policy
+        pkg = pkg.set_dir("foo_dir", bazdir, update_policy='existing')
+        assert 'foo_dir' in pkg.keys()
+        assert existing_bar_size == pkg['foo_dir/bar'].size
```
akarve, line 690: We should really check metadata and hashes to be certain things haven't changed at all. In fact, copy the PackageEntry and check for equality?

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
...
+
+        # incoming update policy
+        pkg = pkg.set_dir("foo_dir", bazdir)
+        assert 'foo_dir' in pkg.keys()
+        assert existing_bar_size != pkg['foo_dir/bar'].size
```
akarve, line 695: See above comments. Let's be more exhaustive; not just checking size.

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
...
+
+        # verify non existing update policy raises value error
+        with pytest.raises(ValueError):
+            pkg.set_dir("foo_dir", foodir, update_policy='invalid_policy')
+
+        # Test for s3 src dir
+        with patch('quilt3.packages.list_object_versions') as list_object_versions_mock:
+            pkg = Package()
```
akarve, line 703: Good thing to test! the Package() construction should take place right before you use it. So move this down and your mock.return_value up.

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
...
+
+            list_object_versions_mock.return_value = ([
```
akarve, line 705: Might as well return more than one object since this is a set_dir call.

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +666,61 @@ def test_s3_set_dir(self):
...
+                dict(Key='foo/a.txt', VersionId='xyz', IsLatest=True, Size=10),
+            ], [])
+
+            pkg.set_dir('', 's3://bucket/foo/', meta={'name': 'test_meta'})
+            assert pkg['a.txt'].get() == 's3://bucket/foo/a.txt?versionId=xyz'
+
+            list_object_versions_mock.return_value = ([
```
akarve, line 712: It's cleaner to do this in a separate test or in a loop. You can do `with patch` and loop over the various update policies.

api/python/tests/integration/test_packages.py
akarve: If you haven't already, do a reall list_object_versions call and check your mock against that. Try to match all fields.

akarve: Since you are mocking an S3 call, we should be mocking this with boto3, so use `self.s3_stubber.add_response`

api/python/tests/integration/test_packages.py
akarve: The test should check the entire state of the in-memory package is my main point; so whatever it takes to get there.

api/python/tests/integration/test_packages.py
cosmic-byte: I'm replicating what we already have for testing [s3_set_dir](https://github.com/quiltdata/quilt/blob/bd1a3fa69f3a3332e69d7ba46b16692497cbeb0c/api/python/tests/integration/test_packages.py#L504). This time just adding update policy

akarve: That's not really correct on a few levels: first because we should be using boto3, and second because when you have the same test with small variations we should probably be using [parameterize](https://github.com/quiltdata/quilt/blob/master/lambdas/thumbnail/tests/test_thumbnail.py#L37) or similar. Have a look. I think it's worth rewriting the test for `set_dir` entirely and doing it with parameterize. But you'll have to feel it out by trying it.

akarve: Actually, skip parameterize and rewriting the set_dir test for now, but definitely use boto3 in your new test. If that fails you can do it the way Kevin did in `set_dir`.

cosmic-byte: Looking at this again, I think it's best to leave it as it is (the way Kevin did in `set_dir`). 
Reason being that `list_object_versions_mock ` is for `quilt3.packages.list_object_versions` which is a [recursive method](https://github.com/quiltdata/quilt/blob/4d03796b433e693443a2f535dfe526211d6fff17/api/python/quilt3/data_transfer.py#L581) in data_trannsfer. So it doesn't have any direct call with s3 boto call.

api/python/quilt3/packages.py
```diff
@@ -748,13 +748,21 @@ def set_dir(self, lkey, path=None, meta=None):
...
 
         Returns:
             self
 
         Raises:
             When `path` doesn't exist
+            ValueError: if the 'update_policy' is not in PACKAGE_UPDATE_POLICY set.
         """
+        if update_policy not in PACKAGE_UPDATE_POLICY:
+            raise ValueError(f"Update policy should be one of {str(PACKAGE_UPDATE_POLICY)}, not {update_policy}")
```
sir-sigurd, line 764:
```suggestion
            raise ValueError(f"Update policy should be one of {PACKAGE_UPDATE_POLICY}, not {update_policy!r}")
```

api/python/quilt3/packages.py
```diff
@@ -748,6 +748,10 @@ def set_dir(self, lkey, path=None, meta=None):
             path(string): path to scan for files to add to package.
                 If None, lkey will be substituted in as the path.
             meta(dict): user level metadata dict to attach to lkey directory entry.
+            update_policy(str): can be either `incoming` (default) or `existing`.
+                If `incoming`, then we overwrite existing matching keys.
+                If `existing`, then we skip entry for matching keys.
```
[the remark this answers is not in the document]
sir-sigurd: @akarve
actually we might want to use `` `'existing'` `` here.

api/python/quilt3/packages.py
```diff
@@ -748,13 +748,21 @@ def set_dir(self, lkey, path=None, meta=None):
...
+                If 'incoming', whenever logical keys match, always take the new entry from set_dir.
+                If 'existing', whenever logical keys match, retain existing entries
+                and ignore new entries from set_dir.
 
         Returns:
             self
 
         Raises:
             When `path` doesn't exist
+            ValueError: if the 'update_policy' is not in PACKAGE_UPDATE_POLICY set.
```
sir-sigurd, line 761:
```suggestion
            ValueError: if the `update_policy` is not in PACKAGE_UPDATE_POLICY set.
```
I think we shouldn't refer to `PACKAGE_UPDATE_POLICY` here.
@akarve could you help with wording?

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +672,70 @@ def test_s3_set_dir(self):
 
             list_object_versions_mock.assert_called_with('bucket', 'foo/')
 
+    def test_set_dir_update_policy(self):
+        """Verify building a package with update policy. """
+
+        # Test for local src dir
+        path = pathlib.Path(__file__).parent
+        data_dir = pathlib.Path(path, "data")
```
akarve, line 680: Delete above line and 
```suggestion
        data_dir = pathlib.Path(__file__).parent / "data"
```

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +672,70 @@ def test_s3_set_dir(self):
...
+        pkg = Package()
+        pkg = pkg.set_dir("/", data_dir, meta={'name': 'test_meta'})
+        assert 'nested' in pkg.keys()
+        assert pkg.meta == {'name': 'test_meta'}
+        existing_one_entry = pkg['nested/one.txt']
+
+        nested_dir = pathlib.Path(data_dir, "nested2")
+        nested_dir.mkdir(parents=True, exist_ok=True)
+        with open(nested_dir / 'one.txt', 'w') as fd:
```
akarve, line 689: we shouldn't modify existing assets to avoid side-effects; i'm not even sure we clean this up in tearDown. You can instead add a file to data as part of this file if you need it.

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +672,70 @@ def test_s3_set_dir(self):
...
+            fd.write(fd.name)
+
+        # existing update policy
```
akarve, line 692: it's cleaner to use pytest.parameterize and make for multiple invocations of this unit test function with different parameters: one with "existing", one with "incoming", etc. See https://github.com/quiltdata/quilt/blob/master/lambdas/thumbnail/tests/test_thumbnail.py#L46. This avoids side-effects, is closer to how things are called in the real world, etc.

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +672,70 @@ def test_s3_set_dir(self):
...
+        assert 'nested' in pkg.keys()
+        assert existing_one_entry != pkg['nested/one.txt']
+
+        # verify non existing update policy raises value error
+        expected_err = f"Update policy should be one of {str(PACKAGE_UPDATE_POLICY)}, not 'invalid_policy'"
+        with pytest.raises(ValueError) as e:
+            pkg.set_dir("nested", data_dir, update_policy='invalid_policy')
+        assert expected_err in str(e.value)
+
+        # remove created directory
```
akarve, line 708: same here; let's not modify the data directory inside the test (again you can add data outside the test) and/or change directories, etc.

api/python/tests/integration/test_packages.py
akarve: this is fine but you need to check that these mocks fired; see `.assert_has_calls` and check `.call_count == N`

api/python/tests/integration/test_packages.py
```diff
@@ -666,6 +672,70 @@ def test_s3_set_dir(self):
...
+            assert pkg['a.txt'].get() == 's3://bucket/foo/a.txt?versionId=xyz'
+            assert pkg['x/y.txt'].get() == 's3://bucket/foo/x/y.txt?versionId=null'
+
+            list_object_versions_mock.return_value = ([
+                dict(Key='bar/a.txt', VersionId='abc', IsLatest=True, Size=10),
+                dict(Key='bar/x/y.txt', VersionId='null', IsLatest=True, Size=10),
+                dict(Key='bar/z.txt', VersionId='123', IsLatest=False, Size=10),
+            ], [])
+
+            # existing update policy
```
akarve, line 729: once you parameterize you won't have this duplicated code. same for existing/incoming blocks above. whenever we have very similar code like that parameterize let's us consolidate the logic—very easy to maintain and extend as well.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
 
         with pytest.raises(QuiltException, match='Found multiple matches'):
             Package.resolve_hash(pkg_name, LOCAL_REGISTRY, hash_prefix)
+
+
+# The following tests were moved out of the PackageTest class to enable parametrization.
+# see (https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses)
+@pytest.mark.parametrize(
+    'update_policy, expected_size, expected_meta, expected_hash',
+    [('incoming', 6, {'name': 'incoming'}, None), ('existing', 1, {'name': 'existing'}, None)]
```
akarve, line 1425: let's put this on four lines

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+
+    data_dir = pathlib.Path(__file__).parent / "data"
+    nested_dir = data_dir / 'nested'
+    nested_dir_2 = data_dir / 'nested2'
+    pkg = Package()
+    pkg = pkg.set_dir("/", nested_dir, meta={'name': 'test_meta'})
+    assert 'one.txt' in pkg.keys()
+    assert pkg.meta == {'name': 'test_meta'}
+
+    pkg = pkg.set_dir("/", nested_dir_2, meta={'name': update_policy}, update_policy=update_policy)
```
akarve, line 1438: it's confusing to use the update_policy in the meta as well. and since we are testing a proper merge, we really need to have _at least_ two children in the directories we are merging.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+    assert 'one.txt' in pkg.keys()
+    assert pkg['one.txt'].size == expected_size
+    assert pkg['one.txt'].hash == expected_hash
+    assert pkg.meta == expected_meta
+
+    # verify non existing update policy raises value error
+    expected_err = f"Update policy should be one of {str(PACKAGE_UPDATE_POLICY)}, not 'invalid_policy'"
```
akarve, line 1445: ok but you can just check for a substring like, 'Update policy should be'; you don't have to match the whole message.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+    with pytest.raises(ValueError) as e:
```
akarve, line 1446: this should be it's own test parameter, i.e. a third case in your parameterize, or standalone test marked with xfail. it doesn't make sense to test this for every test case in parameterize

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+                dict(Key='bar/x/y.txt', VersionId='null', IsLatest=True,
+                     Size=10),
+                dict(Key='bar/z.txt', VersionId='123', IsLatest=False, Size=10),
+            ], [])
+
+        pkg.set_dir('', 's3://bucket/bar', update_policy=update_policy)
+        assert pkg['a.txt'].get() == expected_a_url
+        assert pkg['x/y.txt'].get() == expected_xy_url
+        calls = [call('bucket', 'foo/'), call('bucket', 'bar/')]
+        list_object_versions_mock.assert_has_calls(calls)
```
akarve, line 1485: you can just inline the calls here, easier to read

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+    [('incoming', 6, {'name': 'incoming'}, None), ('existing', 1, {'name': 'existing'}, None)]
+)
+def test_set_dir_update_policy(update_policy, expected_size, expected_meta, expected_hash):
+    """Verify building a package with update policy. """
+
+    data_dir = pathlib.Path(__file__).parent / "data"
+    nested_dir = data_dir / 'nested'
+    nested_dir_2 = data_dir / 'nested2'
+    pkg = Package()
+    pkg = pkg.set_dir("/", nested_dir, meta={'name': 'test_meta'})
```
akarve, line 1434: that reminds me, we should have a test parameter where there's no policy, we use the default kwarg and the test proves it does whatever the default policy is. and it's not really necessary to test metadata here. so do this as a clean parameterized test.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+@pytest.mark.parametrize(
+    'update_policy, expected_size, expected_meta, expected_hash',
+    [('incoming', 6, {'name': 'incoming'}, None), ('existing', 1, {'name': 'existing'}, None)]
+)
+def test_set_dir_update_policy(update_policy, expected_size, expected_meta, expected_hash):
+    """Verify building a package with update policy. """
+
+    data_dir = pathlib.Path(__file__).parent / "data"
+    nested_dir = data_dir / 'nested'
+    nested_dir_2 = data_dir / 'nested2'
```
akarve, line 1432: initialize this right before you use it instead of here

api/python/tests/integration/test_packages.py
akarve, line 1434: let's not assign the results of `set_dir`. use one `pkg` reference throughout.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+    pkg = Package()
+    pkg = pkg.set_dir("/", nested_dir, meta={'name': 'test_meta'})
+    assert 'one.txt' in pkg.keys()
+    assert pkg.meta == {'name': 'test_meta'}
+
+    pkg = pkg.set_dir("/", nested_dir_2, meta={'name': update_policy}, update_policy=update_policy)
+    assert 'one.txt' in pkg.keys()
+    assert pkg['one.txt'].size == expected_size
+    assert pkg['one.txt'].hash == expected_hash
+    assert pkg.meta == expected_meta
```
akarve, line 1442: you're doing the right stuff here but our test folders should really have multiple files so that we are testing and actual merge. and your tests should check that some keys changed and others did not.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+        ('existing', 's3://bucket/foo/a.txt?versionId=xyz', 's3://bucket/foo/x/y.txt?versionId=null'),
+        ('incoming', 's3://bucket/bar/a.txt?versionId=abc', 's3://bucket/bar/x/y.txt?versionId=null')
+    ]
+)
+def test_set_dir_update_policy_s3(update_policy, expected_a_url, expected_xy_url):
+    with patch('quilt3.packages.list_object_versions') as list_object_versions_mock:
+        list_object_versions_mock.return_value = (
+            [
+                dict(Key='foo/a.txt', VersionId='xyz', IsLatest=True, Size=10),
+                dict(Key='foo/x/y.txt', VersionId='null', IsLatest=True,
```
akarve, line 1463: one line?

api/python/tests/integration/test_packages.py
akarve: and remember you need to check the call_count as well. this assert will succeed if it gets called 40 times, even though that's a bug.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+                     Size=10),
+                dict(Key='foo/z.txt', VersionId='123', IsLatest=False, Size=10),
+            ], [])
```
akarve, line 1466: easier to read if each arg and closing base on their own lines

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1416,70 @@ def test_resolve_hash(self):
...
+    nested_dir = data_dir / 'nested'
+    nested_dir_2 = data_dir / 'nested2'
+    pkg = Package()
+    pkg = pkg.set_dir("/", nested_dir, meta={'name': 'test_meta'})
+    assert 'one.txt' in pkg.keys()
+    assert pkg.meta == {'name': 'test_meta'}
+
+    pkg = pkg.set_dir("/", nested_dir_2, meta={'name': update_policy}, update_policy=update_policy)
+    assert 'one.txt' in pkg.keys()
+    assert pkg['one.txt'].size == expected_size
```
akarve, line 1440: You can just check the file contents: https://docs.quiltdata.com/api-reference/package

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,68 @@ def test_resolve_hash(self):
 
         with pytest.raises(QuiltException, match='Found multiple matches'):
             Package.resolve_hash(pkg_name, LOCAL_REGISTRY, hash_prefix)
+
+
+# The following tests were moved out of the PackageTest class to enable parametrization.
+# see (https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses)
+@pytest.mark.parametrize(
+    'update_policy, expected_one_byte',
+    [
```
akarve, line 1433: this is much better! thanks for cleaning this up. a few small things then we can merge:
1. we need to test nested directories (e.g. check what happens in nested/sub)
2. add a test where we don't provide an update_policy and show that it behaves just like `'incoming'`

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,68 @@ def test_resolve_hash(self):
...
+
+        list_object_versions_mock.return_value = (
+            [
+                dict(Key='bar/a.txt', VersionId='abc', IsLatest=True, Size=10),
+                dict(Key='bar/x/y.txt', VersionId='null', IsLatest=True, Size=10),
+                dict(Key='bar/z.txt', VersionId='123', IsLatest=False, Size=10),
+            ],
+            []
+        )
+        pkg.set_dir('', 's3://bucket/bar', update_policy=update_policy)
```
akarve, line 1487: we need some set_dir calls that are not on the root, so `.set_dir('some/dir/in/pkg'...)` for example

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,68 @@ def test_resolve_hash(self):
...
+    [
+        ('existing', 's3://bucket/foo/a.txt?versionId=xyz', 's3://bucket/foo/x/y.txt?versionId=null'),
+        ('incoming', 's3://bucket/bar/a.txt?versionId=abc', 's3://bucket/bar/x/y.txt?versionId=null')
+    ]
+)
+def test_set_dir_update_policy_s3(update_policy, expected_a_url, expected_xy_url):
+    with patch('quilt3.packages.list_object_versions') as list_object_versions_mock:
+        list_object_versions_mock.return_value = (
+            [
+                dict(Key='foo/a.txt', VersionId='xyz', IsLatest=True, Size=10),
```
akarve, line 1467: to be more realistic we really don't want to have the exact same files and folders in `foo` and `bar`. you can think of test coverage in this way: test when SOME keys overlap, when no keys overlap, when all keys overlap.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,68 @@ def test_resolve_hash(self):
...
             Package.resolve_hash(pkg_name, LOCAL_REGISTRY, hash_prefix)
+
+
+# The following tests were moved out of the PackageTest class to enable parametrization.
+# see (https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses)
+@pytest.mark.parametrize(
+    'update_policy, expected_one_byte',
+    [
+        ('incoming', b'one'),
+        ('existing', b'1')
```
akarve, line 1435: In addition to the default argument, you can add an xfail test case, e.g., `"bad_policy"` [like this](https://github.com/quiltdata/quilt/blob/master/lambdas/thumbnail/tests/test_thumbnail.py#L57).

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,95 @@ def test_resolve_hash(self):
...
+
+    nested_dir_2 = data_dir / 'nested2'
+    if update_policy:
+        pkg.set_dir("/", nested_dir_2, update_policy=update_policy)
+    else:
+        pkg.set_dir("/", nested_dir_2)
+    assert pkg['one.txt'].get_bytes() == expected_one_byte
+    assert pkg['two.txt'].get_bytes() == b'two'
+    assert pkg['three.txt'].get_bytes() == b'three'
+    assert pkg['sub/three.txt'].get_bytes() == b'3'
```
akarve, line 1459: why not also check 'sub/two.txt'?

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,95 @@ def test_resolve_hash(self):
...
+
+    # test set_dir for sub a directory
```
akarve, line 1461: we should do this as part of parameterize :)

akarve: reason: it's the same code as above. so add a parameter like `sub_dir: bool`.

api/python/tests/integration/test_packages.py
```diff
@@ -526,6 +531,15 @@ def test_s3_set_dir(self):
 
             list_object_versions_mock.assert_called_with('bucket', 'foo/')
 
+    def test_set_dir_wrong_update_policy(self):
+        """Verify non existing update policy raises value error."""
+        data_dir = pathlib.Path(__file__).parent / "data"
```
sir-sigurd, line 536: Any reason to not use `DATA_DIR` here?

sir-sigurd (approved):

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,98 @@ def test_resolve_hash(self):
...
+    assert 'three.txt' not in pkg.keys()
+    assert pkg.meta == {'name': 'test_meta'}
+
+    nested_dir_2 = data_dir / 'nested2'
+    if update_policy:
+        pkg.set_dir(target_dir, nested_dir_2, update_policy=update_policy)
+    else:
+        pkg.set_dir(target_dir, nested_dir_2)
+    assert pkg['one.txt'].get_bytes() == expected_one_byte
+    assert pkg['sub/two.txt'].get_bytes() == expected_two_byte
```
akarve, line 1466: Almost. We should be using target_dir to construct these strings, e.g. `pkg[target_dir + 'two.txt']`. You want to write test so that it works and runs all the tests for _any_ value of `target_dir`.

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1424,98 @@ def test_resolve_hash(self):
...
+    assert pkg['sub/three.txt'].get_bytes() == expected_three_byte
+    if target_dir == '/':
```
akarve, line 1468: If instead you do `target_dir=''` in parameterize then everything just works, including `p.set_dir("", whatever)`. And a few points of coverage you are missing:
* `sub/` is an existing directory in `nested`; you should really show that it exists before doing the `set_dir`, because what we want to prove is that the package does the right thing based on what's on disk; at the very least you can make a comment about what you expect to be in `nested/sub` and then prove that it gets captured
* The tests prove that certain keys have expected values, but how do we know that we haven't added new keys? We don't. Add something like `assert set(pkg.keys()) == set(["known", "good", "inputs"])`.

Here's how to think about this whole unit test. You are merging A, B:
* Prove A and B are in a given initial state
* Do the merge
* Prove that only the expected adds/modifies/deletes happened

api/python/tests/integration/test_packages.py
cosmic-byte: nice catch

api/python/tests/integration/test_packages.py
```diff
@@ -1410,3 +1423,106 @@ def test_resolve_hash(self):
...
+
+    nested_dir_2 = DATA_DIR / 'nested2'
+    if update_policy:
+        pkg.set_dir(target_dir, nested_dir_2, update_policy=update_policy)
+    else:
+        pkg.set_dir(target_dir, nested_dir_2)
+    assert set(pkg.keys()) == expected_keys
+
+    target_dir = target_dir.strip("/")
+    if target_dir:
```
akarve, line 1469: ideally we wouldn't need this branch because the parametrize expectations are all self-contained, but this is fine for now.

PR #1712, opened 2020-07-20
Make delete_url() follow S3 semantics as docstring says

sir-sigurd:
Before this change FileNotFoundError was raised for non-existing path


sir-sigurd:
This change is already included in https://github.com/quiltdata/quilt/pull/1678, just wanted to merge it separately.

PR #1714, opened 2020-07-21
Package select lambda

kevinemoore:
## Description
An API endpoint to see a folder-like view of package manifests. The lambda uses S3 Select to read only logical keys from manifests and Pandas to extract logical folder views.

## TODO
- [x] Unit Tests


lambdas/package/test/test_package_browse.py
```diff
@@ -0,0 +1,194 @@
...
+        Mocks to tests calls to S3 Select
+        """
+
+        logical_keys = [
+            "foo.csv",
+            "bar/file1.txt",
+            "bar/file2.txt",
+            "bar/baz/file3.txt",
+            "bar/baz/file4.txt"
+        ]
```
sir-sigurd, line 31: Let's add some unicode here.

lambdas/package/test/test_package_browse.py
```diff
@@ -0,0 +1,194 @@
...
+        Test that the S3 Select response is parsed
+        into the correct top-level folder view.
+        """
+        df, _ = load_df(self.s3response)  # pylint: disable=invalid-name
+        assert isinstance(df, pd.DataFrame)
+
+        folder = get_logical_key_folder_view(df)
+        assert len(folder) == 2
+        assert 'foo.csv' in folder
+        assert 'bar/' in folder
```
sir-sigurd, line 163:
```suggestion
        assert sorted(folder) == [
            'bar/',
            'foo.csv',
        ]
```

lambdas/package/test/test_package_browse.py
```diff
@@ -0,0 +1,194 @@
...
+import boto3
+import pandas as pd
+import responses
+
+from ..index import call_s3_select, get_logical_key_folder_view, load_df
+
+
+class TestPackageBrowse(TestCase):
+    """
+    Unit tests for thhe Package API endpoint.
```
sir-sigurd, line 18:
```suggestion
    Unit tests for the Package API endpoint.
```

lambdas/package/index.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def load_df(s3response):
+    """
+    Read a streaming response from s3 select into a
+    Pandas DataFrame
+    """
+    buffer = io.StringIO()
+    for event in s3response['Payload']:
+        if 'Records' in event:
```
sir-sigurd, line 49: It looks like we have to check that `End` is received: https://botocore.amazonaws.com/v1/documentation/api/latest/reference/eventstream.html

lambdas/package/index.py
```diff
@@ -0,0 +1,144 @@
...
+
+def call_s3_select(s3_client, bucket, key, prefix):
+    """
+    Call S3 Select to read only the logical keys from a
+    package manifest that match the desired folder path
+    prefix
+    """
+    sql_stmt = "SELECT s.logical_key from s3object s"
+    if prefix:
+        sql_stmt += f" WHERE s.logical_key LIKE ('{prefix}%')"
```
sir-sigurd, line 96: I think we should use [`SUBSTRING()`](https://docs.aws.amazon.com/AmazonS3/latest/dev/s3-glacier-select-sql-reference-string.html#s3-glacier-select-sql-reference-substring), instead of `LIKE` to filter on the prefix since the prefix itself could contain `%` or `_`. We could also use `SUBSTRING()` to remove prefix from output rows.

lambdas/package/index.py
```diff
@@ -0,0 +1,144 @@
+"""
+Provide the head of a (potentially gzipped) file in S3. Stream to limit
```
nl0, line 2: this comment should be adjusted probably

lambdas/package/index.py
kevinemoore: Good suggestion. Does the new version look like what you had in mind?

lambdas/package/index.py
```diff
@@ -0,0 +1,210 @@
...
+            stats = event['Stats']['Details']
+        elif 'End' in event:
+            # End event indicates that the request finished successfully
+            end_event_received = True
+
+    if not end_event_received:
+        raise IncompleteResultException("Error: Received an incomplete response from S3 Select.")
+
+    buffer.seek(0)
+    df = pd.read_json(buffer, lines=True)  # pylint: disable=invalid-name
```
sir-sigurd, line 74:
```suggestion
    df = pd.read_json(buffer, lines=True)
```
We have it disabled: https://github.com/quiltdata/quilt/blob/835f41ee13433550f508466604ae1ba53c5984ea/pylintrc#L15.

lambdas/package/index.py
```diff
@@ -0,0 +1,210 @@
...
+
+def call_s3_select(s3_client, bucket, key, prefix):
+    """
+    Call S3 Select to read only the logical keys from a
+    package manifest that match the desired folder path
+    prefix
+    """
+    sql_stmt = f"SELECT SUBSTRING(s.logical_key, {len(prefix) + 1}) AS logical_key FROM s3object s"
+    if prefix:
+        sql_stmt += f" WHERE SUBSTRING(s.logical_key, 1, {len(prefix)}) = '{prefix}'"
```
sir-sigurd, line 134:
```suggestion
        quote_str = lambda s: "'%s'" % prefix.replace("'", "''")
        sql_stmt += f" WHERE SUBSTRING(s.logical_key, 1, {len(prefix)}) = {quote_str(prefix)}"
```

lambdas/package/index.py
```diff
@@ -0,0 +1,210 @@
...
+def load_manifest_detail(s3response):
+    """
+    Read a streaming response from s3 select into a dict
+    """
+    end_event_received = False
+    response_data = None
+    for event in s3response['Payload']:
+        if 'Records' in event:
+            records = event['Records']['Payload'].decode()
+            response_data = json.loads(records)
```
sir-sigurd, line 87: Probably this is unlikely, but [it can contain only partial record](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html#RESTSelectObjectAppendix-responses):

> Records message: Can contain a single record, partial records, or multiple records.

So I think we have to iterate till the end here too.

lambdas/package/index.py
```diff
@@ -0,0 +1,210 @@
...
+    return s3_client
+
+
+def call_s3_select(s3_client, bucket, key, prefix):
+    """
+    Call S3 Select to read only the logical keys from a
+    package manifest that match the desired folder path
+    prefix
+    """
+    sql_stmt = f"SELECT SUBSTRING(s.logical_key, {len(prefix) + 1}) AS logical_key FROM s3object s"
```
sir-sigurd, line 132: This will crash if `prefix is None`, we should handle this case either here or in `lambda_handler()`.

lambdas/package/index.py
sir-sigurd: Yes.

lambdas/package/index.py
kevinemoore: You're right. We could have single quotes in logical keys. Most SQL libraries I've used have parameter quoting built-in to the API, but I haven't found one for S3 Select. What other characters could we have in logical keys that would break these queries?

lambdas/package/index.py
kevinemoore: Prefixes and objects are now returned separately in the folder/prefix view.

lambdas/package/index.py
kevinemoore: You mean that we would see a payload with only a subset of a single row?

sir-sigurd: That's not clear for me from reading the mentioned doc if it's possible for this specific case, but I think doesn't make much sense to not handle it in a common way.

kevinemoore:
@sir-sigurd I'd like to add a top-level test that calls `lambda_handler`, but to do it, I'll need to figure out a way to mock the call to s3 select. I haven't found a way to do that when the s3 client is created dynamically. Do you have a suggestion for how to setup that mock?

lambdas/package/index.py
sir-sigurd: > Most SQL libraries I've used have parameter quoting built-in to the API, but I haven't found one for S3 Select.

Me too :disappointed: .

> What other characters could we have in logical keys that would break these queries?

This API call fails for chars in range `\x01-\x1f`, except '\t' and '\n', but we might want to forbid them for local keys, since I expect them to not work in other S3 API calls using XML. All other chars seem to work.

sir-sigurd:
@kevinemoore 

Does this work for you?
```
orig__getattribute__ = botocore.client.BaseClient.__getattribute__
mocked = unittest.mock.Mock()

def new__getattribute__(self, name):
    if name == 'select_object_content':
        return mocked
    return orig__getattribute__(self, name)
    
with unittest.mock.patch('botocore.client.BaseClient.__getattribute__', new__getattribute__):
    session = boto3.Session()
    s3_client = session.client('s3')    
    s3_client.select_object_content(
        Bucket='bucket',
        Key='1',
        InputSerialization={"JSON": {"Type": "LINES"}},
        OutputSerialization={"JSON": {}}, 
        Expression="",
        ExpressionType="SQL",
    )
    print(mocked.mock_calls)
```

kevinemoore:
> Don't forget to add CircleCI job.

Done.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+SCHEMA = {
+    'type': 'object',
+    'properties': {
+        'bucket': {
+            'type': 'string'
+        },
+        'manifest': {
+            'type': 'string'
+        },
+        'access_key': {
```
akarve, line 25: @nl0 is security any better if these secrets go in the headers instead of the query parameters? I am thinking of server logs.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+    'required': ['bucket', 'manifest', 'access_key', 'secret_key'],
+    'additionalProperties': False
+}
+
+
+class IncompleteResultException(Exception):
+    """
+    Exception indicating an incomplete response
+    (e.g., from S3 Select)
+    """
```
akarve, line 50: I think we want a `pass` in here.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+
+def buffer_s3response(s3response):
+    """
+    Read a streaming response from s3 select into a
+    StringIO buffer
+    """
+    response = io.StringIO()
+    end_event_received = False
+    stats = None
+    for event in s3response['Payload']:
```
akarve, line 61: I would use this logic in the indexer lambda (also relies on S3 select, so you could put this in shared uitls.py? `get_s3_select_data` or something.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+            # End event indicates that the request finished successfully
+            end_event_received = True
+
+    if not end_event_received:
+        raise IncompleteResultException("Error: Received an incomplete response from S3 Select.")
+    response.seek(0)
+    return response
+
+
+def get_logical_key_folder_view(df):
```
akarve, line 79: this is a confusing name. i think this is about uniquing. which is all we are doing here. the name led me to look for prefixes, wonder why no prefix is a special case vs `/foo/bar/` etc.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+    session = boto3.Session(
+        aws_access_key_id=aws_access_key_id,
+        aws_secret_access_key=aws_secret_access_key,
+        aws_session_token=aws_session_token
+    )
+    s3_client = session.client('s3')
+    return s3_client
+
+
+def call_s3_select(s3_client, bucket, key, logical_key_prefix, detail=False):
```
akarve, line 107: might as well start using type annotations in lambda: `bucket: string`, etc. and then `-> RETURN_TYPE`.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+            sql_stmt += f" WHERE SUBSTRING(s.logical_key, 1, {prefix_length}) = '{prefix}'"
+
+    print(sql_stmt)
+    response = s3_client.select_object_content(
+        Bucket=bucket,
+        Key=key,
+        ExpressionType='SQL',
+        Expression=sql_stmt,
+        InputSerialization={
+            'JSON': {'Type': 'DOCUMENT'},
```
akarve, line 136: This should be `'Type': LINES`? I'm really confused how this even works because you're querying a manifest.

lambdas/package/test/test_package_browse.py
```diff
@@ -0,0 +1,264 @@
...
+        expected_sql += f" WHERE SUBSTRING(s.logical_key, 1, 4) = '{prefix}'"
+        expected_args = {
+            'Bucket': bucket,
+            'Key': key,
+            'Expression': expected_sql,
+            'ExpressionType': 'SQL',
+            'InputSerialization': {
+                'CompressionType': 'NONE',
+                'JSON': {'Type': 'DOCUMENT'}
+                },
```
akarve, line 162: outdent

.gitignore
```diff
@@ -26,6 +26,9 @@ cov_html/*
 # Pytest cache
 api/python/.pytest_cache/*
 
+# Setup for lambda testing
+lambdas/setup.py
```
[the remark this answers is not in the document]
kevinemoore: This lets me import the lambda code in a test notebook, which is extremely useful. Is there another way to do it?

lambdas/package/index.py
kevinemoore: We don't need it and pylint actually flagged the pass as a violation.

lambdas/package/index.py
kevinemoore: It used to have a prefix until I went with @sir-sigurd's suggestion to have s3 select pull out the prefix with SUBSTRING. It's more than deduplication though. It really is parsing out a folder view. I could drop the `logical_key` part of the name because it could be applied to any list of file paths.

lambdas/package/index.py
kevinemoore: Seems to work fine either way, but I'll change it.

lambdas/package/index.py
kevinemoore: How about `file_list_to_folder`?

kevinemoore:
> @kevinemoore
> 
> Does this work for you?
> 
> ```python
> orig__getattribute__ = botocore.client.BaseClient.__getattribute__
> mocked = unittest.mock.Mock()
> 
> def new__getattribute__(self, name):
>     if name == 'select_object_content':
>         return mocked
>     return orig__getattribute__(self, name)
>     
> with unittest.mock.patch('botocore.client.BaseClient.__getattribute__', new__getattribute__):
>     session = boto3.Session()
>     s3_client = session.client('s3')    
>     s3_client.select_object_content(
>         Bucket='bucket',
>         Key='1',
>         InputSerialization={"JSON": {"Type": "LINES"}},
>         OutputSerialization={"JSON": {}}, 
>         Expression="",
>         ExpressionType="SQL",
>     )
>     print(mocked.mock_calls)
> ```

Using the same technique, I ended up patching `boto3.Session.client`.

lambdas/package/index.py
kevinemoore: Done.

lambdas/package/index.py
```diff
@@ -0,0 +1,186 @@
...
+import io
+import json
+
+import boto3
+import pandas as pd
+
+from t4_lambda_shared.decorator import api, validate
+from t4_lambda_shared.utils import get_default_origins, make_json_response
+
+S3_DOMAIN_SUFFIX = '.amazonaws.com'
```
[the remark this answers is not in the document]
kevinemoore: It's actually not used here so I'll remove it.

.circleci/config.yml
```diff
@@ -49,6 +49,9 @@ workflows:
       - test-lambda:
           name: test-lambda-thumbnail
           path: "thumbnail"
+      - test-lambda:
+          name: test-lambda-package
+          path: "package"
```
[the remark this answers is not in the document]
kevinemoore: Now `pkgselect`.

.gitignore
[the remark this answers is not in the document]
akarve: you could not check this in and just have local .gitignore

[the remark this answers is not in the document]
kevinemoore: Done.

lambdas/package/index.py
nl0: well, maybe the headers are a little less obvious, but this data gets exposed anyways, so i'm not sure we'll gain anything by using headers. tho it's quite easy to implement, so we can do it if you say so

lambdas/pkgselect/index.py
```diff
@@ -0,0 +1,160 @@
...
+            'type': 'string'
+        },
+        'prefix': {
+            'type': 'string'
+        },
+        'logical_key': {
+            'type': 'string'
+        }
+    },
+    'required': ['bucket', 'manifest', 'access_key', 'secret_key'],
```
nl0, line 38: `session_token` should be required i believe

PR #1715, opened 2020-07-21
Fix access counts querying by setting AllowQuotedRecordDelimiter to true

nl0:


akarve:
I will test this. Do we understand why it affected some stacks and not others (with the same catalog version)? The intermittent nature seems odd.

akarve:
This is good, can we also show the "truncated" version of the preview with an Expand button, as we do in search results?

PR #1716, opened 2020-07-21
Trim overview

nl0:
### Catalog / bucket overview:

- Show no more than 7 entries initially + "show more" button for loading more entries (up to 20)
- expandable previews

### Search lambda:
- reduce sample size from 100 to 20 files
- dont fetch source (set `_source = False` bc `_source = []` doesnt seem to do the trick)


akarve:
This is good, can we also show the "truncated" version of the preview with an Expand button, as we do in search results?

nl0:
lambda tests are now failing tho, so i could use some help here

PR #1718, opened 2020-07-22
Add deprecation warning for PackageEntry.physical_keys

sir-sigurd:


api/python/quilt3/packages.py
```diff
@@ -302,6 +302,11 @@ def physical_keys(self):
         """
         Deprecated
         """
+        warnings.warn(
+            "PackageEntry.physical_keys is deprecated, use PackageEntry.psysical_key instead.",
```
nl0, line 306:
```suggestion
            "PackageEntry.physical_keys is deprecated, use PackageEntry.physical_key instead.",
```

PR #1725, opened 2020-07-24
Index packages in ElasticSearch (via manifest notifications)

akarve:


lambdas/es/indexer/index.py
```diff
@@ -108,8 +117,107 @@ def infer_extensions(key, ext):
...
+            Key=key,
+            Expression=SELECT_PACKAGE_META,
+            ExpressionType="SQL",
+            InputSerialization={"JSON": {"Type": "LINES"}},
+            OutputSerialization={"JSON": {}}
+        )
+        for event in raw["Payload"]:
+            if "Records" in event:
+                data = event["Records"]["Payload"]
+                return data
```
sir-sigurd, line 142: Probably this is unlikely, but [it can contain only partial record](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html#RESTSelectObjectAppendix-responses):

> Records message: Can contain a single record, partial records, or multiple records.

So I think we have to iterate till the end [like in this example](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/eventstream.html).

lambdas/es/indexer/index.py
```diff
@@ -108,8 +117,107 @@ def infer_extensions(key, ext):
...
+            key,
+            size,
+            None,
+            etag=etag,
+            s3_client=s3_client,
+            version_id=version_id
+        ).strip()
+    except ValueError as err:
+        print(f"Unexpected manifest pointer file: s3://{bucket}/{key}: {err}")
+        return False
```
sir-sigurd, line 188: Consider rewriting like this:
```suggestion
    err_msg = None
    try:
        manifest_timestamp = int(pointer_file)
    except ValueError as err:
        err_msg = f"Unexpected manifest pointer file: s3://{bucket}/{key}: {err}"
    else:
        if not 1451631600 <= manifest_timestamp <= 1767250800:
            err_msg = f"Invalid manifest pointer s3://{bucket}{key}"
    if err_msg:
        print(err_msg)
        return False

    package_hash = get_plain_text(
        bucket,
        key,
        size,
        None,
        etag=etag,
        s3_client=s3_client,
        version_id=version_id,
    ).strip()
```

lambdas/es/indexer/index.py
```diff
@@ -108,8 +117,107 @@ def infer_extensions(key, ext):
...
+            ext=ext,
+            handle=pointer_prefix[len(POINTER_PREFIX_V1):],
+            key=manifest_key,
+            last_modified=last_modified,
+            package_hash=package_hash,
+            comment=str(first_dict.get("message", "")),
+            metadata=json.dumps(first_dict.get("user_meta", {}))
+        )
+        return True
+    except (json.JSONDecodeError, botocore.exceptions.ClientError) as exc:
```
sir-sigurd, line 210: Under what conditions `botocore.exceptions.ClientError` is raised?

lambdas/es/indexer/index.py
```diff
@@ -108,8 +117,107 @@ def infer_extensions(key, ext):
...
+            DocTypes.PACKAGE,
+            bucket=bucket,
+            etag=etag,
+            ext=ext,
+            handle=pointer_prefix[len(POINTER_PREFIX_V1):],
+            key=manifest_key,
+            last_modified=last_modified,
+            package_hash=package_hash,
+            comment=str(first_dict.get("message", "")),
+            metadata=json.dumps(first_dict.get("user_meta", {}))
```
sir-sigurd, line 207: What values do we want to put in the index when `message` and/or `user_meta` are `null`?

lambdas/es/indexer/test/test_index.py
```diff
@@ -62,6 +70,30 @@
     },
     "userIdentity": {"principalId": "EXAMPLE"}
 }
+# typical response to head_object or get_object
+OBJECT_RESPONSE = {
+    'ResponseMetadata': ANY,
+    'AcceptRanges': 'bytes',
+    'LastModified': datetime.datetime(2019, 5, 30, 23, 27, 29, tzinfo=tzutc()),
```
sir-sigurd, line 77: We might want to use `dateutil.tz.UTC` or maybe even `datetime.timezone.utc` instead of `tzutc()`.

lambdas/es/indexer/index.py
akarve: They both have default values of `''`. We could do something fancier and make the field nullable in ES but I don't think it's necessary since checking for empty string values should be fast.

lambdas/es/indexer/index.py
akarve: I'm sorry to say I think all boto3 errors back out to ClientError 😬

lambdas/es/indexer/test/test_index.py
akarve: That string comes directly from the boto3 API, so I'm must matching what boto3 does.

lambdas/es/indexer/index.py
akarve: OK, I'm going to factor out all the common logic with Kevin's PR (select, stream handling); but probably in a clean PR.

lambdas/es/indexer/index.py
```diff
@@ -108,8 +117,107 @@ def infer_extensions(key, ext):
...
+    wait=wait_exponential(multiplier=2, min=4, max=30),
+    retry=(retry_if_exception(should_retry_exception))
+)
+def select_manifest_meta(s3_client, bucket: str, key: str):
+    """use s3 select to quickly extract line 1 of manifest
+        returns:
+            * first line (if found), None if not found or error
+    """
+    try:
+        raw = s3_client.select_object_content(
```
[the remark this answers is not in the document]
akarve: I'll do that in a clean PR

lambdas/es/indexer/index.py
akarve: Actually, done in this PR.

PR #1729, opened 2020-07-29
Support conversion of PDF to JPEG via thumbnail lambda

akarve:


lambdas/thumbnail/index.py
```diff
@@ -218,58 +232,77 @@ def lambda_handler(request):
...
-        data = thumbnail_bytes.getvalue()
-
-        # Create metadata object
-        info = {
-            'original_size': orig_size,
-            'thumbnail_format': thumbnail_format,
-            'thumbnail_size': thumbnail_size,
-        }
+            thumbnail_format = "JPEG" if input_ == "pdf" else "PNG"
+        # TODO: copy fonts, libraries, binaries to Lambda, set LIB and font ENV VARS
```
JacksonMaxfield: Just curious, in my own work a lot of "document type file readers" require Java to be installed on the image, is that the case here as well?

lambdas/thumbnail/index.py
```diff
@@ -218,58 +232,75 @@ def lambda_handler(request):
...
-        data = thumbnail_bytes.getvalue()
-
-        # Create metadata object
-        info = {
-            'original_size': orig_size,
-            'thumbnail_format': thumbnail_format,
-            'thumbnail_size': thumbnail_size,
-        }
+            thumbnail_format = "JPEG" if input_ == "pdf" else "PNG"
+        if input_ == "pdf":
```
JacksonMaxfield, line 246: Just curious, does reading the PDF require Java to be installed on the image? A lot of document type file readers have Java deps in my experience.

lambdas/thumbnail/index.py
```diff
@@ -218,58 +232,75 @@ def lambda_handler(request):
...
+                    # respect width but not necessarily height (to preserve aspect ratio)
+                    size=(size[0], None)
+                )
+            except (
+                    PDFInfoNotInstalledError,
+                    PDFPageCountError,
+                    PDFSyntaxError,
+                    PopplerNotInstalledError
+            ) as exc:
+                return make_json_response(500, {'error': str(exc)})
```
JacksonMaxfield, line 261: I have been thinking about failed thumbnails for a bit. It may be nice to have a default image when any thumbnail render fails. So instead of just showing the text of the filename it shows some static image of "Could not render preview" or similar.

lambdas/thumbnail/tests/test_thumbnail.py
```diff
@@ -30,28 +32,42 @@ def _make_event(query, headers=None):
...
+        {"size": "w1024h768", "input": "pdf", "page": 4},
+        "pdf-page4-1024w.jpeg", None, [1024, 1450],
+        marks=pytest.mark.poppler
+    ),
+    pytest.param(
+        "MUMmer.pdf",
+        {"size": "w256h256", "input": "pdf", "page": 8},
+        # pdf ignores the height dimension
+        "pdf-page8-256w.jpeg", None, [256, 363],
+        marks=pytest.mark.poppler
```
JacksonMaxfield, line 56: Neato

lambdas/thumbnail/index.py
akarve: Good idea. We could do that in the catalog. CC @nl0.

nl0: backlogged

PR #1730, opened 2020-07-31
Move logic for calling S3 select to a shared function

kevinemoore:
## Description

Multiple lambdas will use s3 select to read manifests in s3 so move querying logic into a single shared function in t4_lambda_shared.


lambdas/shared/t4_lambda_shared/utils.py
```diff
@@ -105,3 +105,30 @@ def buffer_s3response(s3response):
         raise IncompleteResultException("Error: Received an incomplete response from S3 Select.")
     response.seek(0)
     return response
+
+
+def call_s3_select(
```
akarve, line 110: `select_from_manifest`

lambdas/shared/t4_lambda_shared/utils.py
```diff
@@ -105,3 +105,30 @@ def buffer_s3response(s3response):
...
+    key: str,
+    sql_stmt: str
+) -> io.StringIO:
+    """
+    Call S3 Select to read only the logical keys from a
+    package manifest that match the desired folder path
+    prefix
+    """
+
+    print(sql_stmt)
```
akarve, line 122: In prod?

lambdas/shared/t4_lambda_shared/utils.py
```diff
@@ -105,3 +105,30 @@ def buffer_s3response(s3response):
         raise IncompleteResultException("Error: Received an incomplete response from S3 Select.")
     response.seek(0)
     return response
+
+
+def call_s3_select(
+    s3_client: str,
```
akarve, line 111: You could do only s3_client as positional, or s3_client, bucket, and key as positional, the rest should be required kwargs so:

```
s3_client: str,
*,
bucket
key,
etc,
```

lambdas/shared/t4_lambda_shared/utils.py
kevinemoore: Yes. That will be very helpful in the lambda logs, unless there's a better way to log the SQL?

akarve, line 122: Fine but you really need to give this some context:
```suggestion
    print(f"utils.py: manifest_select: {sql_stmt})
```

lambdas/pkgselect/index.py
```diff
@@ -72,47 +72,13 @@ def get_s3_client(
...
 
 
-def call_s3_select(
-    s3_client: str,
-    bucket: str,
-    key: str,
-    logical_key_prefix: str,
-    detail=False
-):
+def fix_quotes(key: str):
```
akarve, line 75: Dima already has something for this in access_counts, just factor that out into shared as well:
https://github.com/quiltdata/quilt/blob/master/lambdas/access_counts/index.py#L34

PR #1732, opened 2020-08-01
Public/anon access to pkgselect

kevinemoore:
## Description
Allow access to the `pkgselect` endpoint for anonymous users only for public packages (manifest files that are publicly readable).

### DIFF
--- a/lambdas/pkgselect/index.py
+++ b/lambdas/pkgselect/index.py
@@ -130,6 +130,7 @@ def lambda_handler(request):
 
         # Use the default S3 client configuration
         s3_client = boto3.client('s3')
+        response = s3_client.head_object(Bucket=bucket, Key=key)
     else:
         return make_json_response(
             400,
--- a/lambdas/pkgselect/index.py
+++ b/lambdas/pkgselect/index.py
@@ -6,6 +6,7 @@ import json
 import os
 
 import boto3
+import botocore
 from botocore import UNSIGNED
 from botocore.client import Config
 import pandas as pd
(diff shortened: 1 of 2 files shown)


lambdas/pkgselect/index.py
```diff
@@ -88,14 +104,46 @@ def lambda_handler(request):
...
+    elif (
+        allow_anonymous_access and
+        access_key is None and
+        secret_key is None and
+        session_token is None
+    ):
+        # Test to see if the target key is publicly accessible. If not, the call
+        # below will raise and exception and return a 403 response
+        anons3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
+        response = anons3.head_object(Bucket=bucket, Key=key)
```
akarve, line 134: this assignment is never used, so don't assign

lambdas/pkgselect/index.py
```diff
@@ -88,14 +104,46 @@ def lambda_handler(request):
...
 
-    # Create an s3 client using the provided credentials
-    s3_client = get_s3_client(aws_access_key_id, aws_secret_access_key, aws_session_token)
+        # Use the default S3 client configuration
+        s3_client = boto3.client('s3')
+    else:
+        return make_json_response(
+            400,
```
akarve, line 140:
```suggestion
            401,
```

lambdas/pkgselect/index.py
akarve: it's much cleaner to try/except here and specifically deal with 403s. i know if you hit an error that the whole function will bomb but what if it's just a throttling error and not permissions related?

PR #1733, opened 2020-08-02
Add Okta support

dimaryaz:

PR #1735, opened 2020-08-03
sso: adjust styling, handle okta popup closing

nl0:
- add okta logo
- adjust layout of sso buttons
- handle okta popup closing by the user


nl0:
#1733

catalog/app/containers/Auth/SSOOkta.js
```diff
@@ -49,6 +50,12 @@ export default ({ mutex, next }) => {
...
     const url = `${oktaDomain}/oauth2/v1/authorize${query}`
     const popup = window.open(url, 'quilt_okta_popup', 'width=300,height=400')
+    const timer = setInterval(() => {
+      if (popup.closed) {
+        clearInterval(timer)
+        handleFailure({ error: 'popup_closed_by_user' })
+      }
+    }, 500)
     popup.focus()
     window.addEventListener('message', ({ origin, data }) => {
```
dimaryaz, line 60: I realized I need to remove the event listener, too. I'll fix it in my PR

catalog/app/containers/Auth/SSOOkta.js
```diff
@@ -31,7 +32,7 @@ export default ({ mutex, next }) => {
   const { push: notify } = Notifications.use()
   const { urls } = NamedRoutes.use()
 
-  const handleClick = () => {
+  const handleClick = React.useCallback(() => {
```
dimaryaz, line 35: Just wondering, when is `React.useCallback` needed? When you pass the function as an attribute to a React element?

Does `handleClick` in `SSOGoogle.js` need it?

nl0: it's needed when a function is used as a prop, correct (to minimize rerenders) or as a dependency for another useCallback

> Does handleClick in SSOGoogle.js need it?

it's a factory returning a callback, so we cant do anything with it in this simple way (some memoization would help, but i don't think it's worth it there)

catalog/app/containers/Auth/SSOOkta.js
nl0: right

PR #1739, opened 2020-08-07
Count pages in PDFs via countPages=true param

akarve:
* `countPages=true` returns `body["info"]["page_count"]` in the response. This call is not efficient (it must render the entire PDF, pressuring memory and increasing latency. Nevertheless, pdftoppm provides no facility for just counting (and not rendering) pages.

### DIFF
--- a/lambdas/thumbnail/index.py
+++ b/lambdas/thumbnail/index.py
@@ -260,7 +260,10 @@ def lambda_handler(request):
     resp = requests.get(url)
     if resp.ok:
         try:
-            thumbnail_format = SUPPORTED_BROWSER_FORMATS.get(imageio.get_reader(resp.content), "PNG")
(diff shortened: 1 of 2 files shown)


lambdas/thumbnail/index.py
```diff
@@ -249,39 +254,52 @@ def lambda_handler(request):
     input_ = request.args.get('input', 'image')
     output = request.args.get('output', 'json')
     page = request.args.get('page', 1)
```
nl0, line 256: @akarve the page should be parsed, otherwise there's an error:
```
'2' is not of type 'integer'

Failed validating 'type' in schema['properties']['page']:
    {'minimum': 1, 'type': 'integer'}

On instance['page']:
    '2'
```

PR #1740, opened 2020-08-10
PDF preview

nl0:
Preview PDF files via thumbnail endpoint

TODO:

- [x] pick up and parse info header (`X-Quilt-Info`) from the response once it's exposed
- [x] styling polish


nl0:
@akarve rtm

PR #1741, opened 2020-08-10
Multi-thread download of large files

sir-sigurd:
## TODO ##

- [x] add tests

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -1,3 +1,6 @@
+import math
+import os
+import stat
 from collections import defaultdict, deque
 from codecs import iterdecode
 from concurrent.futures import ThreadPoolExecutor
@@ -285,7 +288,7 @@ def _upload_file(ctx, size, src_path, dest_bucket, dest_key):
             ctx.run(upload_part, i, start, end)
 
 
-def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
+def _download_file(ctx, size, src_bucket, src_key, src_version, dest_path):
     dest_file = pathlib.Path(dest_path)
     if dest_file.is_reserved():
         raise ValueError("Cannot download to %r: reserved file name" % dest_path)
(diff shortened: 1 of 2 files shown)


api/python/quilt3/data_transfer.py
```diff
@@ -245,19 +256,16 @@ def _upload_file(ctx, size, src_path, dest_bucket, dest_key):
...
         lock = Lock()
-        remaining = len(chunk_offsets)
+        remaining = len(chunk_ranges)
         parts = [None] * remaining
 
         def upload_part(i, start, end):
             nonlocal remaining
             part_id = i + 1
-            with OSUtils().open_file_chunk_reader(src_path, start, end-start, [ctx.progress]) as fd:
+            with OSUtils().open_file_chunk_reader(src_path, start, end - start + 1, [ctx.progress]) as fd:
```
kevinemoore, line 268: What is the purpose of adding the +1 here? Was that a bug in the old code? Or, is there a reason that that this change requires an interval end 1 byte farther?

api/python/quilt3/data_transfer.py
```diff
@@ -295,20 +302,50 @@ def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
...
-    with open(dest_path, 'wb') as fd:
-        while True:
-            chunk = body.read(64 * 1024)
-            if not chunk:
-                break
-            fd.write(chunk)
-            ctx.progress(len(chunk))
+    chunk_ranges = (
+        get_chunk_ranges(size)
+        if is_regular_file and size >= s3_transfer_config.multipart_threshold and os.name != 'nt' else
```
kevinemoore, line 321: What is the significance of 'nt'?

api/python/quilt3/data_transfer.py
```diff
@@ -214,6 +216,15 @@ def check_head_object_works_for_client(s3_client, params):
 UPLOAD_ETAG_OPTIMIZATION_THRESHOLD = 1024
 
 
+def get_chunk_ranges(file_size):
+    adjuster = ChunksizeAdjuster()
+    chunk_size = adjuster.adjust_chunksize(s3_transfer_config.multipart_chunksize, file_size)
+    return [
+        (start, min(start + chunk_size, file_size) - 1)
```
dimaryaz, line 223: Very minor, but I think `(start, min(start + chunk_size, file_size))` would be better: pretty much everything in Python - ranges, lengths, etc. include the start, but not the end.

This will also answer @kevinemoore's question about `end - start + 1`.

api/python/quilt3/data_transfer.py
dimaryaz: See my comment above; I think the `end` value should be adjusted.

api/python/quilt3/data_transfer.py
dimaryaz: Multi-threaded downloads (supposedly) don't work on Windows - so it falls back to single-threaded.

api/python/quilt3/data_transfer.py
```diff
@@ -295,20 +302,50 @@ def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
 
     dest_file.parent.mkdir(parents=True, exist_ok=True)
 
+    with dest_file.open('wb', buffering=0) as f:
```
dimaryaz, line 305: Why `buffering=0`?

api/python/quilt3/data_transfer.py
```diff
@@ -295,20 +302,50 @@ def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
...
+        fileno = f.fileno()
+        is_regular_file = stat.S_ISREG(os.stat(fileno).st_mode)
```
dimaryaz, line 307: `dest_file.is_file()`

api/python/quilt3/data_transfer.py
```diff
@@ -295,20 +302,50 @@ def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
...
+            body = resp['Body']
+            chunk_size = 64 * 1024
+            while True:
+                chunk = body.read(chunk_size)
+                if not chunk:
+                    break
+                ctx.progress(chunk_f.write(chunk))
 
-    ctx.done(PhysicalKey.from_path(dest_path))
+        if not next(remaining_counter):
```
dimaryaz, line 344: That's kind of a strange way to count... Just use a variable? Also, need a lock. See the other upload/download methods; something like:
```
            with lock:
                remaining -= 1
                done = remaining == 0
```

sir-sigurd: > That's kind of a strange way to count.

I'm pretty sure it's atomic.

api/python/quilt3/data_transfer.py
sir-sigurd: I presume that calling `os.stat()` with a file descriptor is faster than with a path, since the latter requires path resolving.

dimaryaz: Ok sure

api/python/quilt3/data_transfer.py
dimaryaz: Hm... Is it guaranteed? The only thing I can find is: if it's implemented in C, then it'll hold the GIL, so it will be atomic - and I'm guessing `next` on a `reversed`, `range`, etc. will be a C function. But, it looks like an implementation detail.

In any case, I think the cost of acquiring a lock is pretty minimal compared to the download of the part - so it's not worth risking data corruption even if we're 99% sure it's ok. (And who knows if someone decides to run this code using Jython or whatever else...)

api/python/quilt3/data_transfer.py
sir-sigurd: And it turns that it works on Windows, I had a wrong understanding of shared file usage on Windows.

dimaryaz: Oh nice!

(I tried it using Wine on Linux, and it worked... but didn't want to assume that it works on real Windows.)

api/python/quilt3/data_transfer.py
```diff
@@ -295,20 +298,63 @@ def _download_file(ctx, src_bucket, src_key, src_version, dest_path):
...
+                chunk = body.read(s3_transfer_config.io_chunksize)
+                if not chunk:
+                    break
+                ctx.progress(chunk_f.write(chunk))
 
-    ctx.done(PhysicalKey.from_path(dest_path))
+        with remaining_counter_lock:
+            remaining_counter -= 1
+            if not remaining_counter:
+                ctx.done(PhysicalKey.from_path(dest_path))
```
dimaryaz, line 354: Don't call `ctx.done` while holding the lock. Just do:
```
        with remaining_counter_lock:
            remaining_counter -= 1
            done = remaining_counter == 0
        if done:
            ctx.done(PhysicalKey.from_path(dest_path))
```

PR #1742, opened 2020-08-10
Expose custom header, X-QUILT-INFO

akarve:
Since this endpoint is binary, this is the first time we're sending metadata back in the headers. I'm kinda surprised that this even works or is needed since `Expose-Headers` is for preflight checks, but this is what @nl0 asked for :)

### DIFF
--- a/lambdas/shared/t4_lambda_shared/decorator.py
+++ b/lambdas/shared/t4_lambda_shared/decorator.py
@@ -12,7 +12,7 @@ from jsonschema import Draft4Validator, ValidationError
 
 GZIP_MIN_LENGTH = 1024
 GZIP_TYPES = {'text/plain', 'application/json'}
-# Used, e.g., for binayr responses when metadata belongs in headers, not body
+# Used, e.g., for binary responses when metadata belongs in headers, not body
 QUILT_INFO_HEADER = 'X-Quilt-Info'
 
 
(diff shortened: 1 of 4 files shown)


lambdas/shared/t4_lambda_shared/decorator.py
```diff
@@ -12,6 +12,8 @@
 
 GZIP_MIN_LENGTH = 1024
 GZIP_TYPES = {'text/plain', 'application/json'}
+# Used, e.g., for binayr responses when metadata belongs in headers, not body
```
akarve, line 15:
```suggestion
# Used, e.g., for binary responses when metadata belongs in headers, not body
```

lambdas/shared/t4_lambda_shared/decorator.py
```diff
@@ -73,7 +75,14 @@ def wrapper(event, _):
...
                     'access-control-allow-origin': '*',
                     'access-control-allow-methods': 'OPTIONS,HEAD,GET,POST',
-                    'access-control-allow-headers': request.headers.get('access-control-request-headers', ''),
+                    'access-control-allow-headers': (
+                        request.headers.get('access-control-request-headers', '')
+                    ),
+                    # for preflight checks, not sure we need it for header to work?
+                    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+                    'access-control-expose-headers': (
+                        request.headers.get('access-control-request-headers', '')
```
akarve, line 84: Alternatively, is this what we want?
```suggestion
                        f"*, Authorization, {QUILT_INFO_HEADER}"
```

nl0: this looks good

PR #1744, opened 2020-08-11
Include size in pkgselect responses

kevinemoore:
## Description
Add size to the pkgselect response. The implementation still does most of the work in s3select and pandas (C), but could be slow for large responses (e.g., a folder with many objects).

## TODO
- [x] Unit Tests
- [ ] Performance testing


lambdas/pkgselect/index.py
```diff
@@ -56,10 +56,16 @@ def file_list_to_folder(df: pd.DataFrame) -> dict:
...
-    except AttributeError:
+        groups = df.groupby(df.logical_key.str.extract('([^/]+/?).*')[0], dropna=True)
+        folder = groups.agg(
+            size=('size', 'sum'),
+            physical_key=('physical_key', 'first')
+        )
+        folder.reset_index(inplace=True)
+        folder.rename(columns={0: 'logical_key'}, inplace=True)
+        prefixes = folder[folder.logical_key.str.contains('/')].to_dict(orient='records')
+        objects = folder[~folder.logical_key.str.contains('/')].to_dict(orient='records')
```
akarve, line 67: What happens in keys with successive `//` to both the regex in the groupby and here?

lambdas/pkgselect/index.py
```diff
@@ -172,7 +178,8 @@ def lambda_handler(request):
         # Call s3 select to fetch only logical keys matching the
         # desired prefix (folder path)
         prefix_length = len(prefix) if prefix is not None else 0
-        sql_stmt = f"SELECT SUBSTRING(s.logical_key, {prefix_length + 1}) AS logical_key FROM s3object s"
+        sql_stmt = f"SELECT SUBSTRING(s.logical_key, {prefix_length + 1}) AS logical_key"
+        sql_stmt += ", s.\"size\", s.physical_keys[0] as physical_key FROM s3object s"
```
akarve, line 182:
```suggestion
        sql_stmt = (
            f"SELECT SUBSTRING(s.logical_key, {prefix_length + 1}) AS logical_key"
            ", s.\"size\", s.physical_keys[0] as physical_key FROM s3object s"
        )
```

lambdas/pkgselect/index.py
```diff
@@ -56,10 +56,16 @@ def file_list_to_folder(df: pd.DataFrame) -> dict:
...
-        folder = pd.Series(df.logical_key.dropna().str.extract('([^/]+/?).*')[0].unique())
-        prefixes = folder[folder.str.endswith('/')].sort_values().tolist()
-        objects = folder[~folder.str.endswith('/')].sort_values().tolist()
-    except AttributeError:
+        groups = df.groupby(df.logical_key.str.extract('([^/]+/?).*')[0], dropna=True)
+        folder = groups.agg(
+            size=('size', 'sum'),
+            physical_key=('physical_key', 'first')
+        )
+        folder.reset_index(inplace=True)
```
akarve, line 64: add comment on what this does

lambdas/pkgselect/index.py
```diff
@@ -56,10 +56,16 @@ def file_list_to_folder(df: pd.DataFrame) -> dict:
...
+        folder.rename(columns={0: 'logical_key'}, inplace=True)
+        prefixes = folder[folder.logical_key.str.contains('/')].to_dict(orient='records')
```
akarve, line 66: I think these are called `common_prefixes` in the S3 API language

lambdas/pkgselect/index.py
akarve: We might as well send back the physical key as a list for consistency with manifest format? i.e. then JS can reuse any manifest-processing code.

nl0: @akarve i think it's actually cleaner this way (with single physical_key)

nl0:
@kevinemoore can you return meta for the package / prefix as well?

nl0 (changes requested):
1. let's not return `physical_key`s for prefixes, bc they are not relevant (they actually point to random objects under the prefix) and waste bandwidth

2. when trying to query an object the api throws `local variable 'result' referenced before assignment`

3. should we return the whole meta object here or just `user_meta`? @akarve

nl0 (approved):
it works now

akarve:
> 3. should we return the whole meta object here or just `user_meta`? @akarve
Whole object

lambdas/pkgselect/index.py
```diff
@@ -56,10 +56,20 @@ def file_list_to_folder(df: pd.DataFrame) -> dict:
     lambda).
     """
     try:
-        folder = pd.Series(df.logical_key.dropna().str.extract('([^/]+/?).*')[0].unique())
-        prefixes = folder[folder.str.endswith('/')].sort_values().tolist()
-        objects = folder[~folder.str.endswith('/')].sort_values().tolist()
-    except AttributeError:
+        groups = df.groupby(df.logical_key.str.extract('([^/]+/?).*')[0], dropna=True)
+        folder = groups.agg(
```
akarve, line 60: Does this automatically skip undefined values?

lambdas/pkgselect/index.py
```diff
@@ -172,7 +182,8 @@ def lambda_handler(request):
         # Call s3 select to fetch only logical keys matching the
         # desired prefix (folder path)
         prefix_length = len(prefix) if prefix is not None else 0
-        sql_stmt = f"SELECT SUBSTRING(s.logical_key, {prefix_length + 1}) AS logical_key FROM s3object s"
+        sql_stmt = f"SELECT SUBSTRING(s.logical_key, {prefix_length + 1}) AS logical_key"
+        sql_stmt += ", s.\"size\", s.physical_keys[0] as physical_key FROM s3object s"
```
akarve, line 186:
```suggestion
        sql_stmt = (
            f"SELECT SUBSTRING(s.logical_key, {prefix_length + 1}) AS logical_key"
            ", s.\"size\", s.physical_keys[0] as physical_key FROM s3object s"
        )
```

lambdas/pkgselect/index.py
kevinemoore: // is illegal in logical keys isn't it?

lambdas/pkgselect/index.py
kevinemoore: Yes.

PR #1746, opened 2020-08-11
Fix Okta

dimaryaz:
Company name isn't sufficient: there may be multiple auth servers for a given domain.
Use a base URL instead.


catalog/app/containers/Auth/SSOOkta.js
```diff
@@ -36,19 +36,18 @@ export default function SSOOkta({ mutex, next }) {
     if (mutex.current) return
     mutex.claim(MUTEX_POPUP)
 
-    const oktaDomain = `https://${cfg.oktaCompanyName}.okta.com`
     const nonce = Math.random().toString(36).substr(2)
     const state = Math.random().toString(36).substr(2)
     const query = NamedRoutes.mkSearch({
       client_id: cfg.oktaClientId,
-      redirect_uri: window.location.origin,
+      redirect_uri: `${window.location.origin}/auth/implicit/callback`, // TODO(dima): remove the path
```
dimaryaz, line 43: (There's no reason to have a path here - just need to wait for BT guys to update their config.)

PR #1747, opened 2020-08-12
Use a logging decorator controlled by an environment variable for verbosity in indexer lambda

akarve:


lambdas/es/indexer/index.py
```diff
@@ -383,10 +384,12 @@ def make_s3_client():
     return boto3.client("s3", config=configuration)
 
 
+@logger()
 def handler(event, context):
     """enumerate S3 keys in event, extract relevant data, queue events, send to
     elastic via bulk() API
     """
+    logger_ = getLogger('quilt-lambda')
```
kevinemoore, line 392: This line is also inside the decorator. I assume there the logger_ variable is only scoped to inside the decorator itself. That does force users to get the logger name right. Can we put that name in .consts or make a call to always return the right logger?

akarve: I'll make the name a constant. The logging class guarantees to always return the same logger if given the same name. In retrospect we don't really need a decorator for that, though I'm going to leave that part in for now.

PR #1750, opened 2020-08-13
Fix missing index in document_queue, add test coverage for .append, more logging

akarve:


lambdas/es/indexer/document_queue.py
```diff
@@ -145,7 +149,7 @@ def append(
 
     def _append_document(self, doc):
         """append well-formed documents (used for retry or by append())"""
-        if doc["content"]:
+        if "content" in doc:
```
akarve, line 152:
```suggestion
        if doc.get("content"):
```

PR #1755, opened 2020-08-19
WIP: Skip the username in Okta sign-ins

dimaryaz:
There are lots of open questions, though:
- The sign-up page still asks for a username (and a password, depending on the config) - do we want that?
- What if the stack is configured to require a password?
- Do we want the same for Google sign-in?

### DIFF
--- a/catalog/app/containers/Auth/SSOOkta.js
+++ b/catalog/app/containers/Auth/SSOOkta.js
@@ -1,4 +1,3 @@
-import { push } from 'connected-react-router/esm/immutable'
 import invariant from 'invariant'
 import * as React from 'react'
 import { FormattedMessage as FM } from 'react-intl'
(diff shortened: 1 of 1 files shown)


akarve:
> * The sign-up page still asks for a username (and a password, depending on the config) - do we want that?
That should only show when Quilt auth is allowed.
> * What if the stack is configured to require a password?
Then the router always sends them to sign in; what am I missing?
> * Do we want the same for Google sign-in?
Yes it should be standard for any SSO config to skip username. OK for separate PR if you choose.

dimaryaz:
I meant, right now, if password sign in is allowed, then the registry requires you to set a password during signup - which would break the SSO flow of skipping the signup screen. (Of course, if you're using Okta, then you probably don't want password sign in enabled, but still, need to make a decision.)

akarve:
> I meant, right now, if password sign in is allowed, then the registry requires you to set a password during signup - which would break the SSO flow of skipping the signup screen. (Of course, if you're using Okta, then you probably don't want password sign in enabled, but still, need to make a decision.)

OK, for now since Okta is ONLY stand-alone we can go with this change. Out of curiosity, what happens in the database if we skip the username flow?

For the Quilt+Google case there actually we should actually need no sign-up for off-domain users, because the admin sets their username when he invites them?

PR #1763, opened 2020-08-28
Package search + fast package listing

nl0:
### TODO:

- [x] box icon from FA for packages

- [x] search lambda and indexer adjustments (@akarve)

  - [x] use relevant fields for package search

  - [x] add revision info to the index

  - [x] expose relevant fields in the response (add to `_source`): `comment`, `handle`, `hash`, `tags`, `metadata_string` (?), `revision` (?)

### DIFF
--- a/catalog/app/containers/Bucket/requests.js
+++ b/catalog/app/containers/Bucket/requests.js
@@ -51,10 +51,6 @@ const catchErrors = (pairs = []) =>
 const withErrorHandling = (fn, pairs) => (...args) =>
   fn(...args).catch(catchErrors(pairs))
 
-const mapAllP = R.curry((fn, list) => Promise.all(list.map(fn)))
-
-const mergeAllP = (...ps) => Promise.all(ps).then(R.mergeAll)
-
 const promiseProps = (obj) =>
   Promise.all(Object.values(obj)).then(R.zipObj(Object.keys(obj)))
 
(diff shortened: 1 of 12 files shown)


catalog/app/containers/Bucket/requests.js
```diff
@@ -727,20 +717,49 @@ export const listPackages = withErrorHandling(
...
-      )
-      .then(R.unnest)
+
+    const body = {
+      query: mkFilterQuery(filter),
+      aggs: {
+        packages: {
+          composite: {
+            // the limit is configured in ES cluster settings (search.max_buckets)
+            size: 10000,
```
nl0, line 727: @akarve FYI this is the hard limit on the total number of packages to be displayed in the catalog

akarve: Do yo mean that you will never issue a request with a size larger than this? Does this affect pagination if the results set is larger than 10K?

nl0: > Do yo mean that you will never issue a request with a size larger than this?

no, it means ES cant create more than 10k "buckets" (==packages) without adjusting configuration, so we can't paginate further than that

> Does this affect pagination if the results set is larger than 10K?

well, yes and no. the "initial" result set (created by the "composite" aggregation) is sorted by handle A to Z no matter what and truncated to this limit (10k). additional sorting (by timestamp, if enabled) and pagination happens after that.
so, to sum it up, pagination is not affected when using default order (handle A to Z) and may be affected when ordering by timestamp

nl0:
@akarve yes, we use debounce

nl0:
nope, i'm mistaken. ima add debounce

PR #1764, opened 2020-08-30
Package search action, upgrade search endpoint

akarve:
* Use the more powerful query_string (instead of simple_query_string) for 'search' and 'packages' actions
* Fix ES version 😬 to 6.x, not 7.x
* Add a 'packages' action with client-customizable body
* Try dropping terminate_after since it might be harming search quality (and not helping performance?)
* Bump query timeout to 30 sec

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -13,7 +13,8 @@ from elasticsearch import Elasticsearch, RequestsHttpConnection
 from t4_lambda_shared.decorator import api
 from t4_lambda_shared.utils import get_default_origins, make_json_response
 
-MAX_QUERY_DURATION = '15s'
+DEFAULT_SIZE = 1_000
+MAX_QUERY_DURATION = '30s'
 NUM_PREVIEW_IMAGES = 100
 NUM_PREVIEW_FILES = 20
 COMPRESSION_EXTS = ['.gz']
@@ -50,23 +51,51 @@ def lambda_handler(request):
     """
 
     action = request.args.get('action')
-    indexes = request.args.get('index')
-    terminate_after = os.getenv('MAX_DOCUMENTS_PER_SHARD')
+    user_body = request.args.get('body', {})
+    user_fields = request.args.get('fields', [])
+    user_indexes = request.args.get('index', "")
+    user_size = request.args.get('size', DEFAULT_SIZE)
+    user_source = request.args.get('_source', [])
+    terminate_after = None  # see if we can skip os.getenv('MAX_DOCUMENTS_PER_SHARD')
 
-    if action == 'search':
+    if action == 'packages':
         query = request.args.get('query', '')
-        body = {
+        body = user_body or {
             "query": {
-                "simple_query_string": {
+                "query_string": {
                     "analyze_wildcard": True,
+                    "lenient": True,
                     "query": query,
-                    "fields": ['content', 'comment', 'key_text', 'meta_text']
+                    # see enterprise/**/bucket.py for mappings
+                    "fields": user_fields or [
+                        # package
+                        'comment', 'handle', 'handle_text^2', 'metadata_string', 'tags'
+                    ]
                 }
             }
         }
-        # TODO: should be user settable; we should proably forbid `content` (can be huge)
+        _source = user_source
+        size = user_size
+    elif action == 'search':
+        query = request.args.get('query', '')
+        body = {
+            "query": {
+                "query_string": {
+                    "analyze_wildcard": True,
+                    "lenient": True,
+                    "query": query,
+                    # see enterprise/**/bucket.py for mappings
+                    "fields": [
+                        # object
+                        'content', 'comment', 'ext', 'key_text^2', 'meta_text',
+                        # package
+                        'comment', 'handle', 'handle_text^2', 'metadata_string', 'tags'
+                    ]
+                }
+            }
+        }
         _source = ['key', 'version_id', 'updated', 'last_modified', 'size', 'user_meta']
-        size = 1000
+        size = DEFAULT_SIZE
     elif action == 'stats':
         body = {
             "query": {"match_all": {}},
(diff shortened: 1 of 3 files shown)


lambdas/es/indexer/test/test_index.py
```diff
@@ -301,7 +304,9 @@ def _make_event(
                 "handle": "pkg/usr",
                 "key": "foo",
                 "last_modified": "not_an_object",
+                "pointer_file": "1598026253",
```
akarve, line 307:
```suggestion
```

lambdas/search/tests/test_search.py
```diff
@@ -227,6 +227,43 @@ def test_post_process_stats_some_gz(self):
...
+            content_type='application/json',
+            match_querystring=True,
+        )
+
+        event = self._make_event(query)
+        resp = lambda_handler(event, None)
+        assert resp['statusCode'] == 200
+        assert json.loads(resp['body']) == {'results': 'blah'}
+
+
```
akarve, line 266: whitespace
```suggestion
```

lambdas/search/index.py
```diff
@@ -50,23 +51,51 @@ def lambda_handler(request):
...
+                        # object
+                        'content', 'comment', 'ext', 'key_text^2', 'meta_text',
+                        # package
+                        'comment', 'handle', 'handle_text^2', 'metadata_string', 'tags'
+                    ]
                 }
             }
         }
-        # TODO: should be user settable; we should proably forbid `content` (can be huge)
         _source = ['key', 'version_id', 'updated', 'last_modified', 'size', 'user_meta']
```
nl0, line 97:
```suggestion
        _source = ['key', 'version_id', 'updated', 'last_modified', 'size', 'user_meta', 'comment', 'handle', 'hash', 'tags', 'metadata_string', 'pointer_file']
```
@akarve this data is required to display the search cards

lambdas/search/index.py
```diff
@@ -50,23 +51,51 @@ def lambda_handler(request):
...
+            "query": {
+                "query_string": {
+                    "analyze_wildcard": True,
+                    "lenient": True,
+                    "query": query,
+                    # see enterprise/**/bucket.py for mappings
+                    "fields": user_fields or [
+                        # package
+                        'comment', 'handle', 'handle_text^2', 'metadata_string', 'tags'
+                    ]
```
nl0, line 73: @akarve we have to group the results by handle

lambdas/search/index.py
```diff
@@ -13,7 +13,8 @@
 from t4_lambda_shared.decorator import api
 from t4_lambda_shared.utils import get_default_origins, make_json_response
 
-MAX_QUERY_DURATION = '15s'
+DEFAULT_SIZE = 1_000
+MAX_QUERY_DURATION = '30s'
```
kevinemoore, line 17: Since API gateway's timeout is 30s, we should make sure Elastic returns with a few seconds to spare. Maybe try 25s instead?

lambdas/search/index.py
```diff
@@ -50,23 +51,51 @@ def lambda_handler(request):
...
+            }
+        }
+        _source = user_source
+        size = user_size
+    elif action == 'search':
         query = request.args.get('query', '')
         body = {
             "query": {
-                "simple_query_string": {
+                "query_string": {
```
kevinemoore, line 83: Changing from simple_query_string to query_string will cause Elastic to throw an exception if users submit bad queries. https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html The fault-tolerant aspect of simple_query_string is probably a better user experience.

lambdas/search/index.py
```diff
@@ -141,12 +170,14 @@ def lambda_handler(request):
...
+    to_search = f"{user_indexes},{index_overrides}" if index_overrides else user_indexes
+
     result = es_client.search(
-        to_search,
-        body,
+        index=to_search,
+        body=body,
         _source=_source,
         size=size,
+        # try turning this off to consider all documents
```
kevinemoore, line 180: Looks good as long as we avoid very large shards, which we should now be doing.

lambdas/search/index.py
akarve, line 17:
```suggestion
MAX_QUERY_DURATION = '25s'
```

lambdas/search/index.py
akarve: Sure, you can do the groupby (or aggregations) in the body right?

lambdas/search/index.py
akarve: Backlogged for better search errors

lambdas/search/index.py
akarve: oh, i can add a default for that sure but i'm letting you control the body for this action so you can sort/etc.

nl0: oh i've just noticed that the whole body can be customized -- it should work

nl0 (approved):

lambdas/search/index.py
```diff
@@ -50,23 +51,59 @@ def lambda_handler(request):
...
-                "simple_query_string": {
+                "query_string": {
                     "analyze_wildcard": True,
+                    "lenient": True,
                     "query": query,
-                    "fields": ['content', 'comment', 'key_text', 'meta_text']
+                    # see enterprise/**/bucket.py for mappings
+                    "fields": [
+                        # object
+                        'content', 'comment', 'ext', 'key_text^2', 'meta_text',
```
akarve, line 95:
```suggestion
                        'content', 'comment', 'ext', 'key_text', 'meta_text',
```

lambdas/search/tests/test_search.py
```diff
@@ -7,9 +7,10 @@
 from unittest.mock import patch
 from urllib.parse import urlencode
 
+import pytest
```
akarve, line 10:
```suggestion
```

PR #1765, opened 2020-08-30
Send pointer_file field to ES so that JS client can resolve packages

akarve:


lambdas/es/indexer/test/test_index.py
```diff
@@ -301,7 +304,9 @@ def _make_event(
                 "handle": "pkg/usr",
                 "key": "foo",
                 "last_modified": "not_an_object",
+                "pointer_file": "1598026253",
```
akarve, line 307:
```suggestion
```

PR #1783, opened 2020-09-03
Use temporary files and unique names for scratch files in Lambda

akarve:
Without this, lambda invocations can stomp on one another's work, corrupting any shared files

### DIFF
--- a/lambdas/shared/t4_lambda_shared/preview.py
+++ b/lambdas/shared/t4_lambda_shared/preview.py
@@ -5,7 +5,7 @@ from io import BytesIO
 import math
 import os
 import re
-from uuid import uuid4
+import tempfile
 import zlib
 
 import fcsparser
@@ -27,7 +27,7 @@ CATALOG_LIMIT_LINES = 512  # must be positive int
 ELASTIC_LIMIT_BYTES = int(os.getenv('DOC_LIMIT_BYTES') or 10_000)
 ELASTIC_LIMIT_LINES = 100_000
 MAX_PREVIEW_ROWS = 1_000
-TEMP_DIR = "/tmp"
+TEMP_DIR = "/tmp/"
 # common string used to explain truncation to user
 TRUNCATED = (
     'Rows and columns truncated for preview. '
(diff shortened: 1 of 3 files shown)


lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -26,7 +27,7 @@
 ELASTIC_LIMIT_BYTES = int(os.getenv('DOC_LIMIT_BYTES') or 10_000)
 ELASTIC_LIMIT_LINES = 100_000
 MAX_PREVIEW_ROWS = 1_000
-TEMP_DIR = "/tmp"
+TEMP_DIR = "/tmp/"
```
akarve, line 30: this is lambda so always linux so we can get away with non-windows path-seps

lambdas/shared/tests/test_preview.py
```diff
@@ -80,7 +80,7 @@ def test_fcs(self):
                 'has_warnings': True,
             },
         }
-        with TemporaryDirectory() as tmp_dir, patch('t4_lambda_shared.preview.TEMP_DIR', tmp_dir):
+        with TemporaryDirectory() as tmp_dir, patch('t4_lambda_shared.preview.TEMP_DIR', tmp_dir + os.path.sep):
```
akarve, line 83: wow this is ugly but more correct
```suggestion
        with TemporaryDirectory() as tmp_dir, patch('t4_lambda_shared.preview.TEMP_DIR', os.path.join(tmp_dir, "", ""):
```

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -26,7 +27,7 @@
 ELASTIC_LIMIT_BYTES = int(os.getenv('DOC_LIMIT_BYTES') or 10_000)
 ELASTIC_LIMIT_LINES = 100_000
 MAX_PREVIEW_ROWS = 1_000
-TEMP_DIR = "/tmp"
+TEMP_DIR = os.path.join('/tmp/', '')
```
akarve, line 30:
```suggestion
TEMP_DIR = os.path.join('/tmp', '')
```

dimaryaz, line 30: `TEMP_DIR` shouldn't be needed at all - `NamedTemporaryFile` already uses `/tmp` by default.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -75,20 +76,22 @@ def extract_fcs(file_, as_html=True):
...
-    # FCS files typically ~1MB
-    with open(file_path, 'wb') as real_file:
-        real_file.write(file_.read())
-    try:
-        meta, data = fcsparser.parse(file_path, reformat_meta=True)
-    except Exception as first:  # pylint: disable=broad-except
+    # FCS files typically < 500MB (Lambda disk)
+    # per Lambda docs we can use tmp/*, OK to overwrite
+    with tempfile.NamedTemporaryFile(prefix=TEMP_DIR) as tmp:
+        inbound = file_.read()
```
dimaryaz, line 83: If they can be up to 500MB, this should really de done in a loop. Or maybe `shutil.copyfileobj` will do the right thing.

akarve: we have 1GB+ in our lambdas but you're right, no reason to be sloppy with memory.

kevinemoore (approved):

PR #1785, opened 2020-09-03
Add OneLogin support

dimaryaz:
A lot of copying and pasting here, but with some very subtle differences. Probably good to refactor this some day.


catalog/app/utils/OneLogin.js
```diff
@@ -0,0 +1,81 @@
...
+
+export function useOneLogin({ clientId, baseUrl }) {
+  return React.useCallback(
+    () =>
+      new Promise((resolve, reject) => {
+        const nonce = Math.random().toString(36).substr(2)
+        const state = Math.random().toString(36).substr(2)
+        const query = NamedRoutes.mkSearch({
+          client_id: clientId,
+          redirect_uri: `${window.location.origin}/post-message.html`,
```
dimaryaz, line 24: OneLogin doesn't support `postMessage` - only redirects - so we redirect to our own URL that calls `postMessage`.

This URL will need to be whitelisted by the customer, so needs to not ever change. Should probably look pretty, too.

Questions:
- What should the path be?
- What's the right way to configure `node` to use a custom path? I'm guessing same thing as we're doing for `__embed`, but I'm wondering if there's an easier solution.

catalog/app/containers/Auth/onelogin-logo.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
nl0, line 1: @dimaryaz pls optimize this svg with `svgo` (`npx svgo --multipass  -i app/containers/Auth/onelogin-logo.svg`)

catalog/static/post-message.html
```diff
@@ -0,0 +1,8 @@
+<!doctype html>
+<html lang="en">
+  <body>
+    <script>
+      window.opener.postMessage(window.location.hash, window.location.origin);
```
nl0, line 5: if using postMessage, i think we should use some `type` field to help the consuming code filter messages and ensure it's the proper one

catalog/app/utils/OneLogin.js
nl0: > so we redirect to our own URL that calls postMessage.

well, the postMessage isnt required, actually -- callback page can be empty and we can just spy on the popup location changes and then, when it changes to our callback route, just grab the location directly from the popup

> Should probably look pretty, too.

not that important, imo

> What should the path be?

well, i often call it `auth-callback` or smth similar, but i guess it's no big deal, up to you

> What's the right way to configure node to use a custom path?

`res.sendFile` should do the trick if you decide to use a prettier path

nl0: @dimaryaz ping me if you need any help tweaking the node server

nl0 (commented):
looks good in general, some inline comments

catalog/app/utils/OneLogin.js
dimaryaz: > well, the postMessage isnt required, actually -- callback page can be empty and we can just spy on the popup location 
changes and then, when it changes to our callback route, just grab the location directly from the popup

Yeah, I've thought about that - but then the main page would have to poll, while postMessage is just one line of code.

Theoretically, we could do the redirect to our own page for _all_ OIDC providers, and not even both using e.g. Google's SDK. Don't know if it's worth it, though.

> well, i often call it `auth-callback` or smth similar, but i guess it's no big deal, up to you

Ok

> `res.sendFile` should do the trick if you decide to use a prettier path

So same as `__embed`? I can make that work.

catalog/app/containers/Auth/onelogin-logo.svg
dimaryaz: Ok

catalog/static/post-message.html
dimaryaz: Ok

nl0 (approved):

PR #1786, opened 2020-09-05
Drag-n-drop package creation (RC)

nl0:


kevinemoore (dismissed):
A couple quick changes:
1) Validate that the package name is valid (has a / in it). I'll add a similar check to the API, but better to enforce it in the UI too.
2) Add s3:// to the bucket name (s3://bucket) for the registry argument (alternatively, we can change that parameter to "bucket" if you prefer)

akarve:
> A couple quick changes:
> 
> 1. Validate that the package name is valid (has a / in it). I'll add a similar check to the API, but better to enforce it in the UI too.
> 2. Add s3:// to the bucket name (s3://bucket) for the registry argument (alternatively, we can change that parameter to "bucket" if you prefer)

1. we should validate package name via API as that logic is going to change and is nontrivial; so it's cleaner if the frontend doesn't need the logic
2. we should keep exactly the same name as the API but hint text will clear things up (and we can put "S3 bucket" in help text

api/python/quilt3/backends/base.py
```diff
@@ -138,8 +138,10 @@ def push_manifest(self, pkg_name: str, top_hash: str, manifest_data: bytes):
         put_bytes(manifest_data, self.manifest_pk(pkg_name, top_hash))
```
akarve, line 138:
```suggestion
        """returns: timestamp to support catalog drag-and-drop => browse"""
        put_bytes(manifest_data, self.manifest_pk(pkg_name, top_hash))
```

akarve (approved):

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -0,0 +1,845 @@
...
+                              totalProgress.percent < 100
+                                ? totalProgress.percent * 0.9
+                                : undefined
+                            }
+                          />
+                          <M.Box pl={1} />
+                          <M.Typography variant="body2" color="textSecondary">
+                            {totalProgress.percent < 100
+                              ? 'Uploading files'
+                              : 'Creating a manifest'}
```
akarve, line 809:
```suggestion
                              : 'Writing package manifest'}
```

PR #1787, opened 2020-09-06
Search: better error message, fix previews in global search + bulletproof downloads (auto-refresh, binary/octet-stream content-type) + expose storage class header in s3-proxy

nl0:


akarve (approved):
Tested key use cases from title on staging. Will deploy.

PR #1788, opened 2020-09-07
Multi-thread hashing of large S3 files

sir-sigurd:
## Description

```
Before:

In [375]: quilt3.data_transfer.calculate_sha256([quilt3.util.PhysicalKey('ai2-datasets', 'arc/ARC-V1-Feb2018.zip', None)], [680841265])
Hashing: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 681M/681M [04:46<00:00, 2.38MB/s]
Out[375]: ['6d2d5ab50b2ceec6ba5f79c921be77cf2de712ea25a2b3f4fff3acc101cecfa0']

After:

In [2]: quilt3.data_transfer.calculate_sha256([quilt3.util.PhysicalKey('ai2-datasets', 'arc/ARC-V1-Feb2018.zip', None)], [680841265])
Hashing: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 681M/681M [01:09<00:00, 9.86MB/s]
Out[2]: ['6d2d5ab50b2ceec6ba5f79c921be77cf2de712ea25a2b3f4fff3acc101cecfa0']
```

## TODO

- [ ] finer progress bar updates
- [ ] test performance with multiple files
- [ ] finer concurrency control
- [ ] tests


dimaryaz (commented):
My biggest concern here is introducing `asyncio` for just one function, while using regular threadpools everywhere else.

I think `asyncio` is great and we should probably switch everything to it - but it's a fairly big and risky change, so we should do it separately, not as part of a performance fix for one function. We should think about proper design, and possibly even expose async functions to the user. But, that will require lots of testing. For now, I think it's best to fix the download the "old" way.

One potential problem with parallel downloads: because hashing has to be done sequentially and cannot be parallelized (for a given file), there is a possibility of running out of memory. If the network is very fast and CPU cannot keep up with the download, this can potentially load every chunk of the file into memory at the same time.

akarve:
I talked a bit more with Dima about this and we're OK with `asyncio` if that's your choice; we just need to test it and unit test it carefully as it's a new dependency for us.

api/python/quilt3/data_transfer.py
```diff
@@ -856,6 +861,76 @@ def calculate_sha256(src_list: List[PhysicalKey], sizes: List[int]):
     return _calculate_sha256_internal(src_list, sizes, [None] * len(src_list))
 
 
+async def _calculate_hash_s3_get_parts(s3_client_provider, s3_executor, s3_pending_parts_semaphore, src, size):
```
akarve, line 864: consider `, *,` to make the last ~3 args mandatory kwargs, thus avoiding bad positional calls

api/python/quilt3/data_transfer.py
```diff
@@ -856,6 +861,76 @@ def calculate_sha256(src_list: List[PhysicalKey], sizes: List[int]):
...
+        (None,)
+    )
+
+    s3_client = s3_client_provider.find_correct_client(S3Api.GET_OBJECT, src.bucket, params)
+
+    def read_to_queue(part_number, q):
+        logger.debug('%r part %s: download started', src, part_number)
+        if part_number is not None:
+            start = part_number * part_size
+            end = min(start + part_size, size) - 1
```
akarve, line 887: please add a comment on why -1, is `Range=` exclusive, etc.

api/python/quilt3/data_transfer.py
```diff
@@ -856,6 +861,76 @@ def calculate_sha256(src_list: List[PhysicalKey], sizes: List[int]):
...
+
+    async def get_part(part_number, q):
+        await s3_pending_parts_semaphore.acquire()
+        try:
+            logger.debug('%r part %s: download enqueued', src, part_number)
+            await loop.run_in_executor(s3_executor, read_to_queue, part_number, q)
+        except Exception:
+            s3_pending_parts_semaphore.release()
+            raise
+
```
akarve, line 908: please add comment on where "normal case" release takes place (below)

api/python/quilt3/data_transfer.py
```diff
@@ -867,20 +942,29 @@ def _calculate_sha256_internal(src_list, sizes, results):
...
-    lock = Lock()
-
-    with tqdm(desc="Hashing", total=total_size, unit='B', unit_scale=True, disable=DISABLE_TQDM) as progress:
-        def _process_url(src, size):
-            hash_obj = hashlib.sha256()
-            if src.is_local():
+    s3_client_provider = S3ClientProvider()
+    # This controls how many parts can be stored in the memory.
+    # This includes the ones that are being downloaded or hashed.
+    s3_max_pending_parts = s3_transfer_config.max_request_concurrency * 4
```
akarve, line 948: why * 4?

api/python/quilt3/data_transfer.py
```diff
@@ -867,20 +942,29 @@ def _calculate_sha256_internal(src_list, sizes, results):
...
+
+    async def main():
+        loop = asyncio._get_running_loop()
+
+        s3_pending_parts_semaphore = asyncio.BoundedSemaphore(s3_max_pending_parts)
+        with tqdm(desc="Hashing", total=total_size, unit='B', unit_scale=True, disable=DISABLE_TQDM) as progress, \
```
akarve, line 954: why not `unit=MB`? that's what we do today

akarve (commented):
Tested on large complex package and got the same top-hash (nice work); a few comments inline

api/python/quilt3/data_transfer.py
```diff
@@ -32,6 +34,9 @@
 MAX_FIX_HASH_RETRIES = 3
 
 
+logger = logging.getLogger('quilt3')
```
akarve, line 37: can we make this a global object in util?

akarve:
I got the following error in Jupyter (can you repro in iPython as well)? A lot of our users are using `quilt3` in Jupyter/iPython so we'll have to consider that:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-11-91e1a0c98333> in <module>
      2     "sequencer/GUID",
      3     message="Initial instrument data upload",
----> 4     registry=f"s3://{buckets['raw']}"
      5 )

~/code/quilt/api/python/quilt3/telemetry.py in decorated(*args, **kwargs)
    129             ApiTelemetry.report_api_use(self.api_name, ApiTelemetry.session_id)
    130 
--> 131             results = func(*args, **kwargs)
    132             # print(f"{len(ApiTelemetry.pending_reqs)} request(s) pending!")
    133 

~/code/quilt/api/python/quilt3/packages.py in push(self, name, registry, dest, message, selector_fn)
   1257                 )
   1258 
-> 1259         self._fix_sha256()
   1260 
   1261         pkg = self.__class__()

~/code/quilt/api/python/quilt3/packages.py in _fix_sha256(self)
    886             sizes.append(entry.size)
    887 
--> 888         results = calculate_sha256(physical_keys, sizes)
    889         exc = None
    890         for entry, obj_hash in zip(self._incomplete_entries, results):

~/code/quilt/api/python/quilt3/data_transfer.py in calculate_sha256(src_list, sizes)
    859     if not src_list:
    860         return []
--> 861     return _calculate_sha256_internal(src_list, sizes, [None] * len(src_list))
    862 
    863 

~/anaconda3/envs/de/lib/python3.7/site-packages/tenacity/__init__.py in wrapped_f(*args, **kw)
    327         @_utils.wraps(f)
    328         def wrapped_f(*args, **kw):
--> 329             return self.call(f, *args, **kw)
    330 
    331         def retry_with(*args, **kwargs):

~/anaconda3/envs/de/lib/python3.7/site-packages/tenacity/__init__.py in call(self, fn, *args, **kwargs)
    407             retry_object=self, fn=fn, args=args, kwargs=kwargs)
    408         while True:
--> 409             do = self.iter(retry_state=retry_state)
    410             if isinstance(do, DoAttempt):
    411                 try:

~/anaconda3/envs/de/lib/python3.7/site-packages/tenacity/__init__.py in iter(self, retry_state)
    354             and isinstance(retry_state.outcome.exception(), TryAgain)
    355         if not (is_explicit_retry or self.retry(retry_state=retry_state)):
--> 356             return fut.result()
    357 
    358         if self.after is not None:

~/anaconda3/envs/de/lib/python3.7/concurrent/futures/_base.py in result(self, timeout)
    426                 raise CancelledError()
    427             elif self._state == FINISHED:
--> 428                 return self.__get_result()
    429 
    430             self._condition.wait(timeout)

~/anaconda3/envs/de/lib/python3.7/concurrent/futures/_base.py in __get_result(self)
    382     def __get_result(self):
    383         if self._exception:
--> 384             raise self._exception
    385         else:
    386             return self._result

~/anaconda3/envs/de/lib/python3.7/site-packages/tenacity/__init__.py in call(self, fn, *args, **kwargs)
    410             if isinstance(do, DoAttempt):
    411                 try:
--> 412                     result = fn(*args, **kwargs)
    413                 except BaseException:
    414                     retry_state.set_exception(sys.exc_info())

~/code/quilt/api/python/quilt3/data_transfer.py in _calculate_sha256_internal(src_list, sizes, results)
   1005                 results[i] = await f
   1006 
-> 1007     util.asyncio_run(main())
   1008     return results
   1009 

~/anaconda3/envs/de/lib/python3.7/asyncio/runners.py in run(main, debug)
     32     if events._get_running_loop() is not None:
     33         raise RuntimeError(
---> 34             "asyncio.run() cannot be called from a running event loop")
     35 
     36     if not coroutines.iscoroutine(main):

RuntimeError: asyncio.run() cannot be called from a running event loop
```

api/python/quilt3/data_transfer.py
dimaryaz: I think tqdm adds prefixes automatically, so it should be `M`.

api/python/quilt3/data_transfer.py
```diff
@@ -19,6 +20,7 @@
 from botocore.exceptions import ClientError, ConnectionError, HTTPClientError, ReadTimeoutError
 import boto3
 from boto3.s3.transfer import TransferConfig
+from quilt3 import util
```
dimaryaz, line 23: Quilt imports should go below third party ones. Also, just import individual functions; see the `from .util import ...` below

dimaryaz (commented):
Regarding Jupyter: that's my concern about adding `asyncio` :-\ We don't know what's going to break.

akarve:
> Regarding Jupyter: that's my concern about adding `asyncio` :-\ We don't know what's going to break.

This [might be solved](https://stackoverflow.com/questions/47518874/how-do-i-run-python-asyncio-code-in-a-jupyter-notebook) in the new Jupyter. Still seems kinda strict to require a specific version of Jupyter.

How different exactly would this code like with `threading` primitives?

api/python/quilt3/data_transfer.py
sir-sigurd: > unit_scale: bool or int or float, optional
If 1 or True, the number of iterations will be printed with an appropriate SI metric prefix (k = 10^3, M = 10^6, etc.) [default: False]. If any other non-zero number, will scale total and n.

api/python/quilt3/data_transfer.py
sir-sigurd: We already have code like this, I'd like to deduplicate it a later PR.

api/python/quilt3/data_transfer.py
```diff
@@ -874,6 +881,88 @@ def calculate_sha256(src_list: List[PhysicalKey], sizes: List[int]):
...
+            yield from iter(q.get, None)
+            f.result()
+            logger.debug('%r part %s: processed', src, part_number)
+        except GeneratorExit:
+            if f.cancel():
+                logger.debug('%r part %s: cancelled', src, part_number)
+            else:
+                stopped_event.set()
+        finally:
+            if not f.cancelled():
```
kevinemoore, line 941: Why is the semaphore release only called if not `f.cancelled`?

kevinemoore: Is that so the hashing of that part will be retried?

kevinemoore (approved):
I'm not confident in my ability to review these changes, but I don't see any issues. It would be great to get @dimaryaz to take a look.

akarve (approved):

api/python/quilt3/data_transfer.py
akarve: noted; it should be a named constant; might even make sense to adjust it based on available memory

api/python/quilt3/data_transfer.py
sir-sigurd: Future can be cancelled only before being runned, so semaphore wasn't acquired when `f.cancelled()` is true.

dimaryaz: It's still not great to have `acquire` and `release` in different places - if nothing else, it's hard to prove that the logic is correct.

Theoretically, if `logger.debug('%r part %s: download enqueued', src, part_number)` throws an exception, then `release` will be called without an `acquire`. It's pretty difficult to make that happen, so kind of a negligible risk, but still...

dimaryaz (approved):
It's still a bit more complicated than I'd like, but I think it's correct, so don't want to block on that. We can improve it in the future.

PR #1793, opened 2020-09-10
Cli push meta

cosmic-byte:


api/python/quilt3/main.py
```diff
@@ -422,6 +428,11 @@ def create_parser():
         help="The commit message for the new package",
         type=str,
     )
+    optional_args.add_argument(
+        "--meta",
+        help="Sets package level metadata",
```
akarve, line 433:
```suggestion
        help="Sets package-level metadata",
```

api/python/tests/test_cli.py
```diff
@@ -41,5 +42,38 @@ def test_push(self):
                 main.main(('push', '--dir', tmp_dir, name))
 
                 mocked_package_class.assert_called_once_with(quilt3.Package)
-                mocked_set_dir.assert_called_once_with('.', tmp_dir)
+                mocked_set_dir.assert_called_once_with('.', tmp_dir, meta=None)
```
akarve, line 45: instead of changing this call we should have a default falsy value for `meta={}` for example; it might even get rid of the `if meta` special case in main.py.

api/python/tests/test_cli.py
```diff
@@ -41,5 +42,38 @@ def test_push(self):
...
                 mocked_push.assert_called_once_with(name, registry=None, dest=None, message=None)
+
+
+@pytest.mark.parametrize(
+    'arg, meta, expected_set_dir_count, expected_push_count, expected_meta',
+    [
+        (None, None, 1, 1, None),
+        ('--meta', '{invalid: meta}', 0, 0, {}),
```
akarve, line 53: this should xfail with QuiltException?

api/python/tests/test_cli.py
```diff
@@ -41,5 +42,38 @@ def test_push(self):
...
+        (Path(tmp_dir) / 'foo').touch()
+        (Path(tmp_dir) / 'bar').mkdir()
+        (Path(tmp_dir) / 'bar' / 'baz')
+
+        with mock.patch('quilt3.Package.__new__', return_value=pkg) as mocked_package_class,\
+             mock.patch.object(pkg, 'set_dir', wraps=pkg.set_dir) as mocked_set_dir, \
+             mock.patch.object(pkg, 'push') as mocked_push:
+
+            if arg:
+                main.main(('push', '--dir', tmp_dir, name, arg, meta))
```
akarve, line 73: where does --regsitry come from? fine if there's a default but we should note the same

api/python/tests/test_cli.py
cosmic-byte: we can't set a default for `meta` in main.py because 
```
meta = json.loads(meta)
```
expects a `str` not a `dict` nor a `NoneType`.  Also, it will still require changing the expected call param you pointed at above.

api/python/tests/test_cli.py
cosmic-byte: It doesn't fail since QuiltException is already handled in [main.py](https://github.com/quiltdata/quilt/blob/aed61e374043355d15c9aaa3e01e9a9418fb703a/api/python/quilt3/main.py#L439)

akarve: sure we should still check that the exception is thrown. one way is patch error handler and make sure it gets called.

api/python/tests/test_cli.py
cosmic-byte: I don't understand the question though. This test is for --meta data and not --registry and also registry is optional.

akarve: this is ok as is but you can note in a comment that registry is optional

api/python/quilt3/main.py
```diff
@@ -422,6 +428,14 @@ def create_parser():
         help="The commit message for the new package",
         type=str,
     )
+    optional_args.add_argument(
+        "--meta",
+        help="""
+            Sets package-level metadata.
+            Format: A valid json string with double-quote key names ie. '{"foo": "bar"}'
```
akarve, line 435:
```suggestion
            Format: A json string with keys in double quotes '{"key": "value"}'
```

akarve (approved):

PR #1799, opened 2020-09-13
Bucket/Code: fix crash when clicking on already selected btn

nl0:


nl0:
@akarve ima merge this since it's quite trivial

PR #1804, opened 2020-09-17
es/indexer: Drop fatal 40X failures on the floor

akarve:
Importantly, we choose to skip the dead letter queue in select cases where we are unlikely to recover from the error, so as to avoid failing the batch. We found a case where list_object_versions may return an object that will never be readable (e.g. due to a foreign owner).


kevinemoore (approved):
This looks like a good change to me. Ideally, we'd capture keys that were not indexed in a better place than just the lambda logs. I don't think the DLQ is any better and we'll end up with more correctly indexed files this way so it seems like a clear win.

PR #1806, opened 2020-09-17
Package metadata display

nl0:
### TODO
- [x] partially expand (tease) meta section

### DIFF
--- a/catalog/app/components/JsonDisplay/JsonDisplay.js
+++ b/catalog/app/components/JsonDisplay/JsonDisplay.js
@@ -29,7 +29,7 @@ function PrimitiveValue({ children, ...props }) {
 
 function PrimitiveEntry({ name, value, topLevel = true, ...props }) {
   return (
-    <Box display="flex" alignItems="center" {...props}>
+    <Box display="flex" {...props}>
       {!topLevel && <IconBlank />}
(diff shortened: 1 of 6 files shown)


akarve (commented):
Looks good; can we add a teased metadata panel to the package root or do you want to do that in a separate change set?

nl0:
> can we add a teased metadata panel to the package root

@akarve it's already there (should work with the recent fix)

akarve (commented):
Cool and I'm approving but please make this change first (if it's not a big deal). Instead of displaying the number of top-level keys can you auto-expand the top-level group (example below; that's how it should start)? it will be much more discoverable that way. and, if possible, instead of showing the number of keys in an unexpanded section, we should show the key names themselves (if it's easy).

Screen Shot 2020-09-21 at 18 52 30

nl0:
> Cool and I'm approving but please make this change first (if it's not a big deal). Instead of displaying the number of top-level keys can you auto-expand the top-level group (example below; that's how it should start)? it will be much more discoverable that way. and, if possible, instead of showing the number of keys in an unexpanded section, we should show the key names themselves (if it's easy).

@akarve done

akarve (approved):

PR #1807, opened 2020-09-18
Case-insensitive package filtering

nl0:


akarve (approved):
looks good but can you up the debounce time? i think we should skeletonize immediately when input starts but wait 250 to 500 msec before flushing the query to ES

PR #1811, opened 2020-09-21
Add PDF to overview

nl0:
Add PDF to overview


catalog/app/containers/Bucket/requests.js
```diff
@@ -224,6 +224,7 @@ const SAMPLE_EXTS = [
   '.xlsx',
   '.ipynb',
   '.md',
+  '.pdf',
```
akarve, line 227: It's a bit outside of this diff but I think we should consider `*.EXT*` matching against `ext:` (instead of exact match for a few reasons):
* for a file like `foo/bar/baz.wb.scene` the extension is actually `wb.scene`
* What about, like, gzipped json? which would be `*json.gz`.

I know this can be expensive but the limit should save us and it should be strictly more useful. e.g. it's common for AI buckets to be only `.json.gz` or something similar.

nl0: fixed

akarve (approved):

PR #1812, opened 2020-09-21
Search: smarter retries and better error messages

nl0:
- [x] smarter search retries
- [x] extract elastic error details
- [x] left-align text


akarve:
For all retry cases (I still need to update the endpoint) can you send `is_retry=X` where X is an integer? (in which case I will try some remediation on the backend, like skipping documents, etc.)? If the user hard reloads the page, X goes back to zero so we don't need to be fancy about trackin gstate.

The Server Error message title should be `Search error (CODE)` for all messages. Underneath we should say:
"ElasticSearch had trouble with that query. The cluster may be busy indexing new documents. Try again later or click RETRY to try a simplified version of your query." And the button should say "RETRY SIMPLIFIED QUERY".

nl0:
@akarve ready for a review

catalog/app/components/SearchResults/SearchResults.js
```diff
@@ -24,8 +24,10 @@ import usePrevious from 'utils/usePrevious'
 import * as requests from 'containers/Bucket/requests'
 
 const PER_PAGE = 10
-const ES_V = '6.7'
-const ES_REF = `https://www.elastic.co/guide/en/elasticsearch/reference/${ES_V}/query-dsl-query-string-query.html#query-string-syntax`
+const ES_V = '6.8'
```
nl0, line 27: @akarve what's the correct version?

akarve (approved):

PR #1815, opened 2020-09-22
Bucket overview: show total package count

nl0:
#


akarve (changes requested):
I think we need to search `BUCKET*` in order for the package count to come back non-zero. Right now I'm seeing `prod/search?index=BUCKET&action=stats`. Also there's a weird thing, maybe it's Chrome IDK, where I see two requests to stats, both are 200s, but I can only see the response for one of them.

nl0:
@akarve 
> I think we need to search BUCKET* in order for the package count to come back non-zero

yep, that worked, thanks

> Right now I'm seeing prod/search?index=BUCKET&action=stats. Also there's a weird thing, maybe it's Chrome IDK, where I see two requests to stats, both are 200s, but I can only see the response for one of them.

it's the preflight (OPTIONS) request for CORS, it doesnt have a response body

akarve (approved):

PR #1821, opened 2020-09-24
UploadDialog: in-browser hashing + warn on exceeding max size

nl0:
UploadDialog: in-browser hashing


kevinemoore (approved):
I tested this change and everything works fine. It's a big improvement to check the max size before uploading and it's very convenient to be able to remove individual files and immediately know if the new package is over or under the limit.

PR #1824, opened 2020-09-25
Admin/Buckets: re-index & repair + adjust SNS ARN input

nl0:
### TODO

- [x] wordsmith @akarve

### DIFF
--- a/catalog/app/constants/style.js
+++ b/catalog/app/constants/style.js
@@ -16,7 +16,7 @@ const appPalette = {
   },
   warning: {
     main: colors.yellow[200],
-    dark: colors.yellow[800],
+    dark: colors.yellow[900],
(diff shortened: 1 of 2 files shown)


kevinemoore (approved):
Now that we have a re-index button, we could really use a "refresh" button for the bucket grid (e.g., to see if indexing is finished).

nl0:
@akarve i've updated the UI to use checkbox + text field and it works (tho backend is still failing with 500 when trying to add `quilt-nl0-stage-2` bucket with blank sns arn; seems like this bug is specific to that bucket, since adding other buckets works). i've also adjusted re-index-related copy a little, but pls have another look anyways.

akarve (approved):

PR #1833, opened 2020-10-06
Follow-up to #1829: Simplify condition

sir-sigurd:


lambdas/es/indexer/document_queue.py
```diff
@@ -118,11 +118,9 @@ def append(
...
             if not handle or not package_hash or not pointer_file:
                 raise ValueError("missing required argument for package document")
-            if (
-                package_stats
-                and not isinstance(package_stats, dict)
-                or isinstance(package_stats, dict)
-                and any(k not in package_stats for k in ['total_files', 'total_bytes'])
+            if not (
+                package_stats is None
+                or isinstance(package_stats, dict) and {'total_files', 'total_bytes'}.issubset(package_stats)
```
sir-sigurd, line 123: This also makes the exception to be raised on non-dict falsy values which I think was expected.

akarve (approved):

PR #1838, opened 2020-10-07
Catalog, Search: add help dropdown to the index landing page

fiskus:
Search Help has a slightly different look for Header and OpenLanding. I made `NavBar/Help.js` more abstract (in terms of styles) and reused for OpenLanding. Maybe I should move it to another place, ex. `components/SearchHelp.js`

demo


catalog/app/containers/NavBar/Help.js
```diff
@@ -247,11 +238,17 @@ function DocsExternalLink() {
   )
 }
 
-export default function Help({ className, onQuery, ...props }) {
+function normalizeSyntaxItem(s) {
+  const multilineToLine = s.replace(/\s/g, '')
+  const trailingSpace = ' '
```
nl0, line 243: i dont think it makes sense to instantiate this variable every time the function gets called, just `return "{mulitilineToLine} "` (wrong quotes bc of markdown syntax) or even make it into a oneliner

nl0 (changes requested):
The changes look good overall, but i think the huge rounded corners look a little out-of-place. i'd rather go for the standard border-radius, like this:
localhost_3000_(Laptop 2x)

catalog/app/containers/NavBar/Help.js
fiskus: The performance cost is insignificant here in my opinion, the transpiller even can remove these variables. But I think it's simpler to read code when all things have some explanations through variable names.

fiskus: For example, in the future when I will see `{multiline} ` I can think that this extra space is a mistyping

fiskus: Converted to oneliner

fiskus:
> the huge rounded corners look a little out-of-place. i'd rather go for the standard border-radius

Removed too much rounded corners as you propose

nl0 (approved):
haven't tested it after the most recent changes, but i believe it should be ok

docs/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ## Catalog, Lambdas
 * [Added] Retry logic for failed queries, minimize load on ES for sample, images
+* [Added] Search help dropdown for the index landing page ([1838](https://github.com/quiltdata/quilt/pull/1838))
```
nl0, line 17: i believe the pr number should be prefixed with a `#` by convention

fiskus: Thanks!

PR #1840, opened 2020-10-07
Add changelog file

sir-sigurd:


docs/CHANGELOG.md
```diff
@@ -0,0 +1,33 @@
...
+* Performance improvements for build and push
+
+## Catalog & Lambdas:
+* PDF previews
+* Browse full package contents (no longer limited to 1000 files)
+* Indexing and search package-level metadata
+* Fixed issue with download button for certain text files
+* FCS files: content indexing and preview
+* Catalog sign-in with email (or username)
+* Catalog support for sign-in with Okta
```
sir-sigurd, line 33: These are the notes from the latest [release](https://github.com/quiltdata/quilt/releases/tag/v3.2.0).
If we want, we could populate it from all previous releases doing this:
```bash
pip install PyGithub
```
```python
from github import Github
g = Github()
repo = g.get_repo('quiltdata/quilt')
releases = [r for r in r.get_releases() if not r.prerelease]
data = '\n\n'.join([
    f'# {r.tag_name.lstrip("v")} - {r.published_at.date()} - {r.title}' + (f'\n{r.body.rstrip()}' if r.body else '')
    for r in releases
])
with open('docs/CHANGELOG.md', 'w') as f:
    f.write(data)
```

akarve (approved):
Looks good but, if there are any manual steps requires, be sure to add them in https://github.com/quiltdata/quilt/blob/master/docs/pull_request_template.md so that devs have a reminder. For instance, it is much more important for us to name PRs correctly.

Should we add [feature], [performance] etc. when we create the new PR?

I would also really like to link each feature to docs, or have the option to include the PR description? That can be for the backlog if it is too complicated. But for example, the `meta` feature we should really have an example like `quilt3 push --meta`. Again, I can live without this now but let's fix this by having a PR template that is so good the developer writes the release notes for us.

docs/CHANGELOG.md
```diff
@@ -0,0 +1,367 @@
...
+# unreleased - YYYY-MM-DD
+## Python API
+* [performance] Improve performance of hashing of S3 files
+  ([#1816](https://github.com/quiltdata/quilt/issues/1816))
+
+## CLI
+* [feature] Add `meta` argument to the push command
+  ([#1793](https://github.com/quiltdata/quilt/issues/1793))
+
+## Catalog
```
sir-sigurd, line 19: @nl0 any suggestions on the format?

fiskus: I don't know how often the catalog is deployed. And is it deployed on its own or alongside with labdas.

I can suggest only:

## [2020-12-31 12:48]
* [feature] New shiny feature
* [fix] Fix ugly bug

Time of the day is relevant if there are multiple deployment throughout the day.

nl0: since there's no separate release process for the catalog, we should keep all the changes under one version, as it were before, and the current format (`x.y.z - YYYY-MM-DD`) works fine for this imo.

we should agree on prefixes (types of changes) tho, like, i dont think `feature` is a good prefix in this context, i'd rather start with the types suggested by [keep a changelog](https://keepachangelog.com/en/1.0.0/#how), and maybe added `Performance` -- it makes perfect sense (maybe some other types will come up later).

nl0 (approved):

PR #1843, opened 2020-10-07
package list: show "push" button when no packages

nl0:
## TODO

- [x] wordsmith if necessary @akarve 

quilt-bucket-onboarding

### DIFF
--- a/catalog/app/containers/Bucket/PackageList.js
+++ b/catalog/app/containers/Bucket/PackageList.js
@@ -6,7 +6,6 @@ import { useHistory, Link } from 'react-router-dom'
 import * as M from '@material-ui/core'
 import { fade } from '@material-ui/core/styles'
 
-import Message from 'components/Message'
 import Skeleton from 'components/Skeleton'
 import Sparkline from 'components/Sparkline'
 import * as AWS from 'utils/AWS'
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Bucket/PackageList.js
```diff
@@ -427,12 +426,26 @@ export default function PackageList({
...
+                  variant="contained"
+                  color="primary"
+                  onClick={openUpload}
+                  startIcon={<M.Icon>add</M.Icon>}
+                >
+                  Push package
+                </M.Button>
+                <M.Box pt={2} />
+                <M.Typography>
+                  Learn how to{' '}
```
akarve, line 442:
```suggestion
                  Or{' '}
```

catalog/app/containers/Bucket/PackageList.js
```diff
@@ -427,12 +426,26 @@ export default function PackageList({
...
+                  <StyledLink href={EXAMPLE_PACKAGE_URL} target="_blank">
+                    create a package
```
akarve, line 444:
```suggestion
                    push a package
```

catalog/app/containers/Bucket/PackageList.js
```diff
@@ -427,12 +426,26 @@ export default function PackageList({
...
+                  </StyledLink>{' '}
+                  with Quilt Python API
```
akarve, line 446:
```suggestion
                  with the Quilt Python API.
```

akarve (approved):
Good start; we should merge but also backlog: if the user clicks Push Package, uses DND, and pushes, but then doesn't "browse" the package and closes window, then I think the list behind remains empty? Not sure if/how we can fix this but it might be a confusing UX.

PR #1844, opened 2020-10-07
Catalog: Fix inability to select bucket

fiskus:
## Description

Click outside search box or search help is handled by ClickAwayListener API. It was fired even when there is no real "outside" (no expanded/focused search box and no opened search help).

Now "click away" event fires when search box is focused or search help is opened.

### DIFF
--- a/catalog/app/containers/NavBar/Search.js
+++ b/catalog/app/containers/NavBar/Search.js
@@ -117,8 +117,15 @@ function SearchBox({
     wrapper: wrapperCls,
     ...classes
   } = useStyles()
+
+  const onClickAway = React.useCallback(() => {
+    if (expanded || helpOpened) {
+      onHelpClose()
+    }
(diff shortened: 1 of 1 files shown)


akarve (approved):
I confirm that this fixes the bug, but a few notes:
* Search bar now disappears when bucket dropdown gets focus. Is this intentional?
* Why didn't this affect the bucket dropdown at the default route `/`?
* Please add a preflight test for this after merge.

fiskus:
>     * Search bar now disappears when bucket dropdown gets focus. Is this intentional?

I can observe the same behaviour on open.quiltdata.com now. Based upon code, it was intended.

>     * Why didn't this affect the bucket dropdown at the default route `/`?

There are two implementations of high-level components for these components' wrapper. Global and Bucket, they slightly differ from one another in terms of state.

nl0 (commented):
lgtm

PR #1845, opened 2020-10-08
Refresh package list after pushing a package

nl0:


nl0:
follow-up to #1843

catalog/app/containers/Bucket/PackageList.js
```diff
@@ -386,7 +391,7 @@ export default function PackageList({
 
   return (
     <>
-      <UploadDialog bucket={bucket} open={uploadOpen} onClose={closeUpload} />
+      <UploadDialog {...{ bucket, refresh, open: uploadOpen, onClose: closeUpload }} />
```
fiskus, line 394: I think that names were implied to be "**openUpload**" and "closeUpload"

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -219,11 +221,12 @@ function FilesInput({ input, meta, uploads, setUploads, errors = {} }) {
     setUploads(R.dissoc(path))
   }
 
+  const onInputChange = input.onChange
```
fiskus, line 224: `input.onChange` used in onDrop callback too. Maybe you want to use `onInputChange` there too

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -758,6 +762,11 @@ export default function UploadDialog({ bucket, open, onClose }) {
           meta: getMetaValue(meta),
         },
       })
+      if (refresh) {
+        // wait for ES index to receive the new package data
+        await new Promise((resolve) => setTimeout(resolve, ES_LAG))
+        refresh()
```
fiskus, line 768: Hm... I would expect, that `refresh` reloads page or reloads some component. If it reloads data, I would use `reFetch`, or `reloadData`, or `retryFetching`

fiskus (approved):
I didn't dig into `Data.use` yet

catalog/app/containers/Bucket/PackageList.js
nl0: `uploadOpen` is a state, more like `uploadDialogIsOpen`, and `closeUpload` is an action/verb

catalog/app/containers/Bucket/UploadDialog.js
nl0: well, it _refreshes_ the data, so i think it's ok semantically)

catalog/app/containers/Bucket/UploadDialog.js
nl0: k

catalog/app/containers/Bucket/UploadDialog.js
fiskus: Of course. Just sound my thoughts :)

akarve (approved):

PR #1851, opened 2020-10-10
Revision list from ES

nl0:
@akarve this one includes lambda changes


catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -328,9 +203,159 @@ function Revision({ bucket, name, id, hash, stats, message, modified, counts })
...
+  time: {
+    ...t.typography.body1,
+    fontWeight: t.typography.fontWeightMedium,
+    lineHeight: t.typography.pxToRem(20),
+    whiteSpace: 'nowrap',
+  },
+  msg: {
+    WebkitBoxOrient: 'vertical',
+    WebkitLineClamp: 2,
+    display: '-webkit-box',
```
fiskus, line 276: Did you mean `display: 'flex'`?

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -144,141 +147,41 @@ const useRevisionStyles = M.makeStyles((t) => ({
...
-function Revision({ bucket, name, id, hash, stats, message, modified, counts }) {
-  const classes = useRevisionStyles()
-  const { urls } = NamedRoutes.use()
+function RevisionLayout({ link, msg, meta, hash, stats, counts }) {
+  const classes = useRevisionLayoutStyles()
   const t = M.useTheme()
   const xs = M.useMediaQuery(t.breakpoints.down('xs'))
   const sm = M.useMediaQuery(t.breakpoints.down('sm'))
-  // eslint-disable-next-line no-nested-ternary
   const sparklineW = xs ? 176 : sm ? 300 : 400
```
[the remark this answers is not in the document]
nl0: prettier think it's ok, and i agree (as long as there's no extra logic, just chained ternaries)

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -144,141 +147,41 @@ const useRevisionStyles = M.makeStyles((t) => ({
...
   const sparklineH = xs ? 32 : 48
-  const dateFmt = xs ? 'MMM d yyyy - h:mmaaaaa' : 'MMMM do yyyy - h:mma'
+  console.log('layout', { xs, sm, sparklineW, sparklineH })
```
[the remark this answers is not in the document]
nl0: yeah, good catch

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -144,141 +147,41 @@ const useRevisionStyles = M.makeStyles((t) => ({
...
-                  <M.Box height={20} display="flex" alignItems="center">
-                    <Skeleton borderRadius="borderRadius" height={16} width="90%" />
-                  </M.Box>
-                )}
-              </>
-            ),
-          },
-          message,
-        )}
+      <M.Box py={1} pl={2} pr={xs ? 2 : Math.ceil(sparklineW / 8 + 1)}>
```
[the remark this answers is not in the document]
nl0: yeah probably

catalog/app/containers/Bucket/PackageRevisions.js
nl0: nope, it's a hack (sorta) to only show two lines of text

fiskus: I would consider moving this pattern to the theme's "mixin" and use it like `t.spacing`. For example:

```js
// theme
createMuiTheme({
  trimLines(linesCount) => ({
     WebkitBoxOrient: 'vertical',
     WebkitLineClamp: 2,
     display: '-webkit-box',
  }),
})
```

```js
// component
makeStyles((t) => {
  msg: {
    ...t.trimLines(2),
  },
})
```

lambdas/search/index.py
```diff
@@ -66,11 +66,14 @@ def lambda_handler(request):
     user_indexes = request.args.get('index', "")
     user_size = request.args.get('size', DEFAULT_SIZE)
     user_source = request.args.get('_source', [])
+    user_from = request.args.get('from')
```
akarve, line 69:
```suggestion
    # 0-indexed starting position (for pagination)
    user_from = request.args.get('from')
```

lambdas/search/index.py
```diff
@@ -66,11 +66,14 @@ def lambda_handler(request):
...
     terminate_after = None  # see if we can skip os.getenv('MAX_DOCUMENTS_PER_SHARD')
 
     if not user_indexes or not isinstance(user_indexes, str):
         raise ValueError("Request must include index=<comma-separated string of indices>")
 
+    es_from = None
```
akarve, line 75: you don't need this because variables declared inside the if statement are valid in the whole function

lambdas/search/index.py
```diff
@@ -197,6 +201,7 @@ def lambda_handler(request):
         body=body,
         _source=_source,
         size=size,
+        from=es_from,
```
akarve, line 204: it looks like `_from` and not `from` is the right kwarg? see docs: https://elasticsearch-py.readthedocs.io/en/6.3.1/api.html#elasticsearch.Elasticsearch.search

Also we should default to `0` not `None`?

lambdas/search/index.py
```diff
@@ -91,6 +94,7 @@ def lambda_handler(request):
             raise ValueError("'packages' action searching indexes that don't end in '_packages'")
         _source = user_source
         size = user_size
+        es_from = user_from
```
akarve, line 97: I think I understand why we did this, but is there ever a case where the frontend sends `from=` and we shouldn't use it? If so it's better to fail way up top, i.e. `raise ValueError("...") if from and action != 'packages'`

akarve: last thing is let's add unit test coverage for this. should be easy to hack existing tests. i can show you next mtg. it's super constructive if you can modify the ES lambdas in general, so thanks for tackling this.

catalog/app/containers/Bucket/PackageRevisions.js
nl0: yeah, good idea. done

lambdas/search/index.py
nl0: it's `from_`, and yes, we should default to `0` probably

lambdas/search/index.py
nl0: done

lambdas/search/index.py
nl0: i think it's safe to accept `from_` in all cases. now trying to fix the tests

nl0: the tests seem ok

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -89,115 +86,35 @@ function RevisionInfo({ revision, bucket, name, path }) {
...
+                      {' | '}
+                      {dateFns.format(r.modified, 'MMMM do yyyy - h:mma')}
+                    </>
+                  }
+                  secondary={
+                    <span className={classes.secondaryText}>
+                      <span className={classes.line}>
+                        {r.message || <i>No message</i>}
+                      </span>
+                      <br />
```
fiskus, line 111: You already have `.line` class and it's `nowrap`ed, I think you can add `display: block` instead of `<br />`

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -6,14 +6,15 @@ import { Link } from 'react-router-dom'
 import * as M from '@material-ui/core'
 import { fade } from '@material-ui/core/styles'
 
+import JsonDisplay from 'components/JsonDisplay'
 import Skeleton from 'components/Skeleton'
 import Sparkline from 'components/Sparkline'
 import AsyncResult from 'utils/AsyncResult'
 import * as AWS from 'utils/AWS'
-import * as BucketConfig from 'utils/BucketConfig'
+// import * as BucketConfig from 'utils/BucketConfig'
```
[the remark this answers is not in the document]
fiskus: Comments become outdated easily and can be more harmful than useful.

If this is a working code, I would suggest to uncomment it and hide behind false condition like:

```js
const isFeatureFlagEnabled = false

if (isFeatureFlagEnabled) {
  enableNewWorkInProgressFeature()
}
```

If this is not working code, I think it should rather be deleted or replaced by comment with pseudocode (but Asana is better place for it :)).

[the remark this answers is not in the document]
fiskus: I understand that is too much for that PR and don't suggest to change this right now, because it's already a pattern. I just discuss the idea

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -144,141 +147,40 @@ const useRevisionStyles = M.makeStyles((t) => ({
...
 
-function Revision({ bucket, name, id, hash, stats, message, modified, counts }) {
-  const classes = useRevisionStyles()
-  const { urls } = NamedRoutes.use()
+function RevisionLayout({ link, msg, meta, hash, stats, counts }) {
+  const classes = useRevisionLayoutStyles()
   const t = M.useTheme()
   const xs = M.useMediaQuery(t.breakpoints.down('xs'))
   const sm = M.useMediaQuery(t.breakpoints.down('sm'))
-  // eslint-disable-next-line no-nested-ternary
```
fiskus, line 183: `no-nested-ternary` rule is still a thing. Esling shows warning on this

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -328,9 +202,155 @@ function Revision({ bucket, name, id, hash, stats, message, modified, counts })
...
         width={sparklineW}
       >
-        {AsyncResult.case(
+        {counts({ sparklineW, sparklineH })}
+      </M.Box>
+    </M.Paper>
+  )
+}
+
+function RevisionSkel() {
```
fiskus, line 211: The full name `RevisionSkeleton` would be more consistent

fiskus (approved):

akarve (approved):

catalog/app/containers/Bucket/PackageRevisions.js
nl0: yeah, i think it's ok to see those warnings and be aware of that

PR #1853, opened 2020-10-13
Fix UnseekableStreamError during upload

sir-sigurd:
## Description

When boto's `put_object()` retries it calls `seek(0)` of Body, which calls progress callback with a negative value. `tqdm.update()` supports negative values since 4.32.

## TODO

Use this section for work-in-progress pull requests. If you're using this section,
please tag your PR "WIP". 

- [x] Unit Tests
- [x] Add a [changelog](CHANGELOG.md) entry

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -168,20 +168,22 @@ class S3ClientProvider:
                 event_name, signal_transferring,
                 unique_id='datatransfer-transferring')
 
+    def _build_client(self, get_config):
+        session = self.get_boto_session()
+        return session.client('s3', config=get_config(session))
+
     def _build_standard_client(self):
-        boto_session = self.get_boto_session()
-
-        config = None
-        if boto_session.get_credentials() is None:
-            config = Config(signature_version=UNSIGNED)
-
-        s3_client = boto_session.client('s3', config=config)
+        s3_client = self._build_client(
+            lambda session:
+                Config(signature_version=UNSIGNED)
+                if session.get_credentials() is None
+                else None
+        )
         self.register_signals(s3_client)
         self._standard_client = s3_client
 
     def _build_unsigned_client(self):
-        boto_session = self.get_boto_session()
-        s3_client = boto_session.client('s3', config=Config(signature_version=UNSIGNED))
+        s3_client = self._build_client(lambda session: Config(signature_version=UNSIGNED))
         self.register_signals(s3_client)
         self._unsigned_client = s3_client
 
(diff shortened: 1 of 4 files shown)


sir-sigurd:
Might make sense to move the tests to a separate PR, if they look messy.

api/python/quilt3/data_transfer.py
```diff
@@ -168,20 +168,22 @@ def register_signals(self, s3_client):
...
+        session = self.get_boto_session()
+        return session.client('s3', config=get_config(session))
 
-        s3_client = boto_session.client('s3', config=config)
+    def _build_standard_client(self):
+        s3_client = self._build_client(
+            lambda session:
+                Config(signature_version=UNSIGNED)
+                if session.get_credentials() is None
+                else None
```
akarve, line 180: I was wondering if `session.client('s3', config=None)` would work (it does) and I think it's worth linking the docs in a comment: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client

api/python/tests/test_data_transfer.py
```diff
@@ -509,6 +513,58 @@ def side_effect(operation_name, *args, **kwargs):
                 data_transfer.copy_file_list([(src, dst, size)])
 
 
+class Success(Exception):
+    pass
+
+
+class S3UploadProgressTest(unittest.TestCase):
+    def setUp(self):
+        self.s3_client = boto3.client('s3', config=botocore.client.Config(signature_version=botocore.UNSIGNED))
```
akarve, line 522: Nice-to-have (not required) `config=None` case

docs/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 * [Performance] Improve performance of hashing of S3 files
   ([#1816](https://github.com/quiltdata/quilt/issues/1816),
    [#1788](https://github.com/quiltdata/quilt/issues/1788)
+* [Fixed] Bump a minimum required version of tqdm.
+  This fixes a crash (`UnseekableStreamError`) when retries during upload.
```
akarve, line 16:
```suggestion
  Fixes a tqdm crash (`UnseekableStreamError`) during upload retry.
```

akarve (approved):

api/python/quilt3/data_transfer.py
sir-sigurd: Don't understand, could you elaborate on that?

akarve: Yes. The reader knows that config=None can happen but without docs can't know if that call is legal or will crash boto

PR #1856, opened 2020-10-16
Package update dialog

nl0:
## TODO

- [x] Add a [changelog](CHANGELOG.md) entry
- [x] lots of things


nl0:
this is mostly working now. ready for initial review / testing

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+    alignItems: 'center',
+    background: t.palette.background.paper,
+    display: 'flex',
+    '&:not(:last-child)': {
+      borderBottomStyle: 'solid',
+      borderBottomWidth: '1px',
+      borderColor: 'inherit',
+    },
+    '&$added': {
+      background: M.colors.green[100],
```
fiskus, line 122: Maybe `t.palette.{success,warning,error}` are more appropriate

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+    : 'Drop files here or click to browse'
+
+  const pipeValue = useMemoEq([onInputChange, value], () => (...fns) =>
+    R.pipe(...fns, onInputChange)(value),
+  )
+
+  const onDrop = React.useCallback(
+    (files) => {
+      if (disabled) return
+      pipeValue(
```
fiskus, line 213: Please, provide some helpful hint on what is going on here (with named constant, for example). I should keep in mind useMemoEq, R.pipe, R.evolve, spread functions, it's too much when I don't know what the result should be.

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+
+      <div className={classes.dropzoneContainer}>
+        <div
+          {...getRootProps({
+            className: cx(
+              classes.dropzone,
+              isDragActive && !disabled && classes.active,
+              !!error && classes.dropzoneErr,
+              !error && warn && classes.dropzoneWarn,
+            ),
```
fiskus, line 326: I think it's more declarative:
```
{
 [classes.active]: isDragActive && !disabled,
 [classes.dropzoneErr]: !!error,
}
```

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+    onInputChange(meta.initial)
+    setUploads({})
+  }, [disabled, onInputChange, setUploads, meta.initial])
+
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({ onDrop })
+
+  const totalProgress = React.useMemo(() => getTotalProgress(uploads), [uploads])
+
+  const computedEntries = useMemoEq(value, ({ added, deleted, existing }) => {
+    const p1 = Object.entries(existing).map(([path, { size }]) => {
```
fiskus, line 263: Don't understand what does the `p` stand for. Path?

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+        const a = added[path]
+        return { type: 'modified', path, size: a.size, delta: a.size - size }
+      }
+      return { type: 'unchanged', path, size }
+    })
+    const p2 = Object.entries(added).reduce((acc, [path, { size }]) => {
+      if (path in existing) return acc
+      return acc.concat({ type: 'added', path, size })
+    }, [])
+    return R.sortBy((x) => [TYPE_ORDER.indexOf(x.type), x.path], p1.concat(p2))
```
fiskus, line 277: Shouldn't `sortBy` use a function that returns a number, not an array?

fiskus: `[1, 2] < [1, 2]` works, but I think an explicit number is more comprehensible and less error prone

fiskus (commented):
Partially read, will review more later

catalog/app/containers/Bucket/requests.js
```diff
@@ -1051,6 +1053,59 @@ export const getRevisionData = async ({
...
+      R.fromPairs,
+    )
+    return { meta, entries }
+  } catch (e) {
+    if (e instanceof errors.ManifestTooLarge) throw e
+    // eslint-disable-next-line no-console
+    console.log('loadManifest error:')
+    // eslint-disable-next-line no-console
+    console.error(e)
+    throw e
```
fiskus, line 1105: Seems like you can do `console.log` for `ManifestTooLarge` too, and remove condition

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+  const value = input.value || { added: {}, deleted: {}, existing: {} }
+  const disabled = meta.submitting || meta.submitSucceeded
+  const error = meta.submitFailed && meta.error
+
+  const onInputChange = input.onChange
+
+  const totalSize = React.useMemo(
+    () => Object.values(value.added).reduce((sum, f) => sum + f.size, 0),
+    [value.added],
+  )
```
fiskus, line 193: `inputValue` in `PackageCreateDialog` and `{ added, deleted, existing }` in `PackageUploadDialog` have different data format, based on `totalSize` function. Maybe things would be simpler if they have the same data format

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+    ? `Total size of new files exceeds recommended maximum of ${readableBytes(
+        PD.MAX_SIZE,
+      )}`
+    : 'Drop files here or click to browse'
+
+  const pipeValue = useMemoEq([onInputChange, value], () => (...fns) =>
+    R.pipe(...fns, onInputChange)(value),
+  )
+
+  const onDrop = React.useCallback(
```
fiskus, line 210: Seems like files droparea should be standalone component reusing — but not re-implementing — by PackageDialog and PackagUpdateDialog

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+  }, [disabled, onInputChange, setUploads, meta.initial])
+
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({ onDrop })
+
+  const totalProgress = React.useMemo(() => getTotalProgress(uploads), [uploads])
+
+  const computedEntries = useMemoEq(value, ({ added, deleted, existing }) => {
+    const p1 = Object.entries(existing).map(([path, { size }]) => {
+      if (path in deleted) {
+        return { type: 'deleted', path, size }
```
fiskus, line 265: This `type` "deleted" and `deleted` key in `action` object are the same. It's not a coincidence, they are the same on a purpose. In this case it's better to use `const` to avoid mistypings (and for better IDE support, and for future Typescript):

```js
const DELETED = 'deleted'
// ---
return { type: DELETED }
// ---
const action = { [DELETED]: { handler: rm } }
```

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+    }),
+    { total: 0, loaded: 0 },
+  ),
+  (p) => ({
+    ...p,
+    percent: p.total ? Math.floor((p.loaded / p.total) * 100) : undefined,
+  }),
+)
+
+function DialogForm({
```
fiskus, line 415: Seems like it should be named `PackageUpdateDialog`

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+
+export function usePackageUpdateDialog({ bucket, name, revision, onExited }) {
+  const s3 = AWS.S3.use()
+
+  const [isOpen, setOpen] = React.useState(false)
+  const [wasOpened, setWasOpened] = React.useState(false)
+  const [exited, setExited] = React.useState(!isOpen)
+  const [exitValue, setExitValue] = React.useState(null)
+  const [success, setSuccess] = React.useState(false)
+  const [key, setKey] = React.useState(1)
```
fiskus, line 846: It's not a AWS Key, it's not a React key, it's not an object key, right? It's something from `utils`, and it looks like not implemented yet?
Maybe use a more specific name?

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+    setWasOpened(false)
+    setKey(R.inc)
+  }, [setWasOpened, setKey])
+
+  const handleExited = React.useCallback(() => {
+    setExited(true)
+    setExitValue(null)
+    if (onExited) {
+      const shouldRefreshManifest = onExited(exitValue)
+      if (shouldRefreshManifest) refreshManifest()
```
fiskus, line 877: I didn't follow, how can I refresh manifest. What actions in a browser should I make?

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -428,21 +432,56 @@ export default function PackageRevisions({
...
+      if (res && res.pushed && res.pushed.name === name) {
+        setKey(R.inc)
+        return true
+      }
+      return false
+    },
+    [name, setKey],
+  )
+
+  const updateDialog = usePackageUpdateDialog({ bucket, name, onExited })
```
fiskus, line 459: Why do you use `component({ prop1, prop2 })` instead of regular `<Component prop1={prop1} prop2={prop2} />`? Because of memoizing?

fiskus (commented):
1. Dialog opens with name, files and **metadata** from the previous revision, but not workflow. Maybe we should select the previous workflow by default too.

2. There is a small bug. When I push a new revision, I see a "Browse package" button. I pushed it, but this dialog appeared again: 
dialog-twice

nl0:
> 1. Dialog opens with name, files and **metadata** from the previous revision, but not workflow. Maybe we should select the previous workflow by default too.
> 2. There is a small bug. When I push a new revision, I see a "Browse package" button. I pushed it, but this dialog appeared again:

@fiskus both these issues have been fixed

fiskus:
"Push complete" dialog. Slightly different issue:
1. Open package
2. Push new revision
3. See "Push complete" dialog, and push "Browse package"
4. It's OK, new package page
5. Click on "Push revision" and see "Push complete" dialog

nl0:
> "Push complete" dialog. Slightly different issue:
> 
> 1. Open package
> 2. Push new revision
> 3. See "Push complete" dialog, and push "Browse package"
> 4. It's OK, new package page
> 5. Click on "Push revision" and see "Push complete" dialog

@fiskus should be fixed now

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+
+const DialogState = tagged([
+  'Closed',
+  'Loading',
+  'Error',
+  'Form', // { manifest, workflowsConfig }
+  'Success', // { name, revision }
+])
+
+export function usePackageUpdateDialog({ bucket, name, revision, onExited }) {
```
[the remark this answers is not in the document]
nl0: `onClose` (request to close the dialog), `onExit` (exit transition started) and `onExited` (exit transition completed) are three different hooks exposed by the Dialog component, they all have different meaning. here i need `onExited` specifically to be able to run my callback after the dialog has exited completely (the transition has completed and it is hidden), so i can change dialog's content and it's not visible.

> there is no Dialog outside usePackageUpdateDialog

well, `usePackageUpdateDialog` has the word `Dialog` in it, so i guess it's ok if we use the same prop name, doesnt seem too leaky to me :)

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -24,42 +20,20 @@ import * as s3paths from 'utils/s3paths'
...
-    ...input,
-    ...rest,
-  }
-  return <M.TextField {...props} />
-}
-
 const useFilesInputStyles = M.makeStyles((t) => ({
   root: {
-    marginTop: t.spacing(2),
+    marginTop: t.spacing(3),
```
fiskus, line 36: Now all form sections have different margin :)

Header
32px
Name
24px
Commit message
32px
Files
24px
Metadata
16px
Workflow
40px
Buttons

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -428,21 +432,56 @@ export default function PackageRevisions({
...
+        px={{ xs: 2, sm: 0 }}
+        display="flex"
+      >
         <M.Typography variant="h5" ref={scrollRef}>
-          {name}
+          <StyledLink to={urls.bucketPackageDetail(bucket, name)}>{name}</StyledLink>{' '}
+          revisions
         </M.Typography>
+        <M.Box flexGrow={1} />
+        <M.Button
```
fiskus, line 476: It could be standalone component `ButtonPushRevisions`, reused by `PackageTree` page

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -24,42 +20,20 @@ import * as s3paths from 'utils/s3paths'
 import { readableBytes } from 'utils/string'
 import * as validators from 'utils/validators'
 
-import SelectWorkflow from './SelectWorkflow'
+import * as PD from './PackageDialog'
```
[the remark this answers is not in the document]
nl0: i think it's very inconvenient to use the long module names in jsx, e.g. `<PackageDialog.FormSkeleton />`, and also there's not too many abbreviations in this file, so i believe it's ok

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+import StyledLink from 'utils/StyledLink'
+import pipeThru from 'utils/pipeThru'
+import * as s3paths from 'utils/s3paths'
+import { readableBytes } from 'utils/string'
+import tagged from 'utils/tagged'
+import useMemoEq from 'utils/useMemoEq'
+import * as validators from 'utils/validators'
+
+import * as PD from './PackageDialog'
+import * as ERRORS from './errors'
```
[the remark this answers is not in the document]
nl0: > It's confusing that ERRORS in PackageUpdateDialog are similar to exceptions, and ERRORS in PackageCreateDialog are final-form errors.

well, i think we should use better names then :)

> I think we should rename all errors in './errors' to NotFoundError, ManifestTooLargeError, and import them explicitly by name. Outside of this PR, of course

i'm not a big fan of importing by name when it's more than one thing, namespace imports seem more convenient

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -499,19 +529,47 @@ export default function PackageTree({
...
+    [name, revision, setRevisionKey, setRevisionListKey],
+  )
+
   return (
     <FileView.Root>
-      {!!bucketCfg && <ExposeLinkedData {...{ bucketCfg, bucket, name, revision }} />}
+      {!!bucketCfg && (
+        <ExposeLinkedData
+          {...{ bucketCfg, bucket, name, revision }}
+          key={`links:${revisionKey}`}
```
fiskus, line 554: As I understand correctly, `revision` change is not enough to refresh this component, and you need to explicitly change React `key` for that

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,991 @@
...
+                </span>
+              )}
+              {!!stats.deleted.count && (
+                <span className={classes.headerFilesDeleted}>
+                  {' -'}
+                  {stats.deleted.count} ({readableBytes(stats.deleted.size)})
+                </span>
+              )}
+              {warn && <M.Icon style={{ marginLeft: 4 }}>error_outline</M.Icon>}
+            </>
```
fiskus, line 333: I would make it a component `<FilesStats addedCount={stats.added.count} deleted={states.deleted.count} />`

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: they are too deep, and here we need something faint for the background

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,991 @@
...
+          contents,
+          meta: PD.getMetaValue(meta),
+          workflow: PD.getWorkflowApiParam(workflow.slug),
+        },
+      })
+      onSuccess({ name, revision: res.timestamp })
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.log('error creating manifest', e)
+      // TODO: see if we should use e.message
```
fiskus, line 595: Backend can send Schema validation errors

Screenshot from 2020-11-23 09-03-03

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,991 @@
...
+        <M.Typography>Please use Quilt CLI to edit this package.</M.Typography>
+      </>
+    ),
+  ],
+  [
+    R.T,
+    () => (
+      <>
+        <M.Typography variant="h6" gutterBottom>
+          Unexpeceted error
```
fiskus, line 799: Typo: Unexpec**e**ted

catalog/app/containers/Bucket/errors.js
```diff
@@ -29,6 +29,16 @@ export class FileNotFound extends BucketError {}
...
 
+export class ManifestTooLarge extends BucketError {
+  constructor(props) {
+    super(
+      `Package manifest at s3://${props.bucket}/${props.key} is too large: ${props.actualSize} (max size: ${props.maxSize})`,
+      props,
+    )
+  }
+}
+ManifestTooLarge.displayName = 'ManifestTooLarge'
```
fiskus, line 40: We should use a different name for this, to stop confusing ESLint :)

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: i actually dont have ideas how to approach this. it seems quite digestible to me (ofc -- i wrote it), and i'm not sure how to clarify it. do you have any suggestions?

fiskus (approved):
Main thing: PackageUpdateDialog is missing Schema validation errors from the backend.

Other stuff: arguable maintainable code patterns

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: more like "part", it's just an intermediary var for the first part of the computation

catalog/app/containers/Bucket/PackageUpdateDialog.js
fiskus: At least, comment on what are input values and output results, please.
Better: self-documenting code with auxiliary constants.

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: well, we cant convert a string (`x.path`) to a number, and seems like array comparison works fine in js, e.g.
`[1, 'aaa'] < [1, 'aab']`

nl0: an alternative would be to use `sortWith`:
```javascript
R.sortWith([
  R.ascend(R.pipe(R.prop('type'), R.indexOf(R.__, TYPE_ORDER))),
  R.ascend(R.prop('path'))
])
```
but that seems more verbose and less readable

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,933 @@
...
+  filesContainerErr: {
+    borderColor: t.palette.error.main,
+  },
+  filesContainerWarn: {
+    borderColor: t.palette.warning.dark,
+  },
+  added: {},
+  modified: {},
+  deleted: {},
+  unchanged: {},
```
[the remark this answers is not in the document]
nl0: these are used for compound selectors below (`'&$added':` etc). `unchanged` is unused, i'm keeping it for consistency

catalog/app/containers/Bucket/requests.js
nl0: `ManifestTooLarge` is expected and we show the appropriate message in the UI, so i see no point in logging it to the console

fiskus: Actually, I think `console.log` shouldn't be here at all. It's a cosmetic thing and should be on the last step of handling error

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: well, these dialogs use different data and slightly different display for the files input, so i dont think we can make things simpler by somehow converting different data to the same format, it doesnt make sense (correct me if i dont get your comment right).
however, we may somehow unify the files input components and parameterize the resulting component to have different display for plain file upload (package creation) vs diff-like file list editing (package update), but this will actually make things more complex, not simpler (in Rich Hicky sense).
i'm open to further refactoring the dialogs to reuse more code in some near future.

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: yep, i thought about this, but it will require some parameterization to handle plain file list vs diff-like file list, and i dont think i want to make this refactoring now

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: yep, and those names are also used by style rules and in destructuring patterns, where it's not very convenient to use constants. i don't think it's the real problem and i dont like this approach (using constants for strings e.g. `const NAME = 'name'`) in general, it seems more like a crutch. btw i believe typescript would help with this, bc it has constant string types (not sure how to call it properly)

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: `PackageUpdateDialog` is the module name, and in the context of the module, this component represents the form ui for this dialog (vs error, success and progress), not the whole dialog

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: it's not from utils and it is implemented. this is used as a sort of cache-busting, a way to trigger component re-renders and data re-fetching.

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: manifest is refetched when you push a new revision while viewing the "latest" revision of the package

catalog/app/containers/Bucket/PackageRevisions.js
nl0: i dont quite get the question. i use a hook here, bc i need to be able to access some trigger functions i.e. `open` and `close`. it's not very convenient to expose triggers when using a component (you need to use refs and stuff), so using hook here makes more sense.

catalog/app/containers/Bucket/PackageCreateDialog.js
nl0: actually, i tried to unify the spacings by this change. the problem is almost all the inputs have their own margins, so we have to adjust accordingly

catalog/app/containers/Bucket/PackageTree.js
nl0: if revision is "latest", it wont change after the push, so we need to explicitly "cache-bust" things

catalog/app/containers/Bucket/PackageRevisions.js
nl0: well, not much to reuse here. also, PackageTree and PackageRevisions use differently size buttons

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: doesnt seem like that much logic to make it standalone

catalog/app/containers/Bucket/errors.js
nl0: what's wrong with it? i didnt see any errors / warnings from eslint

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: i'll take a look

catalog/app/containers/Bucket/errors.js
fiskus: ESLint throws a warning on this because ESLint is confusing this class with React component:
```
class ManifestTooLarge {
  static displayName = 'ManifestTooLarge'
}
```

catalog/app/containers/Bucket/PackageRevisions.js
fiskus: >  i need to be able to access some trigger functions i.e. open and close

I see now. It's an answer

catalog/app/containers/Bucket/PackageUpdateDialog.js
fiskus: Hmm, `shouldRefreshManifest` is always `undefined` or `false` for me. I couldn't find a case with `true`

catalog/app/containers/Bucket/PackageCreateDialog.js
fiskus: To be clear:
Margin between "Commit message" and "Files" became 32px ("Files" margin-top 24px + "Commit message" margin-bottom 8px)
Margin between "Files" and "Metadata" stayed at 24px

I think they should be equal 24px both (or 32px both)

catalog/app/containers/Bucket/PackageUpdateDialog.js
fiskus: As I see from code, `key` fed to `requests.loadManifest` through `useData`. But `loadManifest` don't expect `key` argument. That's why I said, that it looks like not implemented.

fiskus: `key` is used by `useMemoEq`, right?

catalog/app/containers/Bucket/PackageUpdateDialog.js
fiskus: Found that case, finally:

1. `PackageRevisions` page
2. Push new revision
3. **Close "Push complete" dialog**

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: > As I see from code, key fed to requests.loadManifest through useData. But loadManifest don't expect key argument. That's why I said, that it looks like not implemented.

it's not used by the requests function, but it is used by the `useData` hook itself which remembers the parameters dict and re-fetches when it's changed

> And also because there is TODO note about key in useData

there's no custom key extraction feature, so the whole parameters dict is used by default

> key is used by useMemoEq, right?

yep: https://github.com/quiltdata/quilt/blob/master/catalog/app/utils/Data.js#L33

catalog/app/containers/Bucket/errors.js
nl0: ok, noted, but this is used in our base error class, so changing it throughout the app is quite involved. btw i didnt see that warning, maybe because warnings dont abort commit, and i only run eslint as a commit hook)

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: input values and output results of `onDrop` or `pipeValue`?

fiskus: `onDrop` API is clear. It has typescript type hints

I will dig into `pipeValue` and `useMemoEq` later, don't bother at this moment :)

catalog/app/containers/Bucket/PackageCreateDialog.js
nl0: you're right, i'll fix that

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: `pipeValue` pipes the current `value` thru a list of functions (like `pipe`) and then sends that to `onInputChange`.
`useMemoEq` is similar to `React.useMemo`, but uses deep equality checks (`R.equals`) and has a slightly different api.

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: fixed

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -428,21 +432,56 @@ export default function PackageRevisions({
...
+          variant="contained"
+          color="primary"
+          style={{ marginTop: -3, marginBottom: -3 }}
```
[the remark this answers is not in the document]
nl0: why? we have a jss plugin that converts unitless values to px (except for line-height ofc, and other unitless rules)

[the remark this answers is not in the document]
fiskus: For clarity, to think less on code when reading

akarve (approved):
This looks quite good. A bit of wordsmithing inlined. I am also thinking that we should eliminate the + icon and only have the buttons "PUSH PACKAGE" and "REVISE PACKAGE" but that can happen in a separate change.

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -0,0 +1,990 @@
...
+        </div>
+        <M.Box flexGrow={1} />
+        {meta.dirty && (
+          <M.Button
+            onClick={revertFiles}
+            disabled={disabled}
+            size="small"
+            endIcon={<M.Icon fontSize="small">restore</M.Icon>}
+          >
+            Revert files
```
akarve, line 343:
```suggestion
            Undo changes
```

nl0:
> This looks quite good. A bit of wordsmithing inlined. I am also thinking that we should eliminate the + icon and only have the buttons "PUSH PACKAGE" and "REVISE PACKAGE" but that can happen in a separate change.

k done. should i also change the update dialog heading to "Revise package" and submit button label to "Revise"?

PR #1858, opened 2020-10-20
Remove signInRedirect and signOutRedirect

dimaryaz:
They haven't been used since the days of the django auth server.

Bytes are expensive; let's not waste them.


akarve (approved):

akarve:
Adding @nl0 as reviewer for coverage of JS codebase expertise

nl0:
well, actually, this was used on open to redirect users to their profile after signing in

PR #1859, opened 2020-10-21
attempt to fix search lambda

nl0:


lambdas/search/index.py
```diff
@@ -67,14 +67,14 @@ def lambda_handler(request):
     user_size = request.args.get('size', DEFAULT_SIZE)
     user_source = request.args.get('_source', [])
     # 0-indexed starting position (for pagination)
-    user_from = request.args.get('from', 0)
+    user_from = int(request.args.get('from', 0))
```
sir-sigurd, line 70: Does it come as `str`?

sir-sigurd (commented):
Tests are needed.

lambdas/search/index.py
nl0: i'm not sure, but since the invariant below is failing when `from=0`, i can deduce that it's not an `int`

nl0:
> Tests are needed.

well, they are present (maybe not very high quality, i agree)

lambdas/search/index.py
akarve: It does. For some reason, the way we do query parameter processing the values always come as strings.

akarve (approved):

sir-sigurd:
I think we should modify this somehow to ensure that it comes the same way as in prod:
https://github.com/quiltdata/quilt/blob/03695b07d0a779c9b072cfe9dc75cbd8608a6db3/lambdas/search/tests/test_search.py#L120
Something like `dict(urllib.parse.parse_qsl(urllib.parse.urlencode(query)))` could work.
We have similar `_make_event()` functions in tests of other lambdas, they should be modified as well.

PR #1863, opened 2020-10-22
Encase ` — ` with spaces

fiskus:
## Description

Added spaces around `&mdash`. It was stuck to the closest words.

Also linting.

Then:
1 then

Now:
2 now


nl0 (changes requested):
I believe English typography rules dont recommend spacing the em dash: https://en.wikipedia.org/wiki/Dash

fiskus:
You're right, sorry for bothering

PR #1867, opened 2020-10-26
JSON Editor and Schema validation

fiskus:
# Description

JSON editor implemented as a separate component. Root components are `components/JsonEditor/JsonEditor.js` and `components/JsonEditor/State.js`

Editor is a list of two-column tables. Every table is a nested level of JSON.

The hierarchy of components is the following:
 1. `JsonEditor` and `State` are root components
 2. `Column`, it's a two-column table actually. Represents one nested level of JSON
 3. `Row`, `AddRow`/`AddArrayItem`. Represents key/value pair or a control to add new rows
 4. `Cell`. Represents key or value wrapper
 5. `Input` and `Preview`. Edit and show key or value
 6. `Note` is like MUI InputAdornment for both `Input` and `Preview`

For a table implementation, I used `react-table`. It's almost useless in this context, to be honest, and I think we should get rid of it later.

## UI/UX feedback

- [x] 1. show a placeholder / skeleton for the json editor while the schema is loading
- [x] 2. menus on key cells dont open on required fields, should we hide them?
- [x] 3. num, str, etc. helper strings dont seem intuitive enough, i'd rather go for more user-friendly captions e.g. Number, String, etc
- [x] 4. the schema gets refetched and json editor flashes when switching btw key-value and text modes
- [x] 5. keys can't be edited after being entered -- that's a deliberate decision, but seems a little confusing, so maybe we can revisit it moving forward
- [x] 6. "click to edit" native browser tooltip overlaps our "value should be ..." tooltip
- [x] 7. backticks around paths in warnings dont seem necessary, maybe just use conventional code treatment or smth
- [x] 8. the form shouldn't be submitted when pressing enter while editing a key or a value
- [x] 9. something should be done with "schema id ignored" warnings in the console
- [x] 10. are we sure we dont need any placeholders? imo the ui looks a little empty and confusing without them, with just the empty white rectangles (placeholder text doesnt get anywhere as the actual data, it's just for display purposes) @akarve 
- [x] 11. value field cant be focused if the relevant key field is empty -- this is a deliberate decision, but seems a little confusing
- [x] 12. when adding an item to an array there's a react key conflict error in the console
- [x] 13. consider making breadcrumbs clickable, and maybe adding a root segment to go back to the top
- [x] 14. pressing escape while a field is focused should not close the dialog
- [x] 15. there's a strange placeholder / skeleton at the top-left of the packages screen
- [ ] 16. warn on duplicate keys

## Code review feedback

- [x] 1. why do you use `i18nMsgs` dicts everywhere? doesn't seem like this adds any value / developer convenience. consider removing them and using inline strings instead
- [x] 2. do not add empty lines everywhere between styles (classes), since brackets and indentation seem enough for making the structure clear and readable
- [x] 3. there's a possible bug with the `anchorRef`
- [x] 4. remove the flag and the old code
- [ ] 5. avoid using lodash when we can use ramda or vanilla js
- [x] 6. name constants in `UPPER_CASE`
- [x] 7. sort imports by module name alphabetically
- [x] 8. cache compiled schema for perf
- [x] 9. avoid (re)instantiating objects in function bodies, either move them out or memoize

plus inline comments


docs/Walkthrough/Working with the Catalog.md
```diff
@@ -26,6 +26,28 @@ Catalogs also enable you to search the contents of your bucket. We support both
 
 ![](../imgs/catalog_search.png)
 
+### Upload package
```
fiskus, line 29: @akarve 
Docs need your attention and quite probably rewrite.

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -937,3 +1074,34 @@ export default function UploadDialog({ bucket, open, onClose, refresh }) {
...
+        {...{
+          bucket,
+          open,
+          onClose,
+          refresh,
+          workflowsConfig,
+        }}
+      />
+    ),
+    Err: () => (
```
nl0, line 1094: i think you should handle and log the error in the request function and just return null in case of error, that will allow you to just handle the `Ok` case and avoid repeating `<UploadDialog {...{ ... }} />` stuff

catalog/app/containers/Bucket/requests.js
```diff
@@ -10,9 +10,57 @@ import * as Resource from 'utils/Resource'
...
+      workflows: [getNoWorkflow({})],
+    }
+  }
+
+  const { workflows } = data
+
+  const workflowsList = Object.keys(workflows).map((slug) =>
+    parseWorkflow(slug, workflows[slug], data),
+  )
+  // TODO: use const Symbol for None workflow name
```
nl0, line 55: this seems like a good idea

catalog/app/containers/Bucket/requests.js
```diff
@@ -59,6 +107,8 @@ export const bucketListing = ({ s3, bucket, path = '', prefix }) =>
 
 const MAX_BANDS = 10
 
+/* eslint-disable no-console */
```
nl0, line 110: i don't think this is a good idea, bc console calls that were accidentally left will be ignored

catalog/app/containers/Bucket/requests.js
```diff
@@ -212,6 +260,56 @@ export const bucketStats = async ({ req, s3, bucket, overviewUrl }) => {
...
+  } catch (e) {
+    console.log('Unable to fetch')
+    console.error(e)
+  }
+
+  throw new Error('Schema is unavailable')
+}
+
+export const workflowsList = async ({ s3, bucket }) => {
+  const configPath = '.quilt/workflows/config.yml'
```
nl0, line 283: let's move this constant out of the function body

catalog/app/containers/Bucket/requests.js
```diff
@@ -212,6 +260,56 @@ export const bucketStats = async ({ req, s3, bucket, overviewUrl }) => {
...
+    return await s3
+      .getObject({
+        Bucket: bucket,
+        Key: configPath,
+      })
+      .promise()
+      .then((r) => parseWorkflows(r.Body.toString('utf-8')))
+  } catch (e) {
+    console.log('Unable to fetch')
+    console.error(e)
```
nl0, line 307: maybe we should just return null here instead of throwing an exception?

catalog/app/utils/json-schema.js
```diff
@@ -0,0 +1,84 @@
+import * as R from 'ramda'
+import isArray from 'lodash/isArray'
+import isNumber from 'lodash/isNumber'
+import isObject from 'lodash/isObject'
+import isString from 'lodash/isString'
+import isUndefined from 'lodash/isUndefined'
+
+function isSchemaArray(optSchema) {
+  return R.prop('type', optSchema) === 'array'
```
nl0, line 9: we can make this significantly less verbose: `const isSchemaArray = R.propEq('type', 'array')`

catalog/app/utils/json-schema.js
```diff
@@ -0,0 +1,84 @@
...
+function isSchemaBoolean(optSchema) {
+  return R.prop('type', optSchema) === 'boolean'
+}
+
+function isSchemaNull(optSchema) {
+  return R.prop('type', optSchema) === 'null'
+}
+
+export function isSchemaEnum(optSchema) {
+  return Boolean(R.prop('enum', optSchema))
```
nl0, line 33: `R.propSatisfies('enum', Boolean)`

catalog/app/utils/json-schema.js
```diff
@@ -0,0 +1,84 @@
...
+
+function isSchemaBoolean(optSchema) {
+  return R.prop('type', optSchema) === 'boolean'
+}
+
+function isSchemaNull(optSchema) {
+  return R.prop('type', optSchema) === 'null'
+}
+
+export function isSchemaEnum(optSchema) {
```
nl0, line 32: it's a little strange that only some of these test functions are exported

catalog/app/utils/json-schema.js
```diff
@@ -0,0 +1,84 @@
...
+  if (!optSchema) return false
+  return ['anyOf', 'oneOf', 'not', 'allOf'].some((key) => optSchema[key])
+}
+
+function isSchemaReference(optSchema) {
+  return Boolean(R.prop('$ref', optSchema))
+}
+
+export function isNestedType(optSchema) {
+  return isSchemaArray(optSchema) || isSchemaObject(optSchema)
```
nl0, line 50: `const isNestedType = R.either(isSchemaArray, isSchemaObject)`

catalog/app/utils/json-schema.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export function schemaTypetoHumanString(optSchema) {
```
nl0, line 53: should be "schemaType**To**HumanString" i believe.
also, this can be written in point-free style: `const schemaTypeToHumanString = R.cond([ ... ])`

catalog/app/utils/json-schema.js
```diff
@@ -0,0 +1,84 @@
...
+    // NOTE: enum and const can be string too,
+    //       that's why they are first
+    [R.prop('type'), () => R.take(3, optSchema.type)],
+    [isSchemaCompound, () => 'comp'],
+    [isSchemaReference, () => '$ref'],
+    [R.T, () => 'none'],
+  ])(optSchema)
+}
+
+export function doesTypeMatchToSchema(value, optSchema) {
```
nl0, line 68: `to` is unnecessary here, i believe => `doesTypeMatchSchema`

catalog/app/utils/yaml.js
```diff
@@ -0,0 +1,10 @@
+import yaml from 'js-yaml'
+
+export default function parseYaml(inputStr) {
+  try {
+    return yaml.safeLoad(inputStr, 'utf8')
+  } catch (error) {
+    console.error(error)
+    return ''
```
nl0, line 8: do we really need an empty string here? smth like `null` or even `undefined` seems more appropriate imo

nl0 (changes requested):
looks good generally, but there are a few issues, bike-shedding mostly:

1. why do you use `i18nMsgs` dicts everywhere? doesn't seem like this adds any value / developer convenience;

2. i think we should not add empty lines everywhere btw styles (classes), since brackets and indentation seem enough for making the structure clear and readable;

3. there's a possible bug with the `anchorRef`

4. remove the flag and the old code

5. warn on duplicate keys

6. avoid using lodash when we can use ramda or vanilla js

7. name constants in `UPPER_CASE`

8. sort imports by module name alphabetically

9. cache compiled schema for perf

10. avoid (re)instantiating objects in function bodies, either move them out or memoize

i'll add this list to the description

catalog/app/components/JsonEditor/AddArrayItem.js
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+
```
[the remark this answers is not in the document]
fiskus: I don't know, I tried to mimic your style here. Then all external dependencies should be in one group? Then all internal dependencies with aliases, then all relative dependencies, right? Three groups divided by an empty line?

catalog/app/components/JsonEditor/AddRow.js
```diff
@@ -0,0 +1,67 @@
...
+    border: `1px solid ${t.palette.grey[400]}`,
+    width: t.spacing(36),
+  },
+}))
+
+export default function AddRow({ columnPath, onAdd, onExpand }) {
+  const classes = useStyles()
+
+  const onChange = React.useCallback(
+    (_, __, value) => {
```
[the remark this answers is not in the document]
fiskus: > doesnt prettier or eslint barf on underscore var names
No, underscore prefixes or suffixes are forbidden, not entire variable name consisted of underscore :)
`_1`, `_2` are much better, thanks

catalog/app/components/JsonEditor/Column.js
```diff
@@ -0,0 +1,161 @@
+import * as React from 'react'
+import isArray from 'lodash/isArray'
+import { useTable } from 'react-table'
```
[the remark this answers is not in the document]
fiskus: Actually, I sort with `:sort`. But it sorts by imported name

catalog/app/components/JsonEditor/Column.js
```diff
@@ -0,0 +1,161 @@
+import * as React from 'react'
+import isArray from 'lodash/isArray'
```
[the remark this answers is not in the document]
fiskus: I used these a lot. What you think if I will refactor it later and use `R.is` insead of lodash's `isArray`, `isObject` and so on?

catalog/app/components/JsonEditor/AddArrayItem.js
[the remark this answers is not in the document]
nl0: yep, smth like this:
1. all "node builtins" like `path`
2. npm-installed deps (first unscoped, then scoped)
3. local
4. relative

maybe assets should be somewhere at the bottom as well, use your judgement for any special / edge cases

catalog/app/components/JsonEditor/AddRow.js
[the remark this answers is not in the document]
nl0: > underscore prefixes or suffixes are forbidden

in this case we may need to tweak eslint config to allow `_N`-style names

catalog/app/components/JsonEditor/Column.js
```diff
@@ -0,0 +1,161 @@
...
+    (...params) => {
+      setHasNewRow(true)
+      onAddRow(...params)
+    },
+    [onAddRow],
+  )
+
+  return (
+    <div className={classes.root}>
+      {Boolean(columnPath.length) && (
```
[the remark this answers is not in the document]
fiskus: I prefer `Boolean(x)` instead of `!!x` for the reason it reads as more intentional. Comparing "I want not not some value" and "I want boolean representation of some value".

catalog/app/utils/json-schema.js
fiskus: I export functions that are used outside. If I export every function here it would be harder to refactor and maintain, because I should find if this function is used somewhere else.

catalog/app/components/JsonEditor/State.js
```diff
@@ -0,0 +1,236 @@
...
+import isUndefined from 'lodash/isUndefined'
+import toNumber from 'lodash/toNumber'
+
+export const ColumnIds = {
+  Key: 'key',
+  Value: 'value',
+}
+
+export const Actions = {
+  ChangeType: 'change_type',
```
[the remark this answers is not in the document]
fiskus: This value used for React `key`. `Symbol` has `toString` method, but throws an exception on `${symbol}`, and I decided to use simple strings

catalog/app/components/JsonEditor/State.js
```diff
@@ -0,0 +1,236 @@
+import * as R from 'ramda'
+import * as React from 'react'
+import Ajv from 'ajv'
+import isArray from 'lodash/isArray'
+import isUndefined from 'lodash/isUndefined'
+import toNumber from 'lodash/toNumber'
+
+export const ColumnIds = {
```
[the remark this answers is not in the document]
fiskus: I think it's a good practice to store such things in constants. Text editor's autocompletion helps to avoid mistypings.

I used CamelCase because this contstat meaning is enum, actually. And enums in typescript are CamelCased.

catalog/app/utils/json-schema.js
fiskus: I've just tried this. It throws error when `optSchema` is undefined. Same as `R.has` and `R.propEq`:

`R.prop('enum', undefined)` --> `undefined`
`R.has('enum', undefined)` --> throws error
`R.propEq('type', 'array', undefined)` --> throws error
`R.propSatisfies('enum', Boolean, undefined)` --> throws error

catalog/app/components/JsonEditor/AddArrayItem.js
```diff
@@ -0,0 +1,38 @@
...
+import * as M from '@material-ui/core'
+
+const i18nMsgs = {
+  button: 'Add list item',
```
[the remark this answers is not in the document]
fiskus: The reason is to split content from its meaning. It's the main point of i18n before localization.
"Add list item" - it's a content. `i18nMsgs.button` - is a meaning of this content. If you change content to "Add array item" it still has the same meaning: "i18n message for button".

It's like config value for component UI.

catalog/app/components/JsonEditor/AddRow.js
```diff
@@ -0,0 +1,67 @@
...
+import Cell from './Cell'
+import { Actions, ColumnIds, EmptyValue } from './State'
+
+const useStyles = M.makeStyles((t) => ({
+  inputCell: {
+    border: `1px solid ${t.palette.grey[400]}`,
+    padding: 0,
+    width: t.spacing(20),
+  },
+
```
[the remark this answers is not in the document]
fiskus: Ok, it's a habit. I used to this style https://github.com/airbnb/css#formatting

catalog/app/components/JsonEditor/Breadcrumbs.js
```diff
@@ -0,0 +1,59 @@
...
+    marginRight: t.spacing(2),
+  },
+}))
+
+export default function Breadcrumbs({ items, onBack }) {
+  const classes = useStyles()
+
+  const ref = React.useRef()
+  React.useEffect(() => {
+    ref.current.scrollIntoView()
```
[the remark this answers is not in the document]
fiskus: Added dependency to call it once

catalog/app/components/JsonEditor/Cell.js
```diff
@@ -0,0 +1,214 @@
...
+import Preview from './Preview'
+import { Actions, ColumnIds, EmptyValue } from './State'
+
+function CellMenu({ anchorRef, menu, onClose, onClick }) {
+  if (!menu.length) return null
+
+  return (
+    <M.Menu anchorEl={anchorRef.current} onClose={onClose} open>
+      {menu.map((subList) =>
+        subList.key === 'divider' ? (
```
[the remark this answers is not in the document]
fiskus: I don't need it anymore, removed

catalog/app/components/JsonEditor/Column.js
```diff
@@ -0,0 +1,161 @@
...
+import AddArrayItem from './AddArrayItem'
+import AddRow from './AddRow'
+import Breadcrumbs from './Breadcrumbs'
+import Cell from './Cell'
+import Row from './Row'
+import { ColumnIds, EmptyValue } from './State'
+
+const useStyles = M.makeStyles((t) => ({
+  root: {
+    background: '#fff',
```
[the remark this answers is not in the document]
fiskus: Replaced to `t.palette.common.white`, but maybe `t.palette.background.paper` is better

catalog/app/components/JsonEditor/Column.js
```diff
@@ -0,0 +1,161 @@
...
+}
+
+export default function Table({
+  columnPath,
+  data,
+  onAddRow,
+  onCollapse,
+  onExpand,
+  onMenuAction,
+  updateMyData,
```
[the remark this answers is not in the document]
fiskus: It's a function wrapper for `onChange` subscription. Just adds some `setState` calls inside the component. I use `onChangeInternal` for that case usually.

`updateMyData` is the name for `react-table` API method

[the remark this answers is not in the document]
fiskus: I renamed prop to `onChange`, internal helper function renamed to `onChangeInternal`. Method `updateMyData` now used for `react-table` only

catalog/app/components/JsonEditor/Column.js
```diff
@@ -0,0 +1,161 @@
...
+      setHasNewRow(false)
+      updateMyData(...params)
+    },
+    [updateMyData],
+  )
+
+  const tableInstance = useTable({
+    columns,
+    data: data.items,
+
```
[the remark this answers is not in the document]
fiskus: They split data properties, UI properties and event subscription property. I removed empty lines

catalog/app/components/JsonEditor/Input.js
```diff
@@ -0,0 +1,127 @@
...
+        // no default
+      }
+    },
+    [onBlur],
+  )
+
+  return (
+    <M.InputBase
+      autoFocus
+      inputRef={inputRef}
```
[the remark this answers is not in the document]
fiskus: Used for menu, not anymore

catalog/app/components/JsonEditor/Preview.js
```diff
@@ -0,0 +1,115 @@
...
+  menu: {
+    marginLeft: 'auto',
+  },
+}))
+
+function formatValuePreview(v) {
+  if (v === EmptyValue || isUndefined(v)) return ''
+
+  if (isArray(v)) {
+    return `[ ${v.map((item) => formatValuePreview(item))} ]`
```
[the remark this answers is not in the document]
fiskus: > can we use the point-free notation here? e.g. v.map(formatValuePreview)

`formatValuePreview` had two arguments earlier, and I got into the famous js wtf like `['1','2','3'].map(parseInt)`

catalog/app/containers/Bucket/requests.js
fiskus: Please take a look https://github.com/quiltdata/quilt/pull/1882

catalog/app/components/JsonEditor/AddRow.js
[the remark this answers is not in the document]
fiskus: "allowFunctionParams" param for "no-underscore-dangle" is true by default. `_1`, `_2` are legit

catalog/app/components/JsonEditor/State.js
```diff
@@ -0,0 +1,236 @@
...
+      setErrors(validateOnSchema(newData, schema))
+      return newData
+    },
+    [data, schema],
+  )
+
+  const makeAction = React.useCallback(
+    (contextFieldPath, columnId, actionItem) => {
+      switch (actionItem.action) {
+        case Actions.RemoveField:
```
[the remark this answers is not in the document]
fiskus: I don't know, maybe. Elaborate, please.

catalog/app/utils/json-schema.js
fiskus: This is some inconsistency in Ramda. `R.prop` is safe for any value, `R.propEq` throws error when there is no object.
I need `undefined` here, and highlighted this with `opt` prefix

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -867,12 +963,53 @@ export default function UploadDialog({ bucket, open, onClose, refresh }) {
                     isEqual={R.equals}
                   />
 
-                  <RF.Field
-                    component={MetaInput}
-                    name="meta"
-                    validate={validateMeta}
-                    isEqual={R.equals}
-                  />
+                  <Data
```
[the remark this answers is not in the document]
fiskus: How do you see final result? Like that:

```jsx
SchemaFetcher = ({ children, schemaId }) => (
  <Data fetch params>
    {AsyncResult.case({
      Ok: (schema) => children(schema, validateMeta(schema)),
      _: () => <Skeleton />
    })}
  </Data>
)

// ---

<SchemaFetcher schemaId>
  {(schema, validate) => <RF.Field schema={schema} validate={validate} />}
</SchemaFetcher>
```

catalog/app/components/JsonEditor/State.js
```diff
@@ -0,0 +1,236 @@
...
+    return str
+  }
+}
+
+const initialSortCounter = 10000000
+
+// TODO: rename and move to utils/json-schema
+//       name hint: schema path, associated with original object
+function getSchemaPath(objPath) {
+  return objPath.reduce((memo, key) => memo.concat(['properties'], key), [])
```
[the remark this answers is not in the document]
fiskus: They are not equivalent:
```
const ramdaPath = R.chain(R.prepend('properties'))
const schemaPath = (objPath) => objPath.reduce((memo, key) => memo.concat(['properties'], key), [])
ramdaPath(['optList']) // [ 'properties', 'o', 'p', 't', 'L', 'i', 's', 't' ]
schemaPath(['optList']) // [ 'properties', 'optList' ]
```

catalog/app/components/JsonEditor/Column.js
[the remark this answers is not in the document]
fiskus: Please, elaborate on why you don't like lodash in this case, because there are a several reasons that I can imagine. And they affect code differently.

Would you accept this:
Module `utils/types.js` with a number of functions `isNull`, `isArray` and so on. Internally, every function will be has no external dependency (like `isNull(x) => x === null`) or maybe `Ramda` dependency

catalog/app/components/JsonEditor/State.js
```diff
@@ -82,14 +79,16 @@ function getColumn(obj, columnPath, sortOrder, schema) {
     R.pathOr({}, schemaPath.concat('properties'), schema),
   )
 
+  const sortFunc = R.sort((a, b) => a.sortIndex - b.sortIndex)
```
nl0, line 82: fyi this can be expressed as `R.sortBy(R.prop('sortIndex'))` or `R.sort(R.ascend(R.prop('sortIndex')))`

catalog/app/components/JsonEditor/State.js
```diff
@@ -63,13 +60,13 @@ function mapKeys(objectOrArray, callback, schemaKeys) {
     (memo, key) => (memo.includes(key) ? memo : memo.concat(key)),
     Object.keys(objectOrArray),
   )
-  return keys
-    .sort((a, b) => getSortIndex(a) - getSortIndex(b))
-    .map((key) => callback(objectOrArray[key], key, schemaSort[key]))
+  const sortFunc = R.sort((a, b) => getSortIndex(a) - getSortIndex(b))
+  const mapFunc = R.map((key) => callback(objectOrArray[key], key, schemaSort[key]))
+  return R.compose(sortFunc, mapFunc)(keys)
```
nl0, line 65: fyi there's a convenience helper in `utils/pipeThru` that helps to express such cases in a slightly more readable and intuitive way (imo), like this:
```javascript
return pipeThru(keys)(
  R.map((key) => callback(objectOrArray[key], key, schemaSort[key])),
  R.sortBy(getSortIndex),
) 
```

ofc pipeline operator would make this even better

catalog/app/components/JsonEditor/State.js
```diff
@@ -120,15 +117,20 @@ function convertType(value, typeOf) {
   }
 }
 
+const EmptySchema = {}
```
nl0, line 120: let's use lower camelCase or uppercase for this, so it looks like a regular var or a constant, and reserve upper CamelCase for classes, components and types

nl0 (changes requested):

catalog/app/components/JsonEditor/Column.js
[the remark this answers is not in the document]
nl0: yeah, i think it would be better if module names came first, e.g. `from "module" import stuff`, but whatever, it's js

catalog/app/components/JsonEditor/Column.js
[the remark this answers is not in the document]
nl0: > Please, elaborate on why you don't like lodash in this case

well, the main reason is we already use ramda and it has most of this functionality, so i see no point in importing extra code that does the same.

as for the utils module, idk but i think it's not necessary, bc `isNull` is such a trivial thing that i dont think we should abstract it away, just simple `(x) => x === null` or `R.equals(null)` seems good enough. other cases like `isObject` and `isString`, for example, are trivial as well with ramda, it's just `R.is(Object)`, `R.is(String)` and so on -- this, again, seems like it's good enough.

`isArray` is a special case, because we have a built-in function for this -- `Array.isArray`, so we don't need to use any abstractions or introduce any dependencies for this.

however, i think using `lodash` is legit when we want some special logic that's unique to it (missing in ramda)

catalog/app/components/JsonEditor/Column.js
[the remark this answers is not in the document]
nl0: ok, but if you think the reason is good enough, you can keep them

catalog/app/components/JsonEditor/Column.js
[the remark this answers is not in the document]
nl0: it makes perfect sense, but i think `!!` is a widespread convention, almost an idiom, so i'm quite used to it

catalog/app/components/JsonEditor/Preview.js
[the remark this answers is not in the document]
nl0: yep, i thought about that. `Ramda#map` has no such problem

catalog/app/components/JsonEditor/State.js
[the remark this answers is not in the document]
nl0: ok, i see the point

catalog/app/components/JsonEditor/State.js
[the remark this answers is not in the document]
nl0: oh man... sorry, i've tested it on single-letter arrays (e.g. `['a', 'b', 'c']`) and it worked

catalog/app/containers/Bucket/requests.js
nl0: i did, and we'll get to it soon-ish, but it seems outside of this pr's context

catalog/app/utils/json-schema.js
nl0: i see

catalog/app/containers/Bucket/UploadDialog.js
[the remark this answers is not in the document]
nl0: something like this:
```javascript
function SchemaFetcher({ children, bucket, schemaUrl }) {
  const s3 = AWS.S3.use()
  const data = useData(requests.metadataSchema, { s3, bucket, schemaUrl })
  const res = React.useMemo(
    () => AsyncResult.mapCase(
      {
        Ok: (schema) => ({ schema, validate: validateMeta(schema) }),
      },
      data.result,
    ),
    [data.result],
  )
  return children(res)
}

// usage
<SchemaFetcher {...{ bucket, schemaUrl }}>
  {AsyncResult.case({
    Ok: ({ schema, validate }) => <RF.Field schema={schema} validate={validate} />,
    _: () => <Skeleton />,
  })}
</SchemaFetcher>
```

catalog/app/components/JsonEditor/State.js
```diff
@@ -0,0 +1,238 @@
...
+      return newData
+    },
+    [data, schemaValidator],
+  )
+
+  const makeAction = React.useCallback(
+    (contextFieldPath, actionItem) => {
+      switch (actionItem.action) {
+        case ACTIONS.REMOVE_FIELD:
+          return removeField(contextFieldPath, actionItem)
```
nl0, line 174: `actionItem` seems unused by the `removeField` fn

catalog/app/components/JsonEditor/Cell.js
```diff
@@ -110,19 +103,24 @@ export default function Cell({
...
   value: initialValue,
 }) {
-  const classes = useClasses()
+  const classes = useStyles()
 
   const menuAnchorRef = React.useRef(null)
   const [value, setValue] = React.useState(initialValue)
 
+  const [menuAnchorEl, setMenuAnchorEl] = React.useState(null)
+  React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
fiskus: I moved Menu to the same hierarchy level where the actual ref is

catalog/app/components/JsonEditor/Column.js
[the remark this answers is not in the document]
fiskus: I created a backlog task for that in the Code Quality section

catalog/app/components/JsonEditor/Input.js
```diff
@@ -0,0 +1,127 @@
+import * as React from 'react'
+import * as R from 'ramda'
+import isObject from 'lodash/isObject'
+import isString from 'lodash/isString'
```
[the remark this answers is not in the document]
fiskus: Task created

catalog/app/components/JsonEditor/State.js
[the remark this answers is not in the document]
fiskus: We decided in Slack to do it later (maybe)

catalog/app/components/JsonEditor/ButtonMenu.js
```diff
@@ -17,13 +17,61 @@ const useStyles = M.makeStyles((t) => ({
...
+  note,
+  onClick,
+  onMenuClose,
+  onMenuSelect,
+}) {
   const classes = useStyles()
 
+  const ref = React.useRef(null)
+
+  const hasMenu = React.useMemo(() => !!menu.length, [menu])
```
nl0, line 73: not very important, but anyways, `useMemo` isnt required here, because the computation (`!!`) is simple and result is a primitive value

catalog/app/components/JsonEditor/Input.js
```diff
@@ -77,6 +93,19 @@ export default function Input({
   const [valueStr, setValueStr] = React.useState(() =>
     getNormalizedValueStr(originalValue, data.valueSchema),
   )
+  const [firstTime, setFirstTime] = React.useState(true)
```
nl0, line 96: i believe `useState` is not the appropriate hook for this, we should rather use `useRef`

nl0: or you could even try not using this variable at all, just call `useLayoutEffect` with an empty dependency array -- this way it will only be called on mount (similar to `componentDidMount`).
also, why are you using `useLayoutEffect` instead of regular `useEffect` here?

catalog/app/components/JsonEditor/Input.js
```diff
@@ -87,18 +116,13 @@ export default function Input({
   )
 
   const onBlur = React.useCallback(() => {
-    if (columnId !== COLUMN_IDS.KEY || isString(value)) {
+    if (columnId !== COLUMN_IDS.KEY || typeof value === 'string') {
```
nl0, line 119: `typeof value === 'string'` is unsafe -- it fails if the value is, for example, `new String('')`. `R.is(String, value)` is safer. When I told that we should use inline checks instead of `lodash.isX`, I meant the equality checks (`=== null`, `=== undefined), not type checks, bc `typeof` frequently behaves counter-intuitively.

catalog/app/components/JsonEditor/Preview.js
```diff
@@ -44,37 +44,82 @@ const useStyles = M.makeStyles((t) => ({
...
+        {Object.keys(value).join(', ')}
+        <span className={classes.rbrace}>&#125;</span>
+      </span>
+    )
   }
 
-  return v.toString()
+  if (value === null) return 'null'
+
+  if (typeof value === 'string') return <span>&quot;{value}&quot;</span>
```
nl0, line 103: `R.is(String, value)` is safer

catalog/app/components/JsonEditor/State.js
```diff
@@ -48,50 +45,42 @@ export function validateOnSchema(optSchema) {
...
+  required,
+  valueSchema: item,
+  sortIndex,
+  type: getSchemaType(item),
+})
 
-function assocObjValue(objPath, value, obj) {
-  return R.assocPath(objPath, value, obj)
-}
+const assocObjValue = R.assocPath
```
nl0, line 76: do we really need this alias?

catalog/app/components/JsonEditor/State.js
```diff
@@ -48,50 +45,42 @@ export function validateOnSchema(optSchema) {
...
-export function getJsonDictValue(objPath, jsonDict) {
-  return R.prop(serializeAddress(objPath), jsonDict)
-}
+export const getJsonDictValue = (objPath, jsonDict) =>
+  R.prop(serializeAddress(objPath), jsonDict)
 
-export function getObjValue(objPath, obj) {
-  return R.path(objPath, obj)
-}
+export const getObjValue = R.path
```
nl0, line 81: do we really need this alias?

catalog/app/components/JsonEditor/State.js
```diff
@@ -118,13 +109,10 @@ function moveObjValue(oldObjPath, key, obj) {
...
-function dissocSchemaValue(objPath, jsonDict) {
-  return R.dissoc(serializeAddress(objPath), jsonDict)
-}
+const dissocSchemaValue = (objPath, jsonDict) =>
+  R.dissoc(serializeAddress(objPath), jsonDict)
 
-function dissocObjValue(objPath, obj) {
-  return R.dissocPath(objPath, obj)
-}
+const dissocObjValue = R.dissocPath
```
nl0, line 115: do we really need this alias?

catalog/app/components/JsonEditor/State.js
```diff
@@ -240,30 +228,91 @@ function iterateJsonDict(jsonDict, obj, fieldPath, rootKeys) {
...
+
+  // NOTE: fieldPath is like URL for editor columns
+  //       `['a', 0, 'b']` means we are focused to `{ a: [ { b: %HERE% }, ... ], ... }`
   const [fieldPath, setFieldPath] = React.useState([])
+
+  // NOTE: list of JSON Schema validation errors
   const [errors, setErrors] = React.useState([])
 
+  // NOTE: incremented sortIndex counter,
+  //       it needs to place new fields lower old ones
```
nl0, line 300:
```suggestion
  //       it's required to place new fields below existing ones
```

catalog/app/components/JsonEditor/State.js
```diff
@@ -240,30 +228,91 @@ function iterateJsonDict(jsonDict, obj, fieldPath, rootKeys) {
...
   )
 
-  // TODO: use Set
+  // NOTE: list of root object keys + root schema keys
   const [rootKeys, setRootKeys] = React.useState(() =>
-    mergeSchmaAndObjRootKeys(optSchema, obj),
+    mergeSchemaAndObjRootKeys(schema, obj),
   )
 
+  // NOTE: this data represents table columns showen to user
```
nl0, line 314:
```suggestion
  // NOTE: this data represents table columns shown to user
```

catalog/app/components/JsonEditor/State.js
```diff
@@ -240,30 +228,91 @@ function iterateJsonDict(jsonDict, obj, fieldPath, rootKeys) {
...
+  //       it's a main source of UI data
```
nl0, line 315:
```suggestion
  //       it's the main source of UI data
```

catalog/app/utils/json-schema.js
```diff
@@ -1,82 +1,96 @@
...
 
-function isSchemaReference(optSchema) {
-  return !!R.prop('$ref', optSchema)
-}
+const isSchemaReference = (optSchema) => !!R.prop('$ref', optSchema)
 
 export const isNestedType = R.either(isSchemaArray, isSchemaObject)
 
+const compoundTypeToHumanString = (optSchema, condition, divider) =>
+  R.prop(condition, optSchema)
```
nl0, line 43: i think this can throw an error, bc `R.prop` can return undefined if `optSchema` is undefined (or is it always defined?). also, you could just use the vanilla js `optSchema[condition]` here, bc `R.prop` benefits (currying and not throwing on undefined) dont seem to be required here.

nl0 (approved):
the new changes look mostly good, except for some minor issues

catalog/app/components/JsonEditor/Input.js
fiskus: > why are you using `useLayoutEffect`

I want to be sure that all DOM jobs are completed. I can add a TODO comment here, to replace it with a regular `useEffect`, but I think I should invest more time in testing it

catalog/app/components/JsonEditor/State.js
fiskus: Yes, to have consistently similar functions for an original object and a schema. It's a coincidence, that function's implementation is an alias to `R.assocPath`

catalog/app/utils/json-schema.js
fiskus: It's always defined, but I should expect undefined too. WIl fix it

catalog/app/components/JsonEditor/Input.js
fiskus: Sure, `useEffect(func, [])` - is a right choice, thanks

catalog/app/components/JsonEditor/Input.js
fiskus: > typeof value === 'string' is unsafe -- it fails if the value is, for example, new String('').

Good question for a job interview, right after "What's the result of `[] + {}` and `{} + []`" :)

---

Fixed, thanks

catalog/app/components/JsonEditor/JsonEditor.js
```diff
@@ -0,0 +1,101 @@
...
+
+import Column from './Column'
+import Errors from './Errors'
+import State from './State'
+
+const useStyles = M.makeStyles((t) => ({
+  inner: {
+    display: 'flex',
+    overflowX: 'auto',
+    '-ms-overflow-style': 'none',
```
nl0, line 13: a note for the future: in JSS this can be written as `MsOverflowStyle`. the general rule is it inserts a dash before and lowers every uppercase letter.

catalog/app/components/JsonEditor/JsonEditor.js
```diff
@@ -0,0 +1,101 @@
...
+    'scrollbar-width': 'none',
```
nl0, line 14: same here, `scrollbarWidth` would do the same, but w/o quotes

nl0 (approved):
i think this is good to go. @akarve pls wordsmith the docs (in a new pr probably)

PR #1868, opened 2020-10-26
Zip download btns

nl0:
# Description

Add "download as .zip" buttons to directory and package views

# TODO

- [x] better layout on the package page

### DIFF
--- a/catalog/app/components/BreadCrumbs/BreadCrumbs.js
+++ b/catalog/app/components/BreadCrumbs/BreadCrumbs.js
@@ -29,6 +29,7 @@ export function copyWithoutSpaces(e) {
       .getSelection()
       .toString()
       .replace('<EMPTY>', '')
+      .replace('ROOT', '')
       .replace(/\s*\/\s*/g, '/'),
   )
   e.preventDefault()
--- a/catalog/app/containers/Bucket/Dir.js
+++ b/catalog/app/containers/Bucket/Dir.js
@@ -9,6 +9,7 @@ import * as M from '@material-ui/core'
 import { Crumb, copyWithoutSpaces, render as renderCrumbs } from 'components/BreadCrumbs'
 import AsyncResult from 'utils/AsyncResult'
 import * as AWS from 'utils/AWS'
+import * as Config from 'utils/Config'
 import { useData } from 'utils/Data'
 import * as NamedRoutes from 'utils/NamedRoutes'
 import parseSearch from 'utils/parseSearch'
@@ -81,6 +82,7 @@ export default function Dir({
 }) {
   const classes = useStyles()
   const { urls } = NamedRoutes.use()
+  const { noDownload } = Config.use()
   const history = useHistory()
   const s3 = AWS.S3.use()
   const { prefix } = parseSearch(l.search)
@@ -136,10 +138,12 @@ export default function Dir({
           {renderCrumbs(getCrumbs({ bucket, path, urls }))}
         </div>
         <M.Box flexGrow={1} />
-        <FileView.ZipDownloadForm
-          suffix={`dir/${bucket}/${path}`}
-          label="Download dir as .zip"
-        />
+        {!noDownload && (
+          <FileView.ZipDownloadForm
+            suffix={`dir/${bucket}/${path}`}
+            label="Download directory"
+          />
+        )}
       </M.Box>
 
       <Code gutterBottom>{code}</Code>
--- a/catalog/app/containers/Bucket/Dir.js
+++ b/catalog/app/containers/Bucket/Dir.js
@@ -15,6 +15,7 @@ import parseSearch from 'utils/parseSearch'
 import { getBreadCrumbs, ensureNoSlash, withoutPrefix, up, decode } from 'utils/s3paths'
 
 import Code from './Code'
+import * as FileView from './FileView'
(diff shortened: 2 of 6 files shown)


catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -473,7 +484,7 @@ export default function PackageTree({
 
   const crumbs = React.useMemo(() => {
     const segments = s3paths.getBreadCrumbs(path)
-    if (path !== '') segments.unshift({ label: 'ROOT', path: '' })
+    segments.unshift({ label: 'ROOT', path: '' })
```
fiskus, line 487: `Ramda#prepend` is an immutable alternative. I would use Ramda here as we already use it

catalog/app/components/BreadCrumbs/BreadCrumbs.js
```diff
@@ -29,6 +29,7 @@ export function copyWithoutSpaces(e) {
       .getSelection()
       .toString()
       .replace('<EMPTY>', '')
+      .replace('ROOT', '')
```
fiskus, line 32: It's not clear to me, what is going on here :)

I think that it's a good practice to avoid the use of strings in a such way. It's a sort of Magic Number pattern (https://en.wikipedia.org/wiki/Magic_number_(programming))
With named constants, it would be simpler to understand the intentions of code.

fiskus (approved):

catalog/app/containers/Bucket/PackageTree.js
nl0: the point is i want to mutate the existing variable, rather than coming up with a name for a new one)

catalog/app/components/BreadCrumbs/BreadCrumbs.js
nl0: well, i somewhat agree, but it's just how package root dir and empty prefixes show up in the breadcrumbs

catalog/app/containers/Bucket/PackageTree.js
fiskus: I mean:

```
const emptySegment = { label: 'ROOT', path: '' }
const segments = R.prepend(emptySegment, s3paths.getBreadCrumbs(path))
```
or
```
const emptySegment = { label: 'ROOT', path: '' }
const segments = [emptySegment, ...s3paths.getBreadCrumbs(path)]
```

---

But it's just my opinion, I try to avoid `.push` and `.unshift`

nl0: yeah, it makes sense. unshift made more sense when it was under a condition

akarve (approved):
UX looks good. I know this is without Dima's changes but right now the Download button navigates and we need to make sure that never happens. IDK if that means overriding content-type and/or content-encoding at the NGINX level.

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -300,7 +301,17 @@ function DirDisplay({ bucket, name, revision, path, crumbs }) {
       }))
       return (
         <>
-          <TopBar crumbs={crumbs} />
+          <TopBar crumbs={crumbs}>
+            {hashData.case({
+              Ok: ({ hash }) => (
+                <FileView.ZipDownloadForm
+                  label="Download package as .zip"
```
akarve, line 308: Does this actually download sub-packages, or no?

catalog/app/containers/Bucket/Dir.js
```diff
@@ -145,6 +146,10 @@ export default function Dir({
           {renderCrumbs(getCrumbs({ bucket, path, urls }))}
         </div>
         <M.Box flexGrow={1} />
+        <FileView.ZipDownloadForm
+          suffix={`dir/${bucket}/${path}`}
+          label="Download dir as .zip"
```
akarve, line 151: Instead of the down arrow, let's use the `file-zip` (I added it to our fort-awesome) or equivalent. It's just a zip archive icon. That way we can have less text in the button.

```suggestion
          label="Download directory"
```

catalog/app/containers/Bucket/PackageTree.js
akarve, line 308: Same as above, use .zip/archive icon
```suggestion
                  label="Download package"
```

catalog/app/containers/Bucket/FileView.js
```diff
@@ -29,41 +32,56 @@ export function Meta({ data, ...props }) {
...
+  ))
+}
+
+export function ZipDownloadForm({ suffix, label }) {
+  const { s3Proxy, noDownload } = Config.use()
+  const { token } = redux.useSelector(Auth.selectors.tokens) || {}
+  if (!token || noDownload) return null
+  const action = `${s3Proxy}/zip/${suffix}`
+  return (
+    <form action={action} target="_blank" method="POST" style={{ flexShrink: 0 }}>
```
akarve, line 81: does `target="_self"` get around opening the separate tab. won't `_blank` break embed? see discussion in slack.

catalog/app/containers/Bucket/PackageTree.js
nl0: nope

catalog/app/containers/Bucket/FileView.js
nl0: embed works ok btw

PR #1873, opened 2020-10-30
Preview/loaders/Json: render large vega files

nl0:
Render vega files which are <20 MiB right away, render larger ones after pressing a button

# TODO

### DIFF
--- a/catalog/app/components/Preview/loaders/Json.js
+++ b/catalog/app/components/Preview/loaders/Json.js
@@ -2,19 +2,21 @@ import * as R from 'ramda'
 import * as React from 'react'
 
 import * as AWS from 'utils/AWS'
+import AsyncResult from 'utils/AsyncResult'
 import * as Resource from 'utils/Resource'
 
 import { PreviewData, PreviewError } from '../types'
 import * as Text from './Text'
 import * as utils from './utils'
 
-const MAX_SIZE = 1024 * 1024
+const MAX_SIZE = 20 * 1024 * 1024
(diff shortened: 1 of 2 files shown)


catalog/app/components/Preview/loaders/Json.js
```diff
@@ -58,16 +60,21 @@ function VegaLoader({ handle, children }) {
...
 
 export const detect = R.either(utils.extIs('.json'), R.startsWith('.quilt/'))
 
 export const Loader = function JsonLoader({ handle, children }) {
   return utils.useFirstBytes({ bytes: 256, handle }).case({
     Ok: ({ firstBytes, contentLength }) =>
-      !!detectSchema(firstBytes) && contentLength <= MAX_SIZE ? (
-        <VegaLoader {...{ handle, children }} />
+      detectSchema(firstBytes) ? (
+        <VegaLoader {...{ handle, children, gated: contentLength > MAX_SIZE }} />
```
fiskus, line 77: What is the meaning of "gated"? If a file is too large, then we render it through a different "gate", or it does not fit into our gates?

fiskus (approved):

catalog/app/components/Preview/loaders/Json.js
nl0: it's size-gated, meaning that it stays behind the gate if it's too large, waiting to be let in, idk smth like this lol

fiskus: Got it, thanks

PR #1874, opened 2020-10-30
Metadata quality verification API

sir-sigurd:
# Description
This adds `workflow` parameter to `Package.push()` and `Package.build()` functions.

# TODO

- [x] Decide on `workflow` values for 'don't use workflow' and 'use default'
- [x] Store the used workflow ID and URL with versionID
- [x] Unit tests
- [x] Documentation
    - [x] Update docstrings
    - [x] Document `workflows/config.yml`
    - [x] [Run `build.py`](../gendocs/build.py) for new docstrings
- [x] Move JSON Schema for config validation to its own file


api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
...
+  "$id": "https://quiltdata.com/workflows/config/1",
+  "type": "object",
+  "required": [
+    "version",
+    "workflows"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "version": {
+      "const": 1
```
sir-sigurd, line 12: Should we use `string` instead of `number` for config version?

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
...
+      "type": "string",
+      "description": "The workflow that will be used if no workflow provided by user."
+    },
+    "workflows": {
+      "type": "object",
+      "minProperties": 1,
+      "propertyNames": {
+        "pattern": "^[A-Za-z_-][A-Za-z0-9_-]*$",
+        "maxLength": 64
+      },
```
sir-sigurd, line 29: I don't have clear reasons for these constraints, I added them just in case we'd want to rework this somehow in future: like putting workflows in separate files or something like this.
Does that make sense?

sir-sigurd, line 29: @akarve 
I don't have clear reasons for these constraints, I added them just in case we'd want to rework this somehow in future: like putting workflows in separate files or something like this.
Does that make sense?

api/python/quilt3/workflows/config-1.schema.json
fiskus: My opinion: `strings` for version is better. It's more reliable, and we can use even semver

sir-sigurd: I think it's more flexible, so I made it a `string`. Though I'm not sure how semver could be applied for schemas.

api/python/quilt3/workflows/config-1.schema.json
akarve: It feels overly restrictive since any python or even JS access will quote key names? I'n not against this but it might frustrate users. We had this with quilt packages names: they were very restrictive and users kept bumping into this limit.

api/python/quilt3/packages.py
```diff
@@ -947,10 +977,13 @@ def build(self, name, registry=None, message=None):
             registry: registry to build to
                 defaults to local registry
             message: the commit message of the package
+            %(workflow)s
 
         Returns:
             The top hash as a string.
         """
+        registry = get_package_registry(registry)
+        self._validate_with_workflow(registry=registry, workflow=workflow, message=message)
```
akarve, line 986: Here and elsewhere, shouldn't we be documenting the errors that can be thrown in the docstring?

docs/SUMMARY.md
```diff
@@ -21,6 +21,7 @@
 * [Materialization](Advanced Features/Materialization.md)
 * [Working with Manifests](Advanced Features/Working with Manifests.md)
 * [S3 Select](Advanced Features/S3 Select.md)
+* [Workflows](Advanced Features/workflows.md)
```
akarve, line 24: Very much looking for a .md file that shows an example of workflow file in this PR :)

api/python/quilt3/packages.py
sir-sigurd: Ideally, yes, but it should be well-thought since it becomes the part of the public API, so I'd postpone it until this stuff become more well-defined.

api/python/quilt3/workflows/config-1.schema.json
sir-sigurd: These constraints inspired by [GH workflow config](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_id). docker-compose has [similar ones](https://github.com/docker/compose/blob/675c9674e1701cc22763bab2bcd429ac5b82e102/compose/config/compose_spec.json#L18). The general idea is that in the future making it less restrictive is easier than making it more restrictive, because the latter is a breaking change. Also users have to quote it CLI.

docs/API Reference/Package.md
```diff
@@ -205,6 +205,10 @@ __Arguments__
 * __registry__:  registry to build to
     defaults to local registry
 * __message__:  the commit message of the package
+* __workflow__:  workflow ID or `None` to skip workflow validation.
+    If not specified, the default workflow will be used.
+* __For details see__:  https://docs.quiltdata.com/advanced-usage/workflows
```
sir-sigurd, line 210: Unfortunately our renderer can't into links :disappointed:.

docs/Advanced Features/workflows.md
```diff
@@ -0,0 +1,143 @@
+*New in Quilt 3.3*
```
sir-sigurd, line 1: @akarve 
could you check this?
Config and schemas are in the bucket, so all these commands work and you can play with it if you want.

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
+{
```
sir-sigurd, line 1: @akarve 
I'd like you to review descriptions in this file as well.

api/python/tests/integration/test_workflows.py
```diff
@@ -0,0 +1,345 @@
...
+                metadata_schema: schema-id
+            schemas:
+              schema-id:
+                url: %s
+        ''' % create_local_tmp_schema('{"$ref": "other-schema"}')))
+
+        with pytest.raises(QuiltException, match=r"Currently we don't support \$ref in schema."):
+            self._validate(workflow='w1')
+
+    def test_schema_validation_invalid_meta(self):
```
akarve, line 180: This feels like one of the most important tests; shouldn't there be at least on test against `push` that hits this case?

docs/Advanced Features/workflows.md
```diff
@@ -0,0 +1,143 @@
...
+
+`quilt3` provides API to ensure quality of package metadata. To start use it you need to create
```
akarve, line 3:
```suggestion
`quilt3` provides an API to ensure the quality of package metadata. To get started, create a
```

docs/Advanced Features/workflows.md
```diff
@@ -0,0 +1,143 @@
...
+configuration file in your registry under `.quilt/workflows/config.yml`. For example:
```
akarve, line 4: Shouldn't we provide this as a constant from `quilt3`, or from an API?

docs/Advanced Features/workflows.md
```diff
@@ -0,0 +1,143 @@
...
+    is_message_required: true
+    metadata_schema: top-secret
+schemas:
+ superheroes:
+   url: s3://quilt-sergey-dev-metadata/schemas/superheroes.schema.json
+ top-secret:
+   url: s3://quilt-sergey-dev-metadata/schemas/top-secret.schema.json
+```
+
+With this configuration you can't push anything without specifying workflow:
```
akarve, line 27:
```suggestion
With the above configuration you must specify a workflow before you can push:
```

docs/Advanced Features/workflows.md
```diff
@@ -0,0 +1,143 @@
...
+
+```python
+>>> import quilt3
+>>> quilt3.Package().push('test/package', registry='s3://quilt-sergey-dev-metadata')
+
+QuiltException: Workflow is required, but none specified.
```
akarve, line 33:
```suggestion
QuiltException: Workflow required, but none specified.
```

akarve (approved):

akarve:
I will do more testing once we get this on stage, as well as some wordsmithing on the docs, but overall looks good. Added some comments and suggestions.

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
...
+    "workflows"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "version": {
+      "const": "1"
+    },
+    "is_workflow_required": {
+      "type": "boolean",
+      "description": "Whether package can be pushed with unspecified workflow.",
```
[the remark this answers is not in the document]
sir-sigurd: That's not quite correct: if user doesn't pass workflow and there is default, this default will be used.
It could be something like: "If true, any pushed package must conform to some workflow."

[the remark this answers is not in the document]
akarve: I'm not sure that's the right logic. If `is_workflow_required=true` that means ANY one workflow is required. Automatically falling back on `default` is to magical and may confuse people. If a user fails one workflow and pushes again, that's a separate event against a potentially separate workflow.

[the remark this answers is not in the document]
sir-sigurd: I was saying that if there is default workflow set, user doesn't have to "select and pass a workflow" (do you mean pass a workflow to API?), because that default workflow will be used (implicitly selected) for validation.

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
...
+            "type": "boolean",
+            "description": "Whether commit message is required.",
+            "default": false
+          }
+        }
+      }
+    },
+    "schemas": {
+      "type": "object",
+      "description": "JSON Schemas to be used for metadata validation.",
```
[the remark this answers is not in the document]
sir-sigurd: I committed suggestion, but retained capitalization. In [spec](https://json-schema.org/draft/2019-09/json-schema-core.html) they always use either "JSON Schema" or "schema".

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
...
+      "minProperties": 1,
+      "additionalProperties": {
+        "type": "object",
+        "required": [
+          "url"
+        ],
+        "properties": {
+          "url": {
+            "type": "string",
+            "description": "URL from where the schema will be obtained.",
```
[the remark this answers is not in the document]
sir-sigurd: > Is this in fact a URL?

https://tools.ietf.org/html/rfc3986#section-1.1.3
>    A URI can be further classified as a locator, a name, or both.  The
   term "Uniform Resource Locator" (URL) refers to the subset of URIs
   that, in addition to identifying a resource, provide a means of
   locating the resource by describing its primary access mechanism
   (e.g., its network "location").

> I thought we were requiring the schemas to be in S3?

For local repositories we also support `file:///some/path`. 

> If so we should have a more precise format?

I'm not sure, currently I feel it like more implementation detail rather than design choice.

api/python/setup.py
```diff
@@ -70,6 +70,7 @@ def run(self):
         'tenacity>=5.1.1',
         'tqdm>=4.32',
         'requests_futures==1.0.0',
+        'jsonschema==3.*',
```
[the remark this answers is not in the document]
sir-sigurd: We only support Draft 7 and it's supported since [3.0.0](https://github.com/Julian/jsonschema/blob/main/CHANGELOG.rst#v300).

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -0,0 +1,77 @@
...
+  "$id": "https://quiltdata.com/workflows/config/1",
+  "type": "object",
+  "required": [
+    "version",
+    "workflows"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "version": {
+      "const": "1"
```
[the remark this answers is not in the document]
sir-sigurd: I don't quite understand how it works for schemas :man_shrugging:. I think I need to look for example.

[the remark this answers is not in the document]
akarve: Well at a bare minimum, make it a string?

api/python/quilt3/workflows/config-1.schema.json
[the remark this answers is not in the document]
akarve: Oh, by "pass a workflow" I mean "succeed a workflow". Edited.

docs/API Reference/Package.md
akarve: "into"?

api/python/quilt3/workflows/config-1.schema.json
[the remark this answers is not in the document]
sir-sigurd: It's already a string.

docs/API Reference/Package.md
sir-sigurd: Either way it thinks that everything before `:` is a parameter name. At https://docs.quiltdata.com/v/master/api-reference/package#Package.set:
image

docs/Advanced Features/workflows.md
sir-sigurd: I think that currently there is no good place for it.

api/python/quilt3/workflows/config-1.schema.json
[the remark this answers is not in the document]
sir-sigurd: But still, users don't have to select a workflow if there is the default one, have they?

api/python/tests/integration/test_workflows.py
sir-sigurd: I added more unit tests in test_packages.py.

api/python/quilt3/workflows/config-1.schema.json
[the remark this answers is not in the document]
akarve: Rewrote

docs/Advanced Features/workflows.md
akarve: What about util.py? See https://docs.quiltdata.com/api-reference/cli#constants-see-util-py-for-more

akarve (approved):

PR #1875, opened 2020-10-31
Be sure to search key field on normal searches

akarve:
This allows the user to use wildcards on the key (e.g. */research/*) for a research folder anywhere, and results in higher quality results. Also reduce field boosting to ext: and comment: fields.

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -72,7 +72,7 @@ def lambda_handler(request):
         query = request.args.get('query', '')
         my_fields = user_fields or [
             # object
-            'content', 'comment^2', 'ext^2', 'key_text^2', 'meta_text',
+            'content', 'comment', 'ext', 'key', 'key_text^2', 'meta_text',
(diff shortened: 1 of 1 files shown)


akarve:
@nl0 if the catalog overrides `user_fields` please be sure to add `'key'` as needed to match the logic in this PR.

nl0 (approved):
👍

lambdas/search/index.py
```diff
@@ -72,7 +72,7 @@ def lambda_handler(request):
         query = request.args.get('query', '')
         my_fields = user_fields or [
             # object
-            'content', 'comment^2', 'ext^2', 'key_text^2', 'meta_text',
+            'content', 'comment', 'ext', 'key', 'key_text^2', 'meta_text',
```
akarve, line 75:
```suggestion
            'content', 'comment', 'ext', 'key', 'key_text', 'meta_text',
```

PR #1878, opened 2020-11-02
JsonDisplay: tune rendering perf

nl0:
# Description

This changeset optimizes JsonDisplay rendering perf by only using one instance of the `useStyles` hook on the top level and avoiding dynamic style injection that came with `Box`es usage.

# TODO

### DIFF
--- a/catalog/app/components/JsonDisplay/JsonDisplay.js
+++ b/catalog/app/components/JsonDisplay/JsonDisplay.js
@@ -1,17 +1,42 @@
+import cx from 'classnames'
 import * as R from 'ramda'
 import * as React from 'react'
 import * as M from '@material-ui/core'
 
-const IconBlank = () => <M.Box pr={2.5} />
+const useStyles = M.makeStyles((t) => ({
+  root: {
+    fontFamily: t.typography.monospace.fontFamily,
+    fontSize: t.typography.body2.fontSize,
+    overflow: 'auto',
+    whiteSpace: 'pre',
+    width: '100%',
+  },
+  more: {
+    color: t.palette.text.secondary,
+  },
+  flex: {
+    display: 'flex',
+  },
+  compoundInner: {
+    paddingLeft: t.spacing(2),
+  },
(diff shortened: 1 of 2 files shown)


catalog/app/components/JsonDisplay/JsonDisplay.js
```diff
@@ -91,17 +104,21 @@ function CompoundEntry({
         name ? name.length + 2 : 0,
         4, // braces + spaces
       ])
-    if (availableSpace <= 0 || Array.isArray(value)) return <More keys={entries.length} />
+    if (availableSpace <= 0 || Array.isArray(value))
+      return <More keys={entries.length} classes={classes} />
```
fiskus, line 108: I would prefer to see:

```
if (condition) {
  return result
}
```

then

```
if (condition)
  return result
```

fiskus (approved):
Does it mean, that `M.Box` + styles attributes are always worse in terms of performance than `M.makeStyles` + `M.Box`? If this is the case, I would consider to refactor all `M.Box` + attributes occurences.
I prefer `makeStyles` + div, even for more maintainable code. But if it's a more performant choice, it would be a double win :)

nl0:
> Does it mean, that M.Box + styles attributes are always worse in terms of performance than M.makeStyles + M.Box? If this is the case, I would consider to refactor all M.Box + attributes occurences.
I prefer makeStyles + div, even for more maintainable code. But if it's a more performant choice, it would be a double win :)

@fiskus it matters only in specific cases when it's slow). the reason i'm using Box is mostly developer convenience and flexibility (like, if your root element is a Box and you spread the `...props` to it, it could be very handy to customize the layout of the component by the consumer without writing any class definitions)

catalog/app/components/JsonDisplay/JsonDisplay.js
nl0: me too. but this change was introduced by prettier (i hate it already tbh) and i simply missed it

PR #1882, opened 2020-11-02
Discuss: Use logger module instead of console.log

fiskus:
* Added `logger` module
* Replaced `console.{log,error,warn,debug,dir}` to logger method calls

The main purpose is to move `console.log` calls to some module, and:
* remove `// eslint-disabl-next-line` workarounds
* we can pump this module later, send logs to console or log services

I didn't touch files where I was forced to fix other lint issues to keep PR clear.


akarve:
Works for me but up to @nl0

PR #1887, opened 2020-11-04
Enable SSL certificate verification in nginx

dimaryaz:
The fact that it's not the default is kind of insane.

### DIFF
--- a/s3-proxy/Dockerfile
+++ b/s3-proxy/Dockerfile
@@ -71,6 +71,8 @@ RUN set -x \
     && apt-get remove -y curl git gcc make libssl-dev libpcre3-dev zlib1g-dev \
     && apt-get autoremove -y
 
+RUN cat /etc/ssl/certs/*.pem > /etc/nginx/certs.pem
+
 RUN mkdir /var/cache/nginx/
 COPY nginx.conf /root/nginx.conf.tmpl
 COPY run-nginx.sh /root/run-nginx.sh
--- a/s3-proxy/nginx.conf
+++ b/s3-proxy/nginx.conf
@@ -28,13 +28,12 @@ http {
 
         client_max_body_size 8m;
 
-        # if ($http_x_forwarded_proto = 'http') {
-        #     return 400;
-        # }
-
         location ~ ^/(?<s3_region>[^/]+)/(?<s3_bucket>[^/]+)(/(?<s3_path>.*))? {
             proxy_http_version 1.1;
             proxy_buffering off;
+            proxy_ssl_verify on;
+            proxy_ssl_verify_depth 5;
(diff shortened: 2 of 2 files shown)


s3-proxy/nginx.conf
```diff
@@ -28,13 +28,12 @@ http {
...
-        # if ($http_x_forwarded_proto = 'http') {
-        #     return 400;
-        # }
-
         location ~ ^/(?<s3_region>[^/]+)/(?<s3_bucket>[^/]+)(/(?<s3_path>.*))? {
             proxy_http_version 1.1;
             proxy_buffering off;
+            proxy_ssl_verify on;
+            proxy_ssl_verify_depth 5;
+            proxy_ssl_trusted_certificate /etc/nginx/certs.pem;
```
akarve, line 36: I'm guessing it requires a single file, hence the `cat`

s3-proxy/Dockerfile
```diff
@@ -71,6 +71,8 @@ RUN set -x \
     && apt-get remove -y curl git gcc make libssl-dev libpcre3-dev zlib1g-dev \
     && apt-get autoremove -y
 
+RUN cat /etc/ssl/certs/*.pem > /etc/nginx/certs.pem
```
akarve, line 74: Main concern here is that `/etc/ssl/certs/*.pem` falls out of date. If this comes from the base Debian image then I guess it's OK?

akarve (approved):

s3-proxy/nginx.conf
dimaryaz: Yep

s3-proxy/Dockerfile
akarve: This is a good case for rebuilding this container in CI. Like when CAs get compromised these lists all change and depending on a bad list could be bad. Can we maybe get "latest" list from a trusted Python package instead?

dimaryaz: These are the same certs that are hardcoded in every OS and browser, so they practically never change.

(And if they did get out of date... that would break everything else that touches the network, too, not just nginx.)

akarve: Sure and I'm not super worried about this but it has caused Python libraries to get deprecated and issue security warnings. Your base image isn't version pinned so that tag will update in place?

dimaryaz: It is version pinned, I think - "buster" refers to some version. We don't want anything to change without us knowing about it.

But I meant, there's no point in worrying about the certs. That's like asking, "What if they switch over to http/3? Nginx will break!" Yeah, but so will the registry, Python, your browser, and so on. In fact, you won't be able to update the image cause Docker also relies on those certs.

The existing certs should work fine for the next 10 years or so, so no point in worrying right now.

PR #1890, opened 2020-11-05
Pin mod_zip to recent 1.2.0 release

sir-sigurd:
# Description
It doesn't contain any significant changes, so this is made to make version explicit.

Also fixed cleanup.


dimaryaz (approved):
Oh nice! It's convenient that their last release was 12 days ago, after 10 years of no releases...

PR #1893, opened 2020-11-10
Adjust code samples for packages

nl0:
### TODO
- [x] dont use `path`, revert to the old syntax
- [x] strip traling slash
- [x] escape arguments

### DIFF
--- a/catalog/app/containers/Bucket/PackageTree.js
+++ b/catalog/app/containers/Bucket/PackageTree.js
@@ -191,7 +191,7 @@ function PkgCode({ data, bucket, name, revision, path }) {
             import quilt3
             # browse
             p = quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
-            # download
+            # download (be mindful of large packages)
             p = quilt3.Package.install("${name}"${pathPy}${hashPy}, registry="s3://${bucket}", dest=".")
           `,
         },
--- a/catalog/app/containers/Bucket/PackageTree.js
+++ b/catalog/app/containers/Bucket/PackageTree.js
@@ -178,8 +178,7 @@ function ExposeLinkedData({ bucketCfg, bucket, name, revision }) {
 function PkgCode({ data, bucket, name, revision, path }) {
   const code = data.case({
     Ok: ({ hash }) => {
-      const pathPy = path ? `, path="${path}"` : ''
-      const pathCli = path ? ` --path="${path}"` : ''
+      const nameWithPath = JSON.stringify(s3paths.ensureNoSlash(`${name}/${path}`))
       const hashDisplay = revision === 'latest' ? '' : hash.substring(0, 10)
       const hashPy = hashDisplay && `, top_hash="${hashDisplay}"`
       const hashCli = hashDisplay && ` --top-hash ${hashDisplay}`
@@ -192,14 +191,14 @@ function PkgCode({ data, bucket, name, revision, path }) {
             # browse
             p = quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
             # download (be mindful of large packages)
-            p = quilt3.Package.install("${name}"${pathPy}${hashPy}, registry="s3://${bucket}", dest=".")
+            p = quilt3.Package.install(${nameWithPath}${hashPy}, registry="s3://${bucket}", dest=".")
           `,
         },
         {
           label: 'CLI',
           hl: 'bash',
           contents: dedent`
-            quilt3 install ${name}${pathCli}${hashCli} --registry s3://${bucket} --dest .
+            quilt3 install ${nameWithPath}${hashCli} --registry s3://${bucket} --dest .
           `,
         },
       ]
--- a/catalog/app/containers/Bucket/PackageTree.js
+++ b/catalog/app/containers/Bucket/PackageTree.js
@@ -178,7 +178,8 @@ function ExposeLinkedData({ bucketCfg, bucket, name, revision }) {
 function PkgCode({ data, bucket, name, revision, path }) {
   const code = data.case({
     Ok: ({ hash }) => {
-      const nameWithPath = path ? `${name}/${path}` : name
+      const pathPy = path ? `, path="${path}"` : ''
+      const pathCli = path ? ` --path="${path}"` : ''
       const hashDisplay = revision === 'latest' ? '' : hash.substring(0, 10)
       const hashPy = hashDisplay && `, top_hash="${hashDisplay}"`
       const hashCli = hashDisplay && ` --top-hash ${hashDisplay}`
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -178,7 +178,8 @@ function ExposeLinkedData({ bucketCfg, bucket, name, revision }) {
 function PkgCode({ data, bucket, name, revision, path }) {
   const code = data.case({
     Ok: ({ hash }) => {
-      const nameWithPath = path ? `${name}/${path}` : name
+      const pathPy = path ? `, path="${path}"` : ''
+      const pathCli = path ? ` --path="${path}"` : ''
```
fiskus, line 182: If we use camelcase consistently even for acronyms then should name `parseJSON` as `parseJson`, `ReactDOM` as `ReactDom`, and so on. This style is used by Google, for example

But I think `pathCLI` is more consistent with our code

fiskus (commented):
Can you provide open.quiltdata.com link for testing?

nl0:
> Can you provide open.quiltdata.com link for testing?

it's not deployed yet, you can only test it locally

catalog/app/containers/Bucket/PackageTree.js
nl0: yep, i agree, and i try to keep it consistent, but some things (specifically `*JSON` and `ReactDOM`) are just too deeply ingrained

fiskus:
> > Can you provide open.quiltdata.com link for testing?
> 
> it's not deployed yet, you can only test it locally

I understand, but can you give me a link on open.quiltdata.com? I'll open it locally with your branch

nl0:
> > > Can you provide open.quiltdata.com link for testing?
> > 
> > 
> > it's not deployed yet, you can only test it locally
> 
> I understand, but can you give me a link on open.quiltdata.com? I'll open it locally with your branch

sure, you can use this path for example: `/b/ai2-datasets/packages/ai2/arc/tree/latest/.quiltignore`
but it may not work bc of permissions

nl0:
@fiskus just browse any package on staging

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -189,16 +190,16 @@ function PkgCode({ data, bucket, name, revision, path }) {
           contents: dedent`
             import quilt3
             # browse
-            quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
+            p = quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
             # download
-            quilt3.Package.install("${nameWithPath}"${hashPy}, registry="s3://${bucket}", dest=".")
+            p = quilt3.Package.install("${name}"${pathPy}${hashPy}, registry="s3://${bucket}", dest=".")
```
fiskus, line 195: Maybe we should use a different python variable name here, to make it safer to copy-paste and get both variables in REPL:

```
l = Package.browse()
p = Package.install()
```

fiskus (approved):

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -189,16 +190,16 @@ function PkgCode({ data, bucket, name, revision, path }) {
           contents: dedent`
             import quilt3
             # browse
-            quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
+            p = quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
             # download
```
akarve, line 194:
```suggestion
            # download (be mindful of large packages)
```

catalog/app/containers/Bucket/PackageTree.js
akarve: Good catch. I think it's OK to use `p =` for both, since users will likely only do one of them and they both return the a Package object.

akarve (approved):

sir-sigurd:
I recalled that I already suggested using the new syntax for sub-packages, but we decided not to do this, because if the user uses the new syntax but has the old quilt3, he gets an error, and if the user has the new quilt3 and uses the old syntax, he receives a warning.
I'm not sure if enough time has passed to show only the new syntax, or we should display both, or do nothing.
@akarve what do you think?

akarve (approved):
Per discussion use the new `--path` syntax.

sir-sigurd (commented):
The change is fine, but the snippet works only for single files, but not dirs, because of `/` at the end:
image
```
$ quilt3 install test/filename-with-quotes --path="dir2/" --registry s3://quilt-sergey-dev-metadata --dest .
Invalid key 'dir2/'. A package entry key cannot contain a file or folder named '.' or '..' in its path.
```
The old version didn't work as well.
This one works:
```
quilt3 install test/filename-with-quotes --path="dir2" --registry s3://quilt-sergey-dev-metadata --dest .
```
I'm not yet sure if we should allow trailing `/` in Python API/CLI.

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -178,7 +178,7 @@ function ExposeLinkedData({ bucketCfg, bucket, name, revision }) {
 function PkgCode({ data, bucket, name, revision, path }) {
   const code = data.case({
     Ok: ({ hash }) => {
-      const nameWithPath = path ? `${name}/${path}` : name
+      const nameWithPath = JSON.stringify(s3paths.ensureNoSlash(`${name}/${path}`))
```
fiskus, line 181: `ensureNoTrailingSlash`, I suppose

fiskus: Maybe `encodeURIComponent` is more appropriate than `JSON.stringify`?

fiskus (approved):

catalog/app/containers/Bucket/PackageTree.js
nl0: > ensureNoTrailingSlash

it would be more precise, but the function is called just `ensureNoSlash`

> Maybe encodeURIComponent is more appropriate than JSON.stringify?

we need simple special char escaping here to use the escaped strings in python and shell, not uri encoding

catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -189,9 +189,9 @@ function PkgCode({ data, bucket, name, revision, path }) {
           contents: dedent`
             import quilt3
             # browse
-            quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
-            # download
-            quilt3.Package.install("${nameWithPath}"${hashPy}, registry="s3://${bucket}", dest=".")
+            p = quilt3.Package.browse("${name}"${hashPy}, registry="s3://${bucket}")
+            # download (be mindful of large packages)
+            p = quilt3.Package.install(${nameWithPath}${hashPy}, registry="s3://${bucket}", dest=".")
```
sir-sigurd, line 194:
```suggestion
            quilt3.Package.install(${nameWithPath}${hashPy}, registry="s3://${bucket}", dest=".")
```
`Package.install()` returns nothing since a40d4b05ffa8a7cd44451dfee3aaddb88c70304c.

PR #1894, opened 2020-11-10
Search: show retry btn for syntax errors

nl0:


catalog/app/components/SearchResults/SearchResults.js
```diff
@@ -613,6 +613,9 @@ export const handleErr = (retryUrl) => (e) => (
           supported query syntax
         </StyledLink>
         .
+        {!!retryUrl && (
+          <> You can also click RETRY to try a simplified version of your query.</>
```
fiskus, line 617: Just curious, can we replace `<></>` with quotes? Or it doesn't work such way?

nl0: yep, in this case we can, bc there's no nested react elements

akarve (approved):

PR #1899, opened 2020-11-11
Respect terminate_after for all searches

akarve:
* Open was acting slow, probably because sooo many documents per shard
* We should always respect the template-provided terminate_after so that searches finish
* In a future PR we can add some kind of setting to force a deep search but we will still be up against 29 sec API gateway limits

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -67,6 +67,7 @@ def lambda_handler(request):
             raise ValueError("'packages' action searching indexes that don't end in '_packages'")
         _source = user_source
         size = user_size
+        terminate_after = None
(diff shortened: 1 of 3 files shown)


lambdas/search/index.py
```diff
@@ -67,7 +67,6 @@ def lambda_handler(request):
             raise ValueError("'packages' action searching indexes that don't end in '_packages'")
         _source = user_source
         size = user_size
-        terminate_after = None
```
kevinemoore, line 70: This will potentially skip package results. Do we really want that?

kevinemoore (approved):

PR #1907, opened 2020-11-12
Tech Debt: Eliminate lodash type checkers

fiskus:


catalog/app/components/JsonEditor/Note.js
```diff
@@ -27,12 +21,15 @@ const useStyles = M.makeStyles((t) => ({
...
 function getTypeAnnotationFromValue(value, schema) {
   return R.cond([
-    [isArray, () => 'array'],
-    [isObject, () => 'object'],
-    [isString, () => (R.propOr([], 'enum', schema).includes(value) ? 'enum' : 'string')],
-    [isNumber, () => 'number'],
-    [isBoolean, () => 'boolean'],
-    [isNull, () => 'null'],
+    [Array.isArray, () => 'array'],
+    [R.is(Object), () => 'object'],
```
nl0, line 25: let's put this check at the very bottom, just above the catch-all, bc `R.is(Object, new String('')) === true` (it's an edge case and probably there's zero chance to run into it, but anyways)

nl0 (approved):
all good except for one minor comment

PR #1909, opened 2020-11-12
Fix Schema placeholders for deep nested objects

fiskus:
Created a more sophisticated way of getting "Object.keys()" for a current nesting level of object + schema:

Example

Object: `{ "a": { "b": "c" }}`
Schema: `{ "properties": { "a": { "properties": { d, e } }}}`
Path to object: `["a"]`
Result: `[ "b", "d", "e" ]`

---

Vocabulary:
* `objValue` is a link to an object or array inside object
* `schemaItem` is a link to schema property inside Schema

### DIFF
--- a/catalog/app/components/JsonEditor/State.js
+++ b/catalog/app/components/JsonEditor/State.js
@@ -198,14 +198,36 @@ function getJsonDictItem(jsonDict, obj, parentPath, key) {
   }
 }
 
-function getObjKeys(obj, objPath, rootKeys) {
+const noKeys = []
+
+function getObjValueKeys(objValue) {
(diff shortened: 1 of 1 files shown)


nl0 (approved):
not sure what's happening here, but it looks ok

fiskus:
> not sure what's happening here, but it looks ok

I have these data structures:
* Object
* Schema
* Columns -- it's a UI data, represented table columns
* Path -- the path to object like in `Ramda.path`

To render every column of the UI table, I need the list of keys for the current column, i.e. keys for Object and Schema for a given Path. This PR creates a way to get Object keys for a given Path, Schema keys (`properties`) for a given Path, and merges them

PR #1912, opened 2020-11-13
JSON editor UX improvements, part 1

fiskus:
# Description

* Navigate with Tab through editable cells only
* Added description to a type help tooltip
* Required key is bolder now

### DIFF
--- a/catalog/app/components/JsonEditor/Cell.js
+++ b/catalog/app/components/JsonEditor/Cell.js
@@ -76,11 +76,6 @@ export default function Cell({
 
   const [value, setValue] = React.useState(initialValue)
 
-  const isEditable = React.useMemo(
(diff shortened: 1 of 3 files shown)


catalog/app/components/JsonEditor/Cell.js
```diff
@@ -178,7 +179,7 @@ export default function Cell({
     <div
       className={cx(classes.root, { [classes.disabled]: !isEditable })}
       role="textbox"
-      tabIndex={0}
+      tabIndex={isEditable ? null : 0}
```
nl0, line 182: you sure this should be `null` for editable fields and `0` for non-editable, not the other way around?

nl0 (approved):
looks good, just one question

catalog/app/components/JsonEditor/Cell.js
fiskus: :(

PR #1922, opened 2020-11-18
WIP: Workflows unit tests

fiskus:
# Description

* Move Workflows data "as is" but to the directory designated for data
* Add tests for Workflows data

### CR feedback
- [ ] explicitly name units of work, i.e. ".parse" instead of "Workflows config"
- [ ] adjust `it`s to start with "should ..." and have more specific formulations


catalog/app/data/workflows.js
```diff
@@ -0,0 +1,70 @@
...
+  return {
+    description: workflow.description,
+    isDefault: workflowSlug === data.default_workflow,
+    name: workflow.name,
+    schema: parseSchema(workflow.metadata_schema, data.schemas),
+    slug: workflowSlug,
+  }
+}
+
+export default function parseWorkflows(workflowsYaml) {
```
nl0, line 51: making this export default doesn't make much sense to me, i'd rather keep it a regular named export (e.g. `parse`)

catalog/app/data/workflows.spec.js
```diff
@@ -0,0 +1,108 @@
+import parseWorkflows, {
+  emptyWorkflowsConfig,
+  workflowNotAvaliable,
+  workflowNotSelected,
+} from './workflows'
+
+describe('Workflows config', () => {
```
nl0, line 7: cascade of `describe`s and `it`s should form a proper sentence, e.g.
```javascript
describe('workflows.parse', () => {
  describe('when given an empty config file', () => {
    beforeEach(() => {
      // set things up
    })
    it('should return empty config', () => {
      // assertions
    })
    // it('should do something or be something', () => {})
  })
})
```
so it reads like "workflows.parse when given an empty config file should return empty config".
`describe` sets context (specify subject or condition), `it` defines an invariant

catalog/app/data/workflows.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  it('empty file', () => {
+    const config = parseWorkflows('')
+    expect(config).toBe(emptyWorkflowsConfig)
+    expect(config.workflows[0].slug).toBe(workflowNotAvaliable)
```
nl0, line 11: i think this assertion is redundant, because the previous one already guarantees this

catalog/app/data/workflows.spec.js
```diff
@@ -0,0 +1,108 @@
+import parseWorkflows, {
+  emptyWorkflowsConfig,
+  workflowNotAvaliable,
+  workflowNotSelected,
+} from './workflows'
+
+describe('Workflows config', () => {
+  it('empty file', () => {
+    const config = parseWorkflows('')
+    expect(config).toBe(emptyWorkflowsConfig)
```
nl0, line 10: iirc, `.toBe` check for reference equality, but semantically it makes more sense to check for structural / value equality here (`.toEqual` iirc)

catalog/app/data/workflows.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    expect(config.workflows[0].slug).toBe(workflowNotAvaliable)
+  })
+
+  it('no workflows', () => {
+    const data = `
```
nl0, line 15: we can use `dedent` module here to make these strings look a little nicer

nl0 (changes requested):
i think this is a good idea, but i have some suggestions

##### 1. API
i think we should import the whole module and use it as a namespace to add context to its members, e.g.
```
import * as workflows ...

workflows.parse(...)
```
see inline comments for more detail.

##### 2. Module location
creating `data` dir makes sense, but we didnt have it and all similar stuff lives in `utils`. maybe now is not the right time to introduce new structure to avoid more "lava layer" thing, tho i wont insist on this.

##### 3. Test cases structure
please see a note on structuring the tests

catalog/app/data/workflows.js
fiskus: It's the main thing for this module. It is the module, and other stuff is just auxiliary.

The same thing like `import React, { useState, useCallback } from 'react'`

catalog/app/data/workflows.spec.js
fiskus: It guarantees it since `workflowNotAvaliable` is a part of `emptyWorkflowsConfig`. You can break it, and this test to stop you from doing this

catalog/app/data/workflows.spec.js
fiskus: Yes, it's reference equality, and I wanted to check it for reference equality

catalog/app/data/workflows.spec.js
fiskus: Maybe. But I think tests should have no extra stuff as much as possible.

catalog/app/data/workflows.js
```diff
@@ -0,0 +1,70 @@
+import * as R from 'ramda'
+
+import yaml from 'utils/yaml'
+
+/* TODO:
+   ```
+   class Workflows {
```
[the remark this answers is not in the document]
fiskus: Ok

Consider that eslint can't catch mistyping with *-imports:

```js
import * as workflows from 'utils/workflows'
workfow.pars() // no error at compile time, error in runtime
```

```js
import { pars } from '../../../utils/workflows'
// eslint import error (relative paths also helpful when text editor or IDE can't use webpack config)
```

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -658,8 +659,8 @@ function SchemaFetcher({ children, schemaUrl }) {
 }
 
 const getWorkflowApiParam = R.cond([
-  [R.equals(requests.workflowNotAvaliable), R.always(undefined)],
-  [R.equals(requests.workflowNotSelected), R.always(null)],
+  [R.equals(workflows.workflowNotAvaliable), R.always(undefined)],
```
nl0, line 662:
```suggestion
  [R.equals(workflows.notAvaliable), R.always(undefined)],
```

catalog/app/containers/Bucket/UploadDialog.js
```diff
@@ -658,8 +659,8 @@ function SchemaFetcher({ children, schemaUrl }) {
...
+  [R.equals(workflows.workflowNotSelected), R.always(null)],
```
nl0, line 663:
```suggestion
  [R.equals(workflows.notSelected), R.always(null)],
```

catalog/app/data/workflows.js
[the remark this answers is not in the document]
nl0: yep, good point. i think typescript catches those

catalog/app/data/workflows.spec.js
nl0: well i think readability is one of the most important aspects here, and one tiny extra dependency doesnt seem like a problem. i understand your point about avoiding adding complexity to the tests, but imo it's negligible in this case.

catalog/app/data/workflows.spec.js
nl0: but i don't think that's right, bc it doesn't matter if it returns the exact object defined in the module (it's an implementation detail), what really matters here is the value / structure of the returned object

nl0 (changes requested):
I propose the following approach for structuring unit tests we can use all across the codebase consistently:

```javascript
// The top-level describe should contain full module name, so it's easy to identify the module in test runner output
describe('utils/workflow', () => {
  // The next level should contain all the units of work we're about to test
  describe('.parse', () => {
    // The next level may contain the assertions for simpler cases:
    it('should ... when ...', () => {
      // setup
      ...
      // assetions
      expect(...)
    })

    // ... or nested groupings for more complex cases or cases where we need several assertions for one set-up / condition
    describe('when ...', () => {
      // setup
      ...

      beforeEach(() => {
        // more setup
        ...
      })

      it('should ...', () => {
        // one assertion / set of assertions
        ...
      })

      it('should ...', () => {
        // another assertion / set of assertions
        ...
      })
    })
  })

  describe('<another unit / part of the public API>', ...)
}
```

By a widely adopted convention, `it`s should start with "should".

Also, let's try avoiding generic formulations like 'should be correct' or 'should work' as much as possible and instead using some descriptive language e.g. "should return an empty config when given an empty file".

Here's a [decent js unit testing guide](https://github.com/mawrkus/js-unit-testing-guide).

BTW I'd prefer the BDD/Cucumber-style Given/When/Then structure, but I think this discussion is out of scope of the current PR.

catalog/app/utils/workflows.spec.js
```diff
@@ -0,0 +1,108 @@
+import dedent from 'dedent'
+
+import * as workflows from './workflows'
+
+describe('utils/workflows', () => {
+  describe('Workflows config', () => {
+    it('is correct for an empty file', () => {
+      const config = workflows.parse('')
+      expect(config).toBe(workflows.emptyConfig)
```
nl0, line 9: i still believe there should be `.toEqual` here

fiskus:
Will start anew PR with tests only

PR #1930, opened 2020-11-20
Use Select for enums

fiskus:
# Description

* Used Select for enums
* Handled complex enums like `{ "type": "array", "enum": [[1, 2], [10, 20]] }`
* Moved values formatting into `PreviewValue` component, added tooltip with JSON for objects


nl0 (approved):
works and looks really good, tho i've spotted one inconsistency in type display: enum types are rendered in regular font while primitive ones -- in fixed-width font
Screenshot 2020-11-23 at 09 32 20

fiskus:
> enum types -- in fixed-width font

Thanks, fixed

PR #1933, opened 2020-11-24
Workflow is required always, even when no default workflow

fiskus:
# Description

Workflow can be optional for package creation. But UI element is mandatory and should have value always

Screenshot from 2020-11-24 12-37-18


fiskus:
@nl0 fix is ready, but I want to merge it after resolving conflicts when your PR #1856 would be in the master

PR #1934, opened 2020-11-24
Discuss: Coding Standards guide

fiskus:
Guide for formatting and programming practices outside of ESLint scope. Logging rules is crucial to preserve healthy PR culture.

Ideal scenario: don't demand code changes when this isn't covered with this guide.


nl0 (commented):
great idea. i'll try to get to this soonish and capture as many conventions as i can

sir-sigurd:
How about adding CI job to ensure this?

fiskus:
> How about adding CI job to ensure this?

Sure, we should automate everything we can. These rules are for something we can't or didn't yet automate.

nl0:
> How about adding CI job to ensure this?

FYI we already have prettier and eslint in pre-commit hook

fiskus:
Don't understand what's wrong with this PR. But it seems that it's not going to be approved

PR #1935, opened 2020-11-25
Immutable URIs WIP

nl0:
# Description

Add `/uri/*` route to resolve Quilt package URIs as per spec

# TODO

- [x] nice UI
- [x] Unit tests
- [x] use hashes for routing (#1938)

### DIFF
--- a/catalog/app/constants/routes.js
+++ b/catalog/app/constants/routes.js
@@ -81,6 +81,12 @@ export const search = {
     `/search${mkSearch({ q, buckets, p, mode, retry })}`,
 }
 
+// immutable URI resolver
+export const uriResolver = {
+  path: '/uri/:uri(.*)',
+  url: (uri) => `/uri/${uri ? encodeURIComponent(uri) : ''}`,
+}
+
 // bucket
 export const bucketRoot = {
   path: '/b/:bucket',
@@ -114,9 +120,9 @@ export const bucketPackageDetail = {
 export const bucketPackageTree = {
   path: `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/tree/:revision/:path(.*)?`,
   url: (bucket, name, revision, path = '') =>
-    revision === 'latest' && !path
+    (!revision || revision === 'latest') && !path
       ? bucketPackageDetail.url(bucket, name)
-      : `/b/${bucket}/packages/${name}/tree/${revision}/${encode(path)}`,
+      : `/b/${bucket}/packages/${name}/tree/${revision || 'latest'}/${encode(path)}`,
 }
 export const bucketPackageRevisions = {
   path: `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/revisions`,
--- a/catalog/app/constants/routes.js
+++ b/catalog/app/constants/routes.js
@@ -120,9 +120,9 @@ export const bucketPackageDetail = {
 export const bucketPackageTree = {
   path: `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/tree/:revision/:path(.*)?`,
   url: (bucket, name, revision, path = '') =>
-    (!revision || revision === 'latest') && !path
-      ? bucketPackageDetail.url(bucket, name)
-      : `/b/${bucket}/packages/${name}/tree/${revision || 'latest'}/${encode(path)}`,
+    path || (revision && revision !== 'latest')
+      ? `/b/${bucket}/packages/${name}/tree/${revision || 'latest'}/${encode(path)}`
+      : bucketPackageDetail.url(bucket, name),
 }
 export const bucketPackageRevisions = {
   path: `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/revisions`,
--- a/catalog/app/containers/UriResolver/UriResolver.js
+++ b/catalog/app/containers/UriResolver/UriResolver.js
@@ -1,0 +1,139 @@
+import * as React from 'react'
+import { useHistory, Link, Redirect } from 'react-router-dom'
+import * as M from '@material-ui/core'
+
+import Layout from 'components/Layout'
+import * as NamedRoutes from 'utils/NamedRoutes'
+import * as PackageUri from 'utils/PackageUri'
+
+const useStyles = M.makeStyles((t) => ({
+  form: {
+    display: 'flex',
+    marginTop: t.spacing(3),
+  },
+  btn: {
+    marginLeft: t.spacing(2),
+  },
+  linkBox: {
+    marginTop: t.spacing(4),
+    marginLeft: 'auto',
+    marginRight: 'auto',
+  },
+  link: {
+    display: 'block',
+    padding: t.spacing(1.5, 2),
+    '&:hover': {
+      background: t.palette.action.hover,
+    },
+  },
+  field: {
+    ...t.typography.body1,
+    display: 'block',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
+  },
+  name: {
+    fontWeight: t.typography.fontWeightMedium,
+  },
+}))
+
+export default function UriResolver({ match }) {
+  const { urls } = NamedRoutes.use()
+  const history = useHistory()
+  const classes = useStyles()
+
+  const uri = decodeURIComponent(match.params.uri || '')
+  const [parsed, error] = React.useMemo(() => {
+    try {
+      return [uri ? PackageUri.parse(uri) : null]
+    } catch (e) {
+      return [null, e]
+    }
+  }, [uri])
+
+  const [value, setValue] = React.useState(uri)
+
+  const handleChange = React.useCallback(
+    (e) => {
+      setValue(e.target.value)
+    },
+    [setValue],
+  )
+
+  const handleSubmit = React.useCallback(
+    (e) => {
+      e.preventDefault()
+      if (value !== uri) history.push(urls.uriResolver(value))
+    },
+    [value, uri, history, urls],
+  )
+
+  const to =
+    parsed &&
+    urls.bucketPackageTree(
+      parsed.bucket,
+      parsed.name,
+      parsed.hash || parsed.tag,
+      parsed.path,
+    ) + NamedRoutes.mkSearch({ resolvedFrom: uri })
+
+  // automatically redirect if URI is pre-filled with a correct value
+  const [redirect] = React.useState(!!to)
+  if (redirect) return <Redirect to={to} />
+
+  const field = (name, val) => (
+    <span className={classes.field}>
+      <span className={classes.name}>{name}:</span> {val}
+    </span>
+  )
+
+  return (
+    <Layout
+      pre={
+        <M.Box pt={6} pb={4} maxWidth={'656px !important'} component={M.Container}>
+          <M.Typography variant="h4" align="center">
+            Resolve a Quilt package URI
+          </M.Typography>
+
+          <form className={classes.form} onSubmit={handleSubmit}>
+            <M.Input
+              value={value}
+              onChange={handleChange}
+              error={!!error}
+              placeholder="Enter a URI, e.g. quilt+s3://your-bucket#package=user/package@hash"
+              fullWidth
+            />
+            <M.Button
+              className={classes.btn}
+              type="submit"
+              variant="contained"
+              color="primary"
+            >
+              Resolve
+            </M.Button>
+          </form>
+
+          {!!error && (
+            <M.Box mt={2}>
+              <M.Typography color="error">
+                Error parsing URI: {error.msg || `${error}`}
+              </M.Typography>
+            </M.Box>
+          )}
+          {!!parsed && (
+            <M.Paper className={classes.linkBox}>
+              <Link to={to} className={classes.link}>
+                {field('Registry', `s3://${parsed.bucket}`)}
+                {field('Package', parsed.name)}
+                {!!parsed.hash && field('Hash', parsed.hash)}
+                {!!parsed.tag && field('Tag', parsed.tag)}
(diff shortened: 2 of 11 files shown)


catalog/app/constants/routes.js
```diff
@@ -114,9 +120,9 @@ export const bucketPackageDetail = {
 export const bucketPackageTree = {
   path: `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/tree/:revision/:path(.*)?`,
   url: (bucket, name, revision, path = '') =>
-    revision === 'latest' && !path
+    (!revision || revision === 'latest') && !path
```
fiskus, line 123: I've lost. Do we use both `revision`s and `hash`es, or they are the same?

catalog/app/containers/UriResolver/UriResolver.js
```diff
@@ -0,0 +1,139 @@
...
+    parsed &&
+    urls.bucketPackageTree(
+      parsed.bucket,
+      parsed.name,
+      parsed.hash || parsed.tag,
+      parsed.path,
+    ) + NamedRoutes.mkSearch({ resolvedFrom: uri })
+
+  // automatically redirect if URI is pre-filled with a correct value
+  const [redirect] = React.useState(!!to)
```
fiskus, line 82: It's weird that there is no setter.
`const redirect = useMemo(() => !!to, [])` doesn't fit?

catalog/app/containers/UriResolver/UriResolver.js
```diff
@@ -0,0 +1,139 @@
...
+  const field = (name, val) => (
+    <span className={classes.field}>
+      <span className={classes.name}>{name}:</span> {val}
+    </span>
+  )
+
+  return (
+    <Layout
+      pre={
+        <M.Box pt={6} pb={4} maxWidth={'656px !important'} component={M.Container}>
```
fiskus, line 94: :(

catalog/app/utils/PackageUri.js
```diff
@@ -0,0 +1,90 @@
...
+export class PackageUriError extends BaseError {
+  // eslint-disable-next-line react/static-property-placement
+  static displayName = 'PackageUriError'
+
+  constructor(msg, uri) {
+    super(`Invalid package URI (${uri}): ${msg}`, { msg, uri })
+  }
+}
+
+function parsePackageSpec(spec, uri) {
```
fiskus, line 17: Can you make returning value consistent? I mean, return value with the same number of properties. If some property is omitted, make it `null`:

```
  if (spec.includes(':')) {
    return { hash: null, name, tag }
  }
  if (spec.includes('@')) {
    return { hash, name, tag: null }
  }
  return { hash: null, name: spec, tag: null }
```

catalog/app/utils/PackageUri.js
```diff
@@ -0,0 +1,90 @@
...
+// TODO: do we need extra validation for each part (package name, path, registry, etc)?
+export function parse(uri) {
+  const url = parseUrl(uri)
+  if (url.protocol !== 'quilt+s3:') {
+    throw new PackageUriError(
+      `unsupported protocol "${url.protocol}". "quilt+s3:" is currently the only supported protocol.`,
+      uri,
+    )
+  }
+  if (!url.slashes) {
```
fiskus, line 63: `url.parse` is deprecated
https://nodejs.org/dist/latest-v14.x/docs/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost

And actual `URL` doesn't have `slashes` property

catalog/app/constants/routes.js
fiskus: Positive (non-negative :) ) conditions are easier to read:

```
(path && revision !== 'latest')
 ? `/b/${bucket}/packages/${name}/tree/${revision || 'latest'}/${encode(path)}`
 : bucketPackageDetail.url(bucket, name)
```

fiskus (commented):
First shallow reading is done

catalog/app/utils/PackageUri.spec.js
```diff
@@ -0,0 +1,32 @@
+import * as PackageUri from './PackageUri'
+
+// TODO: more coverage
```
fiskus, line 3: I think this comment is obvious and not needed :)

fiskus (commented):
It cost me some brain cycles to realize where can I copy the package's URI. Even I saw it this morning. But I understand this was "quick implementation vs perfect UX" balanced. Maybe we can steal GitHub UI and show copy button near hash.

Screenshot from 2020-11-30 18-42-59
 

Also, I was expecting, that click on the "resolve" button will redirect to the resolved URL.

---

Possible problem: as a user I don't know how to close/hide "Resolved from URI" notification. As a developer, I can see, that some users can bookmark URL with query string or send it to a colleague and will be stuck with this notification.
We can redirect to the canonical URL automatically, or (and) add a close button to notification

---

Issues beyond the task's scope (or not, I don't know): errors are not helpful.

https:///b/fiskus-sandbox-dev/packages/fiskus/2020-11-1?resolvedFrom=quilt%2Bs3%3A%2F%2Ffiskus-sandbox-dev%23package%3Dfiskus%2F2020-11-1
Error: wrong revision, actually: wrong package

http://localhost:3000/b/fiskus-sandbox-dev/packages/fiskus/?resolvedFrom=quilt%2Bs3%3A%2F%2Ffiskus-sandbox-dev%23package%3Dfiskus%2F
Error: nothing here, do you need to log in, and exceptions in console. Actually, no such package, I'm logged in

catalog/app/constants/routes.js
nl0: > Do we use both revisions and hashes, or they are the same?

well, it may be a little confusing, but i think conceptually "revision" is a "package instance" identified by a hash, and in this particular case it makes sense to keep the argument name generic enough ("revision"), bc it can be either a hash or some other thing that identifies a revision, such as a tag (e.g. `latest`) or a timestamp aka pointer file

catalog/app/utils/PackageUri.js
nl0: but why? missing prop access returns undefined, which seems fine

catalog/app/utils/PackageUri.js
nl0: node vs chrome (didnt test other browsers) implementations of w3c URL return different result when trying to parse our URIs, so i decided to fall back to node's 'url' module

catalog/app/utils/PackageUri.spec.js
nl0: it just indicates the work to be done

catalog/app/utils/PackageUri.js
fiskus: You can use `node` implementation of URL.

`const { URL } = require('url')`

---

I don't insist on that, but be prepared. It will shoot your leg one day

nl0:
> It cost me some brain cycles to realize where can I copy the package's URI. Even I saw it this morning. But I understand this was "quick implementation vs perfect UX" balanced. Maybe we can steal GitHub UI and show copy button near hash.

gh ux doesnt fit here exactly, bc a button near a hash is expected to be related to that hash, not to the "canonical URI"

> Also, I was expecting, that click on the "resolve" button will redirect to the resolved URL.

i think i'll implement this and ditch the package box

> Possible problem: as a user I don't know how to close/hide "Resolved from URI" notification. As a developer, I can see, that some users can bookmark URL with query string or send it to a colleague and will be stuck with this notification.
> We can redirect to the canonical URL automatically, or (and) add a close button to notification

i'll add a close btn

> Issues beyond the task's scope (or not, I don't know): errors are not helpful.
>
> https:///b/fiskus-sandbox-dev/packages/fiskus/2020-11-1?resolvedFrom=quilt%2Bs3%3A%2F%2Ffiskus-sandbox-dev%23package%3Dfiskus%2F2020-11-1
> Error: wrong revision, actually: wrong package

well, we may want to add some check to see if the package actually exists (there's at least one revision of that package), but it seems out of scope, you're right

> http://localhost:3000/b/fiskus-sandbox-dev/packages/fiskus/?resolvedFrom=quilt%2Bs3%3A%2F%2Ffiskus-sandbox-dev%23package%3Dfiskus%2F
> Error: nothing here, do you need to log in, and exceptions in console. Actually, no such package, I'm logged in

it's a 404, bc the url doesnt match any route, which seems more or less correct

catalog/app/utils/PackageUri.js
fiskus: To make code more predictable. I need to read the whole function now to know what are possible values (if I'm interested just in form of data). And if I miss some return, I can get unexpected results.

catalog/app/containers/UriResolver/UriResolver.js
nl0: i need this to be executed only once, on construction. `useMemo` does not give us that semantic guarantee and should be used only for optimization

catalog/app/containers/UriResolver/UriResolver.js
```diff
@@ -0,0 +1,103 @@
...
+
+import Layout from 'components/Layout'
+import * as NamedRoutes from 'utils/NamedRoutes'
+import * as PackageUri from 'utils/PackageUri'
+
+const useStyles = M.makeStyles((t) => ({
+  container: {
+    paddingTop: t.spacing(6),
+    paddingBottom: t.spacing(6),
+    maxWidth: '656px !important',
```
fiskus, line 13: You can avoid `!important` by utilizing `maxWidth` attribute:

```jsx
<M.Container maxWidth="sm" />
```

catalog/app/utils/PackageUri.spec.js
```diff
@@ -0,0 +1,179 @@
...
+      )
+    })
+
+    it('should throw on missing name', () => {
+      expect(() => PackageUri.stringify({ bucket: 'bucket' })).toThrowError(
+        /missing "name"/,
+      )
+    })
+
+    it('should throw on both hash and tag', () => {
```
fiskus, line 118: But 
```
PackageUri.parse(`'quilt+s3://bucket-name:latest@intiew'`)
```
(has both hash and tag) is not covered. And it throws error `TypeError: Cannot read property 'replace' of null`

catalog/app/utils/PackageUri.spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+    it('should work for a valid URI w/ hash and path', () => {
+      expect(
+        PackageUri.parse(
+          'quilt+s3://bucket-name#package=quilt/test@abc1&path=sub%2Fpath',
+        ),
+      ).toEqual({
+        bucket: 'bucket-name',
+        name: 'quilt/test',
+        hash: 'abc1',
```
fiskus, line 30: Actually, it is unexpected that `#part` is not hash. It conflicts with https://nodejs.org/dist/latest-v14.x/docs/api/url.html#url_url_hash

I don't know how to resolve it. I think we should consider replacing it with a different word on the project level.

catalog/app/utils/PackageUri.js
```diff
@@ -0,0 +1,98 @@
...
+    throw new PackageUriError('missing slashes between protocol and registry.', uri)
+  }
+  if (url.path) {
+    throw new PackageUriError(
+      'non-bucket-root registries are not supported currently.',
+      uri,
+    )
+  }
+  const bucket = url.host
+  const params = parseQs(url.hash.replace('#', ''))
```
fiskus, line 76: Here. Hash can be `null`

fiskus (approved):
Consider null exception

catalog/app/constants/routes.js
```diff
@@ -114,9 +120,9 @@ export const bucketPackageDetail = {
 export const bucketPackageTree = {
   path: `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/tree/:revision/:path(.*)?`,
   url: (bucket, name, revision, path = '') =>
-    revision === 'latest' && !path
-      ? bucketPackageDetail.url(bucket, name)
-      : `/b/${bucket}/packages/${name}/tree/${revision}/${encode(path)}`,
+    path || (revision && revision !== 'latest')
```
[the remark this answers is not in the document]
nl0: yes, but that clause (`revision !== 'latest'`) is for the cases when `revision` is _not_ `undefined`, i.e. when revision is `undefined` or `"latest"`, we dont append it to the url, but when it's defined and not `"latest"`, we do

catalog/app/containers/UriResolver/UriResolver.js
nl0: nope, because i need specific custom value

fiskus: Why?

nl0: because the placeholder looks best at this width :)

fiskus: In this case I would prefer code clarity over UI look

catalog/app/utils/PackageUri.spec.js
nl0: i've added a test for both hash and tag and fixed that error

catalog/app/utils/PackageUri.spec.js
nl0: URI "hash" and package "hash" are two completely different things, and the format of the URI has been discussed and decided upon in the relevant spec

catalog/app/utils/PackageUri.js
nl0: thanks, fixed

catalog/app/utils/PackageUri.spec.js
fiskus: I understand. But they use the same word and this is confusing. Different things - one word

PR #1936, opened 2020-11-26
Workflow is required always, even when no default workflow

fiskus:
Workflow can be optional for package creation. But UI element is mandatory and should have value always

Screenshot from 2020-11-24 12-37-18

### DIFF
--- a/catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
+++ b/catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
@@ -178,7 +178,7 @@ const useWorkflowInputStyles = M.makeStyles((t) => ({
   },
 }))
 
-export function WorkflowInput({ errors, input, meta, workflowsConfig }) {
+export function WorkflowInput({ input, meta, workflowsConfig, errors = {} }) {
   const classes = useWorkflowInputStyles()
 
   const disabled = meta.submitting || meta.submitSucceeded
@@ -191,7 +191,7 @@ export function WorkflowInput({ errors, input, meta, workflowsConfig }) {
       onChange={input.onChange}
       value={input.value}
       disabled={disabled}
-      error={errors[errorKey]}
+      error={errorKey ? (errors[errorKey] || errorKey) : undefined}
     />
   )
 }
--- a/catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
+++ b/catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
@@ -155,10 +155,11 @@ const useWorkflowInputStyles = M.makeStyles((t) => ({
   },
 }))
 
-export function WorkflowInput({ input, meta, workflowsConfig }) {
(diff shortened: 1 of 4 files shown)


fiskus:
@nl0 **Why `isRequired` removed**
UI data is not identical to config.yml. Config doesn't have "No workflow" option, UI has. UI always have some workflow - some workflow from config.yml or "No workflow". Thus UI always required workflow, and this property is not needed.

catalog/app/containers/Bucket/PackageDialog/SelectWorkflow.js
```diff
@@ -51,6 +62,7 @@ export default function SelectWorkflow({ className, disabled, items, onChange, v
         ))}
       </M.Select>
       <M.FormHelperText>
+        {error && <span className={classes.error}>{error}</span>}
```
nl0, line 65:
```suggestion
        {!!error && <span className={classes.error}>{error}</span>}
```

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -190,6 +191,7 @@ export function WorkflowInput({ input, meta, workflowsConfig }) {
       onChange={input.onChange}
       value={input.value}
       disabled={disabled}
+      error={errors[errorKey]}
```
nl0, line 194:
```suggestion
      error={errorKey ? (errors[errorKey] || errorKey) : undefined}
```

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -178,10 +178,11 @@ const useWorkflowInputStyles = M.makeStyles((t) => ({
   },
 }))
 
-export function WorkflowInput({ input, meta, workflowsConfig }) {
+export function WorkflowInput({ errors, input, meta, workflowsConfig }) {
```
nl0, line 181:
```suggestion
export function WorkflowInput({ input, meta, workflowsConfig, errors = {} }) {
```

nl0:
> **Why `isRequired` removed**
> UI data is not identical to config.yml. Config doesn't have "No workflow" option, UI has. UI always has some workflow - some workflow from config.yml or "No workflow". Thus UI always requires workflow, and this property is not needed.

ok that makes sense

nl0 (approved):

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
fiskus: `errors[errorKey] || errorKey` seems enough, as `{}[undefined]` will be `undefined`

fiskus: But this works for me as well

nl0: > `{}[undefined]` will be `undefined`

well, in theory there may be something under `"undefined"` key (`{ undefined: 1 }[undefined] === 1`)

PR #1937, opened 2020-11-26
Workflows config unit tests

fiskus:
Copied from #1922 and restructured tests

I was forced to move workflows into `app/utils` because `jest` can't handle `requests.js` file

### DIFF
--- a/catalog/app/containers/Bucket/requests.js
+++ b/catalog/app/containers/Bucket/requests.js
@@ -10,62 +10,9 @@ import * as Resource from 'utils/Resource'
 import pipeThru from 'utils/pipeThru'
 import * as s3paths from 'utils/s3paths'
 import tagged from 'utils/tagged'
-import yaml from 'utils/yaml'
+import * as workflows from 'utils/workflows'
 
 import * as errors from './errors'
-
-function parseSchema(schemaSlug, schemas) {
-  return {
-    url: R.path([schemaSlug, 'url'], schemas),
-  }
-}
-
-function parseWorkflow(workflowSlug, workflow, data) {
-  return {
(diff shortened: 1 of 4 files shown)


catalog/app/containers/Bucket/requests.js
```diff
@@ -340,18 +287,18 @@ const WORKFLOWS_CONFIG_PATH = '.quilt/workflows/config.yml'
 export const workflowsList = async ({ s3, bucket }) => {
   try {
     const response = await fetchFile({ s3, bucket, path: WORKFLOWS_CONFIG_PATH })
-    return parseWorkflows(response.Body.toString('utf-8'))
+    return workflows.parse(response.Body.toString('utf-8'))
   } catch (e) {
     if (e instanceof errors.FileNotFound || e instanceof errors.VersionNotFound)
-      return emptyWorkflowsConfig
+      return workflows.emptyConfig
```
nl0, line 293: this return doesnt seem necessary, bc we already have the same statement at the end of the function

nl0 (approved):

catalog/app/containers/Bucket/requests.js
fiskus: It's an "early exit":
If error is `FileNotFound` or `VersionNotFound`, then early exit with `emptyConfig`. Otherwise, **log exception and** exit with `emptyConfig`

fiskus:
I suppose, Jest throws an error when importing `requests` because Quilt frontend imports modules like "react-form/**es**" or "connected-react-router/**esm**", and `jest` don't use Babel for node_modules and can't parse `import`

PR #1938, opened 2020-11-26
Hash-based package routes

nl0:
# Description

Use top_hash in package routes, completely remove timestamps from the UI (timestamp revisions are still resolved, so it's backwards compatible).

# TODO

### DIFF
--- a/catalog/app/components/SearchResults/SearchResults.js
+++ b/catalog/app/components/SearchResults/SearchResults.js
@@ -109,7 +109,7 @@ function ObjectHeader({ handle, showBucket, downloadable }) {
   )
 }
 
-function PackageHeader({ bucket, handle, revision, showBucket }) {
+function PackageHeader({ bucket, handle, hash, showBucket }) {
   const { urls } = NamedRoutes.use()
   return (
     <Heading mb={1}>
@@ -126,16 +126,12 @@ function PackageHeader({ bucket, handle, revision, showBucket }) {
             &nbsp;/{' '}
           </>
         )}
-        <CrumbLink to={urls.bucketPackageTree(bucket, handle, revision)}>
+        <CrumbLink to={urls.bucketPackageTree(bucket, handle, hash)}>
           {handle}
-          {revision !== 'latest' && (
-            <>
-              <M.Box component="span" color="text.hint">
-                @
-              </M.Box>
-              {revision}
-            </>
-          )}
+          <M.Box component="span" color="text.hint">
+            @
+          </M.Box>
+          {R.take(10, hash)}
         </CrumbLink>
       </span>
       <M.Box flexGrow={1} />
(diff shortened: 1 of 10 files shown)


catalog/app/utils/search.js
```diff
@@ -181,7 +179,9 @@ export default async function search({
         throw new SearchError('Timeout')
       }
     }
+    // eslint-disable-next-line no-console
```
fiskus, line 182: Please, review https://github.com/quiltdata/quilt/pull/1882

catalog/app/components/SearchResults/SearchResults.js
```diff
@@ -126,16 +126,12 @@ function PackageHeader({ bucket, handle, revision, showBucket }) {
...
-              <M.Box component="span" color="text.hint">
-                @
-              </M.Box>
-              {revision}
-            </>
-          )}
+          <M.Box component="span" color="text.hint">
+            @
+          </M.Box>
+          {R.take(10, hash)}
```
fiskus, line 134: Such things as `10` over the code are more maintainable as constants in one place. Much better if we would have some function that responsible for `hash => R.take(10, hash)`. In such a case, we would have a single implementation, that could be changed in one place.

fiskus (commented):
I will do manual tests tomorrow

catalog/app/utils/search.js
nl0: i will

fiskus:
Old URL works /b/fiskus-sandbox-dev/packages/fiskus/2011-11-16/tree/1605546363/
I don't know if this technically doable, we can make 301 redirect to new URL b/fiskus-sandbox-dev/packages/fiskus/2011-11-16/tree/67c766d5108b4dc2a7c3b74c8fd9287d880829158603de77bebf73288a7fdc81/

Also, I recall that we talked about short hashes (a la GitHub https://github.com/quiltdata/quilt/pull/1938/commits/d5a76a3), but I don't remember what we decide

nl0:
> I don't know if this technically doable, we can make 301 redirect to new URL

not possible until we implement server-side rendering, tho we can redirect to the more "canonical" hash-based route on the front-end.

@akarve do we want to redirect to the new routes?

> Also, I recall that we talked about short hashes (a la GitHub d5a76a3), but I don't remember what we decide

this will require backend support, so i think we should add this feature later (along with tags support)

fiskus (approved):
Didn't spot any bugs, except for inconsistency with old timestamps (not a bug, intentional)

Screenshot from 2020-11-27 09-51-44

docs/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 ## CLI
 
 ## Catalog, Lambdas
+* [Changed] `top_hash`-based package routes (backwards compatible) ([#1938](https://github.com/quiltdata/quilt/pull/1938))
```
akarve, line 16: This is reasonable behavior but I'm wondering if forwarding to a dedicated `/h/` or whatever hash route is more correct in terms of traceability? Please make a note that /timestamp URLs are not forwarded in any way, but that the user remains at /timestamp.

akarve (approved):
Mostly works in testing. See a few notes in Slack.

akarve (changes requested):
See potential bugs in Slack?

docs/CHANGELOG.md
akarve: To be clear I am totally cool with current behavior (honor both types of routes, no redirect). I am just asking if anything becomes cleaner or safer with two routes.

akarve (approved):

PR #1942, opened 2020-11-30
Remove infinite spinner after failed form submit

fiskus:
`final-form` calls internal `state.blur()` method when `onBlur`, and make field non-`active`. I didn't dig so deep to understand how `active` relates to `meta.validating` property update.

Rough fix for that: don't call `state.blur()` immediately, wait for validation's end


fiskus:
It "blures" on every change…

PR #1943, opened 2020-11-30
Decrease size of Package in-memory representation

sir-sigurd:
# Description
Bechmarked with this:

```python
from sys import getsizeof, stderr
from itertools import chain
from collections import deque
import reprlib
import operator

import quilt3


def total_size(o, handlers=None, f=stderr, verbose=False):
    """ Returns the approximate memory footprint an object and all of its contents.

    Automatically finds the contents of the following builtin containers and
    their subclasses:  tuple, list, deque, dict, set and frozenset.
    To search other containers, add handlers to iterate over their contents:

        handlers = {SomeContainerClass: iter,
                    OtherContainerClass: OtherContainerClass.get_elements}
    """
    dict_handler = lambda d: chain(d, d.items())
    all_handlers = {
        tuple: iter,
        list: iter,
        deque: iter,
        dict: dict_handler,
        set: iter,
        frozenset: iter,
        **(handlers or {}),
    }
    # track which object id's have already been seen
    seen = {}               
    # estimate sizeof object without __sizeof__
    default_size = getsizeof(None)       

    def sizeof(o):
        _id = id(o)
        if _id in seen:       # do not double count the same object
            return 0
        seen[_id] = o
        s = getsizeof(o, default_size)

        if verbose:
            print(s, type(o), reprlib.repr(o), file=f)

        for typ, handler in all_handlers.items():
            if isinstance(o, typ):
                s += sum(map(sizeof, handler(o)))
                break
        return s

    return sizeof(o), seen
    
    
total_size(
    quilt3.Package.browse(
        'aics/pipeline_integrated_single_cell', 
        registry='s3://allencell', 
        top_hash='7fd488f05ec41968607c7263cb13b3e70812972a24e832ef6f72195bdd35f1b2'
    ), 
    handlers={
        quilt3.Package: lambda p: (p.__dict__,), 
        quilt3.packages.PackageEntry: operator.attrgetter(*quilt3.packages.PackageEntry.__slots__),
        quilt3.util.PhysicalKey: operator.attrgetter(*quilt3.util.PhysicalKey.__slots__),
    }
)[0]
```

Before:
```
1066639992
```

After:
```
834581989
```

# TODO

- [x] <del>Unit tests</del>
- [x] <del>Documentation</del>
    - [x] <del>[Run `build.py`](../gendocs/build.py) for new docstrings</del>
- [x] Add code comments

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1,4 +1,5 @@
 import contextlib
+import functools
 import gc
 import hashlib
 import inspect
@@ -355,6 +356,26 @@ class PackageRevInfo:
         self.registry = registry
         self.name = name
         self.top_hash = top_hash
+
+
+class ManifestJSONDecoder(json.JSONDecoder):
+    """
+    Standard json.JSONDecoder reuses same `str` objects for JSON properties, while doing
+    a single `decode()` call.
+    This class also reuses `str` between many `decode()`s.
+    """
+    def __init__(self, *args, **kwargs):
+        @functools.lru_cache(maxsize=None)
+        def memoize_key(s):
+            return s
+
+        def object_pairs_hook(items):
+            return {
+                memoize_key(k): v
+                for k, v in items
+            }
+
+        super().__init__(*args, object_pairs_hook=object_pairs_hook, **kwargs)
 
 
 class Package:
@@ -797,7 +818,7 @@ class Package:
                     disable=DISABLE_TQDM,
                     bar_format='{l_bar}{bar}| {n}/{total} [{elapsed}<{remaining}, {rate_fmt}]',
                 ),
-                loads=json.loads,
+                loads=ManifestJSONDecoder().decode,
             )
             meta = reader.read()
             meta.pop('top_hash', None)  # Obsolete as of PR #130
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,7 @@
 # unreleased - YYYY-MM-DD
 ## Python API
 * [Added] Size of each manifest record is now limited by 1 MB. This constraint is added to ensure that S3 select, Athena and downstream services work correctly. This limit can be overridden with QUILT_MANIFEST_MAX_RECORD_SIZE environment variable. ([#2114](https://github.com/quiltdata/quilt/pull/2114))
+* [Changed] Decrease size of `Package` in-memory representation ([#1943](https://github.com/quiltdata/quilt/pull/1943))
 
 ## CLI
 


api/python/quilt3/packages.py
```diff
@@ -340,6 +341,26 @@ def __init__(self, registry, name, top_hash):
         self.top_hash = top_hash
 
 
+class ManifestJSONDecoder(json.JSONDecoder):
+    """
+    Standard json.JSONDecoder reuses same `str` objects for JSON properties, while doing
+    a single `decode()` call.
```
akarve, line 347: So you mean the standard decoder generates a lot of garbage to be collected? I would say that's *not* reusing str objects, but creating fresh ones.

api/python/quilt3/packages.py
```diff
@@ -340,6 +341,26 @@ def __init__(self, registry, name, top_hash):
...
+    This class also reuses `str` between many `decode()`s.
+    """
+    def __init__(self, *args, **kwargs):
+        @functools.lru_cache(maxsize=None)
```
akarve, line 351: "If maxsize is set to None, the LRU feature is disabled and the cache can grow without bound."—don't we want to bound this ourselves?

akarve: Can you also add a comment as to how exactly this is saving memory? I must not be understanding something, because reading of object_pairs_hook is that this will only save us memory when the same raw object is decoded twice, but that should rarely happen?

sir-sigurd:
```python
In [48]: def test(decoder):
    ...:     data = (
    ...:         '{"key": "value"}',
    ...:         '{"key": "value"}',
    ...:         '{"key": "other value"}',
    ...:     )
    ...:     str_objects = {}
    ...:     for obj in jsonlines.Reader(data, loads=decoder.decode):
    ...:         for item in obj.items():
    ...:             for s in item:
    ...:                 str_objects[id(s)] = s
    ...:     return str_objects
    ...: 

In [49]: test(json.JSONDecoder())
Out[49]: 
{140254112745488: 'key',
 140254112744760: 'value',
 140254114859584: 'key',
 140254114858912: 'value',
 140254114860256: 'key',
 140254113566256: 'other value'}

In [50]: test(ManifestJSONDecoder())
Out[50]: 
{140254113989160: 'key',
 140254113988936: 'value',
 140254113991960: 'value',
 140254105343920: 'other value'}
```
We always have `type` and `value` keys for hash objects that end up in `Package` instance. Also the package from the PR description uses uniform schema for meta of all objects, i.e. there are a lot of keys that could be reused.

nl0 (approved):

PR #1945, opened 2020-11-30
fix non-int indexing of time-lapse data

donovanr:


JacksonMaxfield:
HUGE

akarve:
@donovanr thank you. @JacksonMaxfield can you please approve this PR? can you add a quick unit tests for this? it clearly doesn't break any of the existing tests.

JacksonMaxfield (approved):

JacksonMaxfield:
> @donovanr thank you. @JacksonMaxfield can you please approve this PR? can you add a quick unit tests for this? it clearly doesn't break any of the existing tests.

I personally don't think there are tests needed but your call. Other option is for me to handle adding tests later when I update this to use `aicsimageio==4.0.0`

akarve (changes requested):
Thanks Rory. I think we can take this without unit tests as, if I understand correctly, a non-integer index doesn't even make sense in this context?

Just one additional thing we need, please add to this PR a Changelog entry [here](https://github.com/quiltdata/quilt/blob/master/docs/CHANGELOG.md#catalog-lambdas).

donovanr:
Yeah, the current code would throw an `IndexError` if you ask for the middle time slice of three time points and give it 3/2 = 1.5; this should be a solid upgrade.

Tests for this might be a bit above my paygrade, I didn't see anything in the `lambdas/preview/test/data/` directory that I could use.  It seems like it might be best to merge the PR and then create an broader issue around tests for the thumbnail code.

donovanr:
Should I add the PR to the Changelog before or after it's merged?

akarve:
You can add it before it's merged. i.e. all new PRs have a changelog entry referring to themselves, which is weird but how we do it :)

donovanr:
cool.  i added it to this fork.

akarve (approved):

PR #1947, opened 2020-12-02
Set column types when reading json to pandas

kevinemoore:
# Description

Avoid errors if e.g., logical_keys are numbers when reading them into pandas for pkgselect.


sir-sigurd (changes requested):
The fix seems to be correct, but tests are needed.

lambdas/pkgselect/index.py
```diff
@@ -195,7 +195,14 @@ def lambda_handler(request):
             sql_stmt=sql_stmt
         )
         # Parse the response into a logical folder view
-        df = pd.read_json(result, lines=True)
+        df = pd.read_json(
+            result,
+            lines=True,
+            dtype=dict(
+                logical_key='string',
+                physical_keys='string'
```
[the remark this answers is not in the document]
akarve: Mmm, `physical_keys` is correct. e.g.:

```
{"logical_key": "README.md", "physical_keys": ["s3://quilt-example/akarve/many-revisions/README.md?versionId=lIorkhsOKA0sq.WfMoIjoIByiygJCzfb"], "size": 85, "hash": {"type": "SHA256", "value": "287030c6fad4d43c4d519c0fce9595f472819c2d4b7f2aaeb7b860ea075a8d67"}, "meta": {}}
```

akarve (changes requested):
Please add a [Changelog entry](https://github.com/quiltdata/quilt/blob/master/docs/CHANGELOG.md)
and test case with whatever strings caused this to fail.

lambdas/pkgselect/test/test_pkgselect.py
```diff
@@ -393,3 +398,72 @@ def test_anon_access(self):
...
+        mock_s3 = boto3.client('s3')
+        client_patch = patch.object(
+            mock_s3,
+            'select_object_content',
+            side_effect=[
+                non_string_s3response,
+                self.s3response_meta
+            ]
+        )
+        client_patch.start()
```
sir-sigurd, line 457: Either use `client_patch` as context manager or use `self.addCleanup(client_patch.stop)`, otherwise `client_patch.stop()` won't be called if some of asserts fail and this could make other tests fail (or maybe even give a false positive?).

sir-sigurd (changes requested):

lambdas/pkgselect/test/test_pkgselect.py
```diff
@@ -21,42 +21,47 @@ class TestPackageSelect(TestCase):
...
-                            'Details': {
-                                'BytesScanned': 123,
-                                'BytesProcessed': 123,
-                                'BytesReturned': 123
-                            }
+        return {
+            'Payload': [
+                {
+                    'Records': {
+                        'Payload': bytes
```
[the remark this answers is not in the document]
kevinemoore: Good call.

lambdas/pkgselect/test/test_pkgselect.py
kevinemoore: Good catch.

lambdas/pkgselect/index.py
```diff
@@ -194,9 +195,20 @@ def lambda_handler(request):
...
+        if result is None:
+            response_data = file_list_to_folder(pd.DataFrame())
+        else:
+            # Parse the response into a logical folder view
+            df = pd.read_json(
+                result,
+                lines=True,
+                dtype=dict(
+                    logical_key='string',
+                    physical_key='string'
```
akarve, line 208: In an actual manifest it's `physical_keys`.

sir-sigurd: Isn't this column name in the result set?

kevinemoore: Yes, @sir-sigurd, the column name refers to the DataFrame (result set) and not the manifest.

akarve: Find but why are we changing column names?

docs/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 * [Changed] Preview `*notes` files as plain text ([#1896](https://github.com/quiltdata/quilt/pull/1896))
 * [Changed] Default search operator to "AND" (was "OR") for more precise searches ([#1924](https://github.com/quiltdata/quilt/pull/1924))
 * [Changed] `top_hash`-based package routes (timestamp routes are still supported in the same way) ([#1938](https://github.com/quiltdata/quilt/pull/1938))
+* [Fixed] Empty response from `pkgselect` folder view for packages with all non-string logical_keys or physical_keys.
```
sir-sigurd, line 38: This should be moved to unreleased section in the beginning of the file.

lambdas/pkgselect/test/test_pkgselect.py
sir-sigurd: It looks like you did it in the old test, but not in the new one.

sir-sigurd (changes requested):
Still some minor changes has to be made. Also `catalog/package-lock.json` should be reverted.

sir-sigurd (approved):

PR #1950, opened 2020-12-02
Bucket/PackageDialog: metadata dnd

nl0:
# TODO

### DIFF
--- a/catalog/app/containers/Bucket/PackageCreateDialog.js
+++ b/catalog/app/containers/Bucket/PackageCreateDialog.js
@@ -473,9 +473,7 @@ function PackageCreateDialog({ bucket, open, workflowsConfig, onClose, refresh }
           scroll="body"
           onExited={reset(form)}
         >
-          <M.DialogTitle>
-            {success ? 'Package created' : 'Create package'}
-          </M.DialogTitle>
+          <M.DialogTitle>{success ? 'Package created' : 'Create package'}</M.DialogTitle>
           {success ? (
             <>
               <M.DialogContent style={{ paddingTop: 0 }}>
@@ -564,6 +562,7 @@ function PackageCreateDialog({ bucket, open, workflowsConfig, onClose, refresh }
                           validate={validate}
                           validateFields={['meta']}
                           isEqual={R.equals}
+                          initialValue={PD.EMPTY_META_VALUE}
                         />
                       ),
                       _: () => <PD.MetaInputSkeleton />,
--- a/catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
+++ b/catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
@@ -1,16 +1,20 @@
 import * as R from 'ramda'
 import * as React from 'react'
+import { useDropzone } from 'react-dropzone'
 import * as M from '@material-ui/core'
 import * as Lab from '@material-ui/lab'
 
 import JsonEditor from 'components/JsonEditor'
 import { parseJSON, stringifyJSON, validateOnSchema } from 'components/JsonEditor/State'
 import Skeleton from 'components/Skeleton'
+import * as Notifications from 'containers/Notifications'
 import { useData } from 'utils/Data'
+import Delay from 'utils/Delay'
 import AsyncResult from 'utils/AsyncResult'
 import * as APIConnector from 'utils/APIConnector'
 import * as AWS from 'utils/AWS'
 import pipeThru from 'utils/pipeThru'
+import { readableBytes } from 'utils/string'
 import * as validators from 'utils/validators'
 import * as workflows from 'utils/workflows'
 
@@ -19,6 +23,7 @@ import SelectWorkflow from './SelectWorkflow'
 
 export const MAX_SIZE = 1000 * 1000 * 1000 // 1GB
 export const ES_LAG = 3 * 1000
+export const MAX_META_FILE_SIZE = 10 * 1000 * 1000 // 10MB
 
 export const ERROR_MESSAGES = {
   UPLOAD: 'Error uploading files',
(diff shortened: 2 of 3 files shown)


catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -104,16 +124,18 @@ function mkMetaValidator(schema) {
   return function validateMeta(value) {
     const noError = undefined
 
+    const jsonObjectErr = validators.jsonObject(value.text)
+    if (jsonObjectErr) {
+      return value.mode === 'json'
+        ? jsonObjectErr
+        : [{ message: 'Metadata must be a valid JSON object' }]
```
fiskus, line 131: There is `ERROR_MESSAGES` at the top of the module. I think we can use it to store this string

fiskus (commented):
There is an exception in `validateMeta` on opening CreatePackageDialog

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -268,6 +316,51 @@ export function MetaInput({ schemaError, input, meta, schema }) {
...
+        )
+        return
+      }
+      setLocked(true)
+      readFile(file)
+        .then((contents) => {
+          try {
+            JSON.parse(contents)
+          } catch (e) {
+            notify('The file does not contain valid JSON')
```
fiskus, line 340: Notification is almost invisible. I found it because I seek for it, after reading code.

Is it a temporary solution and do you want to show error inside Form?

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -268,6 +316,51 @@ export function MetaInput({ schemaError, input, meta, schema }) {
...
+          }
+          changeText(contents)
+          // force json editor to re-initialize
+          setJsonEditorKey(R.inc)
```
fiskus, line 344: Consider making a hook for this pattern. 

`const [cacheKey, updateCacheKey] = useReactCache()`

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -287,35 +380,62 @@ export function MetaInput({ schemaError, input, meta, schema }) {
...
-            }[error]
-          }
-          fullWidth
-          multiline
-          rowsMax={10}
-          InputProps={{ classes: { input: classes.jsonInput } }}
-          disabled={disabled}
-        />
-      )}
+      <div {...getRootProps({ className: classes.dropzone })} tabIndex={undefined}>
```
fiskus, line 382: `tabIndex={undefined}`

Don't understand this

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -287,35 +380,62 @@ export function MetaInput({ schemaError, input, meta, schema }) {
...
+            }
+            fullWidth
+            multiline
+            rowsMax={10}
+            InputProps={{ classes: { input: classes.jsonInput } }}
+            disabled={disabled}
+          />
+        )}
+
+        {(isDragActive || locked) && (
```
fiskus, line 415: For me, it's simpler to read when splitting this into two different blocks:

```
{isDragActive && (
  <div overlay>
    Drop file
  </div>
)}

{(locked && !isDragActive) && (
  <CircularProgress />
)}
```

---

Also, mixing two styles:

* isDragActive, isLocked
* dragActive, locked

nl0:
> There is an exception in validateMeta on opening CreatePackageDialog

fixed, pls take another look when you can

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -473,9 +473,7 @@ function PackageCreateDialog({ bucket, open, workflowsConfig, onClose, refresh }
           scroll="body"
           onExited={reset(form)}
         >
-          <M.DialogTitle>
-            {success ? 'Package created' : 'Create package'}
-          </M.DialogTitle>
+          <M.DialogTitle>{success ? 'Package created' : 'Create package'}</M.DialogTitle>
```
fiskus, line 476: I don't understand why my ESLint skipped this

fiskus (approved):

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: i'm not sure there's a lot to gain from `incKey()` vs `setKey(R.inc)`

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: > Notification is almost invisible. I found it because I seek for it, after reading code.

well, it's a default MUI snackbar-style notification, we use those across the app to tell the user something has happened

> Is it a temporary solution and do you want to show error inside Form?

it _is_ shown as a regular field error (tho only after submit, like the rest of the field errors)

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: by default `getRootProps()` has tabIndex=0 and that makes the dropzone focusable with an ugly blue outline, so i unset it

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: > For me, it's simpler to read when splitting this into two different blocks:

this way we have to repeat the wrapper part

> Also, mixing two styles

`isDragActive` comes from the external API, so i have no control over it (well, ofc i can rename it, but i dont think that's necessary)

catalog/app/containers/Bucket/PackageCreateDialog.js
nl0: maybe bc it was a suggestion merged via github web ui

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: i thought that constant is for reusable messages related to server responses mostly

nl0:
@akarve pls wordsmith when you got a minute

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
fiskus: * to not spend time on what is already implemented
* to have business logic and internal technical stuff (dealing with React keys) divided
* to be ready for changes in one place when the environment will change (major React update, migrating to different UI library)

PR #1959, opened 2020-12-04
fixes for workflows doc

sir-sigurd:


docs/Advanced Features/workflows.md
```diff
@@ -68,9 +68,9 @@ Now let's push with `workflow='beta'`:
 QuiltException: Metadata failed validation: 'superhero' is a required property.
 ```
 
-We encounterd another exception because the `beta` workflows specifies
+We encountered another exception because the `beta` workflow specifies
 `metadata_schema: superheroes`.
-Therefore the `test/package` metadata must validated against the
+Therefore, the `test/package` metadata must be validated against the
```
nl0, line 73: @akarve "must validate" or "must be validated"? which one sounds better?

nl0 (approved):

docs/Advanced Features/workflows.md
akarve, line 73:
```suggestion
Therefore, the `test/package` metadata must validate against the
```

akarve (approved):

docs/Advanced Features/workflows.md
```diff
@@ -143,7 +143,7 @@ and specify `workflow=None` in the API:
 Package test/package@06b2815 pushed to s3://quilt-sergey-dev-metadata
 ```
 
-Also `default_workflow` can be set in config to specify which workflow will be used
+Also `default_workflow` can be set in the config to specify which workflow will be used
```
akarve, line 146:
```suggestion
You can also specify a `default_workflow` in `config.yml`.
```

sir-sigurd: I left this up to you, but generally I'd like to be less specific here (we already specified path in the beginning) and do not mention `config.yml` again since we could change the path in the future or maybe even move this configuration out of S3.

PR #1965, opened 2020-12-04
Lint: inline-ignore `no-console` rule

fiskus:


nl0 (approved):
looks good. might as well set this rule to throw errors, bc warnings are ignored on pre-commit

PR #1966, opened 2020-12-04
Lint: resolve ESLint's `react/static-property-placement` rule

fiskus:


nl0 (changes requested):
i think instead we should set this rule config to [static public field](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/static-property-placement.md#when-static-public-field-is-enabled-default)

fiskus:
The point of using public ESLint config is to avoid "bike shedding".
If you prefer "static public field" I can argue that I prefer "property assignment" and vice versa.
If we both prefer "static public field" new developer will argue that he prefer "property assignment".

But if we use airbnb config, we can't argue, it's third-party rule.

nl0:
@fiskus 
> But if we use airbnb config, we can't argue, it's third-party rule.

i dont think so. we use it as a base for our config, and we do override some rules already, and i think we should override this one as well, bc static fields look much cleaner than property assignments

fiskus:
Closed in favour of #1980

PR #1968, opened 2020-12-07
Return initial state on useData call with new params 

fiskus:
`useData` has internal state management. When you pass new params into `useData`
1. state is preserved for a moment
2. returns the previous state
3. then will change according to new params and returns a new result

This leads to the incorrect result, reproduced at #1953

---

Fix contains a workaround: `useData` returns initial state on passing new params

### DIFF
--- a/catalog/app/utils/Data.js
+++ b/catalog/app/utils/Data.js
@@ -5,6 +5,7 @@ import AsyncResult from 'utils/AsyncResult'
 import tagged from 'utils/tagged'
 import useMemoEq from 'utils/useMemoEq'
 import usePrevious from 'utils/usePrevious'
+import usePreviousSync from 'utils/usePreviousSync'
 
 const Action = tagged(['Reset', 'Request', 'Response'])
 
@@ -27,6 +28,7 @@ const reducer = Action.reducer({
 })
 
 export function useData(request, params, { noAutoFetch = false } = {}) {
+  let stateResetingIsInProgress = false
   // TODO: accept custom key extraction fn (params => key for comparison)
   const [state, dispatch] = React.useReducer(reducer, initial)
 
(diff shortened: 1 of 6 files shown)


catalog/app/utils/usePrevious.js
```diff
@@ -1,5 +1,9 @@
 import * as React from 'react'
 
+/**
+ * Fire onChange callback with previous value and memoize current value
+ * TODO: explain why `useEffect`
```
nl0, line 5: onChange callback is fired in useEffect so that one can execute effectful code there. for sync usage it should be just `const prev = usePrevious(...)`.
tho in #1973 i've fixed `usePrevious` to update the stored value synchronously based on this use case (when useEffect doesnt get called due to re-renders caused by setState)

catalog/app/utils/Data.js
```diff
@@ -47,7 +49,13 @@ export function useData(request, params, { noAutoFetch = false } = {}) {
...
   })
 
-  const result = useMemoEq(state, mapResult)
+  usePreviousSync(params, (prev) => {
+    if (!R.equals(params, prev)) {
+      stateResetingIsInProgress = true
+    }
+  })
+
+  const result = useMemoEq(stateResetingIsInProgress ? initial : state, mapResult)
```
nl0, line 58: this is not quite correct, bc if auto-fetch is enabled we should transition directly to `Pending` (skipping `Init`)

catalog/app/utils/usePreviousSync.spec.js
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react'
+import { act, renderHook } from '@testing-library/react-hooks'
+
+import usePreviousSync from './usePreviousSync'
+
+describe('utils/usePreviousSync', () => {
+  it('should return the same value', () => {
```
nl0, line 7: fyi the point of the original hook was to return the previous value, not the same one

PR #1969, opened 2020-12-07
Update revisions after package update when navigating to pushed revision

nl0:


catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -928,12 +928,12 @@ export function usePackageUpdateDialog({ bucket, name, hash, onExited }) {
   const handleExited = React.useCallback(() => {
     setExited(true)
     setSuccess(false)
-    setExitValue(null)
```
fiskus, line 931: So, the primary change is moving this line to the bottom of the function, right?

And, why do change state to ref? I don't mind using ref here, but maybe I'm missing something

fiskus: Also, `exitValue` seems redundant, because it's calculated from `success`:
```
exitValue = success => ({ pushed: success })
```
You can remove it and call `onExited({ pushed: success })`. This way, you'll reduce mutated state data for component

fiskus (approved):
* I don't understand using `ref` over `state`
* There is room for decreasing state data

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: > the primary change is moving this line to the bottom of the function, right?

nope, the primary change is switching to `useRef`

> why did you change state to ref?

bc when the dialog gets unmounted (see DialogWrapper code), the cleanup function gets called with outdated state captured by the closure, so in order for it to access updated state we need to use a mutable ref instead of state

> Also, exitValue seems redundant, because it's calculated from success

it's not redundant, bc `success` is reset on close (and computed state is based on this), and we need to persist `exitValue` up until `onExited` is called

fiskus: > `success` is reset on close

I can find `setSuccess(false)` inside `onExit`. `close` callback doesn't have `setSuccess(false)`

nl0: you're right, seems like i've misread my own code or had some false memory)
fixed now, thanks

fiskus (approved):

PR #1970, opened 2020-12-07
.gitignore: add node_modules and ipynb-styling subrepos for eslint

nl0:
ESLint reuses .gitignore, so we need to add stuff from ipynb-styling there for ESLint to ignore it, bc it hangs up trying to lint some built artefacts

### DIFF
--- a/catalog/.gitignore
+++ b/catalog/.gitignore
@@ -1,3 +1,7 @@
+node_modules
+ipynb-styling/notebook
+ipynb-styling/pygments-css
+
 # Don't check auto-generated stuff into git
 coverage
 build


catalog/.gitignore
```diff
@@ -1,3 +1,7 @@
+node_modules
+ipynb-styling/notebook
```
fiskus, line 2: I don't have this directory, is it your personal stuff?

nl0: these directories are created when building ipynb styles (see readme and makefile in ipynb-styling)

fiskus: :ok_hand:

fiskus (approved):

PR #1973, opened 2020-12-08
Fix data hook

nl0:
Always use up-to-date state for rendering, fixes #1953, supersedes #1968

### DIFF
--- a/catalog/app/utils/Data.js
+++ b/catalog/app/utils/Data.js
@@ -28,7 +28,13 @@ const reducer = Action.reducer({
 
 export function useData(request, params, { noAutoFetch = false } = {}) {
   // TODO: accept custom key extraction fn (params => key for comparison)
-  const [state, dispatch] = React.useReducer(reducer, initial)
+  const [state, setState] = React.useState(initial)
+  const stateRef = React.useRef()
+  stateRef.current = state
+
+  const dispatch = (action) => {
+    setState((stateRef.current = reducer(stateRef.current, action)))
+  }
 
   const fetch = useMemoEq([request, params], () => () => {
     dispatch(Action.Request({ request, params }))
@@ -41,13 +47,13 @@ export function useData(request, params, { noAutoFetch = false } = {}) {
       })
   })
 
-  usePrevious({ params, noAutoFetch }, (prev) => {
-    if (R.equals({ params, noAutoFetch }, prev)) return
+  const prev = usePrevious({ params, noAutoFetch })
+  if (!R.equals({ params, noAutoFetch }, prev)) {
     if (noAutoFetch) dispatch(Action.Reset())
     else fetch()
-  })
+  }
 
-  const result = useMemoEq(state, mapResult)
+  const result = useMemoEq(stateRef.current, mapResult)
 
   const doCase = useMemoEq([result], () => (cases, ...args) =>
     AsyncResult.case(cases, result, ...args),
--- a/catalog/app/utils/usePrevious.js
+++ b/catalog/app/utils/usePrevious.js
@@ -1,10 +1,11 @@
 import * as React from 'react'
 
-export default (value, onChange) => {
+export default function usePrevious(value, onChange) {
   const ref = React.useRef()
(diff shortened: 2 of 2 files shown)


catalog/app/utils/Data.js
```diff
@@ -28,7 +28,13 @@ const reducer = Action.reducer({
 
 export function useData(request, params, { noAutoFetch = false } = {}) {
   // TODO: accept custom key extraction fn (params => key for comparison)
-  const [state, dispatch] = React.useReducer(reducer, initial)
+  const [state, setState] = React.useState(initial)
+  const stateRef = React.useRef()
+  stateRef.current = state
```
fiskus, line 33: I strongly recommend reduce or even remove state from `useData`, not add more

catalog/app/utils/Data.js
```diff
@@ -41,13 +47,13 @@ export function useData(request, params, { noAutoFetch = false } = {}) {
       })
   })
 
-  usePrevious({ params, noAutoFetch }, (prev) => {
-    if (R.equals({ params, noAutoFetch }, prev)) return
+  const prev = usePrevious({ params, noAutoFetch })
```
fiskus, line 50: `prev` is implementation. Please, use more specific name

catalog/app/utils/Data.js
```diff
@@ -28,7 +28,13 @@ const reducer = Action.reducer({
 
 export function useData(request, params, { noAutoFetch = false } = {}) {
   // TODO: accept custom key extraction fn (params => key for comparison)
-  const [state, dispatch] = React.useReducer(reducer, initial)
+  const [state, setState] = React.useState(initial)
+  const stateRef = React.useRef()
+  stateRef.current = state
+
+  const dispatch = (action) => {
+    setState((stateRef.current = reducer(stateRef.current, action)))
```
fiskus, line 36: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/best-practices.js#L266

One day ESLint will catch it as `error`

fiskus (approved):

fiskus:
The main source of bugs is sophisticated, hard to read code. Consider making code simpler if you catch a bug in it

catalog/app/utils/Data.js
nl0: but this is not a "return statement"

catalog/app/utils/Data.js
nl0: well, i would be glad to do that, but how? "make things as simple as possible, but not simpler". seems like the problem requires some level of complexity and i dont see a way to simplify it further, however, i'm open to suggestions

catalog/app/utils/usePrevious.js
```diff
@@ -1,10 +1,11 @@
 import * as React from 'react'
 
-export default (value, onChange) => {
+export default function usePrevious(value, onChange) {
```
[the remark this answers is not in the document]
nl0: why do you think "callback" is better in this context?

catalog/app/utils/Data.js
nl0: it means smth like "previous value of the arguments", so it's not "implementation", it's semantically correct

catalog/app/utils/usePrevious.js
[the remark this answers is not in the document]
fiskus: `onChange` is expected name for `<input />`
`usePrevious` returns previous value asynchronously - it's a natural pattern for nodejs environment. And usually `callback` used for that, so this name is expected (not quite, because callbacks returns error too)

Is your intention was to call function after _changing_ ref, and that's why you choose onChange?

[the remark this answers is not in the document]
nl0: it's not used to return previous value asynchronously -- actually, the value is _returned_ synchronously, but the onChange callback is meant to be called when the value (possibly) changes, so that the caller can compare previous and current values and run some (possibly effectful) code, so it's more like `onPossibleValueChange`

PR #1974, opened 2020-12-08
Limit JsonEditor height, small refactoring

fiskus:
Current behavior:
* JsonEditor renders all columns
* only the last one is in the viewport
* rest columns are hidden by CSS (overflow + scrolls, but scrollbars are invisible)

Changed behavior:
* JsonEditor renders current column only
* JsonEditor is limited by height (enough to fit the whole Dialog into 1920x1080)
* Vertical scrollbar is visible

### DIFF
--- a/catalog/app/components/JsonEditor/JsonEditor.js
+++ b/catalog/app/components/JsonEditor/JsonEditor.js
@@ -67,7 +67,7 @@ function JsonEditor({
   )
 
   const columnData = R.last(columns)
-  const columnPath = R.slice(0, columns.length - 1, fieldPath)
+  const columnPath = R.take(columns.length - 1, fieldPath)
 
   return (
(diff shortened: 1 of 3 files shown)


catalog/app/components/JsonEditor/JsonEditor.js
```diff
@@ -70,27 +66,25 @@ function JsonEditor({
     [changeValue, onChange],
   )
 
+  const columnData = R.last(columns)
+  const columnPath = R.slice(0, columns.length - 1, fieldPath)
```
nl0, line 70: fyi `R.slice(0, x)` is equivalent to `R.take(x)`

nl0 (approved):

catalog/app/components/JsonEditor/JsonEditor.js
fiskus: Thanks, I forget about it

PR #1977, opened 2020-12-10
Prevent possible downloading of truncated file

sir-sigurd:
# Description
If file is modified after we getting its size with `head_object`, the truncated file will be downloaded.

# TODO

- [x] Unit tests
- [x] <del>Documentation</del>
    - [x] <del>[Run `build.py`](../gendocs/build.py) for new docstrings</del>

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -808,7 +808,9 @@ def copy_file(src: PhysicalKey, dest: PhysicalKey, size=None, message=None, call
         if _looks_like_dir(dest):
             dest = dest.join(src.basename())
         if size is None:
-            size, _ = get_size_and_version(src)
+            size, version_id = get_size_and_version(src)
+            if src.version_id is None:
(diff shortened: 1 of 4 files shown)


kevinemoore (approved):
Seems like we should always try to get the version_id when we save an object's size, but that seems to be done by the caller of copy_file. This solution looks good to me.

akarve (approved):

dimaryaz (approved):

api/python/quilt3/data_transfer.py
```diff
@@ -806,7 +806,9 @@ def sanity_check(rel_path):
         if _looks_like_dir(dest):
             dest = dest.join(src.basename())
         if size is None:
-            size, _ = get_size_and_version(src)
+            size, version_id = get_size_and_version(src)
+            if src.version_id is None:
```
kevinemoore, line 810: Should that be `if version_id` instead of `if src.version_id`?

kevinemoore (approved):
Looks like a small typo.

PR #1979, opened 2020-12-10
Run CI jobs for js stuff / catalog

nl0:
- [x] test and build the catalog (eslint temporary disabled)
- [x] track bundle size (TODO @akarve [set up bundlewatch token](https://bundlewatch.io/#/getting-started/using-bundlewatch?id=setup-ci-auth-variables-needed-by-bundlewatch) for this repo)
- [x] codecov


sir-sigurd (approved):
Unfortunately it doesn't display workflow names, and job names are not very distinguishable. 
image

We might want to prefix job names with `js`/`py`, but I'm fine to leave it as is.

catalog/package.json
```diff
@@ -16,7 +16,7 @@
     "analyze:clean": "rimraf stats.json",
     "preanalyze": "npm run analyze:clean",
     "analyze": "cross-env NODE_ENV=production node ./internals/scripts/analyze.js",
-    "bundlesize": "bundlesize -f './build/*.js'",
+    "bundlewatch": "bundlewatch './build/*.js'",
```
fiskus, line 19: Does It use the default size limit? I can't find its configurating

fiskus: I see you comment for Aneesh

fiskus (approved):

nl0:
> Unfortunately it doesn't display workflow names, and job names are not very distinguishable.
> image
> 
> We might want to prefix job names with `js`/`py`, but I'm fine to leave it as is.

i was using `js-test` initially, but then renamed it to `test-catalog` to be consistent with the existing naming scheme). but i also dont mind prefixing things with `py-` / `js-`

catalog/package.json
nl0: it doesnt use any size limits, just tracking the changes (at least it should start doing so after we merge it)

PR #1981, opened 2020-12-11
Fix React key scope

fiskus:
Any bucket with images, for example, `/b/quilt-bio-staging`:

Screenshot from 2020-12-11 10-08-50


nl0 (approved):
good catch

PR #1982, opened 2020-12-11
Backend-specific subclasses of PhysicalKey

sir-sigurd:
# Description
This change is mostly for avoiding annoying `if pk.is_local()` and to ease addition of new backends when/if we decide to do it.

This is a POC, I'll finish it if this change is generally approved.

# TODO

- [ ] Unit tests
- [ ] Documentation
    - [ ] [Run `build.py`](../gendocs/build.py) for new docstrings

### DIFF
--- a/api/python/quilt3/backends/base.py
+++ b/api/python/quilt3/backends/base.py
@@ -2,13 +2,6 @@ import abc
 import operator
 import time
 
-from quilt3.data_transfer import (
-    copy_file,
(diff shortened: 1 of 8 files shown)


akarve (approved):
Important because it enable us to add providers for GCS, Azure, etc.

drernie:
Doesn't seem to have been finished; plan to handle via Python path lib.

PR #1980, opened 2020-12-11
Adjust ESLint config and fix errors

nl0:

PR #1984, opened 2020-12-11
Update js dependencies

nl0:


fiskus (approved):
As for #1980 I suggest merging it after open.quiltdata.com deploy.

Also, after merging, `prettier/prettier` rule will break on `catalog/app/utils/APIConnector.js`

nl0:
> Also, after merging, prettier/prettier rule will break on catalog/app/utils/APIConnector.js

yep, and CI should catch that

PR #1985, opened 2020-12-11
Normalize paths for bundlewatch

nl0:


nl0:
this is quite trivial, so ima merge it

PR #1988, opened 2020-12-14
Preseve formatting for dropMessage

fiskus:
Was:
Screenshot from 2020-12-14 14-41-46

Became:
Screenshot from 2020-12-14 14-42-00

### DIFF
--- a/catalog/app/containers/Bucket/PackageUpdateDialog.js
+++ b/catalog/app/containers/Bucket/PackageUpdateDialog.js
@@ -402,7 +402,7 @@ function FilesInput({ input, meta, uploads, setUploads, errors = {} }) {
               !error && warn && classes.dropMsgWarn,
             )}
           >
-            {label}
+            <M.Typography>{label}</M.Typography>
           </div>
         </div>
         {disabled && (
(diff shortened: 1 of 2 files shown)


catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -402,7 +401,7 @@ function FilesInput({ input, meta, uploads, setUploads, errors = {} }) {
               !error && warn && classes.dropMsgWarn,
             )}
           >
-            {label}
+            <M.Typography variant="body2">{label}</M.Typography>
```
nl0, line 404: couldnt we just use `span`?

nl0 (approved):

catalog/app/containers/Bucket/PackageUpdateDialog.js
fiskus: Yes. Do you prefer `<span>` here?

nl0: well, the change would be smaller, and it also seems a little simpler

PR #1992, opened 2020-12-17
Search results: handle folders

nl0:
Screenshot 2020-12-17 at 18 23 28

### DIFF
--- a/catalog/app/components/SearchResults/SearchResults.js
+++ b/catalog/app/components/SearchResults/SearchResults.js
@@ -33,6 +33,7 @@ const CrumbLink = M.styled(StyledLink)({ wordBreak: 'break-word' })
 
 function ObjectCrumbs({ handle, showBucket = false }) {
   const { urls } = NamedRoutes.use()
+  const isDir = handle.key.endsWith('/')
 
   const crumbs = React.useMemo(() => {
     const all = getBreadCrumbs(handle.key)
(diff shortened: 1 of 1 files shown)


catalog/app/components/SearchResults/SearchResults.js
```diff
@@ -492,6 +500,23 @@ function ObjectHit({ showBucket, hit: { path, versions, bucket } }) {
...
+    path,
+    versions: [v],
+    bucket,
+  },
+}) {
+  const handle = { bucket, key: path }
+  return (
+    <Section>
+      <ObjectHeader {...{ handle, showBucket }} />
+      <Meta meta={v.meta} />
```
fiskus, line 515: I don't see meta for "dir" search query inside quilt-dev-null. How can I find directories with meta?

If there is no meta, ObjectHeader's margin-bottom appends to Section's padding-bottom, and space above and under dir's title are not equal.

catalog/app/components/SearchResults/SearchResults.js
```diff
@@ -85,12 +88,12 @@ function HeaderIcon(props) {
   )
 }
 
-function ObjectHeader({ handle, showBucket, downloadable }) {
+function ObjectHeader({ handle, showBucket, downloadable = false }) {
   return (
     <Heading display="flex" alignItems="center" mb={1}>
```
fiskus, line 93: I think, you need `mb={1}` outside of `ObjectHeader`, and use `mb={0}` for directory without metadata.

fiskus (approved):

catalog/app/components/SearchResults/SearchResults.js
nl0: > How can I find directories with meta?

i'm afraid the only option is creating them)

> If there is no meta, ObjectHeader's margin-bottom appends to Section's padding-bottom, and space above and under dir's title are not equal.

yeah i see, i'll fix that, thx

nl0: fixed

PR #1997, opened 2021-01-12
Handle delete markers for objects in ES

nl0:
- catalog
  - bucket overview
    - filter-out deleted objects from bucket summary (images and sample)
    - fix object count query
  - search results
    - show 'Delete Marker' labels
- search lambda
  - expose `delete_marker` by default in `search` action
  - expose `delete_marker` in `images` and `sample` actions

caveat: images and sample files result sets are filtered from delete markers after querying / limiting, so when delete markers are encountered, the sets will be smaller than the respective limits (`NUM_PREVIEW_IMAGES = 100` and `NUM_PREVIEW_FILES = 20`)

### DIFF
--- a/catalog/app/containers/Bucket/Overview.js
+++ b/catalog/app/containers/Bucket/Overview.js
@@ -757,89 +757,82 @@ function Head({ req, s3, overviewUrl, bucket, description }) {
   const classes = useHeadStyles()
   const isRODA = !!overviewUrl && overviewUrl.includes(`/${RODA_BUCKET}/`)
   const colorPool = useConst(() => mkKeyedPool(COLOR_MAP))
+  const statsData = useData(requests.bucketStats, { req, s3, bucket, overviewUrl })
+  const pkgStatsData = useData(requests.bucketPkgStats, { req, bucket })
   return (
-    <Data fetch={requests.bucketStats} params={{ req, s3, bucket, overviewUrl }}>
-      {(res) => (
-        <M.Paper className={classes.root}>
-          <M.Box className={classes.top}>
-            <M.Typography variant="h5">{bucket}</M.Typography>
-            {!!description && (
-              <M.Box mt={1}>
-                <M.Typography variant="body1">{description}</M.Typography>
-              </M.Box>
-            )}
-            {isRODA && (
-              <M.Box
-                mt={1}
-                position={{ md: 'absolute' }}
-                right={{ md: 32 }}
-                bottom={{ md: 31 }}
-                color="grey.300"
-                textAlign={{ md: 'right' }}
-              >
-                <M.Typography variant="body2">
-                  From the{' '}
-                  <M.Link href={RODA_LINK} color="inherit" underline="always">
-                    Registry of Open Data on AWS
-                  </M.Link>
-                </M.Typography>
-              </M.Box>
-            )}
-            <M.Box mt={{ xs: 2, sm: 3 }} display="flex" alignItems="baseline">
-              <StatDisplay
-                value={AsyncResult.prop('totalBytes', res)}
-                format={readableBytes}
-                fallback={() => '? B'}
-              />
-              <StatDisplay
-                value={AsyncResult.prop('totalObjects', res)}
-                format={readableQuantity}
-                label="Objects"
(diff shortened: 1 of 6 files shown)


nl0:
some python tests are broken for some reason (it has something to do with dependencies)

sir-sigurd:
> some python tests are broken for some reason (it has something to do with dependencies)

PyPI is having [problems](https://status.python.org), but even when deps are installed tests fail [here](https://app.circleci.com/pipelines/github/quiltdata/quilt/2343/workflows/336dbdcc-017d-4048-912b-081591168ba6/jobs/54780).

fiskus (commented):
Can you provide example of how can I test it manually? Or it's too complicated?

I see "Error resolving revision" on `/b/aneesh-ai2-temp/packages/akarve/pkg5` and don't know if it how it supposed to be.

catalog/app/containers/Bucket/Overview.js
```diff
@@ -757,89 +757,82 @@ function Head({ req, s3, overviewUrl, bucket, description }) {
   const classes = useHeadStyles()
   const isRODA = !!overviewUrl && overviewUrl.includes(`/${RODA_BUCKET}/`)
   const colorPool = useConst(() => mkKeyedPool(COLOR_MAP))
+  const statsData = useData(requests.bucketStats, { req, s3, bucket, overviewUrl })
+  const pkgStatsData = useData(requests.bucketPkgStats, { req, bucket })
   return (
-    <Data fetch={requests.bucketStats} params={{ req, s3, bucket, overviewUrl }}>
```
fiskus, line 761: There is only indentation change after removing top level `<Data />`, right?

akarve:
fixed unit tests and did delete_marker filtering in ES instead (or is there some reason why the client needs to do it?). @nl0 please take a look

catalog/app/containers/Bucket/Overview.js
nl0: basically yes, but also the data has been split into two requests

catalog/app/containers/Bucket/Overview.js
```diff
@@ -757,89 +757,82 @@ function Head({ req, s3, overviewUrl, bucket, description }) {
...
-            display="flex"
-            flexDirection={{ xs: 'column', md: 'row' }}
-            alignItems={{ md: 'flex-start' }}
-            position="relative"
+            mt={1}
+            position={{ md: 'absolute' }}
+            right={{ md: 32 }}
+            bottom={{ md: 31 }}
+            color="grey.300"
+            textAlign={{ md: 'right' }}
```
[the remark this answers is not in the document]
nl0: yeah probably you're right

nl0:
@fiskus 

> Can you provide example of how can I test it manually? Or it's too complicated?

hook up to the searchminimal stack and observe:

- no deleted files in overview image grid
- no deleted files in summary (sample)

you can also try uploading and removing files and see that uploading increments total file count and removing (creating a delete marker) does not.

the last thing is the display of delete marker versions in search results: try searching the filename that was deleted and you should see 'Delete Marker' in the version stack.

> I see "Error resolving revision" on `/b/aneesh-ai2-temp/packages/akarve/pkg5` and don't know if it how it supposed to be.

delete markers for packages arent handled in this PR

akarve (approved):

fiskus:
I tried added and deleted files in b/searchminimal on master. Don't see deleted files already.
On this branch I can see deleted files in overview for `/b/aneesh-ai2-temp`

So, I failed to reproduce

nl0:
> I tried added and deleted files in `b/searchminimal` on `master`. Don't see deleted files already.
> On this branch I can see deleted files in overview for `/b/aneesh-ai2-temp`
> 
> So, I failed to reproduce

you should use the `searchminimal` *stack* (that is, api gateway and registry from that stack, bucket doesn't really matter). also, it's probably broken right now if Aneesh's lambda changes are deployed there, so i think we should deploy both latest catalog and lambda changes and test after that.

fiskus:
I  understand, I used searchminimal stack and searchminimal bucket

PR #1998, opened 2021-01-12
Package name validation

fiskus:
* Package name error is shown immediately
* Error message moved to `helperText` (text under Input)
* Warning messages added to package name Input
* Fixed bug with infinite spinner after losing focus

### Create package:
* Package exists: "Package-name exists. Submitting will revise it"

### Copy package:
Name is pre-filled and validating on start
* Successor package exists: "Package-full-name exists. Submitting will revise it"
* Successor package doesn't exist: "Package-full-name will be created"

### Update package:
* Change name and package doesn't exist: "New package will be created"
* Change name but the package with the same name exists already: "Package with this name exists already"

Screenshot from 2021-01-12 19-20-33
Screenshot from 2021-01-12 19-20-52


catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -109,14 +109,41 @@ export function useNameValidator() {
         })
         if (!res.valid) return 'invalid'
       }
-      return undefined
+      // HACK: react-form doesn't change `FormState.validating` on async validation when field loses focus
+      //       but it handles errors: if error is happened, then it's no longer validating
+      return 'noError'
```
fiskus, line 114: It doesn't work. Form can' be submitted with error

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -93,25 +93,61 @@ const readFile = (file) =>
...
+
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const validate = React.useCallback(
     cacheDebounce(async (name) => {
       if (name) {
-        const res = await req({
-          endpoint: '/package_name_valid',
-          method: 'POST',
-          body: { name },
+        const list = await requests.bucketListing({
```
nl0, line 141: i think we should check for package existence here, not for data existence (either by querying the package index or listing the `.quilt/named_packages/$handle` dir)

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -154,23 +190,52 @@ export const getMetaValue = (value) =>
...
-        {validating && <M.CircularProgress size={13} style={{ marginLeft: 8 }} />}
-      </>
-    ),
+    helperText: error || helperText,
+    ...rest,
+  }
+  return <M.TextField {...props} />
+}
+
+export function PackageNameInput({ errors, input, meta, validating, ...rest }) {
```
nl0, line 206: why we need to special-case these inputs?

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -1028,23 +1055,18 @@ function DialogForm({
                   required: 'Enter a package name',
                   invalid: 'Invalid package name',
                 }}
-                margin="normal"
-                fullWidth
```
nl0, line 1032: i believe we should not encapsulate these layout-related props into the specialized component -- they should be set by the calling component when required

nl0 (changes requested):
looks good in general, but we should change package existence checking code to check for package existence rather than data directory existence. plus some minor issues / questions in inline comments

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
fiskus: They have different logic now: name shows error as soon as user type something in (or even on start for Update and Copy), commit message shows error after submitFailed only.

catalog/app/containers/Bucket/PackageUpdateDialog.js
fiskus: Why not?

Now there three levels:
1. RF.Form and props related to final-form
2. CommitInput and NameInput and props related to the visual look
3. Field - an abstract component that can be anything

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
fiskus: Also NameInput has workaround for fixing infinite spinner, CommitMessageInput doesn't have it

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
fiskus: I don't know anything about this, please, provide more details

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: ok makes sense

catalog/app/containers/Bucket/PackageUpdateDialog.js
nl0: ok for now, we'll see if need to change this going forward

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
nl0: see the contents of `.quilt/named_packages` dir: it has all the packages registered in the bucket and has the `$handle/$pointer_file` structure, so to check if the package exists in the bucket you should list objects under `.quilt/named_packages/$handle/` prefix -- if there are any, the package exists, otherwise it does not.

fiskus: I've added method requests.ensurePackageIsPresent, where i check for `.quilt/named_packages/package_name/latest`. `headObject(".quilt/named_packages/package_name")` (just directory, without package revision) suddenly doesn't work

nl0: > I've added method requests.ensurePackageIsPresent, where i check for .quilt/named_packages/package_name/latest.

i think listing revisions is more reliable, bc the `latest` tag may be absent for some reason

> headObject(".quilt/named_packages/package_name") (just directory, without package revision) suddenly doesn't work

that's expected, bc head request only works for objects, and there's no such object, there's only prefix, so you should use `listObjectsV2` for that prefix (you can limit number of keys to 1 to just see if there's _anything_ under that prefix, no matter what)

fiskus: I changed `requests.ensurePackageIsPresent`. Now it uses `request.bucketListing` (which uses `listObjectsV2`). I just check if any file under `.named_packages/package_name`

catalog/app/containers/Bucket/requests.js
```diff
@@ -1235,3 +1235,13 @@ export const objectAccessCounts = ({ s3, analyticsBucket, bucket, path, today })
     today,
     window: 365,
   })
+
+export const ensurePackageIsPresent = async ({ s3, bucket, name }) => {
+  const listOfRevisions = await bucketListing({
```
nl0, line 1240: i think you should just use `s3.listObjectV2` directly, bc you dont need all the things happening in the `bucketListing`, just `listObjectV2({ Bucket, Prefix, MaxKeys: 1 })` and then check `.Contents.KeyCount`

fiskus: done

nl0 (approved):

PR #2000, opened 2021-01-13
Respect delete markers in package-related ES queries

nl0:
### TODO

- [x] handle delete markers for packages in search results

### DIFF
--- a/catalog/app/containers/Bucket/Overview.js
+++ b/catalog/app/containers/Bucket/Overview.js
@@ -758,7 +758,7 @@ function Head({ req, s3, overviewUrl, bucket, description }) {
   const isRODA = !!overviewUrl && overviewUrl.includes(`/${RODA_BUCKET}/`)
   const colorPool = useConst(() => mkKeyedPool(COLOR_MAP))
   const statsData = useData(requests.bucketStats, { req, s3, bucket, overviewUrl })
-  const pkgStatsData = useData(requests.bucketPkgStats, { req, bucket })
+  const pkgCountData = useData(requests.countPackages, { req, bucket })
   return (
     <M.Paper className={classes.root}>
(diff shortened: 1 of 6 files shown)


nl0:
2000 GET BTW

fiskus:
If it's ready for review, could you provide an example of how can I test it manually?

nl0:
@fiskus 

> If it's ready for review, could you provide an example of how can I test it manually?

### 1. Package counts

Package counts in bucket overview should now show the total number of revisions, so you can try revising a package and that should increment that counter. You could also try deleting the revision and see the counter decrement (tho i'm not sure how to do that -- you have to use Quilt CLI or Python API)

### 2. Package and revision listing

Only non-deleted packages should be shown in the package list with correct revision counts (in general, package listing should correspond to the `.quilt/named_packages` dir contents).

The same with the revisions: revision list (both in breadcrumbs dropdown and on the separate page) should not include deleted revisions (the same here: revision listing should correspond to the contents of `.quilt/named_packages/$handle`).

Again, to properly test this you should create and then delete packages / revisions via Quilt Python API / CLI.

Also, currently there's a bug Aneesh was talking about today that prevents inserting all the delete markers into ES index, so it's kinda broken at the moment.

fiskus:
1. I see a package number increment on revising a package with `searchminimal` stack. On staging stack API has 500 error. For me, it's counter-intuitive, that now it's still "123 packages" but actually "123 package versions".

fiskus:
2. `b/aneesh-ai2-temp/packages/` on `staging` stack shows "No packages", on `searchminimal` stack shows deleted package `akarve/ds121` (don't remember exact name). As I see, it's not correct in both cases.

akarve (approved):

PR #2001, opened 2021-01-13
Package Dialog layout: two columns

fiskus:
I moved Intercom widget to the left side because it conflicts with the submit button.

Screenshot from 2021-01-13 22-22-39
Screenshot from 2021-01-13 22-23-10


nl0:
> I moved Intercom widget to the left side because it conflicts with the submit button.

@fiskus maybe just hide it underneath the dialog? (use smaller z-index) bc footer layout takes intercom button placement into account

nl0 (changes requested):
The code looks good, but i think we should use more conventional layout for fullscreen dialogs with an appbar on the top with title and buttons (see [MUI example](https://material-ui.com/components/dialogs/#full-screen-dialogs)). This will also solve the problem with the intercom widget.

Also some comments on the 2-column layout:

- i think we need more spacing between the columns (it should be larger than spacing between section title and contents)

- it would be nice to add a title to the first section (tho i'm not sure which one to use) and align the sections better

spacing

col-layout

fiskus:
> i think we should use more conventional layout for fullscreen dialogs with an appbar on the top with title and buttons (see MUI example).

I don't think it's conventional for desktop, it's more natural for mobile.
---

Note that sections horizontally aligned now. Files bottom is the metadata bottom, when they are empty. FIles bottom is the workflows bottom when they are in maximum height, and workflow is empty. They can't be pixel perfect aligned, because workflows input can have additional description, and because empty Select has less height than filled [1] (I tried to fix it at our side, but I think additional code doesn't worth it)

[1] https://github.com/mui-org/material-ui/issues/21587

fiskus:
> just hide it underneath the dialog? (use smaller z-index)

It seems that intercom doesn't have API for this. I can add `body .intercom-lightweight-app` to `globalStyles.js`

fiskus:
* I've added more space between columns. Unfortunately `<M.grid container />` and `<M.Dialog />` are not frields, there vertical scroll, I used my own styles.
* Added "Main" subheader.
* Made right column vertically flexible:

Screenshot from 2021-01-15 14-33-39
Screenshot from 2021-01-15 14-57-06
Screenshot from 2021-01-15 14-57-35
Screenshot from 2021-01-15 14-58-11

nl0 (approved):
looks good, definitely better than the previous version, tho i'd explore a material-style fullscreen dialog-based approach too (in another PR ofc).

> I don't think it's conventional for desktop, it's more natural for mobile.

it seems like it's quite conventional for material design framework.

anyways, feel free to merge

fiskus:
`<M.Dialog fullscreen />` is where I started

Also, Material Design spec has a note, that fullscreen dialog is mobile only: https://material.io/components/dialogs#full-screen-dialog

nl0:
> `<M.Dialog fullscreen />` is where I started
> 
> Also, Material Design spec has a note, that fullscreen dialog is mobile only: https://material.io/components/dialogs#full-screen-dialog

ok i see

PR #2002, opened 2021-01-14
Package Dialog skeleton

fiskus:
* Separate skeleton component for every part of the form
* Reuse these skeleton components 
* Set margins outside of components
* Rename `PackageDialog/FormSkeleton.js` to `PackageDialog/Skeleton.js`


nl0 (approved):
lgtm

PR #2003, opened 2021-01-15
Filter out deleted files from samples in lambda / ES

nl0:
depends on #2000

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -50,6 +50,7 @@ DELETED_METRIC = {
     },
 }
 
+
 def make_sample_objects_agg(num):
     return {
         'terms': {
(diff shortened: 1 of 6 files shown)


akarve (approved):

lambdas/search/index.py
```diff
@@ -23,6 +23,63 @@
...
+    },
+}
+
+
+def make_sample_objects_agg(num):
+    return {
+        'terms': {
+            'field': 'key',
+            'order': [{'modified': 'desc'}],
+            'size': 1000000,
```
akarve, line 59: Can you add a comment on this? It's way more than the JS client could handle? If we're trying to limit server-side work for performance reasons, tweaking MAX_DOCUMENTS_PER_SHARD will probably be faster.
```suggestion
            'size': 1000000,
```

akarve (approved):

PR #2014, opened 2021-01-19
Re-render JsonEditor on value input

fiskus:
Reduce React state usage inside JsonEditor. Make external value as a source of truth.

JsonEditor still has an internal state: sortOrder for keys and fieldsPath (it's like URL for the entire component), in other aspects internal code become clearer, I hope.


catalog/app/utils/json-schema/json-schema.js
```diff
@@ -126,7 +126,7 @@ export function makeSchemaDefaultsSetter(optSchema) {
     const validate = ajv.compile(schema)
 
     return (obj) => {
-      validate(obj)
+      validate(R.clone(obj))
```
nl0, line 129: i thought the point of this call is to mutate the object and set the defaults -- in this case i believe you should return the cloned object

nl0 (approved):
seems good, some inline comments

catalog/app/utils/json-schema/json-schema.js
fiskus: I think, I made a mistake, will investigate it. Thank you

catalog/app/components/JsonEditor/State.js
```diff
@@ -108,14 +86,15 @@ export function iterateSchema(schema, sortCounter, parentPath, memo) {
       key,
       parentPath,
       required,
-      sortIndex: sortCounter.current,
+
```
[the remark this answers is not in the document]
fiskus: It was :) It helped me refactor, I will remove it

PR #2015, opened 2021-01-19
Indexer: use $key:$version as an id for ES docs

nl0:


lambdas/es/indexer/document_queue.py
```diff
@@ -124,6 +130,8 @@ def append(
             # TODO remove this; it's not meaningful since we use a different index
             # type for object vs. package documents
             "_type": "_doc",
+            # Elastic native keys
```
sir-sigurd, line 133: I'm not sure if this comment adds any value...

lambdas/es/indexer/document_queue.py
```diff
@@ -43,6 +43,12 @@
 RETRY_429 = 3
 
 
+def get_id(key, version_id):
+    """guarantee primary key uniqueness for package delete (marker) documents"""
+    # TODO: use smth like _hash(key) + _hash(version_id) for ids to overcome field length resctriction (512 bytes)
```
sir-sigurd, line 48: Might make sense to add a reference for the length restriction: https://www.elastic.co/guide/en/elasticsearch/reference/7.10/mapping-id-field.html.

lambdas/es/indexer/document_queue.py
```diff
@@ -43,6 +43,12 @@
 RETRY_429 = 3
 
 
+def get_id(key, version_id):
+    """guarantee primary key uniqueness for package delete (marker) documents"""
```
sir-sigurd, line 47: The comment seems a bit misleading since it guarantees uniqueness for all S3 objects.

sir-sigurd (approved):
There are some nitpicks, otherwise LGTM.

lambdas/es/indexer/document_queue.py
nl0:
```suggestion
    # TODO: use smth like _hash(key) + _hash(version_id) for ids to overcome [512 bytes length restriction](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/mapping-id-field.html)
```

sir-sigurd, line 48:
```suggestion
    """
    Generate unique value for every object in the bucket to be used as 
    document `_id`. This value must not exceed 512 bytes in size:
    https://www.elastic.co/guide/en/elasticsearch/reference/7.10/mapping-id-field.html.
    # TODO: both object key and version ID are up to 1024 bytes long, so 
    # we need to use something like `_hash(key) + _hash(version_id)` to 
    # overcome the mentioned size restriction.
    """
```

lambdas/es/indexer/document_queue.py
```diff
@@ -43,6 +43,12 @@
...
+    # TODO: use smth like _hash(key) + _hash(version_id) for ids to overcome [512 bytes length restriction](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/mapping-id-field.html)
```
sir-sigurd, line 48:
```suggestion
    """
    Generate unique value for every object in the bucket to be used as 
    document `_id`. This value must not exceed 512 bytes in size:
    https://www.elastic.co/guide/en/elasticsearch/reference/7.10/mapping-id-field.html.
    # TODO: both object key and version ID are up to 1024 bytes long, so 
    # we need to use something like `_hash(key) + _hash(version_id)` to 
    # overcome the mentioned size restriction.
    """
```

PR #2016, opened 2021-01-19
Populate package name with username prefix

fiskus:
* Set initial name for Create Package dialog as `username/`
* Don't show package name error if it's not modified by the user
* Fix package name validation for Update Package (bug: if you change the revised name and then change back, warning message still presents)


catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -353,6 +356,10 @@ const getTotalProgress = R.pipe(
 
 const defaultNameWarning = ' ' // Reserve space for warning
 
+const selectUser = createStructuredSelector({
```
nl0, line 359: you dont need a structured selector here, you can just use `authSelectors.username` directly

nl0 (approved):

catalog/app/containers/Bucket/PackageCreateDialog.js
fiskus: done

PR #2017, opened 2021-01-19
Fix ES queries and catalog bugs

nl0:
- fix getting total package handle count (affects pagination)
- redirect to the last page when navigating to an out-of-bounds page, fix crash on out-of-bounds pages
- return more data from ES (`took`, `hits.total`, `timed_out`) to aid debugging and benchmarking

### DIFF
--- a/lambdas/search/index.py
+++ b/lambdas/search/index.py
@@ -202,7 +202,12 @@ def lambda_handler(request):
         }
         size = 0
         _source = False
-        filter_path = 'aggregations.objects.buckets.latest.hits.hits._source'
+        filter_path = ','.join([
+            'took',
+            'timed_out',
+            'hits.total',
+            'aggregations.objects.buckets.latest.hits.hits._source'
+        ])
     elif action == 'sample':
         body = {
             'query': {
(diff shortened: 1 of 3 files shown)


lambdas/search/index.py
```diff
@@ -202,7 +202,12 @@ def lambda_handler(request):
         }
         size = 0
         _source = False
-        filter_path = 'aggregations.objects.buckets.latest.hits.hits._source'
+        filter_path = ','.join([
+            'took',
+            'timed_out',
+            'hits.total',
+            'aggregations.objects.buckets.latest.hits.hits._source'
+        ])
```
sir-sigurd, line 210: Does it make sense to deduplicate these fields here and below?
It looks like you don't have to join these strings:
https://elasticsearch-py.readthedocs.io/en/6.3.1/api.html#response-filtering

(also I'd use a tuple instead of a list, but that's nano-optimization)

nl0: as for deduping -- i dont think it's worth it

akarve (approved):

PR #2022, opened 2021-01-20
Bucket admin link

fiskus:
* Add ability to open Edit Dialog by providing URL query parameter
* Add link to Dialog from Bucket Overview

Screenshot from 2021-01-20 11-06-05


catalog/app/containers/Bucket/Overview.js
```diff
@@ -804,6 +812,11 @@ function Head({ req, s3, overviewUrl, bucket, description }) {
             fallback={() => null}
           />
         </M.Box>
+        <RRLink className={classes.settings} to={urls.adminBuckets(bucket)}>
```
nl0, line 815: we should only show this for users with admin privileges

catalog/app/containers/Admin/Buckets.js
```diff
@@ -871,9 +878,23 @@ function CRUD({ buckets }) {
     },
   ]
 
+  const editingBucket = React.useMemo(
+    () => (bucketName ? rows.find(({ name }) => name === bucketName) : null),
```
nl0, line 882: i think it might make sense to redirect to the root page or render error msg inside the dialog if the bucket wasnt found instead of simply ignoring the query param

nl0 (approved):
looks good. it would be nice to make the edit icon into a link instead of doing the redirect inside the click handler, but i suspect it's a little more involved and not required right now

catalog/app/containers/Admin/Buckets.js
fiskus: I made redirect, using react-router-dom's `<Redirect />`

catalog/app/containers/Bucket/Overview.js
fiskus: done

fiskus:
> t would be nice to make the edit icon into a link instead of doing the redirect inside the click handler

It's simple to make an icon a link. But the table's row is also clickable. To make the table row a link I need to change the table's html/css layout.

nl0:
good to merge

PR #2023, opened 2021-01-20
Tune ES queries

nl0:
only package-related queries for now

### DIFF
--- a/catalog/app/containers/Bucket/requests.js
+++ b/catalog/app/containers/Bucket/requests.js
@@ -760,6 +760,28 @@ const NOT_DELETED_METRIC = {
   },
 }
 
+const withCalculatedRevisions = (s) => ({
+  scripted_metric: {
+    init_script: `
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Bucket/requests.js
```diff
@@ -1009,10 +1024,10 @@ export const getPackageRevisions = withErrorHandling(
         },
         aggs: {
           revisions: {
-            composite: {
-              // the limit is configured in ES cluster settings (search.max_buckets)
-              size: 10000,
-              sources: [{ pointer: { terms: { field: 'key' } } }],
+            terms: {
+              field: 'key',
+              size: 1000000,
```
akarve, line 1029: Can you add a comment on this limit? I would also like to document what it means for our users, no more than 1M package instances?

akarve (approved):
You sir are a Wizard :godmode:

PR #2024, opened 2021-01-21
Package Dialog: vertically flexible layout

fiskus:
Fit to any height (600px minimum), don't restrict by exact pixel numbers.

Screenshot from 2021-01-21 18-19-11
Screenshot from 2021-01-21 18-20-23
Screenshot from 2021-01-21 18-19-25


nl0 (approved):

akarve:
🔥

fiskus:
@nl0 I've changed the behavior. Now, I calculate JsonEditor's height and set M.DialogContent height according to it:

```js
  const dialogContentStyles = React.useMemo(
    () => ({
      height: R.clamp(
        430 /* minimal height */,
        window.innerHeight - 200 /* free space for headers */,
        400 /* space to fit other inputs */ + metaHeight,
      ),
      paddingTop: 0,
    }),
    [metaHeight],
  )
```

JsonEditor's height is recalculated on every metadata change

catalog/app/containers/Bucket/PackageCopyDialog.js
```diff
@@ -84,11 +84,21 @@ function DialogTitle({ bucket }) {
...
+  },
   meta: {
+    display: 'flex',
+    flexDirection: 'column',
     marginTop: t.spacing(3),
+    overflowY: 'auto',
   },
 }))
 
+// <M.DialogContent /> doesn't have className prop
```
nl0, line 101: but it has `classes` prop, so you can pass a class using `classes={{ root: className }}`

catalog/app/containers/Bucket/PackageCopyDialog.js
```diff
@@ -178,9 +200,28 @@ function DialogForm({
...
+        setMetaHeight(editorElement.clientHeight)
+      }
+
+      handleNameChange(values.name)
+    },
+    [editorElement, handleNameChange, setMetaHeight],
+  )
+
+  React.useEffect(() => {
+    if (document.body.contains(editorElement)) {
```
nl0, line 220: shouldnt we check if `editorElement` is not null?

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -330,10 +330,17 @@ function FilesInput({
 function DialogSuccess({ bucket, hash, name, onClose }) {
   const { urls } = NamedRoutes.use()
 
+  const dialogContentStyles = React.useMemo(
```
nl0, line 333: seems like this change is unnecessary

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -422,8 +433,21 @@ function PackageCreateDialog({
...
+  const [metaHeight, setMetaHeight] = React.useState(0)
   const classes = useStyles()
 
+  const dialogContentStyles = React.useMemo(
+    () => ({
+      height: R.clamp(
+        420 /* minimal height */,
+        window.innerHeight - 200 /* free space for headers */,
+        400 /* space to fit other inputs */ + metaHeight,
+      ),
```
nl0, line 445: it probably makes sense to move this out and reuse

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -548,13 +572,24 @@ function PackageCreateDialog({
...
+        setMetaHeight(editorElement.clientHeight)
+      }
       if (dirtyFields.name) handleNameChange(values.name)
     },
-    [handleNameChange],
+    [editorElement, handleNameChange, setMetaHeight],
   )
 
+  React.useEffect(() => {
+    if (document.body.contains(editorElement)) {
```
nl0, line 588: same question here re: `editorElement` may be null?

catalog/app/containers/Bucket/PackageDialog/DialogError.js
```diff
@@ -86,7 +87,7 @@ export default function DialogError({ error, skeletonElement, title, onCancel })
     <>
       <M.DialogTitle>{title}</M.DialogTitle>
 
-      <M.DialogContent className={classes.content}>
+      <M.DialogContent style={dialogContentStyles}>
```
nl0, line 90: actually, `className` _is_ supported, tho undocumented: https://github.com/mui-org/material-ui/blob/v4.11.3/packages/material-ui/src/DialogContent/DialogContent.js#L36 (`classes.root` is supported and documented)

catalog/app/containers/Bucket/PackageDialog/DialogLoading.js
```diff
@@ -1,21 +1,16 @@
 import * as React from 'react'
 import * as M from '@material-ui/core'
 
-const useStyles = M.makeStyles(() => ({
-  content: {
-    paddingTop: 0,
-    position: 'relative',
-  },
-}))
+const dialogContentStyles = {
```
nl0, line 4: same here: we can and should use a class for this

catalog/app/containers/Bucket/PackageDialog/DialogSuccess.js
```diff
@@ -6,6 +6,10 @@ import * as M from '@material-ui/core'
 import * as NamedRoutes from 'utils/NamedRoutes'
 import StyledLink from 'utils/StyledLink'
 
+const dialogContentStyles = {
```
nl0, line 9: same here

catalog/app/containers/Bucket/PackageDialog/PackageDialog.js
```diff
@@ -345,13 +351,14 @@ const useMetaInputStyles = M.makeStyles((t) => ({
...
 
 // TODO: warn on duplicate keys
-export function MetaInput({
-  className,
-  schemaError,
-  input: { value, onChange },
-  meta,
-  schema,
-}) {
+export const MetaInput = React.forwardRef((props, ref) => {
```
nl0, line 354: pls use named function here

catalog/app/containers/Bucket/PackageDialog/Skeleton.js
```diff
@@ -13,19 +13,19 @@ export function FilesInputSkeleton({ animate, className }) {
...
 
-export function MetaInputSkeleton({ animate, className }) {
-  return (
-    <div className={className}>
-      <M.Box display="flex" mb={2}>
-        <Skeleton {...{ height: 24, width: 64, animate }} />
-        <M.Box flexGrow={1} />
-        <Skeleton {...{ height: 24, width: 64, animate }} />
-      </M.Box>
+export const MetaInputSkeleton = React.forwardRef(({ animate, className }, ref) => (
```
nl0, line 21: pls used named fn here

catalog/app/containers/Bucket/PackageDialog/Skeleton.js
```diff
@@ -13,19 +13,19 @@ export function FilesInputSkeleton({ animate, className }) {
...
+  <div className={className}>
+    <M.Box display="flex" mb={2}>
+      <Skeleton {...{ height: 24, width: 64, animate }} />
+      <M.Box flexGrow={1} />
+      <Skeleton {...{ height: 24, width: 64, animate }} />
+    </M.Box>
+    <div ref={ref}>
```
nl0, line 28: why is ref forwarded to some inner element instead of the root?

nl0: also, why do we need this div wrapper around the Box?

catalog/app/containers/Bucket/PackageDirectoryDialog.js
```diff
@@ -82,10 +88,23 @@ function DialogForm({
...
   const nameExistence = PD.useNameExistence(successor.slug)
   const [nameWarning, setNameWarning] = React.useState('')
+  const [metaHeight, setMetaHeight] = React.useState(0)
   const classes = useStyles()
 
   const req = APIConnector.use()
 
+  const dialogContentStyles = React.useMemo(
+    () => ({
+      height: R.clamp(
```
nl0, line 98: seeing this for the third time -- definitely should be moved out and reused

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -422,10 +460,18 @@ function DialogForm({
 }
 
 function DialogPlaceholder({ close }) {
+  const dialogContentStyles = React.useMemo(
```
nl0, line 463: pls use a class or keep it inline -- i see no point in using a hook here

catalog/app/containers/Bucket/PackageUpdateDialog.js
```diff
@@ -450,11 +496,17 @@ function DialogError({ error, close }) {
 }
 
 function DialogSuccess({ bucket, name, hash, close }) {
+  const dialogContentStyles = React.useMemo(
```
nl0, line 499: doesnt seem necessary, it's trivial enough to be kept inline imo

nl0 (commented):
the layout logic seems nice, tho i have some code-related questions / comments (inline)

catalog/app/components/JsonEditor/JsonEditor.js
```diff
@@ -24,23 +24,23 @@ const useStyles = M.makeStyles((t) => ({
...
+    changeValue,
+    className,
+    disabled,
+    columns,
+    jsonDict,
+    fieldPath,
+    makeAction,
+    onChange,
+    setFieldPath,
+  } = props
```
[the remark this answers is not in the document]
fiskus: I started with `const Component = forwardRef({ ...props }, ref) => {}` and got additional indentation by prettier because arguments list (restructured props) is too long. So I decided to separate props destructuring.

catalog/app/containers/Bucket/PackageCopyDialog.js
fiskus: I forgot it, thanks

catalog/app/containers/Bucket/PackageCopyDialog.js
fiskus: No, because `document.body` doesn't contain null :)
```js
document.body.contains(null) === false
```

catalog/app/containers/Bucket/PackageDialog/Skeleton.js
fiskus: > why is ref forwarded to some inner element instead of the root?

I'm interested in height of this inner element. This inner element's height is a matter of change. It contains editor rows itself, so I think it's semantically correct. 
We can change `ref` to `editorRef` or something like this.

> why do we need this div wrapper around the Box?

There two Boxes:
```jsx
<div ref={wantToKnowHeightHere}><M.Box /><M.Box/></div>
```

catalog/app/containers/Bucket/PackageDialog/DialogError.js
fiskus: So, I'll use `className` then, if you don't mind?

catalog/app/components/JsonEditor/JsonEditor.js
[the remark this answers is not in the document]
fiskus: Adding named function removes this indentation problem

PR #2026, opened 2021-01-22
Use EncodingType=url for S3 list* requests to handle special chars

nl0:
Currently some weird key names (e.g. '\r\n') crash AWS SDK while performing list* operations, so we need to tell it explicitly to url-encode the keys (and then we need to manually decode them). However those keys still arent properly supported by the catalog -- they render with empty names and arent navigable (some quite involved refactoring required to fully handle this).

### DIFF
--- a/catalog/app/containers/Bucket/requests.js
+++ b/catalog/app/containers/Bucket/requests.js
@@ -33,13 +33,13 @@ export const bucketListing = ({ s3, bucket, path = '', prefix }) =>
       R.applySpec({
         dirs: R.pipe(
           R.prop('CommonPrefixes'),
-          R.map((p) => decodeURI(p.Prefix)),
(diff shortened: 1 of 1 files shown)


fiskus (approved):

sir-sigurd (changes requested):
```python
In [3]: [x['Key'] for x in s3_client.list_objects_v2(Bucket='quilt-dev-null-2')['Contents']]
Out[3]: ['\x01', '\r\n', '    ', '&&&&', '+', '//', 'ы\\ы', 'ыыыы']
```
These keys are displayed like that:
image

It looks like `decodeURIComponent()` should be used instead, space is encoded as `+` and needs additional handling.

sir-sigurd (approved):

PR #2036, opened 2021-01-28
Update webpack to v5, adjust config

nl0:


catalog/package.json
```diff
@@ -44,11 +44,7 @@
     "test:watch": "npm run test:only -- --watch",
     "prettify": "prettier --write"
   },
-  "browserslist": [
-    "last 2 versions",
-    "> 1%",
-    "IE 10"
-  ],
+  "browserslist": ["last 2 versions", "> 1%"],
```
fiskus, line 47: calling `npm install` formats this array to multiline:

```diff
-  "browserslist": ["last 2 versions", "> 1%"],
+  "browserslist": [
+    "last 2 versions",
+    "> 1%"
+  ],
```

Linux, node v15.5.0, npm 7.3.0

catalog/internals/webpack/webpack.prod.babel.js
```diff
@@ -147,4 +44,20 @@ module.exports = require('./webpack.base.babel')({
   performance: {
     assetFilter: (assetFilename) => !/(\.map$)|(^(main\.|favicon\.))/.test(assetFilename),
   },
+
+  // options passed to all instances of HtmlWebpackPlugin
+  html: {
+    minify: {
```
fiskus, line 50: `build` directory has the same size with just `minify: true`. I don't think this fine-tuning worth it

fiskus (approved):

catalog/package.json
nl0:
```suggestion
  "browserslist": [
    "last 2 versions",
    "> 1%"
  ],
```

catalog/internals/webpack/webpack.prod.babel.js
nl0: looks like we can omit this altogether and it will still minify the output in production mode

fiskus: :heavy_plus_sign:

PR #2040, opened 2021-02-01
Directory to package dialog

fiskus:
* Move `FilesInput` from `PackageUpdateDialog` to its own module
* Add disabled state for `FilesInput`
* Add `PackageDirectoryDialog` and re-use `FilesInput` there


catalog/app/containers/Bucket/PackageDialog/DialogError.js
```diff
@@ -47,6 +47,8 @@ const errorDisplay = R.cond([
         </M.Typography>
       </>
     ),
+  ],
+  [
```
nl0, line 51: ohwow

nl0 (approved):
looks good. the only thing i noticed after brief testing was that one can click on the drop area and trigger file selection dialog -- i think we should disable this when the files input is disabled.

PR #2043, opened 2021-02-03
TypeScript: toolchain

nl0:
# Description

- use typescript compiler
- use ts-jest
- ditch babel
- rm local polyfills, use polyfill.io instead
- ditch generators and obsolete scripts

## To backlog

- run CI jobs in parallel (build, test, eslint, prettier)
- use ts-node for server-side stuff and toolchain, migrate that code to ts
- consider serving the polyfills ourselves to not depend on a service outside our control (polyfill.io)


nl0:
@fiskus ready for initial review, i'll write a better description soonish

catalog/app/index.html
```diff
@@ -26,6 +26,7 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
     <!-- styles for Jupyter Notebooks -->
     <link href="/ipynb.css" rel="stylesheet">
+    <script src="https://polyfill.io/v3/polyfill.min.js?features=Intl%2Cdefault%2CResizeObserver%2CBlob%2Cfetch%2Ces2016%2Ces2017%2Ces2018%2Ces2019%2Ces2015"></script>
```
fiskus, line 29: The slow download will slow all stuff below from downloading.
I suppose that the application's javascript will be inserted below and will be downloaded after plolyfill.io.

Maybe, `npm` package is safer?

catalog/internals/webpack/webpack.prod.js
```diff
@@ -1,21 +1,16 @@
 const path = require('path')
-const TerserPlugin = require('terser-webpack-plugin')
+
 const CompressionPlugin = require('compression-webpack-plugin')
+const TerserPlugin = require('terser-webpack-plugin')
```
fiskus, line 4: It seems that TerserPlugin is the default, and there is no need to set it manually

https://webpack.js.org/configuration/optimization/#optimizationminimize

catalog/app/index.html
nl0: > The slow download will slow all stuff below from downloading.

brief testing shows that polyfill.io works really fast (+ caching), and it's probably more reliable than any of our homegrown solutions would be

> I suppose that the application's javascript will be inserted below and will be downloaded after plolyfill.io.

that's the point, yes, to load all the polyfills before executing the app code.

> Maybe, npm package is safer?

safer how?

the main point of using polyfill.io vs polyfills included in the bundle is smaller bundle size and automatic polifylling based on user-agent.
we could run the polyfill service ourselves if we dont trust polyfill.io enough, but this would require running an extra container and i'm not sure we need that

catalog/internals/webpack/webpack.prod.js
nl0: we need to configure parallelism explicitly, bc otherwise it fails in CI

nl0:
btw entry point bundles got ~15% smaller (from ~1.6MB to ~1.35MB)

catalog/app/index.html
fiskus: Polyfill.io will be down one day, and will block our application. If you think, that effort to avoid this is much bigger than that risk, I'm totally fine with that.

nl0: > If you think, that effort to avoid this is much bigger than that risk

yep, i think so, tho we can backlog revisiting the polyfilling setup later

fiskus (approved):

PR #2044, opened 2021-02-03
Avoid unneeded unquoting of S3 events data

sir-sigurd:
# Description

If I'm not missing something bucket names don't need quoting: https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules.
Version id is URL-ready, check `x-amz-version-id` at https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html.
It looks like ETags don't require quoting too, though I haven't found clear contract on it.

### DIFF
--- a/lambdas/es/indexer/index.py
+++ b/lambdas/es/indexer/index.py
@@ -52,7 +52,7 @@ import pathlib
 import re
 from os.path import split
 from typing import Optional
-from urllib.parse import unquote, unquote_plus
+from urllib.parse import unquote_plus
 
 import boto3
 import botocore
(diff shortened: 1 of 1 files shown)


akarve (approved):
There is no ETag contact I can find in AWS docs but in practice it has always been an MD5 or MD5 of MD5s but the contract is looser than that. Anyway this only gets trafficked at the requests level by boto3 which should transparently handle encoding.

Version Id and bucket name I agree are URL safe by definition.

PR #2047, opened 2021-02-04
Fix some of race conditions in util.write_yaml()

sir-sigurd:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation
    - [ ] [Run `build.py`](../gendocs/build.py) for new docstrings

### DIFF
--- a/api/python/quilt3/util.py
+++ b/api/python/quilt3/util.py
@@ -280,9 +280,8 @@ def write_yaml(data, yaml_path, keep_backup=False):
     try:
         if path.exists():
             path.rename(backup_path)
-        if not path.parent.exists():
-            path.parent.mkdir(parents=True)
-        with path.open('w') as config_file:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with path.open('x') as config_file:
             yaml.dump(data, config_file)
     except Exception:     # intentionally wide catch -- reraised immediately.
         if backup_path.exists():


api/python/quilt3/util.py
```diff
@@ -280,9 +280,8 @@ def write_yaml(data, yaml_path, keep_backup=False):
     try:
         if path.exists():
             path.rename(backup_path)
-        if not path.parent.exists():
```
dimaryaz, line 283: The `if path.exists(): path.rename(...)` has the same race condition.

api/python/quilt3/util.py
```diff
@@ -280,9 +280,8 @@ def write_yaml(data, yaml_path, keep_backup=False):
...
-            path.parent.mkdir(parents=True)
-        with path.open('w') as config_file:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with path.open('x') as config_file:
```
dimaryaz, line 284: This will raise an exception if the file exists. Ok, better than two processes writing to the same file, but... now look at the exception handler, and imagine what kind of race conditions happen there :)

dimaryaz (commented):
I'm not sure this function is even worth fixing... My suggestion would be:
- Don't call it from the registry (already done)
- Document that it's not safe to call it concurrently (though... no reasonable user would think it's safe, anyway)
- Maybe rewrite it completely. Right now, it's way over-engineered. Do we really need backups? At most, I'd write to a temp file, then rename it to the config file, so it's atomic.

PR #2051, opened 2021-02-05
TypeScript: migrate app (pt. 1)

nl0:


catalog/app/app.tsx
```diff
@@ -56,7 +56,7 @@ fontLoader('Roboto', 'Roboto Mono').then(() => {
 const ErrorBoundary = composeComponent(
   'ErrorBoundary',
   Sentry.inject(),
-  createBoundary(({ sentry }) => (error, info) => {
+  createBoundary(({ sentry }: { sentry: any }) => (error: any, info: any) => {
```
fiskus, line 59: I should think about it more, but it looks weird right now.
You set "noImplicitAny" to `tsconfig`, then converted `js` to `ts` and set `any` for the value which has exact type.

You need this file as `typescript` because other files depend on it?

catalog/app/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
+import * as M from '@material-ui/core'
+
+import Delay from 'utils/Delay'
+
+interface PlaceholderOwnProps {
```
fiskus, line 6: `interface PlaceholderOwnProps extends M.BoxProps` ?

catalog/app/containers/NavBar/NavBar.tsx
```diff
@@ -19,30 +19,39 @@ import bg from './bg.png'
...
 
-const Item = React.forwardRef((props, ref) => (
-  <M.MenuItem
-    // eslint-disable-next-line no-nested-ternary
-    component={props.to ? Link : props.href ? 'a' : undefined}
-    ref={ref}
-    {...props}
-  />
-))
+interface ItemProps {
```
fiskus, line 32: `interface ItemProps extends M.MenuItemProps` too, I guess

catalog/app/utils/Delay.tsx
```diff
@@ -1,6 +1,12 @@
 import * as React from 'react'
 
-export default function Delay({ ms = 1000, alwaysRender = false, children }) {
+interface DelayProps {
+  ms?: number
+  alwaysRender?: boolean
+  children: (ready?: boolean) => JSX.Element
```
fiskus, line 6: I think, using `React.ReactElement` is better, because `React` is imported here

catalog/app/containers/NavBar/NavBar.tsx
```diff
@@ -19,30 +19,39 @@ import bg from './bg.png'
...
+  to?: string
```
fiskus, line 33: Actually, `to` accepts objects too. And we can set two types union (one OR another) to avoid passing `to` and `href` the same time:
```ts
type BrowserLinkProps = { href: string } & MenuItemProps;
type RouterLinkProps = LinkProps & MenuItemProps;
type ItemProps = BrowserLinkProps | RouterLinkProps;
```

catalog/app/app.tsx
fiskus: I suppose, types are:

```
import { _callOnClient } from '@sentry/minimal'
import { CaptureContext } from '@sentry/types'

sentry: typeof _callOnClient
error: Error
info: CaptureContext
```

catalog/app/app.tsx
```diff
@@ -99,12 +99,12 @@ const intercomUserSelector = (state) => {
...
 }
 
-const sentryUserSelector = (state) => {
+const sentryUserSelector = (state: any) => {
   const { user: u } = Auth.selectors.domain(state)
   return u ? { username: u.current_user, email: u.email } : {}
 }
 
-const render = (messages) => {
+const render = (messages: any) => {
```
fiskus, line 107: My suggestion is `Record<'en', Record<string, string>>`, but I'm not quite sure

fiskus (approved):

catalog/app/app.tsx
nl0: > You set "noImplicitAny" to tsconfig, then converted js to ts and set any for the value which has exact type.

well, it's better be explicit, so we get errors when there's no types, and for now i'm setting untyped things to `any` for compilation to succeed (ima use $TSFixMe for easier tracking)

> Do you need this file as typescript because other files depend on it?

nope, but it's an entry point and it seems logical to start with it

> I suppose, types are:

yep, looks good, but these should be added to the tracking module

catalog/app/utils/Delay.tsx
nl0: i would use ReactNode, but it doesnt compile, react typings for functional components are a little funky (probably for historical reasons)

catalog/app/containers/NavBar/NavBar.tsx
nl0: makes sense, but doesnt typecheck. we'll revisit it someday soon

catalog/app/app.tsx
nl0: yep, looks right. i've added typings for i18n module

PR #2055, opened 2021-02-08
Update js dependencies (non-breaking)

nl0:


fiskus (approved):
There two major updates: tslib and path-browserify

nl0:
> There are two major updates: tslib and path-browserify

@fiskus they don't contain breaking changes for our code (tslib update deprecates older ts versions, path-browserify update deprecate older node api)

PR #2059, opened 2021-02-11
Add React context provider for AWS Athena SDK

fiskus:


catalog/app/utils/AWS/Athena.tsx
```diff
@@ -0,0 +1,31 @@
...
+      ...cfg,
+      ...overrides,
+    },
+    (opts: Athena.ClientConfiguration) => new Athena(opts),
+  )
+
+  return <AthenaContext.Provider value={client}>{children}</AthenaContext.Provider>
+}
+
+export const useAthena = () => React.useContext(AthenaContext)()
```
nl0, line 29: i think we should add `Credentials.use().suspend()` here to ensure the credentials are resolved (like in S3 provider)

nl0 (approved):
lgtm, but we should wait for credentials

fiskus:
I'll find it if ever need in future

PR #2065, opened 2021-02-14
Convert some code under containers/Auth to TS, refactor to not use deprecated utils

nl0:


catalog/app/containers/Auth/Layout.tsx
```diff
@@ -0,0 +1,220 @@
...
+export function Heading(props: M.TypographyProps) {
+  return <M.Typography variant="h4" align="center" {...props} />
+}
+
+interface FieldOwnProps {
+  // TODO: use redux-form / final-form type definitions
+  input: {}
+  meta: {
+    error: any
+    submitError: any
```
fiskus, line 30: It seems that code expects `error` and `submitError` to be string:

```
    error: string | null
    submitError: string | null
```
or
```
    error?: string
    submitError?: string
```

fiskus (approved):

catalog/app/containers/Auth/Layout.tsx
```diff
@@ -0,0 +1,220 @@
...
+const useContainerStyles = M.makeStyles({
+  root: {
+    marginLeft: 'auto',
+    marginRight: 'auto',
+    maxWidth: 300,
+    width: '100%',
+  },
+})
+
+export function Container(props: JSX.IntrinsicElements['div']) {
```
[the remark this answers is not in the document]
nl0: yep, makes sense, i'll adjust that

catalog/app/containers/Auth/Layout.tsx
nl0: yep, it seems that way in this case, but generally redux-form and final form accepts anything as a validation error

nl0:
@fiskus thx for the feedback, i've applied your suggestions and rebased the PR, gonna merge now

PR #2066, opened 2021-02-16
Add a hack for "+" in S3 URIs

dimaryaz:
Workaround for https://forums.aws.amazon.com/thread.jspa?threadID=55746

### DIFF
--- a/s3-proxy/Dockerfile
+++ b/s3-proxy/Dockerfile
@@ -14,12 +14,15 @@ RUN set -x \
     && addgroup --system nginx \
     && adduser --system --disabled-login --ingroup nginx --no-create-home --home /nonexistent --shell /bin/false nginx
 
+COPY s3.patch /root/s3.patch
+
 RUN set -x \
     && apt-get update \
-    && apt-get install -y gettext \
+    && apt-get install -y gettext patch \
     && apt-get install -y curl gcc make libssl-dev libpcre3-dev zlib1g-dev \
     && curl -L http://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz | tar zx \
     && cd nginx-${NGINX_VERSION} \
+    && patch -p1 < /root/s3.patch \
     && mkdir mod_zip \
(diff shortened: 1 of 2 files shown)


s3-proxy/s3.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/core/ngx_string.c b/src/core/ngx_string.c
+index 468e9600..7d8cfc9c 100644
```
akarve, line 2: These first two lines shouldn't be here?

s3-proxy/Dockerfile
```diff
@@ -14,12 +14,15 @@ RUN set -x \
...
+COPY s3.patch /root/s3.patch
+
 RUN set -x \
     && apt-get update \
-    && apt-get install -y gettext \
+    && apt-get install -y gettext patch \
     && apt-get install -y curl gcc make libssl-dev libpcre3-dev zlib1g-dev \
     && curl -L http://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz | tar zx \
     && cd nginx-${NGINX_VERSION} \
+    && patch -p1 < /root/s3.patch \
```
akarve, line 25: minor: with this you don't need an further comments
```suggestion
    && patch -p1 < /root/s3_encode_plus.patch \
```

s3-proxy/s3.patch
akarve: Or do we need `git patch` if we retain these lines? Sorry, I've never seen this kind of patch, usually just `diff - u` output so not familiar.

dimaryaz: Oh, that's just what `git diff` outputs; it's harmless.

But, didn't know about `git apply` - I don't need to install `patch`.

nl0 (approved):
ohwow looks magical

s3-proxy/s3.patch
dimaryaz: Oh wait, no, we don't have `git` here.

PR #2067, opened 2021-02-16
Don't decode and re-encode URLs in mod_zip

dimaryaz:
Workaround for https://forums.aws.amazon.com/thread.jspa?threadID=55746

Decoding the URLs is not really needed to begin with - but Nginx will encode
them in "internal" requests. So, remove the "internal" bit.

(This might prevent mod_zip from accessing internal-only servers,
but we're not using those.)


nl0 (approved):
looks less magical than the previous one)

PR #2069, opened 2021-02-17
Update package name warning texts

fiskus:
* Move `<Code />` to separate component and reuse it for messages
* Make messages identical with the new text proposed by @akarve 

Screenshot from 2021-02-17 16-47-32

### DIFF
--- a/catalog/app/components/Code/Code.tsx
+++ b/catalog/app/components/Code/Code.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react'
+import * as M from '@material-ui/core'
+
+const useCodeStyles = M.makeStyles((t) => ({
+  root: {
+    background: t.palette.grey[300],
+    borderRadius: '2px',
+    color: t.palette.text.primary,
+    fontFamily: (t.typography as $TSFixMe).monospace.fontFamily,
+    padding: '0 3px',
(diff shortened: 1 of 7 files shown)


nl0 (approved):

catalog/app/components/Code/Code.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import * as M from '@material-ui/core'
+
+const useCodeStyles = M.makeStyles((t) => ({
+  root: {
+    background: t.palette.grey[300],
+    borderRadius: '2px',
+    color: t.palette.text.primary,
+    fontFamily: (t.typography as $TSFixMe).monospace.fontFamily,
```
fiskus, line 9: I just not familiar with redefining libraries. Should be something like:

```ts
import * as styles from '@material-ui/styles'

declare module '@material-ui/styles' {
  Typography extends styles.Typography {
    monospace: TypographyStyle
  }
}
```

But I couldn't make it work.

nl0: i think it should be just something like `interface Typography { extra props here }`, see `app/@types/mui.d.ts`, tho not a big deal

PR #2073, opened 2021-02-18
Show search results number

fiskus:
I don't know if it is ok to pass AsyncResult as a prop.

Also, I changed `total` from the number of hits in JSON to `total` responded by API. So, the results number doesn't match with pages number (145684 results should be on 14569 pages, but there are 91 pages only).

Screenshot from 2021-02-18 15-41-06


catalog/app/containers/Bucket/Search.js
```diff
@@ -106,6 +104,18 @@ function QueryInput({ query, bucket, onChange }) {
   )
 }
 
+function SearchStats({ data }) {
+  return data.case({
+    _: () => '',
+    Err: () => '',
```
nl0, line 110: this can be removed since `_` handles the error case as well

catalog/app/containers/Bucket/Search.js
```diff
@@ -290,15 +312,19 @@ function Search({ bucket, query, page, mode, retry }) {
         >
           <QueryInput query={query || ''} bucket={bucket} onChange={handleQueryChange} />
         </M.Box>
-        <M.Box flexGrow={1} display={{ xs: 'none', sm: 'block' }} />
+        <M.Box
+          flexGrow={1}
+          display={{ xs: 'none', sm: 'block' }}
```
nl0, line 317: i think it doesnt make much sense to keep half of the styles inline and move the other half to the class, just use a div and put all its styles into the class

catalog/app/containers/Search/Search.js
```diff
@@ -197,6 +186,18 @@ function BucketSelectDropdown({ buckets, onChange, short = false }) {
   )
 }
 
+function SearchStats({ data }) {
+  return data.case({
+    _: () => '',
+    Err: () => '',
```
nl0, line 192: same here, `_` is enough

catalog/app/containers/Search/Search.js
```diff
@@ -399,7 +412,13 @@ export default function Search({ location: l }) {
                   onChange={handleQueryChange}
                 />
               </M.Box>
-              <M.Box flexGrow={1} display={{ xs: 'none', sm: 'block' }} />
+              <M.Box
+                flexGrow={1}
+                display={{ xs: 'none', sm: 'block' }}
```
nl0, line 417: same here re: Box props vs div + className

catalog/app/containers/Bucket/Search.js
```diff
@@ -290,15 +312,19 @@ function Search({ bucket, query, page, mode, retry }) {
...
+          className={classes.stats}
+        >
+          {query && <SearchStats data={data} />}
```
nl0, line 320: i think we should cast `query` to boolean here to be safe
```suggestion
          {!!query && <SearchStats data={data} />}
```

catalog/app/containers/Search/Search.js
```diff
@@ -399,7 +412,13 @@ export default function Search({ location: l }) {
...
+                className={classes.stats}
+              >
+                {q && <SearchStats data={data} />}
```
nl0, line 420:
```suggestion
                {!!q && <SearchStats data={data} />}
```

nl0 (approved):
looks good, some minor comments

PR #2074, opened 2021-02-19
package events lambda

sir-sigurd:
## TODO
- [x] tests
- [x] finalize schema
- [x] <del>add stack name or ID into event? or create a separate bus for each stack?</del>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,6 +244,7 @@ workflows:
                 - thumbnail
                 - pkgselect
                 - shared
+                - pkgevents
       - test-lambda:
           name: test-lambda-indexer
           path: es/indexer
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -78,7 +78,6 @@ def pkg_created_event(s3_event):
         'Source': 'com.quiltdata',
         'DetailType': 'package-revision',
         'Resources': [
-            bucket_obj['arn'],
             # TODO: add stack ARN?
         ],
         'Detail': json.dumps({
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -9,6 +9,10 @@ event_bridge = boto3.client('events')
 s3 = boto3.client('s3')
 
 
+class PutEventsException(Exception):
+    pass
+
+
 class EventsQueue:
     MAX_SIZE = 10
 
@@ -23,7 +27,9 @@ class EventsQueue:
     def _flush(self):
         events = self._events
         self._events = []
-        event_bridge.put_events(Entries=events)
+        resp = event_bridge.put_events(Entries=events)
+        if resp['FailedEntryCount']:
+            raise PutEventsException(resp)
 
     def flush(self):
         if self:
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -46,7 +46,7 @@ class EventsQueue:
         return bool(self._events)
 
 
-PKG_POINTER_REGEX = re.compile(r'\.quilt/named_packages/([\w-]+/[\w-]+)/(.{10})')
+PKG_POINTER_REGEX = re.compile(r'\.quilt/named_packages/([\w-]+/[\w-]+)/([0-9]{10})')
 
 
 def pkg_created_event(s3_event):
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -64,8 +64,8 @@ def pkg_created_event(s3_event):
     bucket_obj = s3_event_obj['bucket']
     bucket = bucket_obj['name']
     try:
-        resp = s3.get_object(Bucket=bucket, Key=key, Range=f'bytes=0-{EXPECTED_POINTER_SIZE}')
-    except s3.exceptions.NoSuckKey:
+        resp = s3.get_object(Bucket=bucket, Key=key, Range=f'bytes=0-{EXPECTED_POINTER_SIZE - 1}')
+    except s3.exceptions.NoSuchKey:
         logger.warning('pointer is created in bucket %r at %r, but not found', bucket, key)
         return
     if resp['ContentLength'] != EXPECTED_POINTER_SIZE:
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -72,7 +72,7 @@ def pkg_created_event(s3_event):
         'Detail': json.dumps({
             'bucket': bucket,
             'handle': pkg_name,
-            'top_hash': pkg_hash,
+            'topHash': pkg_hash,
         }),
     }
 
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -3,7 +3,6 @@ import json
 import re
 
 import boto3
-from dateutil.parser import isoparse
 
 event_bridge = boto3.client('events')
 s3 = boto3.client('s3')
@@ -63,7 +62,7 @@ def pkg_created_event(s3_event):
     version_id = obj['versionId']
     pkg_hash = s3.get_object(Bucket=bucket, Key=key, IfMatch=etag, VersionId=version_id)['Body'].read().decode()
     return {
-        'Time': isoparse(s3_event['eventTime']),
+        'Time': s3_event['eventTime'],
         'Source': 'quiltdata.pkg',
         'DetailType': 'created',
         'Resources': [
--- a/lambdas/pkgevents/index.py
+++ b/lambdas/pkgevents/index.py
@@ -1,0 +1,84 @@
+import itertools
+import json
+import re
+
+import boto3
+from dateutil.parser import isoparse
+
+event_bridge = boto3.client('events')
+s3 = boto3.client('s3')
+
+
+class EventsQueue:
+    MAX_SIZE = 10
+
+    def __init__(self):
+        self._events = []
+
+    def append(self, event):
+        self._events.append(event)
+        if len(self) >= self.MAX_SIZE:
+            self._flush()
+
+    def _flush(self):
+        events = self._events
+        self._events = []
+        event_bridge.put_events(Entries=events)
+
+    def flush(self):
+        if self:
+            self._flush()
+
+    def __len__(self):
+        return len(self._events)
+
+    def __bool__(self):
+        return bool(self._events)
+
+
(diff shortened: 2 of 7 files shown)


lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
+import itertools
```
akarve, line 1: "classifier" is confusing. prefer "events" or something related.

lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
...
+    bucket_obj = s3_event_obj['bucket']
+    bucket = bucket_obj['name']
+    etag = obj['eTag']
+    version_id = obj['versionId']
+    pkg_hash = s3.get_object(Bucket=bucket, Key=key, IfMatch=etag, VersionId=version_id)['Body'].read().decode()
+    manifest_key = MANIFEST_PREFIX_V1 + pkg_hash
+    resp = s3.head_object(Bucket=bucket, Key=manifest_key)
+    return {
+        'Time': datetime.fromisoformat(s3_event['eventTime'].rstrip('Z')),
+        'Source': 'quiltdata.pkg',
```
akarve, line 69: com.quiltdata? eventbridge.quiltdata?

lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
...
+        return
+    bucket_obj = s3_event_obj['bucket']
+    bucket = bucket_obj['name']
+    etag = obj['eTag']
+    version_id = obj['versionId']
+    pkg_hash = s3.get_object(Bucket=bucket, Key=key, IfMatch=etag, VersionId=version_id)['Body'].read().decode()
+    manifest_key = MANIFEST_PREFIX_V1 + pkg_hash
+    resp = s3.head_object(Bucket=bucket, Key=manifest_key)
+    return {
+        'Time': datetime.fromisoformat(s3_event['eventTime'].rstrip('Z')),
```
akarve, line 68: I think we should forward S3 time stamps unmodified and even preserve their naming conventions and keys: less is more. the more this looks like a standard AWS event, the easier users will understand it.

lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
...
+        'Resources': [
+            # TODO: use object arn
+            bucket_obj['arn'],
+            # TODO: add stack ARN?
+        ],
+        'Detail': json.dumps({
+            'bucket': bucket,
+            'name': pkg_name,
+            'top_hash': pkg_hash,
+            'manifest': {  # TODO: do we need this?
```
akarve, line 80: maybe not?

lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
...
+    bucket = bucket_obj['name']
+    etag = obj['eTag']
+    version_id = obj['versionId']
+    pkg_hash = s3.get_object(Bucket=bucket, Key=key, IfMatch=etag, VersionId=version_id)['Body'].read().decode()
+    manifest_key = MANIFEST_PREFIX_V1 + pkg_hash
+    resp = s3.head_object(Bucket=bucket, Key=manifest_key)
+    return {
+        'Time': datetime.fromisoformat(s3_event['eventTime'].rstrip('Z')),
+        'Source': 'quiltdata.pkg',
+        'DetailType': 'created',
```
akarve, line 70:
```suggestion
        'EventType': 'Package:Created',
```

lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
...
+        'Resources': [
+            # TODO: use object arn
+            bucket_obj['arn'],
+            # TODO: add stack ARN?
+        ],
+        'Detail': json.dumps({
+            'bucket': bucket,
+            'name': pkg_name,
```
akarve, line 78: handle is more consistent with docs
```suggestion
            'handle': pkg_name,
```

lambdas/classifier/app.py
akarve: i would actually consider including the ENTIRE original event under `rawEvent` or something. I know it's not efficient but it is hugely helpful for debuggin and maybe will enable users to do smart/clever things idk.

lambdas/classifier/app.py
akarve, line 70: Need some careful design around event types :) Let's look at the cross product. Consider having `eventType` and `eventAction` or `eventType:eventAction` but the latter is not as clean.

TYPE
* Package
* Object
* Folder
* Cron

ACTION
* Delete
* Create

lambdas/classifier/app.py
sir-sigurd: boto3 wants `datetime` objects here.

lambdas/classifier/app.py
```diff
@@ -0,0 +1,99 @@
...
+    if not '1451631600' <= pointer_name <= '1767250800':
+        return
+    bucket_obj = s3_event_obj['bucket']
+    bucket = bucket_obj['name']
+    etag = obj['eTag']
+    version_id = obj['versionId']
+    pkg_hash = s3.get_object(Bucket=bucket, Key=key, IfMatch=etag, VersionId=version_id)['Body'].read().decode()
+    manifest_key = MANIFEST_PREFIX_V1 + pkg_hash
+    resp = s3.head_object(Bucket=bucket, Key=manifest_key)
+    return {
```
akarve, line 67: Very important: `eventVersion`. I suggest reviewing all S3 fields and copying where appropriate. https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html

lambdas/classifier/app.py
sir-sigurd: Considering that we want to eliminate pointers I'd rather not include any info about them in package events, and adding info about manifest object could beneficial for v1/v2 compatibility.

lambdas/classifier/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM public.ecr.aws/lambda/python:3.8
```
sir-sigurd, line 1: @akarve 

https://www.python.org/downloads/release/python-380/
> Release Date: Oct. 14, 2019

lambdas/classifier/app.py
sir-sigurd: Probably we should consider https://docs.aws.amazon.com/eventbridge/latest/userguide/aws-events.html and look through https://docs.aws.amazon.com/eventbridge/latest/userguide/event-types.html.

akarve: Totally agree. In particular glue catalog events are a good pattern for us:

```
{
    "version": "0",
    "id": "2617428d-715f-edef-70b8-d210da0317a0",
    "detail-type": "Glue Data Catalog Table State Change",
    "source": "aws.glue",
    "account": "123456789012",
    "time": "2019-01-16T18:16:01Z",
    "region": "eu-west-1",
    "resources": [
        "arn:aws:glue:eu-west-1:123456789012:table/d1/t1"
    ],
    "detail": {
        "databaseName": "d1",
        "changedPartitions": [
            "[C.pdf, dir3]",
            "[D.doc, dir4]"
        ],
        "typeOfChange": "BatchCreatePartition",
        "tableName": "t1"
    }
}
```

lambdas/classifier/Dockerfile
akarve: Ok

lambdas/classifier/app.py
akarve: Also, we need an id field? Why not make it a hash of certain fields? We may need to dedupe messages based on this field.

lambdas/classifier/app.py
sir-sigurd: > Actually `eventBridge:quilt`.

https://docs.aws.amazon.com/eventbridge/latest/userguide/aws-events.html

> source

> Identifies the service that sourced the event. All events sourced from within AWS begin with "aws." Customer-generated events can have any value here, as long as it doesn't begin with "aws." **We recommend the use of Java package-name style reverse domain-name strings.**

i.e. it would be like `com.quiltdata` or `com.quiltdata.XXX`.

sir-sigurd: I looked through different event types and it looks like they use different property naming conventions: [some](https://docs.aws.amazon.com/eventbridge/latest/userguide/event-types.html#acd-event-types) use camel case and [others](https://docs.aws.amazon.com/eventbridge/latest/userguide/event-types.html#codepipeline-event-type) use hyphens. There are some recommendations to use camel case ([1](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format), [2](https://jsonapi.org/recommendations/#naming)), so I'm planning to stick to it.

.circleci/config.yml
```diff
@@ -226,6 +226,7 @@ workflows:
                 - thumbnail
                 - pkgselect
                 - shared
+                - pkgevents
```
akarve, line 229: pkgevents is too specific? try events, stackevents, etc?

lambdas/pkgevents/index.py
```diff
@@ -0,0 +1,103 @@
...
+        return
+    if resp['ContentLength'] != EXPECTED_POINTER_SIZE:
+        logger.warning('pointer in bucket %r at %r has %d bytes, but %d bytes expected',
+                       bucket, key, resp['ContentLength'], EXPECTED_POINTER_SIZE)
+        return
+
+    return {
+        'Time': s3_event['eventTime'],
+        'Source': 'com.quiltdata',
+        'DetailType': 'package-revision',
```
akarve, line 79:
```suggestion
        'DetailType': 'Quilt package state change',
```

.circleci/config.yml
akarve: Still feels too narrow; eventbridge or events is recommended to accommodate folder, object, etc. events

lambdas/pkgevents/index.py
```diff
@@ -0,0 +1,102 @@
...
+            self._flush()
+
+    def __len__(self):
+        return len(self._events)
+
+    def __bool__(self):
+        return bool(self._events)
+
+
+PKG_POINTER_REGEX = re.compile(r'\.quilt/named_packages/([\w-]+/[\w-]+)/(.{10})')
```
akarve, line 49: I think 10 chars is way to specific. It doesn't handle pointers, for instance (which I realize we don't respond to, but we will). If you really only care about numeric timestamps then `\d{10}` but again too specific? I would take something more like `(.+)$`.

PR #2082, opened 2021-02-24
Disable throwing on failing codecov upload

fiskus:


sir-sigurd (commented):
We can do this unless they fix the problem soon, but I'm not ready yet.

fiskus:
They said they fixed it https://github.com/codecov/codecov-bash/issues/411#issuecomment-785225047

PR #2085, opened 2021-02-24
Add default username prefix for package name at folder-to-package dialog

fiskus:


catalog/app/containers/Bucket/PackageDialog/PackageDialog.spec.ts
```diff
@@ -0,0 +1,18 @@
+import * as PD from './PackageDialog'
+
+describe('containers/Bucket/PackageDialog/PackageDialog', () => {
+  describe('useUsernamePrefix', () => {
```
nl0, line 4:
```suggestion
  describe('getUsernamePrefix', () => {
```

nl0 (approved):

PR #2090, opened 2021-03-05
Remove packaging from dependencies

sir-sigurd:
Seems to be unused since
2744c1273438dfdbcef7ed20a4944b3b1bbbd382


akarve (approved):
Looks harmless but @kevinemoore why did you include this 2 years ago?

PR #2091, opened 2021-03-05
Add env var to disable cache

sir-sigurd:
# Description

# TODO

- [x] Unit tests
- [x] add unit test for `PackageEntry.get_cached_path()`
- [x] Documentation

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1,3 +1,4 @@
+import contextlib
 import gc
 import hashlib
 import inspect
@@ -6,6 +7,7 @@ import json
 import os
 import pathlib
 import shutil
+import tempfile
 import textwrap
 import time
 import uuid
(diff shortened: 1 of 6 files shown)


akarve (approved):

akarve (approved):

api/python/quilt3/packages.py
```diff
@@ -583,18 +590,30 @@ def _browse(cls, name, registry=None, top_hash=None):
...
+                    # if the download gets interrupted.
+                    tmp_path = local_pkg_manifest.with_suffix('.tmp')
+                    download_manifest(tmp_path)
+                    tmp_path.rename(local_pkg_manifest)
+            else:
+                # This tmp file has to closed before downloading, because on Windows it can't be
+                # opened for concurrent access.
+                with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+                    local_pkg_manifest = tmp_file.name
+                    stack.callback(os.unlink, local_pkg_manifest)
```
dimaryaz, line 613: This seems a little bit over-engineered... Ideally, we'd just download the manifest and immediately deserialize it, without even writing to a temporary file - or create a temporary file, but use its file descriptor instead of opening it again. But I guess that's not supported right now.

Fine for now, but would be nice to clean up some day.

dimaryaz (approved):

PR #2092, opened 2021-03-05
Add env var for transfer concurrency

sir-sigurd:
# Description

# TODO

- [x] Unit tests
- [x] Documentation

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -45,11 +45,13 @@ from tenacity import (
 )
 from tqdm import tqdm
 
+from . import util
 from .session import create_botocore_session
 from .util import DISABLE_TQDM, PhysicalKey, QuiltException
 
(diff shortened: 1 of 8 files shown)


api/python/quilt3/data_transfer.py
```diff
@@ -52,6 +52,15 @@
 MAX_FIX_HASH_RETRIES = 3
 
 
+def get_int_from_env(var_name):
+    val = os.getenv(var_name)
+    if val:
+        return int(val)
+
+
+MAX_CONCURRENCY = get_int_from_env('QUILT_TRANSFER_MAX_CONCURRENCY') or 10
```
akarve, line 61: um this will fail for negative values, so we should check for the same in `get_int_from_env`

akarve (changes requested):

akarve (approved):

PR #2095, opened 2021-03-10
allow callable for dest parameter of Package.push()

sir-sigurd:
# Description

# TODO

- [x] Unit tests
- [x] Documentation

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1035,16 +1035,17 @@ class Package:
         registry = get_package_registry(registry)
 
         self._set_commit_message(message)
-
         self._fix_sha256()
-        manifest = io.BytesIO()
-        self._dump(manifest)
 
         top_hash = self.top_hash
+        self._push_manifest(name, registry, top_hash)
+        return top_hash
+
+    def _push_manifest(self, name, registry, top_hash):
+        manifest = io.BytesIO()
+        self._dump(manifest)
         self._timestamp = registry.push_manifest(name, top_hash, manifest.getvalue())
         self._origin = PackageRevInfo(str(registry.base), name, top_hash)
-
-        return top_hash
 
     @ApiTelemetry("package.dump")
     def dump(self, writable_file):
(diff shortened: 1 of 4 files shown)


api/python/quilt3/packages.py
```diff
@@ -1352,7 +1367,7 @@ def selector_fn(*args):
             # Copy the datafiles in the package.
             physical_key = entry.physical_key
 
-            new_physical_key = dest_parsed.join(logical_key)
+            new_physical_key = dest_fn(logical_key, entry, top_hash)
```
sir-sigurd, line 1370: Not passing package name here since it's already known before push.

api/python/quilt3/packages.py
```diff
@@ -1323,15 +1329,21 @@ def selector_fn(*args):
                     f"'build' instead."
                 )
 
-        if dest is None:
-            dest_parsed = registry_parsed.join(name)
+        if callable(dest):
+            def dest_fn(*args, **kwargs):
+                return registry_parsed.join(dest(*args, **kwargs))
```
akarve, line 1334: I'm not sure if I'm reading this right, but if the user provides a dest callable then we should give them control of the entire string, i.e. they might want to set the physical key to any registry/bucket. Here it looks like we restrict to registry= value?

sir-sigurd: reworked it

api/python/quilt3/packages.py
```diff
@@ -1387,19 +1412,19 @@ def physical_key_is_temp_file(pk):
...
 
             if user_is_configured_to_custom_stack():
                 navigator_url = get_from_config("navigator_url")
 
                 print(f"Successfully pushed the new package to "
-                      f"{catalog_package_url(navigator_url, dest_parsed.bucket, name, tree=False)}")
+                      f"{catalog_package_url(navigator_url, registry.base.bucket, name, tree=False)}")
             else:
-                dest_s3_url = str(dest_parsed)
+                dest_s3_url = str(registry.base)  # FIXME
```
sir-sigurd, line 1427: Not sure what should be printed here.

api/python/quilt3/packages.py
```diff
@@ -1277,6 +1283,8 @@ def push(self, name, registry=None, dest=None, message=None, selector_fn=None, *
         Args:
             name: name for package in registry
             dest: where to copy the objects in the package
+                This must be either S3 URI in bucket specified by registry or a callable that takes
+                logical_key, package_entry and top_hash and returns S3 URI.
```
akarve, line 1287:
```suggestion
                Must be either an S3 URI prefix in the registry bucket, or a callable that takes
                logical_key, package_entry, and top_hash and returns S3 URI. S3 URIs start with s3://$bucket/$path/.
```

api/python/quilt3/packages.py
akarve: Well the package still has a destination, so we should print the value of registry=, which I think is what you have?

akarve (approved):
Added some suggested text improvements.

api/python/quilt3/packages.py
```diff
@@ -1277,6 +1283,8 @@ def push(self, name, registry=None, dest=None, message=None, selector_fn=None, *
         Args:
             name: name for package in registry
             dest: where to copy the objects in the package
+                Must be either an S3 URI prefix in the registry bucket, or a callable that takes
+                logical_key, package_entry, and top_hash and returns S3 URI. S3 URIs start with s3://$bucket/$path/.
```
sir-sigurd, line 1287:
```suggestion
                logical_key, package_entry, and top_hash and returns S3 URI. S3 URIs format is s3://$bucket/$key.
```

PR #2104, opened 2021-03-15
Refactor spreadsheets tests

fiskus:
* Move repeated unit-tests code to functions
* Make JSON converted from spreadsheets more explicit: replace undefined with null, fill empty values with nulls (`[,,]` → `[null, null]`)
* Add `xlsx` files and try to avoid its specifics: remove empty arrays, don't test for Dates
---
Parsing Excel dates creates timezoned dates (+4 hours), but then they are treated like UTC:
1. Input to Excel — 2020-01-01
2. Unit-test output ­— 2019-12-12

I didn't find where the root of this problem yet, and ignored testing dates for now.


catalog/app/utils/spreadsheets/spreadsheets.ts
```diff
@@ -27,7 +34,16 @@ export function parseSpreadsheet(
...
   })
-  return rowsToJson(transpose ? R.transpose(rows) : rows)
+  const maxSize = rows.reduce((memo, row) => R.max(memo, row.length), 0)
+  return pipeThru(rows)(
+    R.without([[]]),
+    R.map((row: any[]) => {
+      const nullsTail = R.repeat(null, maxSize - row.length)
+      return R.concat(row, nullsTail)
+    }),
+    R.when(() => transpose, R.transpose),
```
nl0, line 44: since the computation doesnt depend on the actual values, smth like this seems more logical: `transpose ? R.transpose : R.identity,`
tho it doesnt make much difference

nl0 (approved):

PR #2105, opened 2021-03-15
Hide Bucket's tab

fiskus:
* rename workflowsList to workflowsConfig
* update "js-yaml"
* load `catalog/config.yml` and hide tabs and buttons if the preference set


api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -112,6 +112,32 @@
           }
         }
       }
+    },
```
sir-sigurd, line 115: I should reuse it from the shared folder...

catalog/app/utils/bucketPreferences.ts
```diff
@@ -0,0 +1,110 @@
...
+    // eslint-disable-next-line no-console
+    console.log('Unable to fetch')
+    // eslint-disable-next-line no-console
+    console.error(e)
+    throw e
+  }
+}
+
+export function useBucketPreferences(bucket: string): BucketPreferences | null {
+  // FIXME: fetch only on bucket change
```
fiskus, line 99: I want to have simple API:

```tsx
const preferences = usePreferences(bucket) // fetches document only once, then fetches on bucket change
return (
  <div>
    {preferences.flag && <span>Show flagged content</span>}
  </div>
)
```

@nl0 Is there a way to do this using existing code from `app/utils`? Or, maybe, this API doesn't fit with you?

nl0: > Is there a way to do this using existing code from app/utils?

yep.
we can use something similar to `utils/BucketConfig` to achieve this or create some provider component and render it in the `Bucket` component, so that every component under it could pull it from the context. ping me in slack if you want to discuss this further.

catalog/app/containers/Bucket/BucketNav.tsx
```diff
@@ -0,0 +1,84 @@
...
+      minWidth: 120,
+    },
+  })),
+  RC.withProps({ component: Link }),
+  M.Tab,
+)
+
+interface BucketNavProps {
+  bucket: string
+  section: string | boolean
```
nl0, line 25: this probably should be a union of specific sections we have

catalog/app/utils/bucketPreferences.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  return {
+    ui: {
+      actions: R.mergeRight(defaultPreferences.ui.actions, data?.ui?.actions || {}),
+      nav: R.mergeRight(defaultPreferences.ui.nav, data?.ui?.nav || {}),
+    },
+  }
+}
+
+const BUCKET_PREFERENCES_PATH = '.quilt/catalog/config.yml'
```
nl0, line 67: it's still not 100% clear which extension should be used: we already use `yml` for workflows, but `yaml` is a recommended option

shared/schemas/workflows.yml.json
```diff
@@ -112,6 +112,32 @@
           }
         }
       }
+    },
+    "ui": {
```
nl0, line 116: this doesnt seem relevant anymore

nl0 (approved):
looks ok, some minor comments inline. pls decide if you want to make the config fetching happen once per bucket here or in a clean PR

catalog/app/utils/bucketPreferences.ts
fiskus: It's recommended, just like /dʒɪf/ is recommended spelling of "gif" :)
`.yml` is widely used, and I think it's more predictable.

fiskus: I use both. And unconsciously tend to use `.yaml` at formal documents and `.yml` in the wild.
I separate this discussion to Slack

nl0 (approved):

PR #2107, opened 2021-03-16
Save search mode to localStorage and redirect to last used one

fiskus:
1. If user manually select search mode, save it to localStorage
2. If page is opened without `?mode=`, try to restore mode from localStorage


nl0 (approved):
looks good, tho the setting isnt scoped to a bucket (@akarve should we scope this to a bucket, i.e. remember last used mode for every bucket?). also, should we remember mode for the global search?

nl0 (approved):

PR #2108, opened 2021-03-16
Drag'n'drop target visuals

fiskus:
Screenshot from 2021-03-17 12-16-06


catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
```diff
@@ -551,7 +566,12 @@ export const MetaInput = React.forwardRef(function MetaInput(
         </Lab.ToggleButtonGroup>
       </div>
 
-      <div {...getRootProps({ className: classes.dropzone })} tabIndex={undefined}>
+      <div
+        {...getRootProps({
+          className: classes.dropzone,
+        })}
+        tabIndex={undefined}
+      >
```
nl0, line 574: this change doesnt seem necessary (or is it prettier?)

catalog/app/utils/dragging.ts
```diff
@@ -0,0 +1,19 @@
...
+    document.addEventListener('dragover', () => {
+      setDragActive(true)
+    })
+    document.addEventListener('dragleave', () => {
+      setDragActive(false)
+    })
+    document.addEventListener('drop', () => {
+      setDragActive(false)
+    })
+  }, [])
```
nl0, line 16: probably we should remove event listeners in the cleanup function

nl0 (changes requested):
overall looks good, some comments inline, plus not very nice looking gaps appeared around the file list:

borders

catalog/app/utils/dragging.ts
```diff
@@ -3,17 +3,20 @@ import * as React from 'react'
 export default function useDragging() {
   const [isDragActive, setDragActive] = React.useState(false)
 
+  const setActive = React.useCallback(() => setDragActive(true), [])
+  const setInactive = React.useCallback(() => setDragActive(false), [])
```
nl0, line 7: doesnt eslint complain about non-exhaustive dependencies? i know that setState is stable and can be omitted from dep list, but i thought eslint doesnt know this

nl0 (changes requested):
now files border is made invisible and the dashed outline is inset
files-border

also, meta outline pushes the contents when active, and the original border is also hidden, which doesnt look very nice imo, especially at the junctions:
meta-borders

what i propose is just overlaying the outline like this:
border-overlay

catalog/app/utils/dragging.ts
fiskus: Eslint knows this

nl0 (approved):

PR #2110, opened 2021-03-17
Increase size of drop area

fiskus:
Screenshot from 2021-03-17 14-40-52


nl0 (approved):
seems a little bit too much, but whatever

PR #2114, opened 2021-03-17
limit record size of manifest

sir-sigurd:
# TODO

- [x] Unit tests
- [x] Documentation

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1098,7 +1098,7 @@ class Package:
                     'Quilt recommends less than 1MB of metadata per object, '
                     'and less than 1MB of package-level metadata. '
                     'This enables S3 select, Athena and downstream services '
-                    'to work correctly. This limit can be overridden with '
+                    'to work correctly. This limit can be overridden with the '
                     'QUILT_MANIFEST_MAX_RECORD_SIZE environment variable.'
                 )
             return data
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1094,7 +1094,7 @@ class Package:
                 entry_text = 'package metadata' if lk is None else f'entry with logical key {lk!r}'
                 raise QuiltException(
                     f"Size of manifest record for {entry_text} is {encoded_size} bytes, "
-                    f"but it's limited by {MANIFEST_MAX_RECORD_SIZE} bytes. "
+                    f"but must be less than {MANIFEST_MAX_RECORD_SIZE} bytes. "
                     'Quilt recommends less than 1MB of metadata per object, '
                     'and less than 1 MB of package-level metadata. '
                     'This enables S3 select, Athena and downstream services '
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1094,7 +1094,7 @@ class Package:
                 entry_text = 'package metadata' if lk is None else f'entry with logical key {lk!r}'
                 raise QuiltException(
                     f"Size of manifest record for {entry_text} is {encoded_size} bytes, "
-                    f"but it's limited by {MANIFEST_MAX_RECORD_SIZE} bytes."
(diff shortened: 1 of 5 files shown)


api/python/quilt3/packages.py
```diff
@@ -1065,7 +1072,15 @@ def dump(self, writable_file):
         return self._dump(writable_file)
 
     def _dump(self, writable_file):
-        writer = jsonlines.Writer(writable_file)
+        json_encode = json.JSONEncoder(ensure_ascii=False).encode
+
+        def dumps(obj):
+            data = json_encode(obj).encode()
+            if len(data) > MAX_MANIFEST_RECORD_SIZE:
```
sir-sigurd, line 1079: Maybe we should set this limit on the size of user metadata, but then we need to be sure that we have enough space for our own data.

akarve: I think you are doing it right: we should measure the TOTAL size of the row because that's what S3 select will care about.

api/python/quilt3/packages.py
```diff
@@ -1065,7 +1071,15 @@ def dump(self, writable_file):
...
+                raise QuiltException('blah-blah')  # FIXME
```
akarve, line 1079:
```
'Quilt recommends less than 1MB of metadata per object, and less than 1MB of package-level metadata. This enables S3 select, Athena and downstream services to work correctly. You can override this exception as follows...
```

akarve (dismissed):

api/python/quilt3/packages.py
```diff
@@ -1074,7 +1084,26 @@ def dump(self, writable_file):
...
+            if encoded_size > MANIFEST_MAX_RECORD_SIZE:
+                lk = obj.get('logical_key')
+                entry_text = 'package metadata' if lk is None else f'entry with logical key {lk!r}'
+                raise QuiltException(
+                    f"Size of manifest record for {entry_text} is {encoded_size} bytes, "
+                    f"but it's limited by {MANIFEST_MAX_RECORD_SIZE} bytes."
+                    'Quilt recommends less than 1MB of metadata per object, '
+                    'and less than 1MB of package-level metadata. '
+                    'This enables S3 select, Athena and downstream services '
+                    'to work correctly. This limit can be overridden with '
```
akarve, line 1101:
```suggestion
                    'to work correctly. This limit can be overridden with the '
```

akarve (approved):

api/python/quilt3/packages.py
```diff
@@ -1074,7 +1084,26 @@ def dump(self, writable_file):
...
+
+        def dumps(obj):
+            data = json_encode(obj)
+            encoded_size = len(data.encode())
+            if encoded_size > MANIFEST_MAX_RECORD_SIZE:
+                lk = obj.get('logical_key')
+                entry_text = 'package metadata' if lk is None else f'entry with logical key {lk!r}'
+                raise QuiltException(
+                    f"Size of manifest record for {entry_text} is {encoded_size} bytes, "
+                    f"but it's limited by {MANIFEST_MAX_RECORD_SIZE} bytes. "
```
akarve, line 1097:
```suggestion
                    f"but should be less than {MANIFEST_MAX_RECORD_SIZE} bytes. "
```

api/python/quilt3/packages.py
```diff
@@ -1074,7 +1084,26 @@ def dump(self, writable_file):
...
+                    'Quilt recommends less than 1MB of metadata per object, '
```
akarve, line 1098: Maybe 1MB should depend on MANIFEST_MAX_RECORD_SIZE, idk.

api/python/quilt3/packages.py
sir-sigurd: "should" would make sense if it was a warning, but it's an exception. Use "must" instead?

akarve: Yeah, fixed

PR #2115, opened 2021-03-18
add more logging

sir-sigurd:


api/python/quilt3/packages.py
```diff
@@ -51,6 +52,8 @@
     validate_package_name,
 )
 
+logger = logging.getLogger(__name__)
```
akarve, line 55: Mmm, do we really want a new logger for this? Not a big deal but why not do what we do in Lambdas and have a universally-configured logger?

See /Users/karve/code/quilt/lambdas/shared/t4_lambda_shared/utils.py.

akarve (approved):

PR #2118, opened 2021-03-19
Load one file for the list of paths

fiskus:
API for the `fetchFile` is preserved:

```ts
fetchFile({ s3, bucket, path: 'path-to-the-file' })
```

with additional support for the list of paths:

```ts
fetchFile({ s3, bucket, path: ['fallback-path', 'most-probably-existing-file'] })
```


nl0 (changes requested):
the implementation seems counter-intuitive for couple of reasons:
- only the first and the last items of the list are tried, other items possibly existing in the middle are ignored
- last item is tried first, which doesnt make sense to me

proposition: iterate thru the list from start and return the first positive result

catalog/app/containers/Bucket/requests.js
```diff
@@ -218,8 +218,28 @@ export const bucketStats = async ({ req, s3, bucket, overviewUrl }) => {
...
+
+  return (
+    fileExists ||
+    (await ensureObjectIsPresentInCollection({ s3, bucket, key: R.init(key) }))
+  )
+}
+
 const fetchFileVersioned = async ({ s3, bucket, path, version }) => {
-  const versionExists = await ensureObjectIsPresent({
+  const isPresent = Array.isArray(path)
```
nl0, line 239: we could just normalize path (convert it to a list) and use `ensureObjectIsPresentInCollection`

fiskus:
> only the first and the last items of the list are tried, other items possibly existing in the middle are ignored

You probably confused `R.init` with `R.head`. `R.head` returns first element of array, but `R.init` returns _array_ without last element.

I double-checked: all items on the list are tried

nl0:
> > only the first and the last items of the list are tried, other items possibly existing in the middle are ignored
> 
> You probably confused `R.init` with `R.head`. `R.head` returns first element of array, but `R.init` returns _array_ without last element.
> 
> I double-checked: all items on the list are tried

yeah i see now, but still i think it's more logical to use the natural order (try the first item of the array first)

fiskus:
I turned the list of file paths over

catalog/app/containers/Bucket/requests.js
```diff
@@ -218,11 +218,29 @@ export const bucketStats = async ({ req, s3, bucket, overviewUrl }) => {
...
 }
 
+// key is array of keys: string[]
+const ensureObjectIsPresentInCollection = async ({ s3, bucket, keys, version }) => {
+  if (!keys.length) return null
+
+  const fileExists = await ensureObjectIsPresent({
+    s3,
+    bucket,
+    key: R.head(keys),
```
nl0, line 228: `keys[0]`? =)

nl0 (approved):

PR #2121, opened 2021-03-22
Prepopulate date

fiskus:
Treat pre-population of dates as an extension for pre-population of defaults.

When schema has `default` keyword, its value is used instead of `undefined`.
When schema has `{ format: 'date', dateformat: 'yyyy-MM-dd' }`, the current date is used instead of `undefined`.

Implementation:

* Add visual placeholders for defaults and dates
* Prepopulate defaults and default dates on metadata submit

Additional work:

* Convert `json-schema.js` to Typescript 
* Replace Ajv({ useDefaults: true }) with custom implementation because we need to assign dates too

---

Besides type conversion I added functions `scanSchemaAndPrefillValues`, `getDefaultValue` and replaced `makeSchemaDefaultsSetter`

- [x] Unit tests
- [ ] Documentation


catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+    properties: Record<string, JsonSchema>
+    type: string | string[] | JsonSchema[]
+  } & Record<CompoundCondition, JsonSchema[]>
+>
+
+// NOTE: can't use R.prop directly, because R.prop type doesn't allow undefined
+const prop = (
+  property: keyof JsonSchema,
+  optSchema?: JsonSchema,
+): JsonSchema[keyof JsonSchema] => (optSchema ? R.prop(property, optSchema) : undefined)
```
nl0, line 25: i think we can default `optSchema` to `{}` and then just return `optSchema[property]`, and maybe call it `propSafe` or `propPartial` idk

catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export const isSchemaArray = (optSchema?: JsonSchema) =>
+  prop('type', optSchema) === 'array'
+
+export const isSchemaObject = (optSchema?: JsonSchema) =>
+  prop('type', optSchema) === 'object'
+
+const isSchemaString = (optSchema?: JsonSchema) => prop('type', optSchema) === 'string'
```
nl0, line 33: actually, seems like this `prop` helper doesnt give us much since we can just write smth like
```js
const isSchemaString = (schema?: JsonSchema = {}) => schema.type === 'string'
```

nl0: or use optional chaining
```js
const isSchemaString = (schema?: JsonSchema) => schema?.type === 'string'
```

catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+export const isNestedType = R.either(isSchemaArray, isSchemaObject)
+
+function compoundTypeToHumanString(
+  optSchema: JsonSchema,
+  condition: CompoundCondition,
+  divider: string,
+): string {
+  if (!Array.isArray(prop(condition, optSchema))) return ''
+
+  return (optSchema[condition] as JsonSchema[])
```
nl0, line 72: why is it necessary to cast the value to `JsonSchema[]` here? seems like it should be resolved to that automatically. if that's because it's possibly undefined, we could just use a trailing `!`

catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+    //       that's why they are first
+    [
+      R.propOr('', 'type'),
+      () =>
+        Array.isArray(optSchema.type)
+          ? optSchema.type.join('|')
+          : (optSchema.type as string),
+    ],
+    [isSchemaAnyOf, () => compoundTypeToHumanString(optSchema, 'anyOf', '|')],
+    [isSchemaOneOf, () => compoundTypeToHumanString(optSchema, 'oneOf', '&')],
```
nl0, line 95: what's the difference btw `anyOf` and `oneOf` in this context? seems like both these cases could use `|` and leave `&` for `allOf`

catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+  } catch (e) {
+    // TODO: add custom errors
+    return () => [e]
+  }
+}
+
+function scanSchemaAndPrefillValues(
+  callback: (v?: any, s?: JsonSchema) => any,
+  optValue: Record<string, any>,
```
nl0, line 167: why is this `opt`?

catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+      )
+
+    if (schemaItem.items && Array.isArray(valueItem))
+      return R.assoc(
+        key,
+        valueItem.map((v) => scanSchemaAndPrefillValues(callback, v, schemaItem.items)),
+        memo,
+      )
+
+    const preDefinedValue = callback(schemaItem)
```
nl0, line 196: it would be nice to rename `callback` into smth more descriptive, like `getValue` or smth

catalog/app/utils/json-schema/json-schema.ts
```diff
@@ -0,0 +1,221 @@
...
+      return validate.errors || []
+    }
+  } catch (e) {
+    // TODO: add custom errors
+    return () => [e]
+  }
+}
+
+function scanSchemaAndPrefillValues(
+  callback: (v?: any, s?: JsonSchema) => any,
```
nl0, line 166: seems like the signature is incorrect -- first argument should be removed

nl0 (approved):
looks good, some minor comments inline

PR #2122, opened 2021-03-23
Multi folder push

nl0:
# TODO

- [ ] Unit tests
- [ ] Documentation


catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -375,78 +359,42 @@ const useFileStyles = M.makeStyles((t) => ({
...
-          hint: 'Restore',
-          icon: 'undo',
-          handler: handle(FilesAction.Revert),
-        },
-        unchanged: { hint: 'Delete', icon: 'clear', handler: handle(FilesAction.Delete) },
-      }[type]),
-    [type, handle],
-  )
+type FileProps = FileOwnProps &
+  Omit<React.HTMLAttributes<HTMLDivElement>, keyof FileOwnProps>
```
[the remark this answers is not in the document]
fiskus: even this

```tsx
interface FileProps extends React.HTMLAttributes<HTMLDivElement> {
  name: string
  state?: FilesEntryState
  size?: number
  action?: React.ReactNode
}
```

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -536,257 +477,69 @@ const useDirStyles = M.makeStyles((t) => ({
...
+
+type DirProps = DirOwnProps &
+  Omit<React.HTMLAttributes<HTMLDivElement>, keyof DirOwnProps>
+
+const Dir = React.forwardRef<HTMLDivElement, DirProps>(function Dir(
+  {
+    name,
+    state = 'unchanged',
+    active = false,
+    empty = false,
```
fiskus, line 499: As I can see exposing `empty` is redundant because it uses `children.length`

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -845,15 +591,449 @@ function DropzoneMessage({ error, warn, disabled }: DropzoneMessageProps) {
...
+  progress?: {
+    total: number
+    loaded: number
+    percent: number
+  }
+}) {
+  const classes = useLockStyles()
+  return (
+    <div className={classes.root}>
+      {!!progress && (
```
fiskus, line 700: Early exit will make code slightly clearer:

```tsx
  const classes = useLockStyles()

  if (!progress) return <div className={classes.root} />

  return (
    <div className={classes.root}>
      <div className={classes.progressContainer}>
        …
      </div>
    </div>
  )
```

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -845,15 +591,449 @@ function DropzoneMessage({ error, warn, disabled }: DropzoneMessageProps) {
...
+          )}
+        </>
+      )}
+    </div>
+  )
+}
+
+const useFilesContainerStyles = M.makeStyles((t) => ({
+  root: {
+    direction: 'rtl', // show the scrollbar on the left
```
fiskus, line 727: :(

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -845,15 +591,449 @@ function DropzoneMessage({ error, warn, disabled }: DropzoneMessageProps) {
...
+          icon: 'undo',
+          handler: handle(FilesAction.Revert),
+        },
+        unchanged: { hint: 'Delete', icon: 'clear', handler: handle(FilesAction.Delete) },
+      }[state]),
+    [state, handle],
+  )
+
+  const onClick = React.useCallback((e: React.MouseEvent) => {
+    e.stopPropagation()
```
fiskus, line 893: It worth writing down the reason in comments

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -845,15 +591,449 @@ function DropzoneMessage({ error, warn, disabled }: DropzoneMessageProps) {
...
+  const toggleExpanded = React.useCallback(
+    (e) => {
+      e.stopPropagation()
+      setExpanded((x) => !x)
+    },
+    [setExpanded],
+  )
+
+  const onClick = React.useCallback((e: React.MouseEvent) => {
+    e.stopPropagation()
```
fiskus, line 931: It blocks opening OS's file manager, I suppose? I think you should mention it in comment

fiskus: There is `noClick` option for "react-dropzone". Probably, it will help

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -1016,88 +1195,207 @@ export function FilesInput({
...
+              </M.Icon>
+            }
+          >
+            Select {selected < value.length ? 'all' : 'none'}
+          </M.Button>
+        )}
+      </Header>
+
+      <ContentsContainer>
+        <Contents error={!!error} warn={truncated}>
```
fiskus, line 1353: I think it's better to rename `error` prop to something else, like `hasError` or `invalid`

fiskus (approved):
Small UI issues:
* Directory has `cursor: pointer`, file has `cursor: default`
* Hard to see which entry is unselected, probably we can dim unselected one

Screenshot from 2021-03-24 09-37-45

nl0:
@fiskus 

> Directory has cursor: pointer, file has cursor: default

fixed

> Hard to see which entry is unselected, probably we can dim unselected one

yep, good idea, i've implemented that

catalog/app/containers/Bucket/Dir.tsx
```diff
@@ -25,9 +25,34 @@ import Summary from './Summary'
 import { displayError } from './errors'
 import * as requests from './requests'
 
+interface RouteMap {
```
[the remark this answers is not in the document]
nl0: it's the part of the route map relevant to the components in this file

> It's the auxiliary type, not the actual map, isn't it?

well, yeah, kinda

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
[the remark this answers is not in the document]
nl0: the point of this is to exclude props that are not passed to the div from the type to avoid possible conflicts and to articulate that those props arent related to the underlying html element

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: there's a little nuance here: you may have empty children and not want to display the "<EMPTY>" placeholder  -- and that's exactly the case with the peer folders selection -- we dont support showing nested objects, so we just need to render the dir name and not its empty contents

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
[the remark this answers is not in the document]
fiskus: I wrote the same: "use HTMLAttributes from DivElement but re-define name, state size and action", didn't?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: i dont think so bc in this particular case it adds repetition, while the condition just hides some nested elements, which seems idiomatic to me

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: ?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: why? MUI uses similar convention and i don's see what's wrong with it

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: It looks bad in my opinion. I hoped you would realize that after a while, but you stuck with this decision and scale it to another Dialog

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: `// stop click from propagating to parent elements and triggering their handlers`

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: It's harder to read and maintain code if identically named variables have different content

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: the main reason for that is scrollbar on macos blocking the buttons on the right from being clicked, it's not an aesthetics-based decision, but more like a workaround for a usability issue, and i'm open to alternatives if you can provide any.

it looks okayish imo since most of the time the scrollbar is hidden (on macos at least).

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: > It blocks opening OS's file manager, I suppose?

nope, it prevents the click from propagating to parent elements and triggering click handlers there.
adding the comment

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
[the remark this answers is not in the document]
nl0: i was about to say that you cant override a prop of an interface with incompatible type, but seems like i was confusing declaration merging with plain interface extension, so you're probably right

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: well, they have similar meaning

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -259,6 +245,11 @@ export const validateHashingComplete = (state: FilesState) => {
   return undefined
 }
 
+export const validateNonEmptySelection = (state: FilesSelectorState) => {
+  if (state.every(R.propEq('selected', false))) return 'emptySelection'
```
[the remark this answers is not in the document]
nl0: ok

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: The alternative is obvious, so I think you already declined it: show scrollbar at the right side, probably add some space with padding.

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: The next team member will need to add some components above in the hierarchy, and he will have bad times finding why his `onClick` handler doesn't work. Finally, found this `stopPropagation` he will need some glue to refactor this and not to break already existing code. You can help him by providing the comment, explaining the reason behind this architectural decision.

If you don't understand why it's important, don't bother, we can get back to this when the problem becomes real if it does.

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: padding on the right looks as bad as the scrollbar on the left)
one alternative that comes to mind is (ab)using horizontal scrollbar (akin to smth like telegram embedded reader), but this seems even more odd and unconventional for a primarily desktop-oriented (web)app

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: but i'm not arguing here -- i've added the comments)
or do you mean these comments should be more detailed?

fiskus: I'm asking for
```js
// use Ohm's law to calculate voltage,
// see ohm-i.tsx and ohm-r.tsx
const u = i* r
```
rather then
```js
// assign `u` to the `i` multiplied by `r`, so `u` will be the product of `i` and `r` multiplication
const u = i * r
```
---
Again, if you don't understand what I mean, we can get back to this when the real problem occurs

PR #2124, opened 2021-03-25
Paginate pkgselect

kevinemoore:
# Description
Adding `limit` and `offset` parameters to the `pkgselect` lambda to support paging the response (e.g., to avoid the 10MB lambda payload size limit).

# TODO

- [x] Unit tests
- [ ] Documentation


nl0 (approved):
looks good, but it would be really helpful to have the total item count in response to be able to paginate conveniently and efficiently

kevinemoore:
> looks good, but it would be really helpful to have the total item count in response to be able to paginate conveniently and efficiently

Good idea. I added return fields `total` and `returned` to pass back the total results (objects + prefixes) and the number returned in the current call.

nl0 (approved):
👍

PR #2126, opened 2021-03-25
sort packages by modification time by default

nl0:


fiskus:
Sort by name is useful because I can remember every package name and their position on the page. And I can scroll unconsciously to the right position. Like, "I need the "fiskus/sandbox" package, it's on three wheel-scrolls from top".

The point is: I think it would be useful to save my preference for sorting to the localStorage.

fiskus (approved):

akarve (approved):

nl0:
> The point is: I think it would be useful to save my preference for sorting to the localStorage.

makes sense, backlogging this

PR #2130, opened 2021-03-26
Use checkboard for transparent images

fiskus:
Screenshot from 2021-03-26 18-33-28
Screenshot from 2021-03-26 18-33-57


nl0 (approved):
nice. i didnt realize you can set the background for the actual image element

PR #2137, opened 2021-03-29
Package handle template

fiskus:
- [x] Unit tests
- [ ] Documentation

### DIFF
--- a/catalog/app/utils/packageHandle.spec.ts
+++ b/catalog/app/utils/packageHandle.spec.ts
@@ -1,0 +1,57 @@
+import * as packageHandle from './packageHandle'
+
+describe('utils/packageHandle', () => {
+  describe('getValueName', () => {
+    // eslint-disable-next-line no-template-curly-in-string
+    expect(packageHandle.getValueName('${{ username }}')).toBe('username')
+  })
+
+  describe('convertItem', () => {
+    it('should return static string when no template', () => {
+      expect(packageHandle.convertItem('fgsfds')).toBe('fgsfds')
+      expect(packageHandle.convertItem('')).toBe('')
+    })
+
+    it('should return template when template is broken', () => {
+      expect(packageHandle.convertItem('start and ${{ no end')).toBe(
+        'start and ${{ no end',
+      )
+    })
+
+    it('should return converted when template has values', () => {
+      expect(
+        packageHandle.convertItem(
+          // eslint-disable-next-line no-template-curly-in-string
+          'what-${{ username }}-do/make-${{ directory }}-update',
+          { directory: 'staging', username: 'fiskus' },
+        ),
+      ).toBe('what-fiskus-do/make-staging-update')
+      expect(
+        packageHandle.convertItem(
+          // eslint-disable-next-line no-template-curly-in-string
+          '${{ username }}/${{ directory }}',
+          { directory: 'staging', username: 'fiskus' },
+        ),
+      ).toBe('fiskus/staging')
+    })
+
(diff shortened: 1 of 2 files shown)


catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -580,32 +581,56 @@ function PackageCreateDialog({
...
+      initialValues={initialValues}
       subscription={{
-        handleSubmit: true,
-        submitting: true,
-        submitFailed: true,
         error: true,
-        submitError: true,
+        handleSubmit: true,
         hasValidationErrors: true,
-        form: true,
```
fiskus, line 602: Removed `form` only, and sort other properties alphabetically

catalog/app/containers/Bucket/PackageCreateDialog.js
```diff
@@ -580,32 +581,56 @@ function PackageCreateDialog({
...
+        initialValues: true,
```
fiskus, line 625: Also, I added `initialValues`

shared/schemas/workflows.yml.json
```diff
@@ -2,11 +2,27 @@
...
+        },
+        "packages": {
+          "$ref": "#/definitions/handleTemplate"
+        }
+      }
+    }
+  },
   "properties": {
     "version": {
       "const": "1"
```
fiskus, line 28: @sir-sigurd I don't know should I increment version? I've added additional property `package_handle`, probably version should be `1.1` now?

catalog/app/containers/Bucket/PackageDirectoryDialog.tsx
```diff
@@ -258,24 +260,52 @@ function DialogForm({
...
+    (workflow) =>
+      packageHandle.convert(workflow?.packageHandle, 'files', {
+        directory: s3paths.ensureNoSlash(path),
+        username: PD.getUsernamePrefix(username),
+      }),
+    [path, username],
+  )
+
+  const [initialValues, setInitialValues] = React.useState({
+    files: initialFiles,
```
nl0, line 275: this way files initial value wont be updated on input props change (it shouldnt be a problem in practice bc it wont be changing while the dialog is open, but anyways, we should keep that in mind).
btw why did you consolidate all the initial values into one object and put it into state var?
what are the benefits of this vs using separate computed and state vars?

api/python/quilt3/workflows/config-1.schema.json
```diff
@@ -7,6 +7,25 @@
     "workflows"
   ],
   "additionalProperties": false,
+  "definitions": {
```
nl0, line 10: do we have duplicate schema files in python code and catalog?

nl0 (approved):
looks ok, tho i'm not sure about using `initialValues` object persisted in react state vs granular values for individual fields.

catalog/app/containers/Bucket/PackageDirectoryDialog.tsx
fiskus: Just one benefit — it works :)
I couldn't make it working with updating `initialValue` per field, they just don't update

api/python/quilt3/workflows/config-1.schema.json
fiskus: yes

nl0:
probably `workflows` and `packageHandle` stuff belongs to `model`, since it's our domain logic, but ofc we can move this around later as a part of some future refactoring

PR #2139, opened 2021-03-30
pkgpush lambda

sir-sigurd:
# Description

# TODO

- [x] Unit tests
- [x] Unit tests for telemetry

### DIFF
--- a/lambdas/pkgpush/index.py
+++ b/lambdas/pkgpush/index.py
@@ -1,0 +1,360 @@
+import contextlib
+import functools
+import json
+import os
+import tempfile
+import traceback
+from http import HTTPStatus
+
+import boto3
+import requests
+from botocore.exceptions import ClientError
+from jsonschema import Draft7Validator
+
+# Must be done before importing quilt3.
+os.environ['QUILT_DISABLE_CACHE'] = 'true'  # noqa: E402
+import quilt3
+from quilt3.backends import get_package_registry
+from quilt3.backends.s3 import S3PackageRegistryV1
+from quilt3.util import PhysicalKey
+from t4_lambda_shared.decorator import api
+from t4_lambda_shared.utils import get_default_origins, make_json_response
+
+AUTH_ENDPOINT = os.environ['AUTH_ENDPOINT']
+
+PROMOTE_PKG_MAX_MANIFEST_SIZE = int(os.environ['PROMOTE_PKG_MAX_MANIFEST_SIZE'])
+PROMOTE_PKG_MAX_PKG_SIZE = int(os.environ['PROMOTE_PKG_MAX_PKG_SIZE'])
+PROMOTE_PKG_MAX_FILES = int(os.environ['PROMOTE_PKG_MAX_FILES'])
+PKG_FROM_FOLDER_MAX_PKG_SIZE = int(os.environ['PKG_FROM_FOLDER_MAX_PKG_SIZE'])
+PKG_FROM_FOLDER_MAX_FILES = int(os.environ['PKG_FROM_FOLDER_MAX_FILES'])
+
+PACKAGE_ID_PROPS = {
+    'registry': {
+        'type': 'string',
+        'format': 'uri',
+    },
+    'name': {
+        'type': 'string',
+    },
+}
+
+PACKAGE_REV_LOCATION_SCHEMA = {
+    '$schema': 'http://json-schema.org/draft-07/schema#',
+    'id': 'https://quiltdata.com/package-revision-location/1',
+    'type': 'object',
+    'properties': {
+        **PACKAGE_ID_PROPS,
+        'top_hash': {
+            'type': 'string',
+            'pattern': '^[0-9a-f]{64}$',
+        }
+    },
+    'required': [
+        *PACKAGE_ID_PROPS,
+        'top_hash',
+    ],
+}
+
+PACKAGE_BUILD_META_PROPS = {
+    'message': {
+        'type': 'string',
+    },
+    'meta': {
+        'type': 'object',
+    },
+    'workflow': {
+        'type': ['string', 'null']
+    },
+}
+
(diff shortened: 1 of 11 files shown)


lambdas/pkgpush/index.py
```diff
@@ -0,0 +1,309 @@
...
+    dst_registry = get_registry(get_dst(data))
+    src_registry = get_registry(get_src(data))
+    copy_data = _get_successor_params(src_registry, dst_registry).get('copy_data', True)
+
+    try:
+        pkg = get_pkg(src_registry, data)
+        if copy_data:
+            total_size = 0
+            total_files = 0
+            for lk, e in pkg.walk():
```
akarve, line 183: It would be much cleaner to return total_size, and total_files as a tuple from a utility function that takes the package as an argument.

lambdas/pkgpush/index.py
```diff
@@ -0,0 +1,309 @@
...
+    get_schema_validator('PACKAGE_PROMOTE_SCHEMA')(data)
+
+    def get_pkg(src_registry, data):
+        quilt3.util.validate_package_name(data['parent']['name'])
+        manifest_pk = src_registry.manifest_pk(data['parent']['name'], data['parent']['top_hash'])
+        manifest_size, version = quilt3.data_transfer.get_size_and_version(manifest_pk)
+        if manifest_size > PROMOTE_PKG_MAX_MANIFEST_SIZE:
+            raise ApiException(
+                HTTPStatus.BAD_REQUEST,
+                f"Manifest size is {manifest_size}, but max supported size is {PROMOTE_PKG_MAX_MANIFEST_SIZE}"
```
akarve, line 261:
```suggestion
                f"Manifest size of {manifest_size} exceeds supported limit of {PROMOTE_PKG_MAX_MANIFEST_SIZE}"
```

akarve (changes requested):
A few small changes requested. The module comment should also discuss RAM requirements and we should size accordingly.

lambdas/pkgpush/index.py
sir-sigurd: I think it doesn't make sense because we want to fail ASAP and in future we'd want to rework it to fail without loading the whole manifest in the memory.

akarve: I can live with it but wish to point out that you can still fail early even if the iteration happens in a separate function

lambdas/pkgpush/index.py
```diff
@@ -0,0 +1,356 @@
...
+        #       for downloading to fileobj.
+        with tempfile.NamedTemporaryFile() as tmp_file:
+            quilt3.data_transfer.copy_file(
+                manifest_pk,
+                PhysicalKey.from_path(tmp_file.name),
+                size=manifest_size,
+            )
+            pkg = quilt3.Package.load(tmp_file)
+        if any(e.physical_key.is_local() for lk, e in pkg.walk()):
+            raise ApiException(HTTPStatus.BAD_REQUEST, "Parent's manifest contains non S3 physical keys.")
```
akarve, line 320:
```suggestion
            raise ApiException(HTTPStatus.BAD_REQUEST, "Parent's manifest contains non-S3 physical keys.")
```

lambdas/pkgpush/requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/quiltdata/quilt@3db13ee2e451939a297260b55d6119bcf255e669#egg=quilt3&subdirectory=api/python
```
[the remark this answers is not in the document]
sir-sigurd: https://github.com/quiltdata/quilt/pull/2129

lambdas/pkgpush/index.py
sir-sigurd: Well, currently a function where iteration happens is `Package.set_dir()`.

lambdas/pkgpush/index.py
```diff
@@ -0,0 +1,356 @@
...
+
+import boto3
+import requests
+from botocore.exceptions import ClientError
+from jsonschema import Draft7Validator
+
+# Must be done before importing quilt3.
+os.environ.update(
+    QUILT_DISABLE_CACHE='true',
+    QUILT_DISABLE_USAGE_METRICS='true',
```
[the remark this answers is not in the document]
sir-sigurd: Currently it's disabled in registry:
https://github.com/quiltdata/enterprise/blob/efc08a8e88936b009ba92df136ab6ca202d34263/registry/quilt_server/views.py#L331-L333

Do we track requests to registry? Do we want to track each quilt3 API call here? Couldn't we use Cloudwatch metrics for lambda instead?

lambdas/pkgpush/index.py
```diff
@@ -0,0 +1,358 @@
...
+            return f(request)
+        return wrapper
+    return innerdec
+
+
+def setup_telemetry(f):
+    @functools.wraps(f)
+    def wrapper(request):
+        # FIXME: Should we use aws_request_id from lambda context as session id?
+        quilt3.telemetry.ApiTelemetry.reset_session_id()
```
sir-sigurd, line 210: N.B.

lambdas/pkgpush/index.py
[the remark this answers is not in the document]
sir-sigurd: I've created PRs for updating schema:
https://github.com/quiltdata/quilt/pull/2151
https://github.com/quiltdata/enterprise/pull/259

I'm not aware of how we process collected data, so couldn't we use modified [`client_type`](https://github.com/quiltdata/quilt/blob/a1d93771c602c96bceda232072b00c0847ef0e7a/api/python/quilt3/telemetry.py#L17) instead? (something like `quilt3-python-client@lambda`)

akarve (approved):

akarve (approved):

PR #2147, opened 2021-04-03
Pkg select on empty manifest

kevinemoore:
# Description
Don't crash pkgselect for empty manifests.

# TODO

- [x] Unit tests


akarve (approved):

lambdas/pkgselect/index.py
```diff
@@ -81,18 +81,20 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
             axis=1
         ).to_dict(orient='records')
         objects = folder[~folder.logical_key.str.contains('/')].to_dict(orient='records')
-        returned_results = len(prefixes) + len(objects)
-    except AttributeError as err:
-        # Pandas will raise an attribute error if the DataFrame has
-        # no rows with a non-null logical_key. We expect that case if
+
+    except (AttributeError, KeyError) as err:
```
sir-sigurd, line 85: It's very error-prone to have so large `try` clause with such non-specific exceptions handled. Consider rewriting it with either of these:

1. Check that expected columns are present in `DataFrame` (`{'physical_key', 'logical_key', 'size'}.issubset(df.columns)`).
2. Normalizing output of query by using `COALESCE()` (or normalizing `DataFrame` itself?). Also it looks like a row with package metadata isn't used here, so might make sense to filter out it with `WHERE logical_key IS NOT MISSING`.

sir-sigurd (changes requested):

lambdas/pkgselect/index.py
kevinemoore: Good call. It's safer and still fast to check the columns in the input DataFrame.

I suppose we could replace `dropna` in the group_by with a filter in the s3select where clause. That should be a little more efficient, but let's do that in a separate PR.

sir-sigurd (approved):

PR #2148, opened 2021-04-05
JsonEditor remove-row button

fiskus:
* Replace menu consisting of one "Remove" action with "Remove" button
* Convert most of JsonEditor code to TypeScript (except State.js)

Before
1 before

After
2 after


catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
```diff
@@ -470,178 +476,179 @@ interface MetaInputProps {
...
-
-  const handleTextChange = (e: React.ChangeEvent<{ value: string }>) => {
-    changeText(e.target.value)
-  }
+export const MetaInput = React.forwardRef<HTMLDivElement, MetaInputProps>(
+  function MetaInput(
+    { className, schemaError, input: { value, onChange }, meta, schema },
+    ref,
+  ) {
+    const classes = useMetaInputStyles()
```
fiskus, line 484: Below are indentation changes only and removing of `// @ts-expect-error`.

catalog/package-lock.json
```diff
@@ -3605,6 +3616,7 @@
       "dependencies": {
         "anymatch": "~3.1.1",
         "braces": "~3.0.2",
+        "fsevents": "~2.3.1",
```
nl0, line 3619: it's a little strange that npm adds these non-relevant changes to the lockfile, i wonder what's the reason -- changes in some dependency resolution logic btw npm versions or smth

nl0 (approved):
lgtm

catalog/package-lock.json
fiskus: I have `npm@7.3.0` and `node@v15.5.0`. Global packages: 
* @gridsome/cli@0.3.4
* firebase-tools@9.6.1
* lerna@4.0.0
* npm@7.3.0
* svgo@2.2.2
* typescript-language-server@0.5.1
* typescript@4.1.3

---

"fsevents" is `optionalDependency` for "chokidar". Probably, I don't have "fsevents" as a dependency for my global packages, but you have. And that's why I have `fsevents` in package-lock and you don't

nl0: doesnt seem like i have any global packages, but i have npm@7.6.0, so maybe this is the reason. or maybe it's bc the lockfile is maintained by renovate most of the time 🤷

fiskus: I've updated to 7.6.0 and `npm install` removed additional changes

PR #2151, opened 2021-04-06
telemetry changes in client for pkgpush lambda

sir-sigurd:
for #2139

## TODO

- [ ] unit tests


api/python/quilt3/telemetry.py
```diff
@@ -15,19 +15,26 @@
 TELEMETRY_URL = "https://telemetry.quiltdata.cloud/Prod/metrics"
 TELEMETRY_USER_AGENT = "QuiltCli"
 TELEMETRY_CLIENT_TYPE = "quilt3-python-client"
-TELEMETRY_SCHEMA_VERSION = "pyclient-usage-metrics-v1"
+TELEMETRY_SCHEMA_VERSION = "pyclient-usage-metrics-v2"
```
akarve, line 18: We should use semver or schemaversion for this, but not a blocking concern
https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/#schemaver

I think this counts as an ADDITION so `...-v1.0.1`.

The original string is silly but that's a separate problem.

api/python/quilt3/telemetry.py
```diff
@@ -15,19 +15,26 @@
...
 
 DISABLE_USAGE_METRICS_ENVVAR = "QUILT_DISABLE_USAGE_METRICS"
 MAX_CLEANUP_WAIT_SECS = 5
 
 
 class ApiTelemetry:
     session = None
-    session_id = str(uuid.uuid4())
```
akarve, line 26: Just a note, this will have implications for how we unique sessions. Need to note this in the analytics backend. I will show you.

api/python/quilt3/telemetry.py
```diff
@@ -137,6 +145,10 @@ def decorated(*args, **kwargs):
         return decorated
 
 
+ApiTelemetry.reset_session_id()
+ApiTelemetry.set_client_context(None)
```
akarve, line 149: For backward-compatibility with old events, we should make sure this shows up as null in the analytics beacon and not 'None' or something like that.

akarve (changes requested):

api/python/quilt3/telemetry.py
akarve: Per Slack this is still a class variable. Doesn't it still need to be declared?

sir-sigurd: You can't "declare" variable in Python, the deleted line was assignment that is moved to `reset_session_id()`.

sir-sigurd: IMHO as is this class doesn't add much value and it might worth to move all its stuff to the module level.

akarve: I guess it's ok but it seems very weird for initialization to depend on a function call when another LOC when eg report_api_use calls it

akarve: ie much cleaner to init in original location

drernie:
@sir-sigurd Can you address these comments and merge?

drernie:
I guess not. Closing...

PR #2154, opened 2021-04-07
Move package API requests to one file, consolidate naming and internal API

fiskus:
I need this refactoring because I want to be able to validate params for every package API request in one place


catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
+import * as R from 'ramda'
+import * as React from 'react'
+
+import { JsonValue } from 'components/JsonEditor/constants'
+import * as APIConnector from 'utils/APIConnector'
+import { makeSchemaDefaultsSetter, JsonSchema } from 'utils/json-schema/json-schema'
```
nl0, line 6: is `/json-schema/json-schema` intentional?

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
...
+  contents: FileUpload[]
+  target: {
+    bucket: string
+    name: string
+  }
+}
+
+interface UpdatePackageParams extends BasePackageParams {
+  contents: FileUpload[]
+  source: {
```
nl0, line 103: actually, this is not source, but destination -- it tells where to put the manifest. the source is implicit -- update dialog just loads everything from the source manifest to the `contents` array

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
...
+import * as React from 'react'
+
+import { JsonValue } from 'components/JsonEditor/constants'
+import * as APIConnector from 'utils/APIConnector'
+import { makeSchemaDefaultsSetter, JsonSchema } from 'utils/json-schema/json-schema'
+import pipeThru from 'utils/pipeThru'
+import * as workflows from 'utils/workflows'
+
+// "CREATE package" - creates package from scratch to new target
+// "UPDATE package" - creates package from source (existing manifest) to the same target
```
nl0, line 11: > to the same target

that's not necessarily the case, since the destination name can be edited

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
...
+  logical_key: string
+  physical_key: {
+    bucket: string
+    key: string
+    version: string
+  }
+  size: number
+}
+
+interface ManifestBodyBase {
```
nl0, line 32: it's not exactly "Manifest", i'd use a more generic name like "BaseRequestBody", "CreateRequestBody", etc

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
...
+interface Response {
+  top_hash: string
+}
+
+// FIXME: this is copypasted from PackageDialog -- next time we need to TSify utils/APIConnector properly
+interface ApiRequest {
+  <O>(opts: {
+    endpoint: string
+    method?: 'GET' | 'PUT' | 'POST' | 'DELETE' | 'HEAD'
+    body?: {}
```
nl0, line 129: you can parameterize `body` if you want to type-check it

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
...
+  }): Promise<O>
+}
+
+const uploadManifest = (
```
nl0, line 133: i dont think this function should exist
1) its name doesnt reflect all its use cases, it looks like a leaked abstraction;
2) it doesnt have enough logic, i think it's better to just call `req` directly (you can parameterize body and response types to add type checking there)

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
...
+
+const getWorkflowApiParam = R.cond([
+  [R.equals(workflows.notAvailable), R.always(undefined)],
+  [R.equals(workflows.notSelected), R.always(null)],
+  [R.T, R.identity],
+]) as (
+  slug: typeof workflows.notAvailable | typeof workflows.notSelected | string,
+) => string | null | undefined
+
+const createPackage = (
```
nl0, line 163: i think we can reduce duplication and complexity by placing the request invocation inline, like this:
```typescript
export function useCreatePackage() {
  const req: ApiRequest = APIConnector.use()
  return React.useCallback(
    (
      { contents, message, meta, target, workflow }: CreatePackageParams,
      schema: JsonSchema,
    ) =>
      req<Response, CreateRequestBody>({
        endpoint: ENDPOINT_CREATE,
        method: 'POST',
        body: {
          name: target.name,
          registry: `s3://${target.bucket}`,
          message,
          contents,
          meta: getMetaValue(meta, schema),
          workflow: getWorkflowApiParam(workflow.slug),
        }), // or even just cast this object to CreateResponseBody
    [req],
  )
}
```

catalog/app/containers/Bucket/requests/packageUpload.ts
```diff
@@ -0,0 +1,258 @@
+import * as R from 'ramda'
```
nl0, line 1: i think we need to use some more generic name for this file, bc its contents doesnt relate only to uploading, maybe smth like `packageManipulation` or `packageCreation`

nl0 (changes requested):
the general direction is right, but i think we should
1. adjust naming
2. remove unnecessary level of abstraction by calling `req` directly

catalog/app/containers/Bucket/requests/packageUpload.ts
fiskus: `json-schema/json-schema` is a mistake

Also, I'm planning to rename `json-schema` to `jsonSchema`

catalog/app/containers/Bucket/requests/packageUpload.ts
fiskus: btw, do you prefer `destination` instead of `target`? Probably, `destination` is more explicit

nl0: i prefer concise names, so `target` seems better, but `dst` would also do =) doesnt matter that much as long as it's used consistently

catalog/app/containers/Bucket/requests/packageUpload.ts
fiskus: "request body" just means nothing to me, but I'm ok with it since it's inside one module

catalog/app/containers/Bucket/requests/packageUpload.ts
fiskus: It's a place where I can validate the request body with Schema

catalog/app/containers/Bucket/requests/packageUpload.ts
fiskus: Maybe, `requests/package.ts` then?

fiskus:
> remove unnecessary level of abstraction by calling req directly
I can't do this, because my goal was to consolidate requests in one function where I can validate the request body

catalog/app/containers/Bucket/requests/packageUpload.ts
nl0: sounds good

catalog/app/containers/Bucket/requests/packageUpload.ts
nl0: yep, we get the meaning from the context

catalog/app/containers/Bucket/requests/packageUpload.ts
nl0: you can validate it the same way by using type arg for `req`

nl0:
> I can't do this, because my goal was to consolidate requests in one function where I can validate the request body

but requests are different, so are the bodies. you can type-check request body by simply using type arg when calling `req` directly.

nl0:
ok, i see that by "validate" you probably mean some runtime validation, not type-checking, but i still dont quite get it

fiskus:
function uploadManifest(req, s3, workflow, body) {
  const schema = await s3.fetchFile(workflow.objectSchemaurl)
  const error = validateBody(schema)
  if (error.length) throw error
  return req(endpoint, body)
}

function useCreatePackage(body) {
  const req = useAPIConnector()
  const s3 = useS3()
  return uploadManifest(req, s3, workflow.objectSchemaUrl, body)
}

function useCopyPackage(body) {
  const req = useAPIConnector()
  const s3 = useS3()
  return uploadManifest(req, s3, workflow.objectSchemaUrl, body)
}

fiskus:
I don't have working code for the next PR, but I wanted something like this ^

nl0:
ok now i get it, but
1) `uploadManifest` doesnt seem like an appropriate name for this function
2) probably you dont need to type-check request body if you plan to validate it in runtime against a schema

nl0 (approved):

fiskus:
Do you have in mind what name could be appropriate?

nl0:
> Do you have in mind what name could be appropriate?

dunno, smth like `requestWithValidation` maybe, but it's your call since you have better idea of its potential use cases and the logic you want to add there

fiskus:
I can be confident about an object's shape with type-checking, but Schema gives more possibilities like, for example, checking number values (`{ min: 0, max: 256}` for filesize).
Also, Schema can be different for different workflows

nl0:
> I can be confident about an object's shape with type-checking, but Schema gives more possibilities like, for example, checking number values (`{ min: 0, max: 256}` for filesize).
> Also, Schema can be different for different workflows

true, but you must somehow guarantee that the schema is compatible with the type

PR #2157, opened 2021-04-07
Suggestion for #2154

nl0:


nl0:
@fiskus i meant something like this. all the checks are still in place (except for endpoint check, but those are hardcoded and even more safe now, bc there are no invalid endpoint/body pairs)

fiskus:
I need to be able to call validate function on body:

```ts
function makeRequest(body) {
  const error = validate(body)
  if (error) throw error
  return fetch('/api')
}

const requestA = makeRequest({a: 1})
const requestB = makeRequest({b: 1})
```

nl0:
> I need to be able to call validate function on body:
> 
> ```ts
> function makeRequest(body) {
>   const error = validate(body)
>   if (error) throw error
>   return fetch('/api')
> }
> 
> const requestA = makeRequest({a: 1})
> const requestB = makeRequest({b: 1})
> ```

i dont quite get it, could you please elaborate a little?

nl0:
let's move this thread back to #2154

fiskus:
```ts
function uploadManifest(req, s3, workflow, body) {
  const schema = await s3.fetchFile(workflow.objectSchemaurl)
  const error = validateBody(schema)
  if (error.length) throw error
  return req(endpoint, body)
}

function useCreatePackage(body) {
  const req = useAPIConnector()
  const s3 = useS3()
  return uploadManifest(req, s3, workflow.objectSchemaUrl, body)
}

function useCopyPackage(body) {
  const req = useAPIConnector()
  const s3 = useS3()
  return uploadManifest(req, s3, workflow.objectSchemaUrl, body)
}
```

PR #2161, opened 2021-04-09
Document package push limitations in catalog [ci skip]

akarve:


sir-sigurd (approved):

docs/API Reference/limitations.md
```diff
@@ -0,0 +1,30 @@
...
+ Package manifest size (metadata)                           | 100 MB 
+ Package size (data; via promotion or from an S3 directory) | 10 GB  
+ Maximum mumber of files per push                           | 5,000  
+
+
+### API
+
+As of this writing, with sufficient client-side memory, you can comfortably scale
+Quilt packages to one million objects with no practical limit on object size, outside
+of S3's 5 TB per object limit.
```
akarve, line 26:
```suggestion
As of this writing, with sufficient client-side memory, you can comfortable scale
Quilt packages to at least one million objects per package, with no practical limit
on object size (save S3's 5 TB per object limit). A fast network, or better yet an AWS
compute instance in the same region as your Quilt S3 buckets, is recommended.
```

PR #2164, opened 2021-04-13
Copy tweaks

akarve:
@nl0 in the A/B flavors do we send the copy itself so to the analytics beacon so that we can tease out conversion differences? Anything I need to do besides changing the strings?


nl0 (approved):
> in the A/B flavors do we send the copy itself so to the analytics beacon so that we can tease out conversion differences?

yep

> Anything I need to do besides changing the strings?

no, you're good

PR #2171, opened 2021-04-15
Ability to add S3 folders / files to package

nl0:
# Description

# TODO

- [ ] Unit tests
- [ ] Automated tests (e.g. Preflight)
- [ ] Documentation
    - [ ] [Python: Run `build.py`](../gendocs/build.py) for new docstrings
    - [ ] JavaScript: basic explanation and screenshot of new features
- [x] [Changelog](CHANGELOG.md) entry

PR #2180, opened 2021-04-22
add S3 data to existing package

sir-sigurd:
# Description
See https://github.com/quiltdata/quilt/pull/2171 for frontend part.

# TODO

- [x] Unit tests

### DIFF
--- a/lambdas/pkgpush/index.py
+++ b/lambdas/pkgpush/index.py
@@ -17,7 +17,11 @@ from quilt3.backends import get_package_registry
 from quilt3.backends.s3 import S3PackageRegistryV1
 from quilt3.util import PhysicalKey
 from t4_lambda_shared.decorator import ELBRequest, api
-from t4_lambda_shared.utils import get_default_origins, make_json_response, get_quilt_logger
+from t4_lambda_shared.utils import (
+    get_default_origins,
+    get_quilt_logger,
+    make_json_response,
+)
 
 PROMOTE_PKG_MAX_MANIFEST_SIZE = int(os.environ['PROMOTE_PKG_MAX_MANIFEST_SIZE'])
 PROMOTE_PKG_MAX_PKG_SIZE = int(os.environ['PROMOTE_PKG_MAX_PKG_SIZE'])
@@ -406,8 +410,7 @@ def large_request_handler(request_type):
                     s3.delete_object(SERVICE_BUCKET, user_request_key, version_id)
                 except Exception:
                     logger.exception('error while removing user request file from S3')
-                finally:
-                    return result
+                return result
         return wrapper
     return inner
 
--- a/lambdas/pkgpush/index.py
+++ b/lambdas/pkgpush/index.py
@@ -17,7 +17,7 @@ from quilt3.backends import get_package_registry
 from quilt3.backends.s3 import S3PackageRegistryV1
 from quilt3.util import PhysicalKey
 from t4_lambda_shared.decorator import ELBRequest, api
-from t4_lambda_shared.utils import get_default_origins, make_json_response
+from t4_lambda_shared.utils import get_default_origins, make_json_response, get_quilt_logger
 
 PROMOTE_PKG_MAX_MANIFEST_SIZE = int(os.environ['PROMOTE_PKG_MAX_MANIFEST_SIZE'])
 PROMOTE_PKG_MAX_PKG_SIZE = int(os.environ['PROMOTE_PKG_MAX_PKG_SIZE'])
@@ -136,6 +136,9 @@ PACKAGE_CREATE_SCHEMA = {
 # Monkey patch quilt3 S3ClientProvider, so it builds a client using user credentials.
 user_boto_session = None
 quilt3.data_transfer.S3ClientProvider.get_boto_session = staticmethod(lambda: user_boto_session)
+
+
+logger = get_quilt_logger()
 
 
 def get_user_credentials(request):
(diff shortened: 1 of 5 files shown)


lambdas/pkgpush/index.py
```diff
@@ -360,3 +373,96 @@ def get_pkg(src_registry, data):
...
+                hash_ = entry.get('hash')
+                obj_size = entry.get('size')
+                meta = entry.get('meta')
+
+                if hash_ and obj_size:
+                    hash_obj = dict(type='SHA256', value=hash_)
+                    pkg.set(
+                        logical_key,
+                        quilt3.packages.PackageEntry(
+                            physical_key,
```
nl0, line 436: we need to resolve object's versionId somewhere here i guess

lambdas/pkgpush/index.py
```diff
@@ -360,3 +373,96 @@ def get_pkg(src_registry, data):
...
+                            obj_size,
+                            hash_obj,
+                            meta,
+                        )
+                    )
+                else:
+                    pkg.set(logical_key, str(physical_key), meta)
```
nl0, line 443: or probably here. does `pkg.set` do it?

sir-sigurd: yup, it does

lambdas/pkgpush/index.py
```diff
@@ -360,3 +418,127 @@ def get_pkg(src_registry, data):
         pkg_max_size=PKG_FROM_FOLDER_MAX_PKG_SIZE,
         pkg_max_files=PKG_FROM_FOLDER_MAX_FILES,
     )
+
+
+def large_request_handler(request_type):
+    user_request_key = f'user-requests/{request_type}'
```
akarve, line 424: I don't understand how this is thread safe at all. What happens if multiple users issue concurrent requests of the same type? Won't the be reading/deleting the same file in the TemporaryFile loop below?

lambdas/pkgpush/index.py
```diff
@@ -360,3 +418,127 @@ def get_pkg(src_registry, data):
...
+@api(cors_origins=get_default_origins(), request_class=ELBRequest)
+@api_exception_handler
+@auth
+@json_api({'type': 'string', 'minLength': 1, 'maxLength': 1024})
+@large_request_handler('create-package')
+@setup_telemetry
+def create_package(request):
+    json_iterator = map(json.JSONDecoder().decode, (line.decode() for line in request.stream))
+
+    data = next(json_iterator)
```
akarve, line 466: I'd suggest calling this `first_line` or something because the first line is special. `data` makes more sense if you're going to iterate with it.

lambdas/pkgpush/index.py
```diff
@@ -360,3 +418,127 @@ def get_pkg(src_registry, data):
...
+                raise ApiException(HTTPStatus.BAD_REQUEST, f"{entry['physical_key']} is not a valid s3 URL.")
+            if physical_key.is_local():
+                raise ApiException(HTTPStatus.BAD_REQUEST, f"{str(physical_key)} is not in S3.")
+            logical_key = entry['logical_key']
+
+            hash_ = entry.get('hash')
+            obj_size = entry.get('size')
+            meta = entry.get('meta')
+
+            if hash_ and obj_size:
```
akarve, line 501: I think this will mistakenly skip empty files which have already been hashed.

lambdas/pkgpush/tests/test_index.py
```diff
@@ -155,23 +163,25 @@ def _make_event(cls, body, credentials):
...
         }
 
-    def make_request_base(self, params, *, credentials):
+    def make_request_wrapper(self, params, *, credentials, **kwargs):
+        return Request(
+            self._make_event(json.dumps(params), credentials=credentials),
+        )
+
+    def make_request_base(self, data, *, credentials, **kwargs):
         # This is a function before it get wrapped with @api decorator.
```
akarve, line 172:
```suggestion
        # This is a function before it gets wrapped with @api decorator.
```

lambdas/shared/t4_lambda_shared/utils.py
```diff
@@ -15,6 +15,8 @@
 
 PACKAGE_INDEX_SUFFIX = "_packages"
 
+LAMBDA_TMP_SPACE = 512 * 2 ** 20
```
akarve, line 18: AWS says 512 MB not 512 MiB
```suggestion
LAMBDA_TMP_SPACE = 512_000_000
```

lambdas/pkgpush/tests/test_index.py
```diff
@@ -578,3 +588,358 @@ def setup_s3_load_pkg_source(self):
...
+                *self.package_entries,
+            ])
+            assert pkg_response.status_code == 200
+
+    @mock.patch('quilt3.workflows.validate', lambda *args, **kwargs: None)
+    def test_object_level_meta(self):
+        entry1 = {
+            'logical_key': 'obj1',
+            'physical_key': str(self.physical_key),
+            'size': self.file_data_size,
```
akarve, line 750: Need a test case where one and both of these are falsy.

akarve (changes requested):

lambdas/pkgpush/index.py
```diff
@@ -187,19 +244,26 @@ def wrapper(request):
     return wrapper
 
 
+def get_schema_validator(schema):
+    iter_errors = Draft7Validator(schema).iter_errors
+
+    def validator(data):
+        ex = next(iter_errors(data), None)
+        if ex is not None:
+            raise ApiException(HTTPStatus.BAD_REQUEST, ex.message)
```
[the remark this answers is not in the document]
sir-sigurd: Yes, but that's how it currently works and there is no reason to change it right now.

lambdas/pkgpush/index.py
sir-sigurd: Users use the same key to write requests, but they write to the versioned bucket created by us via Cloudformation, so each request get its own ID (version ID of object). A single lambda instance process only a single request at the same time.

lambdas/shared/t4_lambda_shared/utils.py
sir-sigurd: I tried to run `shutil.disk_usage('/tmp/')` and got this:
```
usage(total=551346176, used=892928, free=538333184)
```
```
In [23]: 538333184 / 2 ** 20
Out[23]: 513.39453125
```

I think it might make sense to drop that check and simply try to preallocate space using [`os.posix_fallocate()`](https://docs.python.org/3/library/os.html#os.posix_fallocate) to fail early.

lambdas/pkgpush/index.py
```diff
@@ -120,11 +127,51 @@
 }
 
 
+PACKAGE_CREATE_SCHEMA = {
```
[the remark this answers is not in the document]
sir-sigurd: You could have seen that in the registry :slightly_smiling_face:,

lambdas/shared/t4_lambda_shared/utils.py
sir-sigurd: `os.posix_fallocate()` fails with `PermissionError` in lambda :disappointed: .

PR #2183, opened 2021-04-27
Add HTML meta titles

fiskus:
For all pages except Auth/*

### DIFF
--- a/catalog/app/containers/Bucket/PackageList.js
+++ b/catalog/app/containers/Bucket/PackageList.js
@@ -23,6 +23,7 @@ import { readableQuantity } from 'utils/string'
 import useDebouncedInput from 'utils/useDebouncedInput'
 import usePrevious from 'utils/usePrevious'
 
+import * as HtmlMeta from './HtmlMeta'
 import PackageCreateDialog from './PackageCreateDialog'
 import Pagination from './Pagination'
 import { displayError } from './errors'
@@ -394,6 +395,8 @@ export default function PackageList({
 
   return (
     <>
+      <HtmlMeta.PackageList bucket={bucket} />
+
       <PackageCreateDialog
         {...{ bucket, refresh, open: uploadOpen, onClose: closeUpload }}
       />
--- a/catalog/app/containers/Admin/Buckets.js
+++ b/catalog/app/containers/Admin/Buckets.js
@@ -21,6 +21,7 @@ import { useTracker } from 'utils/tracking'
 import * as validators from 'utils/validators'
 
 import * as Form from './Form'
+import * as HtmlMeta from './HtmlMeta'
 import * as Table from './Table'
 import * as data from './data'
 
@@ -940,6 +941,7 @@ export default function Buckets({ location }) {
   const buckets = Cache.useData(data.BucketsResource, { req })
   return (
     <M.Box mt={2} mb={2}>
+      <HtmlMeta.Buckets />
       <React.Suspense
         fallback={
           <M.Paper>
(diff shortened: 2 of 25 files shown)


nl0 (changes requested):
I think the current approach doesn't work that well bc
1. there's too much repetition (title base is repeated thrice, segment divider is repeated many times inline);
2. title "implementations" are isolated from usage and grouped "by type" (as some people used to group "reducers" and other redux stuff in ye olden days);
3. some code from "containers" imported in "components" -- unless absolutely necessary, this should be avoided, bc containers are more like "composition roots" -- entities composed of simpler components, not exporting lower level stuff for components.

### What i propose

#### 1. Create a single module (under `utils` or `components`) implementing basic reusable logic, e.g.:
```typescript
const BASE = 'Quilt is a versioned data portal for AWS'
const DIVIDER = ' • '

interface MetaTitleProps {
  children?: string | string[]
  base?: string
}

export default function MetaTitle({ children = [], base = BASE }: MetaTitleProps) {
  let segments = [base]
  if (children) {
    if (Array.isArray(children)) {
      segments = [...children, ...segments]
    } else {
      segments.unshift(children)
    }
  }
  
  return (
    <Helmet>
      <title>
        {segments.join(DIVIDER)}
      </title>
    </Helmet>
  )
}
```

#### 2. Use it directly everywhere we need title set, e.g.:
```typescript
<MetaTitle>{['Users and Roles', 'Admin']}</MetaTitle>
<MetaTitle>Resolve a Quilt package URI</MetaTitle>
<MetaTitle>{bucket}</MetaTitle>
<MetaTitle>{[`${name}@${revisionShortened}/${path}`], bucket]}</MetaTitle>
```

fiskus:
Do you think I should create one file for the whole project, or several files: for catalog, for open.quiltdata, for landing?

fiskus:
> some code from "containers" imported in "components"
Can you point out this? I can't find it.

I even didn't change any "*component*"
Screenshot from 2021-05-04 16-11-07

nl0:
> > some code from "containers" imported in "components"
> 
> Can you point out this? I can't find it.
> 
> I even didn't change any "_component_"
> Screenshot from 2021-05-04 16-11-07

oh, i've mistaken a little -- i meant containers/Search and containers/UriResolver -- they are building blocks for containers/App and at the same time they import stuff from containers/App (not exactly a circular dependency, but smells funny imo)

nl0:
> Do you think I should create one file for the whole project, or several files: for catalog, for open.quiltdata, for landing?

seems to me that one globally shared component is enough -- we could use context to configure title base if this is required

catalog/app/containers/Bucket/PackageList.js
```diff
@@ -394,6 +395,8 @@ export default function PackageList({
 
   return (
     <>
+      <MetaTitle>{['Package list', bucket]}</MetaTitle>
```
nl0, line 398:
```suggestion
      <MetaTitle>{['Packages', bucket]}</MetaTitle>
```

catalog/app/utils/MetaTitle.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return [children, base]
+}
+
+export function getTitle(base: string, children?: string | string[]) {
+  return getTitleSegments(base, children).join(DIVIDER)
+}
+
+export default function MetaTitle({ children, base = BASE }: MetaTitleProps) {
+  const titleStr = React.useMemo(() => getTitle(base, children), [base, children])
```
nl0, line 26: this will be re-evaluated every render, since the most common use case is smth like `<MetaTitle>{[a, b]}</MetaTitle>` -- children array is created anew each render and useMemo wont work

nl0 (approved):
lgtm, some minor issues

PR #2184, opened 2021-04-27
Shorten text: Preview parsing errors

fiskus:
Before
Screenshot from 2021-04-27 12-14-33

After
Screenshot from 2021-04-27 12-14-42


sir-sigurd (commented):
The original message looks unclear to me, but the proposed version might be not descriptive enough.
The full description of what has happened is something like: "While generating this preview we tried to parse this file as CSV and extract as much data as possible. These errors were encountered during parsing: ..."

Also there is a message that informs that preview was truncated and I think this message can be found only by accident:
image
Maybe we could use a single approach that handles this message too.

akarve (approved):

fiskus:
I don't know how to resolve issue mentioned by Sergey, filed it to backlog

PR #2185, opened 2021-04-28
Save package sorting mode in localStorage

fiskus:


catalog/app/containers/Bucket/PackageList.js
```diff
@@ -381,7 +393,15 @@ export default function PackageList({
         urls.bucketPackageList(bucket, { filter: filtering.value || undefined, sort }),
       )
     }
-  }, [history, urls, bucket, sort, filtering.value, computedFilter])
+
+    const sortPackagesBy = storage.load()?.sortPackagesBy
+    switch (sortPackagesBy) {
+      case 'modified':
+      case 'name':
+        history.replace(makeSortUrl(sortPackagesBy))
```
nl0, line 401: i dont think we should abuse `useEffect` call used to `// set filter query param on filter input change (debounced)`, but create a separate call with required checks and side effects.
also, probably we should only execute this side effect if sort mode is different from the current one? (`sortPackagesBy !== sort`).
oh, and another thing -- we should get stored value outside the `useEffect` body, otherwise `sortPackagesBy` can easily become stale (not sure if it's ok to access localStorage every render, but at least it's synchronous).

fiskus: 1. moved my code to another useEffect
2. moved sortPackagesBy outside useEffect
3. set condition to not trigger function body

nl0 (approved):
lgtm, tho i iirc @akarve said that maybe we dont want this feature after all, so better confirm with him

fiskus:
Aneesh allowed this feature in private zoom conversation :)

PR #2186, opened 2021-04-29
Local packages prototype

kevinemoore:
# Prototype Only - DO NOT MERGE
This prototype demonstrates a local-mode substitute for package listing that does not rely on a Quilt stack (i.e. does not rely on elasticsearch).


nl0:
superseded by #2619

PR #2187, opened 2021-04-30
Fix "None" shown twice

fiskus:
Screenshot from 2021-04-30 07-40-59

### DIFF
--- a/catalog/app/containers/Bucket/PackageDialog/SelectWorkflow.js
+++ b/catalog/app/containers/Bucket/PackageDialog/SelectWorkflow.js
@@ -29,6 +29,7 @@ export default function SelectWorkflow({
   const classes = useStyles()
 
   const noChoice = items.length === 1
+  const showDisabledNoneOption = !items.find(({ slug }) => slug === workflows.notSelected)
 
   return (
     <M.FormControl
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
```diff
@@ -341,6 +342,14 @@ export function WorkflowInput({
 export const defaultWorkflowFromConfig = (cfg?: workflows.WorkflowsConfig) =>
   cfg && cfg.workflows.find((item) => item.isDefault)
 
+export const useWorkflowValidator = (workflowsConfig?: workflows.WorkflowsConfig) => (
+  workflow?: workflows.Workflow,
```
nl0, line 346: probably we should memoize returned function by `workflowsConfig?.isWorkflowRequired`

catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
```diff
@@ -660,6 +670,13 @@ export function SchemaFetcher({
 
   const selectedWorkflow = workflow || initialWorkflow
 
+  if (!selectedWorkflow) {
+    const error = new Error(`"default_workflow" or "workflow.id" don't exist`)
```
nl0, line 674:
```suggestion
    const error = new Error(`"default_workflow" or "workflow.id" doesn't exist`)
```

nl0 (approved):

PR #2192, opened 2021-05-10
Support custom nav link (incl. admin UI)

nl0:
# Description

1. Support catalog settings file at `$serviceBucket/catalog/settings.json` with the single `customNavLink` field (using schemaless JSON bc it's not supposed to be editable by hand, only written by the catalog code).

2. Add admin section for customizing catalog settings.

Screen Shot 2021-05-10 at 17 19 56
Screen Shot 2021-05-10 at 17 20 06
Screen Shot 2021-05-10 at 17 20 28

# TODO

### DIFF
--- a/catalog/app/containers/Admin/Settings.tsx
+++ b/catalog/app/containers/Admin/Settings.tsx
@@ -12,16 +12,10 @@ import * as validators from 'utils/validators'
 import * as Form from './Form'
 
 const useNavLinkEditorStyles = M.makeStyles((t) => ({
-  section: {
+  actions: {
     alignItems: 'center',
     display: 'flex',
     marginTop: t.spacing(1),
-  },
-  fields: {
-    display: 'flex',
-    flexDirection: 'column',
-    marginRight: t.spacing(2),
-    overflow: 'hidden',
   },
   field: {
     display: 'flex',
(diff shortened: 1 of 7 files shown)


fiskus:
As this URL change affects other users, I think, we can limit link label, because extra long labels break layout.

Possibly we can improve admin UI by sticking buttons "edit"/"remove" to the right, or moving them to the next line.

Suggestion for another PR: I think we need to rename Menu item "Users and buckets" to "Admin" or add every admin page to it. Also, it's good to have note in docs, where can I find link customization.

catalog/app/containers/Admin/Settings.tsx
```diff
@@ -0,0 +1,257 @@
...
+  fieldName: {
+    ...t.typography.body2,
+    flexShrink: 0,
+    fontWeight: t.typography.fontWeightMedium,
+    width: 50,
+  },
+  fieldValue: {
+    ...t.typography.body2,
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
```
fiskus, line 38: If you intended to cut long labels, it doesn't work

catalog/app/utils/CatalogSettings.tsx
```diff
@@ -0,0 +1,78 @@
...
+    console.error(e)
+    return null
+  }
+}
+
+const CatalogSettingsResource = Cache.createResource({
+  name: 'CatalogSettings.config',
+  fetch: fetchSettings,
+  // @ts-expect-error
+  key: () => null,
```
fiskus, line 46: What is a meaning of `key === null`? Don't save to cache?

Probably, instead of explaining, it's better to have `Cache.createResourceOnSomeAdditionalTerms` method

catalog/app/containers/Admin/Settings.tsx
```diff
@@ -0,0 +1,257 @@
...
+        </div>
+      ) : (
+        <div className={classes.section}>
+          <div className={classes.notConfigured}>Not configured</div>
+          <M.Button variant="outlined" color="primary" size="small" onClick={edit}>
+            Add link
+          </M.Button>
+        </div>
+      )}
+      <M.Dialog open={editing} onExited={handleExited} fullWidth>
```
fiskus, line 151: I'd rather want to see this component divided into two pieces: one for viewing and the second for editing

fiskus (approved):

catalog/app/containers/Admin/Settings.tsx
nl0: i forgot `whiteSpace: 'nowrap'`, thx

catalog/app/containers/Admin/Settings.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const classes = useStyles()
+  return (
+    <M.Box mt={2} mb={2}>
+      <M.Paper>
+        <M.Box px={3} pt={2} pb={1}>
+          <M.Typography variant="h6">Catalog Customization</M.Typography>
+        </M.Box>
+        <M.Box px={3} pb={2}>
+          <div className={classes.sectionHeading}>Custom navbar link</div>
+          <React.Suspense fallback={<M.CircularProgress />}>
```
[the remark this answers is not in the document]
nl0: nope, it's a placeholder shown while catalog settings are being loaded

catalog/app/containers/Admin/Settings.tsx
fiskus: Yes. "View"-part is simple, you could declutter it from state management:

```tsx
const NavLinkEdit = () => { state && <Form />}
const NavLinkAdd = () => <button />
const NavLinkView = () => <div />

const NavLinkEditor = () => (
  {noLink ? <NavLinkAdd /> : <NavLinkView />}
  <NavLinkEdit />
)
```

nl0:
> As this URL change affects other users, I think, we can limit link label, because extra long labels break layout.

yep, they do, so we probably should limit the length, or maybe just have max-width set on the navbar items, tho probably the user editing the link sees the changes instantly and should use their judgment to decide whether they want broken navbar layout =)

> Possibly we can improve admin UI by sticking buttons "edit"/"remove" to the right, or moving them to the next line.

next line could work, but moving them to the far right doesnt seem right bc in case of short values the buttons become too far from the related data

> Suggestion for another PR: I think we need to rename Menu item "Users and buckets" to "Admin" or add every admin page to it.

it was that way earlier, but @akarve decided to rename it

> Also, it's good to have note in docs, where can I find link customization.

yep, it would be nice to have the docs for the admin ui =)


Defer to @akarve:

- should we limit label length or let the user (admin) break the navbar layout?
- should we rename Menu item "Users and buckets" to "Admin" or add every admin page to it?

catalog/app/containers/Admin/Settings.tsx
nl0: well, it's all coupled quite tight currently, and i dont see a simple way of splitting it clearly without leaking abstractions, passing a ton of callbacks and state around

fiskus:
> like that?

Yes, I think it's better, because it doesn't depend on label's length and consistent with different links

nl0:
> As this URL change affects other users, I think, we can limit link label, because extra long labels break layout.

how about smth like this?

Screenshot 2021-05-11 at 15 25 00
Screenshot 2021-05-11 at 15 25 32

akarve:
> should we limit label length or let the user (admin) break the navbar layout?
Limit but generous limit, IDK maybe 32

should we rename Menu item "Users and buckets" to "Admin" or add every admin page to it?
> "Admin settings"

A few other things
* It's not must-have but I would very much prefer a data grid for this UI with columns: key, value, text
* This table should be called "Global" or s/th like that

PR #2197, opened 2021-05-14
Athena Queries

fiskus:
* Separate ES and Athena queries to two routes
* Create Athena directory namespace for components
* Almost every request is paginated by AWS (using NextToken), so almost every response is saved at `[prev] = useState`, then next page is a "`prev` + `NextToken` response"
* Data is following: there is a list of workgroups, every workgroup has a list of saved named queries and list of queries executions (history)

- [ ] <del>Unit tests</del> (Don't know what to test here)
- [ ] <del>Documentation</del> (next PR #2199 )


catalog/app/containers/Bucket/Queries/requests/athena.ts
```diff
@@ -0,0 +1,328 @@
...
+  const status = statusData?.QueryExecution?.Status?.State
+  if (status === 'FAILED' || status === 'CANCELLED') {
+    throw new Error(status)
+  }
+
+  if (status === 'SUCCEEDED') {
+    return statusData?.QueryExecution || null
+  }
+
+  return waitForQueryStatus(athena, QueryExecutionId)
```
nl0, line 223: is this function polling for query results in a loop? if true, we probably need some timeout / sleep before concurrent requests.
also, we probably want to use a loop instead of recursion, bc js doesnt have tail call optimization and we can overflow the call stack if it takes too many iterations.

catalog/app/containers/Bucket/Queries/requests/athena.ts
```diff
@@ -0,0 +1,328 @@
...
+): AsyncData<QueryResultsResponse> {
+  const athena = AWS.Athena.use()
+  return useData(
+    fetchQueryResults,
+    { athena, prev, queryExecutionId },
+    { noAutoFetch: !queryExecutionId },
+  )
+}
+
+async function hashQueryBody(queryBody: string, workgroup: string): Promise<string> {
```
nl0, line 272: it could make sense to normalize query body (say, replace all whitespace with a single space, trim the string, maybe something else)

nl0 (approved):
the code looks good (some inline comments), i'm going to test the ui now

catalog/app/components/Skeleton/Skeleton.tsx
```diff
@@ -30,9 +30,9 @@ const useStyles = M.makeStyles((t) => ({
   },
 }))
 
-interface SkeletonProps {
-  className?: string
+interface SkeletonProps extends M.BoxProps {
   animate?: boolean
+  className?: string
```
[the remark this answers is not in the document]
fiskus: Oh, I see now what was happened.
I deleted it actually, then reverted deletion. I need to delete it again

catalog/app/containers/Bucket/Queries/Athena/Results.tsx
```diff
@@ -0,0 +1,111 @@
...
+  onLoadMore?: () => void
+}
+
+function Empty() {
+  return (
+    <M.Paper>
+      <M.Box p={3} textAlign="center">
+        <M.Typography variant="h6">No results for this query</M.Typography>
+        <M.Typography>
+          Select another query execution or execute another query
```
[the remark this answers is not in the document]
fiskus: No, this query doesn't have results. For example `CREATE DATABASE`

catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
```diff
@@ -0,0 +1,597 @@
...
+        handleWorkgroupsLoadMore,
+        queriesData,
+        queryMeta,
+        queryResultsData,
+        queryRunData,
+        workgroupsData,
+        workgroup,
+      }) => (
+        <div>
+          {queryRunData.case({
```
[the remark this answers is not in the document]
fiskus: Like this, right?

```tsx
<State>
  {({queryRunData}) =>
    queryRunData.case({
      Ok: () => <Redirect />
      _: () => (
        <>
          <All />
          <Other />
          <Components />
        </>
      )
    })
  }
</State>

File: catalog/app/containers/Bucket/Queries/requests/athena.ts (hunk 1 above)
Comment (fiskus) [reply catalog/app/containers/Bucket/Queries/requests/athena.ts:223]: There is only one concurrent request because it's a recursion with async/await. And by the same reason, I believe, it doesn't suffer from call stack overflowing

File: catalog/app/containers/Bucket/Queries/Athena/Athena.tsx (hunk 5 above)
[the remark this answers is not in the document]
Comment (nl0) [reply catalog/app/containers/Bucket/Queries/Athena/Athena.tsx]: yeah, that's one way of doing this, another being like this:
```typescript
<State>
  {({queryRunData}) => {
    if (AsyncResult.Ok.is(queryRunData.result)) return <Redirect />
    return (
      <>
        <All />
        <Other />
        <Components />
      </>
    )
  }}
</State>
```

the same, but less nesting

catalog/app/containers/Bucket/Queries/requests/athena.ts
nl0: i meant subsequent, not concurrent. call stack still can overflow when there's too many subsequent requests. and to avoid hammering the backend with one request after another, we may want to add a timeout between subsequent requests.

fiskus:
> 1. UI jumps when select has description

Fixed

> 2 3 5. AsyncResult/State related stuff

I'll create a backlog task for this.

> 4. Table filters

Backlog too

> 6. Store workgroup in querystring

I already thought about it, but I didn't like it. I'll make backlog task to think about it more
> 7. Run query → navigate back to history → no relevant query

Fixed
> 8. Right-aligned text columns

Made them left-aligned

> 9. The completion date is a link to the results

I think it's a natural pattern for facebook/twitter era
> 10. Double border

Fixed
> 11. Missing `id` at table

We can add it on demand
> 12. Missing white space

Fixed

> 13. Less space for tabs

Done

catalog/app/containers/Bucket/Queries/requests/athena.ts
fiskus: I added one-second timeout before making next request

nl0: yep, that's nice, but we still need to convert this to a loop

nl0: something like this would be safer:
```typescript
async function waitForQueryStatus(
  athena: Athena,
  QueryExecutionId: string,
): Promise<Athena.QueryExecution | null> {
  while (true) {
    const statusData = await athena.getQueryExecution({ QueryExecutionId }).promise()
    const status = statusData?.QueryExecution?.Status?.State
    if (status === 'FAILED' || status === 'CANCELLED') {
      throw new Error(status)
    }

    if (status === 'SUCCEEDED') {
      return statusData?.QueryExecution || null
    }

    await wait(1000)
  }
}
```
we can also add a timeout or limit number of requests to avoid infinite loop, but that's up to you

nl0:
here aligning items to the right actually made sense =)

Screenshot 2021-05-20 at 13 16 37

fiskus:
Yes, but I don't want to overengineer and guess column type. It's string right now

catalog/app/containers/Bucket/Queries/requests/athena.ts
fiskus: Done

catalog/app/containers/Bucket/Queries/requests/athena.ts
```diff
@@ -211,19 +211,21 @@ async function waitForQueryStatus(
...
-  const statusData = await athena.getQueryExecution({ QueryExecutionId }).promise()
-  const status = statusData?.QueryExecution?.Status?.State
-  if (status === 'FAILED' || status === 'CANCELLED') {
-    throw new Error(status)
+  let status: string | undefined = 'QUEUED'
+  let queryExecution = null
+  while (status === 'QUEUED' || status === 'RUNNING') {
+    // NOTE: await is used to intetionally pause loop and make requests in series
+    // eslint-disable-next-line no-await-in-loop
+    await wait(1000)
```
nl0, line 219: looks good, but i'd place the timeout after the request, so that the first request is fired right away, and subsequent ones -- following a delay

catalog/app/containers/Bucket/Queries/requests/athena.ts
```diff
@@ -0,0 +1,334 @@
...
+    // NOTE: await is used to intetionally pause loop and make requests in series
+    // eslint-disable-next-line no-await-in-loop
+    const statusData = await athena.getQueryExecution({ QueryExecutionId }).promise()
+    status = statusData?.QueryExecution?.Status?.State
+    queryExecution = statusData?.QueryExecution || null
+    if (status === 'SUCCEEDED') {
+      return queryExecution
+    }
+    // eslint-disable-next-line no-await-in-loop
+    await wait(1000)
```
nl0, line 226: seems like this will wait a second before throwing an error, so we probably need to move throw statement inside the loop as well under some condition. and after this we wont need to check for any conditions in the "while" condition clause (making it `while (true) {`). after this, we wont need `status` and `queryExecution` variables outside of "while", effectively transforming this function into the variant i've posted some time ago).
also, i dont see how the case with `status === undefined` is handled

fiskus: I use you variant now

PR #2199, opened 2021-05-18
Athena docs

fiskus:


akarve (changes requested):
Comments inline

fiskus:
There is no configuration file for Athena. I fetch queries using Athena SDK, not from `.quilt` directory

fiskus:
Actually, when you query with the same idempotency token after a while, Athena creates a new results table. I don't know why, I didn't find proof in the documentation.

Maybe, we can add the button "Reload" on the query results page to explicitly reset the idempotency token?

PR #2203, opened 2021-05-20
Delete package revision

fiskus:
Screenshot from 2021-06-03 11-04-39

- [ ] Unit tests
- [ ] Documentation


fiskus:
>  1. redirect to the package list?

I thought that we should redirect consistently to one page. And if we delete the only revision package has, then the package won't have the revisions list. Also, the deletion takes some time, and I thought it would be confusing to show the revision list with deleted but still present revision.
    
>  2. delete from revision list

I think deletion should be a little difficult. Let user see and inspect revision in details which he going to delete

akarve (changes requested):
Here's my suggested text for the dialog:

*Really delete revision <short_hash> of <package>?*

This package revision will be lost forever. Package deletion does not delete objects in the package, but it does delete all metadata and all records of the contents of this revision. Are you sure you want to delete it?

*CANCEL*, *YES, DELETE IT* with scary color

akarve:
Just to make sure do we have a way to hide this feature via config.yaml? We can't ship without that. Maybe even off by default IDK.

fiskus:
> do we have a way to hide this feature via config.yaml?

Yes, and it's documented here https://github.com/quiltdata/quilt/pull/2203/files#diff-650ecbf0441cc74e0380cd887426bea4693e7f39731f527b4438f5855357aa94

> off by default

Agree.
It's a bit inconsistent with other features but safe. If we decide to turn it on by default later:
1. users that chose to turn it on will not get affected
2. users that used defaults will get the new feature they didn't know about

akarve:
> > do we have a way to hide this feature via config.yaml?
> 
> Yes, and it's documented here https://github.com/quiltdata/quilt/pull/2203/files#diff-650ecbf0441cc74e0380cd887426bea4693e7f39731f527b4438f5855357aa94
> 
> > off by default
> 
> Agree.
> It's a bit inconsistent with other features but safe. If we decide to turn it on by default later:
> 
> 1. users that chose to turn it on will not get affected
> 2. users that used defaults will get the new feature they didn't know about

opt-in is the safest an should be our policy moving forward. as a rule we won't change our minds since 2 is bad :)

akarve (approved):

PR #2210, opened 2021-05-27
Handle no Athena workgroups

fiskus:
Screenshot from 2021-05-27 11-53-21
Screenshot from 2021-05-27 11-54-05

### DIFF
--- a/catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
+++ b/catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
@@ -371,11 +371,10 @@ function State({ children, queryExecutionId }: StateProps) {
     <WorkgroupsFetcher>
       {({ handleWorkgroupsLoadMore, workgroupsData }) =>
         workgroupsData.case({
-          _: (workgroupsDataResult) => {
-            const workgroups = workgroupsDataResult.value
-            return AsyncResult.Init.is(workgroupsDataResult) ||
-              AsyncResult.Pending.is(workgroupsDataResult) ||
-              workgroups?.list?.length ? (
+          _: (workgroupsDataResult) =>
+            AsyncResult.Init.is(workgroupsDataResult) ||
+            AsyncResult.Pending.is(workgroupsDataResult) ||
+            workgroupsDataResult.value?.list?.length ? (
               <QueryResultsFetcher queryExecutionId={queryExecutionId}>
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
```diff
@@ -341,61 +371,65 @@ function State({ children, queryExecutionId }: StateProps) {
...
-                        }
-                      </QueryRunner>
-                    )}
-                  </QueriesFetcher>
-                )
-              }}
-            </QueryResultsFetcher>
-          ),
+          _: ({ value: workgroups }) =>
+            AsyncResult.Init.is(workgroups) || workgroups?.list?.length ? (
```
nl0, line 375: the first argument passed to this function is `AsyncResult`, so `workgroups` cant be `AsyncResult`, bc its a `value` prop of that `AsyncResult`.
something like this should work:
```
 _: (x) =>
  AsyncResult.Init.is(x) || x.value?.list?.length ? (
```

nl0 (approved):
looking good mostly

catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
fiskus: `AsyncResult.Init.is(workgroups)` returns true right now

fiskus: Screenshot from 2021-05-27 17-17-25
Screenshot from 2021-05-27 17-17-36

nl0: 🤔 that's strange, but you can merge if you're confident that it works as expected, tho i'd suggest adding a note to possibly come back to that later

nl0: probably it's bc when result is pending it has previous state as a value

fiskus: I suppose this is more correct:
```tsx
 _: (x) =>
  AsyncResult.Pending.is(x) || x.value?.list?.length ? (
```

nl0: then probably `AsyncResult.Init.is(x) || AsyncResult.Pending.is(x) || x.value?.list?.length ? (` (if i understand the intent of this condition correctly)

PR #2212, opened 2021-05-27
GraphQL

nl0:
# Description

# TODO

- [x] invalidate graphql cache on login / logout
- [x] provide introspected schema to graphcache -- punt
- [x] polish / organize things better
- [x] extract graphql tooling setup into a separate PR and rebase

# A quick rundown:

- set up graphql machinery
- introduce `model` module with types (both static and runtime) defining our model with the help of `io-ts` / `fp-ts`
- use graphql to query buckets all across the catalog (via `utils/BucketConfig` mostly)
- use graphql queries and mutations to manage buckets in admin (also migrate bucket management forms to `final-form`)
- `utils/LinkedData` refactoring started (punted for now)
- ~~`utils/ResourceCache` adjustment to not cleanup cache immediately (fixed some authenticated-related issues)~~
- ~~minor auth logic adjustments~~


fiskus:
Could you provide a list of changes? Is this PR contains
* adding GraphQL to project
* admin/Buckets refactoring and connection with GraphQL

?
Or something else?

nl0:
> Could you provide a list of changes? Is this PR contains
> 
>     * adding GraphQL to project
> 
>     * admin/Buckets refactoring and connection with GraphQL
> 
> 
> ?
> Or something else?

A quick rundown:

- set up graphql machinery
- introduce `model` module with types (both static and runtime) defining our model with the help of `io-ts` / `fp-ts`
- use graphql to query buckets all across the catalog (via `utils/BucketConfig` mostly)
- use graphql queries and mutations to manage buckets in admin (also migrate bucket management forms to `final-form`)
- `utils/LinkedData` refactoring started (punted for now)
- `utils/ResourceCache` adjustment to not cleanup cache immediately (fixed some authenticated-related issues)
- minor auth logic adjustments

shared/graphql/schema.graphql
```diff
@@ -0,0 +1,208 @@
...
+  relevanceScore: Int!
+
+  snsNotificationArn: String
+  lastIndexed: Datetime
+  scannerParallelShardsDepth: Int
+  skipMetaDataIndexing: Boolean
+  fileExtensionsToIndex: [String!]
+}
+
+type Query {
```
fiskus, line 94: This name is too general. Do you have a plan to scale this? How can we add new queries or new Queries (like ES query and Athena query)?
Do you think we need namespaces inside one file or several files for schemas?

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -290,125 +445,120 @@ function BucketFields({ add = false, reindex }) {
...
-    scanner_parallel_shards_depth: get('scannerParallelShardsDepth', Number),
-    sns_notification_arn: get('snsNotificationArn', (v) => {
-      if (v === DO_NOT_SUBSCRIBE_SYM) return DO_NOT_SUBSCRIBE_STR
-      if (typeof v === 'string') return v.trim()
-      return undefined
-    }),
-    skip_meta_data_indexing: get('skipMetaDataIndexing'),
-    delay_scan: get('delayScan'),
+const ADD_MUTATION = `
+  mutation Admin_Buckets_Add($input: BucketAddInput!) {
```
fiskus, line 449: Maybe better to store them in .gql files to leverage syntax highlighting?

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -1,39 +1,177 @@
...
+      ? IO.success(u)
+      : IO.failure(u, c),
+  R.identity,
+)
+
+type FormSpec<Obj extends {}> = {
+  [K in keyof Obj]: (formValues: Record<string, unknown>) => Obj[K]
+}
+
+const editFormSpec: FormSpec<Model.BucketUpdateInput> = {
```
fiskus, line 82: Are you planning to move this kind of stuff (convert one data format to another) to Model?

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -290,125 +445,120 @@ function BucketFields({ add = false, reindex }) {
...
-          true,
-        )
-        push(`Bucket "${b.name}" added`)
-        t.track('WEB', { type: 'admin', action: 'bucket add', bucket: b.name })
-        close()
-      } catch (e) {
-        if (APIConnector.HTTPError.is(e, 409, /Bucket already added/)) {
-          throw new RF.SubmissionError({ name: 'conflict' })
+        const input = R.applySpec(addFormSpec)(values)
+        const res = await add({ input })
```
fiskus, line 492: Maybe use `Admin/requests.ts` and place this request there?

fiskus (commented):
I don't have much to say.

GraphQL stuff is ok. Naming in `schema.graphql` doesn't seem to scale, and I'm expecting to see GraphQL requests outside of React components.

But, honestly, this PR is mostly about adding io-ts and few other functional libraries. I hope it's not just a cool toy you play with but it will improve code. Now I see reducing type annotation but increasing library boilerplate. I need time to get used to these libraries, see more benefits besides reducing type annotations and get productive using them.

shared/graphql/schema.graphql
nl0: this name is graphql convention for the entry point type (https://graphql.org/learn/schema/#the-query-and-mutation-types). to add new queries / mutations, we just add fields to these entry point types (Query and Mutation respectively).
graphql doesnt support namespacing of any kind, unfortunately, -- everything lives in a single scope. we can split the declaration into several files and then merge / build the schema, but this doesnt seem necessary at this point.

catalog/app/containers/Admin/Buckets.tsx
nl0: in this particular case -- no, because it's not reusable -- it converts a local data representation (form data) into more general / reusable / canonical format defined in the schema / model

nl0: however, i do plan to find / write some harness for defining forms (something akin to https://github.com/ForbesLindesay/define-form)

catalog/app/containers/Admin/Buckets.tsx
nl0: that seems like a step backwards, bc
1. graphql provides a clean way to define requests without much boilerplate (i.e. converting data types back and forth and issuing a fetch call or smth), so everything you need to do to define a request is writing a graphql mutation (and a type annotation for it bc currently we dont have automatic type generation or anything like that);
2. i believe that code should be grouped by feature, not by "type" (so, for example, request definitions should be close to the code making these requests, not in a single place shared by the whole app), meaning we shouldnt put all admin requests in one place, bc they dont have anything in common (all the "common" / standardized parts are already in graphql schema and `model` module);

catalog/app/containers/Admin/Buckets.tsx
nl0: graphql syntax can be highlighted even inside strings (when using `gql` template tag at least), but maybe we'll move the queries out eventually for another reason -- it's easier to process these files and compile them into ts / js by webpack or maybe some other tool (inline processing is harder bc typescript doesnt support custom transformers, but it's still somewhat feasible)

nl0:
@fiskus 

> GraphQL stuff is ok. Naming in `schema.graphql` doesn't seem to scale

well, graphql doesnt support namespacing, and it doesnt seem to be a huge problem to come up with the names for everything we have in the model without running into conflicts

> and I'm expecting to see GraphQL requests outside of React components.

could you please elaborate on this?

> But, honestly, this PR is mostly about adding io-ts and few other functional libraries. I hope it's not just a cool toy you play with but it will improve code. Now I see reducing type annotation but increasing library boilerplate. I need time to get used to these libraries, see more benefits besides reducing type annotations and get productive using them.

this pr is focused on graphql and other stuff is just to support that cause.
the reason for io-ts / fp-ts is i needed an effective and convenient solution for runtime type-checking / conversion, and io-ts / fp-ts ecosystem seems like the most comprehensive one. i'll go into more details on that in a separate write-up. without that there would be a lot more boilerplate.

nl0:
this one is ready for merge, waiting for the backend changes

fiskus:
> GraphQL requests in `Admin/requests.ts`

I mean, we used to place requests in `requests.ts` prior to this PR. Also, I think it's good to separate requests from components. Like separating UI components from business logic components etc.

As I see you want to achieve something like self-containing components: component has corresponding logic, styles, and requests all in one file. Ok, if it's your choice, I'm fine with that though I think this leads to components-behemoths that hard to maintain.

fiskus (approved):

nl0:
> As I see you want to achieve something like self-containing components: component has corresponding logic, styles, and requests all in one file. Ok, if it's your choice, I'm fine with that though I think this leads to components-behemoths that hard to maintain.

not necessarily in one file, but i think we should group code by feature (all code related to one feature lives in one place), not by function (like all the requests in one place, all the reducers in one place and such).
we can always move things out of a file if it gets too large, like, make `Users` folder and place all sub-components and related queries / requests there (with some shared code in a sibling folder when necessary), i agree that behemoth components are not the way to go.

nl0:
closing in favor of #2238 and #2240

PR #2214, opened 2021-05-28
Click package PREFIX to filter package tab view

fiskus:
Screenshot from 2021-05-28 18-22-19


catalog/app/containers/Bucket/PackageLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+  name: string
+}
+
+export default function PackageLink({ bucket, name }: PackageLinkProps) {
+  const classes = useStyles()
+  const { urls } = NamedRoutes.use()
+  const [prefix, suffix] = name.split('/')
+  return (
+    <span className={classes.name}>
+      <Link to={urls.bucketPackageList(bucket, { filter: prefix })}>{prefix}</Link>/
```
nl0, line 24: we should set filter to `${prefix}/`

catalog/app/containers/Bucket/RevisionInfo.jsx
```diff
@@ -0,0 +1,192 @@
+import cx from 'classnames'
```
nl0, line 1: were there any significant change to this code or was it just migrated over to TS and moved to a separate file?

nl0 (approved):
looks good, but we should filter by `${prefix}/`, so, for example, after clicking `fiskus/*` prefix we dont have `fiskus-smth/*` in our list

catalog/app/containers/Bucket/RevisionInfo.jsx
fiskus: There no changes to this code at all (it still js btw, not typescript). I moved it to a separate file because I imported it to another component at the first iteration. Then I changed `BucketLink` component but decided to leave `RevisoinInfo` as separate component. Because it looks like a separate component and its easier to collaborate on small components rather then big

PR #2215, opened 2021-05-29
Disable S3 filepicker when no `ui.sourceBuckets` and set default source bucket

fiskus:
Screenshot from 2021-05-29 12-13-26

Spec: https://github.com/quiltdata/specs/pull/16


catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -1316,15 +1324,19 @@ export function FilesInput({
     setS3FilePickerOpen(true)
   }, [])
 
+  const isS3FilePickerEnabled = !!bucket && !!buckets?.length
```
nl0, line 1327: `bucket` is always a string (non-nullable), so what's the point checking `!!bucket`?

catalog/app/constants/urls.js
```diff
@@ -8,6 +8,7 @@ export const blogYC =
 export const caseStudy = 'https://quiltdata.com/docs/quants.pdf'
 export const company = 'https://company.quiltdata.com'
 export const docs = 'https://docs.quiltdata.com'
+export const docs35 = 'https://docs.quiltdata.com/v/master'
```
nl0, line 11: why do we need a different url here?

nl0 (commented):
looks good, but we should decide on setting the default bucket and incorporate that logic into code populating the source buckets

catalog/app/constants/urls.js
fiskus: There is no https://docs.quiltdata.com/catalog/preferences but https://docs.quiltdata.com/v/master/catalog/preferences

When we deploy a new version (v.3.5) first link will become available https://docs.quiltdata.com/catalog/preferences. And we could find all `docs35` and replace them with `docs` (Catalog Preferences, Athena Queries etc.)

nl0: maybe call the variable `docsMaster` then?

fiskus: I don't mind to use `docsMaster`.
My reason was that `docsMaster` is a good name and you don't have the itch to replace it with just `docs`.

fiskus:
Why can we merge it separately from the default buckets feature? This PR hides S3FilePicker if no source buckets. If a user wants to populate a package with the same bucket he needs to set `sourceBuckets` field in config.

nl0:
> Why can we merge it separately from the default buckets feature? This PR hides S3FilePicker if no source buckets. If a user wants to populate a package with the same bucket he needs to set `sourceBuckets` field in config.

well, we need at least some way to predictably set the default bucket. prior to this PR it was always set to the current bucket, after this PR is merged the default bucket is unpredictable (if there's more than 1 in the config)

nl0 (approved):

PR #2222, opened 2021-06-02
Save ES query results

fiskus:
Add "Run and Save" (or "Save results" depending on context) click on which triggers CreatePackageDialog with initial metadata `{ results, query }`


fiskus:
ES queries tab is deprecated. But idea to create files on-the-fly is promising

PR #2225, opened 2021-06-03
JsonEditor: add Autocomplete for Enum

fiskus:
Now instead of handling viewing and editing `EnumSelect` is intended to edit only. Its implementation replaced from `M.Select` to `Lab.Autocomplete`

Screenshot from 2021-06-04 15-02-36


catalog/app/components/JsonEditor/Cell.tsx
```diff
@@ -86,6 +86,8 @@ export default function Cell({
 
   const onKeyPress = React.useCallback(
     (event) => {
+      if (editing) return
+
       if (!editing) {
```
nl0, line 91: given previous statement, the condition will be always true, so `if` isnt required anymore

nl0 (approved):
lgtm

PR #2226, opened 2021-06-07
Increase bucket dropdown height

fiskus:
* make the default menu taller (fits 10 items on 1080p)
* end on a half list item so user keeps scrolling

Screenshot from 2021-06-07 17-23-51

Screenshot from 2021-06-07 17-30-15

### DIFF
--- a/catalog/app/containers/NavBar/BucketSelect.js
+++ b/catalog/app/containers/NavBar/BucketSelect.js
@@ -117,11 +117,18 @@ function CustomPopper({ style: css, ...props }) {
   )
 }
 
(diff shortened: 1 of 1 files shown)


nl0 (approved):
lgtm, tho i dont see how this is implemented:
> end on a half list item so user keeps scrolling

nl0:
also, 52vh may be a little too small on smaller (phone) screens

fiskus:
Solution is based on hard-coded number. The value of this PR is not so strong to bring code smell into codebase

PR #2228, opened 2021-06-15
Catalog: Bucket permissions: GraphQL

nl0:
# Description

# TODO

- [x] schema
- [x] model definition
- [x] wire up admin ui mutations
- [x] write updaters / optimistic results for mutations
- [x] normalize form input values (trim, etc)
- [x] fetch roles for users table via graphql


catalog/app/containers/Admin/RFForm.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react'
+import * as M from '@material-ui/core'
+
+type FieldProps = M.TextFieldProps & {
+  meta: $TSFixMe
+  input: {
+    value: $TSFixMe
+    onChange: (value: $TSFixMe) => void
```
fiskus, line 8: I see, that you are in a hurry :)
So, I'm approving now and review later, even after merging

fiskus (approved):

catalog/app/containers/Admin/RFForm.tsx
nl0: the problem with this is more fundamental -- we need a way to define form typings properly, so every field is strongly typed wrt field value, validator, initial value, errors and such (smth like this probably: https://github.com/ForbesLindesay/define-form/tree/master/packages/react-define-form), otherwise it doesnt make much sense

PR #2237, opened 2021-06-29
Decrease Bytes range downloading for Vega detection

fiskus:
It's a workaround to avoid a Signature Mismatch error.

I think if we wouldn't find out how to fix it properly, it worth using this.


nl0 (approved):
i doubt it will really make a difference. i bet it will be breaking intermittently with any value of `BYTES_TO_SCAN` (well, except the original smallish one), but we can use 64kb as an optimization -- it seems a little more reasonable to load 64kb than 128kb for (potentially) tens of files in search results or a dashboard, for example.

PR #2238, opened 2021-06-29
GraphQL tooling setup

nl0:
extracted from #2212 

- set up graphql client (urql), incl. auth
- set up graphql codegen (custom preset / plugin)
- add schema, generate introspection data and types


catalog/internals/graphql-codegen/quilt-graphql.js
```diff
@@ -0,0 +1,137 @@
+const assert = require('assert')
```
fiskus, line 1: Probably, you can use ES6 imports if you rename file to `quilt-graphql.mjs`

fiskus (approved):
I think you need to add instructions to README, how to setup dev environment with graphql

catalog/internals/graphql-codegen/quilt-graphql.js
nl0: that was the plan, but it doesnt work -- codegen uses `require` to load plugins and node throws an exception saying smth like "use `import` to load mjs modules"

nl0:
> I think you need to add instructions to README, how to setup dev environment with graphql

@fiskus there's no new setup needed other than the usual `npm i`. and when you edit schema or graphql documents, you need to `npm run gql:generate` to generate typescript files for them (i'll add a pre-commit hook and watch mode in a subsequent PR)

PR #2239, opened 2021-06-29
Resolve minor eslint warnings

fiskus:


catalog/app/utils/AWS/Signer.js
```diff
@@ -2,7 +2,7 @@ import SignerV4 from 'aws-sdk/lib/signers/v4'
 import * as React from 'react'
 import * as redux from 'react-redux'
 
-import * as Auth from 'containers/Auth'
+import * as authSelectors from 'containers/Auth/selectors'
```
nl0, line 5: what's the reason for this change?

nl0 (approved):

catalog/app/utils/AWS/Signer.js
fiskus: Cycle dependency:
Layout -> NavBar -> CatalogSettings -> AWS -> Signer,Credentials etc -> Auth -> SignIn, SIgnOut etc -> Layout

nl0: oh i see, thanks for catching this. but how was it working before?

nl0 (approved):

catalog/app/utils/AWS/Signer.js
fiskus: ES6 allows cycle dependencies. So, it's a working code, but it's dangerous and has lint warning

PR #2245, opened 2021-07-01
Multi-column quilt_summarize.json

fiskus:
* Support extended file syntax (`{ path }` instead of string) and list of files
* Re-use same components for the summary on Bucket/Overview and Bucket/Summary


catalog/app/containers/Bucket/requests/requestsUntyped.js
```diff
@@ -361,10 +363,17 @@ export const ensureObjectIsPresent = (...args) =>
     }),
   )
 
+function waitFor(timeout) {
```
nl0, line 366: see `utils/wait`

nl0 (approved):
looks good, some minor comments inline

PR #2247, opened 2021-07-03
Lambdas: deep indexing of Excel files (XLS, XLSX)

akarve:


lambdas/es/indexer/index.py
```diff
@@ -456,6 +457,21 @@ def maybe_get_contents(bucket, key, ext, *, etag, version_id, s3_client, size):
             # if this is not an HTML/catalog preview
             columns = ','.join(list(info['schema']['names']))
             content = trim_to_bytes(f"{columns}\n{body}", ELASTIC_LIMIT_BYTES)
+        elif inferred_ext == ".xls" or inferred_ext == ".xlsx":
```
sir-sigurd, line 460:
```suggestion
        elif inferred_ext in (".xls", ".xlsx"):
```

sir-sigurd (commented):
Are catalog changes intentional?

lambdas/es/indexer/index.py
akarve: > Are catalog changes intentional?

No. Messed up rebase. I'll fix it.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -59,6 +60,27 @@ def decompress_stream(chunk_iterator, compression):
             break
 
 
+def extract_excel(file_, as_html=False):
```
sir-sigurd, line 63: Might make sense to make it consistent with `extract_fcs()` and `extract_parquet()`:
```suggestion
def extract_excel(file_, as_html=True):
```

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -59,6 +60,27 @@ def decompress_stream(chunk_iterator, compression):
...
+        body - html or text version of *first sheet only* in workbook
+        info - metadata
+    """
+    try:
+        first_sheet = pandas.read_excel(file_, sheet_name=0)
+    except XLRDError:
+        first_sheet = pandas.read_excel(file_, sheet_name=0, engine='openpyxl')
+
+    if as_html:
+        html = remove_pandas_footer(first_sheet._repr_html_())  # pylint: disable=protected-access
```
sir-sigurd, line 78: Consider using [`DataFrame.to_html()`](https://pandas.pydata.org/pandas-docs/version/1.1.5/reference/api/pandas.DataFrame.to_html.html). It looks like `df._repr_html_() == df.to_html(notebook=True)`.

@nl0 
I noticed that `df._repr_html_()` outputs inline CSS, don't we do styling on frontend? If we do, we could omit CSS with `df.to_html(notebook=False)`.

nl0: i need to check and see which styles are provided by the backend and which -- by the front-end. ideally we should have all the styles provided by the frontend, yep, but let's backlog this. it would be very helpful to see isolated output of `df.to_html(notebook=True)` and `df.to_html(notebook=False)` for some sample file.

lambdas/shared/t4_lambda_shared/preview.py
akarve: done; for some weird reason did this on purpose thinking that default False was an anti-pattern

lambdas/shared/t4_lambda_shared/preview.py
akarve: i switched to the public method in this diff but there are still other places in the code that use the private one

akarve: (actually switching to `to_html(notebook=True)` breaks a unit test in preview so I'm not gonna mess with it right now)

kevinemoore (approved):

PR #2248, opened 2021-07-05
Vega fallback loader

fiskus:
If a request with "Range: bytes" fails then request full file contents.


catalog/app/components/Preview/loaders/Json.js
```diff
@@ -152,14 +153,41 @@ function JsonLoader({ gated, handle, children }) {
...
     Ok: ({ firstBytes, contentLength }) =>
       detectSchema(firstBytes) ? (
         <VegaLoader {...{ handle, children, gated: contentLength > MAX_SIZE }} />
       ) : (
         <JsonLoader {...{ handle, children, gated: contentLength > MAX_SIZE }} />
       ),
+
+    Err: (err) => {
+      if (!error) {
+        sentry('captureException', error)
```
nl0, line 185: did you mean this?
```suggestion
        sentry('captureException', err)
```

nl0 (commented):
seems like this code will try to load the whole file regardless of its size, and i'm afraid this can be a problem for like 10MB+ files

fiskus:
I added HEAD request to check the file's size

fiskus:
We discussed and decided to move head request BEFORE bytes request, because it makes more sense to make a correct request instead of making an error request and do something to fix it after.

nl0 (changes requested):
as per zoom discussion: we should use head request before scanning the first bytes

fiskus:
I moved `HEAD` request before `Range: bytes` request

nl0 (approved):

PR #2250, opened 2021-07-05
Update @testing-library and jest dependencies

fiskus:


catalog/app/components/Dropzone/Dropzone.spec.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment jsdom
```
fiskus, line 2: https://github.com/facebook/jest/pull/9874/files

https://jestjs.io/docs/configuration#testenvironment-string

catalog/package.json
```diff
@@ -171,8 +171,8 @@
     "@types/react-table": "^7.7.1",
     "@types/react-test-renderer": "^17.0.1",
     "@types/recompose": "^0.30.7",
-    "@typescript-eslint/eslint-plugin": "^4.28.1",
-    "@typescript-eslint/parser": "^4.28.1",
+    "@typescript-eslint/eslint-plugin": "^4.28.2",
+    "@typescript-eslint/parser": "^4.28.2",
```
nl0, line 175: some leftover stuff from the previous pr?

catalog/app/containers/Bucket/viewModes.spec.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment jsdom
```
nl0, line 2: i think we should set this globally in `jest.config.js`, since we're testing a web app and targeting browser-like environment

nl0 (changes requested):
i think we should set jest-environment globally. plus i've noticed some unrelated stuff in package.json diff -- is that intentional?

nl0 (approved):

PR #2251, opened 2021-07-05
Update Typescript

fiskus:
* tslib: 2.2.0 → 2.3.0
* fork-ts-checker-webpack-plugin: 6.2.6 →6.2.12
* ts-loader: 9.1.1 → 9.2.3
* typescript: 4.2.4 → 4.3.5


catalog/package.json
```diff
@@ -130,7 +130,6 @@
     "remarkable": "^1.7.4",
     "reselect": "^4.0.0",
     "sanitize.css": "^11.0.1",
-    "tslib": "^2.2.0",
```
nl0, line 133: this should be kept under "dependencies", bc it's a runtime dependency

nl0 (changes requested):

catalog/package.json
fiskus: But there are no actual "tslib" imports in our code. Is this runtime dependency injected during the building step? "build" directory doesn't have "tslib" either

nl0 (approved):

PR #2258, opened 2021-07-09
Subpackage download, frontend

fiskus:
Waiting for backend

- [x] <del>Unit tests</del> can't write test, because code inside large component and use HTTP request
- [x] <del>Documentation</del> functionality is obvious

### DIFF
--- a/catalog/app/containers/Bucket/PackageTree.js
+++ b/catalog/app/containers/Bucket/PackageTree.js
@@ -322,7 +322,7 @@ function DirDisplay({
             {!noDownload && (
               <FileView.ZipDownloadForm
                 className={classes.button}
-                label={path ? 'Download directory' : 'Download package'}
+                label={path ? 'Download sub-package' : 'Download package'}
                 suffix={downloadPath}
               />
             )}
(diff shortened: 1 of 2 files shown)


catalog/app/containers/Bucket/PackageTree.js
```diff
@@ -318,8 +322,8 @@ function DirDisplay({
             {!noDownload && (
               <FileView.ZipDownloadForm
                 className={classes.button}
-                label="Download package"
-                suffix={`package/${bucket}/${name}/${hash}`}
+                label={path ? 'Download directory' : 'Download package'}
```
nl0, line 325:
```suggestion
                label={path ? 'Download sub-package' : 'Download package'}
```

nl0 (approved):

catalog/app/containers/Bucket/PackageTree.js
fiskus: I think that the user can ask "what do you mean by sub-package", and the answer will be "it's just the directory you currently browsing". So, I think it's better to explicitly articulate what user will get

nl0: yeah, but maybe we should differentiate plain s3 directory from package directory, idk. @akarve wdyt?

PR #2259, opened 2021-07-09
Remove some unused / obsolete code

nl0:


catalog/app/utils/pipeThru.js
```diff
@@ -1,3 +1,4 @@
 import * as R from 'ramda'
 
+// TODO: deprecate, use pipe from fp-ts
```
fiskus, line 3:
```suggestion
// TODO: deprecate, use pipe from fp-ts
/** @deprecated */
```

fiskus: Screenshot from 2021-07-09 15-07-19

fiskus (approved):

catalog/app/utils/pipeThru.js
nl0: nice

PR #2261, opened 2021-07-09
Handle RoleAssigned error while deleting a role

nl0:


fiskus (approved):

sir-sigurd (approved):
LGTM, but IMHO it would look nicer if role being deleted was grayed out instead of instant  disappearing.

nl0:
> LGTM, but IMHO it would look nicer if role being deleted was grayed out instead of instant disappearing.

ok, i'll consider this, tho optimistic updates is a widely adopted pattern. one thing that could improve this experience is adding a field to the Role model indicating that it is assigned to someone and forbidding deletion in this case

PR #2265, opened 2021-07-10
Admin: roles ux improvements

nl0:


fiskus (approved):
I like always visible non-wrapped buttons more, although the table's content becomes wider and there is a horizontal scroll.
I think in further iterations we need to make a table denser (decrease font and padding)

nl0:
> although the table's content becomes wider and there is a horizontal scroll.

i didnt see h-scroll on staging data, but the lines get wrapped, so we should probably set some overflow: hidden / ellipsis (tho it doesnt work with tables without some hackery)

> I think in further iterations we need to make a table denser (decrease font and padding)

yeah probably

catalog/app/containers/Admin/BucketsPermissions.tsx
```diff
@@ -108,7 +111,23 @@ export default function BucketPermissions({
...
   return (
     <div className={className}>
-      <M.Typography variant="h6">Bucket access</M.Typography>
+      <div className={classes.heading}>
+        <M.Typography variant="h6">Bucket access</M.Typography>
+        <M.Tooltip
+          arrow
+          title={
+            <>
+              Admin users can see all the buckets, but can only access the contents of the
```
nl0, line 120: @akarve wordsmith pls

fiskus:
Horizontal scroll is on searchminimal, because there is long "sergey_test_managed" user

PR #2266, opened 2021-07-12
Admin UI docs

nl0:
# TODO

- [ ] complete the documentation @akarve


akarve (approved):
Looks quite good. Made some minor edits and added TODO to provide more detail on the bucket settings panel, which is quite intricate. Also added link to [Preferences.md] which admins will want to know about.

nl0:
> I can't remember if we need to minify them or did you already?

i didnt and i dont think we do, but i can if you think it's necessary

PR #2267, opened 2021-07-12
Bucket/Listing: escape quotes in filenames

nl0:


sir-sigurd (approved):

catalog/app/containers/Bucket/Listing.tsx
```diff
@@ -1057,7 +1057,7 @@ export function Listing({
           toolbar: { truncated, locked, loadMore, items, children: toolbarContents },
           footer: { truncated, locked, loadMore, items },
         }}
-        getRowId={(row) => row.name}
+        getRowId={(row) => row.name.replaceAll("'", "\\'")}
```
fiskus, line 1060: Maybe, `escape(row.name)`?

fiskus (approved):

catalog/app/containers/Bucket/Listing.tsx
nl0: well, first, it's [not recommended](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) to use `escape`, and second, it's a plain string in single quotes, so all the other characters except unescaped single quote _should_ be fine.

PR #2268, opened 2021-07-13
Admin/Roles: fix some lock-related layout issues

nl0:


catalog/app/containers/Admin/Roles.tsx
```diff
@@ -86,7 +86,14 @@ const useStyles = M.makeStyles((t) => ({
     position: 'absolute',
     right: 0,
     top: 64,
-    zIndex: 1, // above Select and Checkbox
+    zIndex: 3, // above Select, Checkbox and sticky table header
+  },
+  title: {
+    '&>*': {
```
fiskus, line 92: I think it's better to use an additional span instead of an asterisk.

fiskus (approved):

catalog/app/containers/Admin/Roles.tsx
nl0: well, there's actually an h2 inside, not a span, and there's no way to change that or target it specifically via a class api, so i guess asterisk is fine, bc the structure is quite predictable. or do you mean add our own span with a class?

PR #2272, opened 2021-07-15
Make `quilt3 catalog` dependencies optional

sir-sigurd:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation


api/python/quilt3/main.py
```diff
@@ -101,6 +100,8 @@ def _launch_local_s3proxy():
     Launches an s3 proxy (via docker)
     on localhost:5002
     """
+    import dns.resolver
```
sir-sigurd, line 103: I didn't add `dnspython` as dependency because this function is actually unused.
https://github.com/quiltdata/quilt/blob/fea619d385a87321c53e164c38fbb953c509a0e1/api/python/quilt3/main.py#L110-L112

It might be a good time to remove this function.

api/python/quilt3/main.py
```diff
@@ -149,7 +150,17 @@ def cmd_catalog(navigation_target=None, detailed_help=False):
...
+    try:
+        from .registry import app  # Delay importing it cause it's expensive.
+    except ModuleNotFoundError as e:
+        if e.name not in (
+            'flask',
+            'flask_cors',
+            'flask_json',
+        ):
+            raise
+        print('Running `quilt3 catalog` requires you to install quilt3[catalog]')
```
akarve, line 162:
```suggestion
        print(To use `quilt3 catalog` you must first install quilt3[catalog]')
```

api/python/quilt3/main.py
akarve: @kevinemoore can we safely remove `_launch_local_s3proxy()`?

akarve (approved):

api/python/quilt3/main.py
kevinemoore: Instead of removing it, I'd rather see us use it to avoid sending local-mode traffic to open.quiltdata.com. I don't believed it's being called currently so we could remove it. Maybe it's safest to remove it temporarily and bring that functionality back when we fix local-mode packages.

kevinemoore (commented):
This approach seems backwards to me. We're making the path towards open-source adoption of Quilt harder to streamline a developer use case. Wouldn't it make more sense to create a `quiltcore` library with minimal dependencies for the few power users that want to tightly control their Python dependencies.

If that's more work, let's do this for now, but plan to streamline the open-source path when we fix packages in the local catalog.

api/python/quilt3/main.py
sir-sigurd: I adjusted message, but removed `optional` since I think it's quite obvious :slightly_smiling_face: .

sir-sigurd:
Closed in favor of https://github.com/quiltdata/quilt/pull/2562.

PR #2277, opened 2021-07-18
adjust roles admin ui

nl0:


catalog/app/containers/Admin/Roles.tsx
```diff
@@ -48,10 +48,17 @@ const columns = [
...
   {
-    id: 'managed',
-    label: 'Managed',
+    id: 'source',
+    label: 'Source',
     getValue: (r: Role) => r.__typename === 'ManagedRole',
-    getDisplay: (value: boolean) => (value ? 'Yes' : 'No'),
+    getDisplay: (value: boolean) =>
+      value ? (
+        <abbr title="This IAM role is created and managed by Quilt">Quilt</abbr>
```
fiskus, line 56: I think we can add something about "per-bucket management"

catalog/app/containers/Admin/Roles.tsx
```diff
@@ -48,10 +48,17 @@ const columns = [
...
+      ) : (
+        <abbr title="This IAM role is provided and managed by you or another administrator">
```
fiskus, line 58: And something about using IAM role by link as-is

fiskus (approved):

akarve (approved):

PR #2284, opened 2021-07-26
Preview: fix re-rendering (and re-requesting) images in markdown files every 5 sec (cleanup interval)

nl0:
dont re-run processing function when its identity changes, since it's
usually inline and changes on every render (instead we should list deps explicitly
when required)


sir-sigurd (commented):
I don't understand what this does but for me it solves the problem of loading the same images again and again.

fiskus (approved):

PR #2290, opened 2021-07-31
Bucket deep indexing settings

nl0:
Screenshot 2021-07-31 at 16 32 48

# TODO

- [x] get defaults and constraints from the backend
- [ ] Documentation

### DIFF
--- a/catalog/app/containers/Admin/Buckets.tsx
+++ b/catalog/app/containers/Admin/Buckets.tsx
@@ -59,7 +59,7 @@ const normalizeExtensions = FP.function.flow(
   Types.decode(Types.fromNullable(IO.string, '')),
   R.replace(/['"]/g, ''),
   R.split(','),
-  R.map(R.trim),
+  R.map(R.pipe(R.trim, R.toLower)),
   R.reject((t) => !t),
   R.uniq,
   R.sortBy(R.identity),
@@ -67,7 +67,7 @@ const normalizeExtensions = FP.function.flow(
     exts.length ? (exts as FP.nonEmptyArray.NonEmptyArray<Types.NonEmptyString>) : null,
 )
 
-const EXT_RE = /\.[0-9a-zA-Z]+/
+const EXT_RE = /\.[0-9a-z_]+/
 
 const validateExtensions = FP.function.flow(normalizeExtensions, (exts) =>
   exts && !exts.every(R.test(EXT_RE)) ? 'validExtensions' : undefined,
--- a/catalog/app/containers/Admin/Buckets.tsx
+++ b/catalog/app/containers/Admin/Buckets.tsx
@@ -19,6 +19,7 @@ import * as Dialogs from 'utils/Dialogs'
 import type FormSpec from 'utils/FormSpec'
 import MetaTitle from 'utils/MetaTitle'
 import * as NamedRoutes from 'utils/NamedRoutes'
+import * as Sentry from 'utils/Sentry'
 import assertNever from 'utils/assertNever'
 import parseSearch from 'utils/parseSearch'
 import { useTracker } from 'utils/tracking'
@@ -98,16 +99,31 @@ const editFormSpec: FormSpec<Model.GQLTypes.BucketUpdateInput> = {
     (s) => s.trim() || 'null',
     Types.decode(Types.withFallback(Types.JsonFromString, null)),
   ),
-  fileExtensionsToIndex: R.pipe(
-    R.prop('fileExtensionsToIndex'),
-    Types.decode(Types.fromNullable(IO.string, '')),
-    R.split(','),
-    R.map(R.trim),
-    R.reject((t) => !t),
-    R.uniq,
-    (exts) =>
-      exts.length ? (exts as FP.nonEmptyArray.NonEmptyArray<Types.NonEmptyString>) : null,
-  ),
+  fileExtensionsToIndex: (values) =>
+    !values.enableDeepIndexing
+      ? []
+      : FP.function.pipe(
+          values.fileExtensionsToIndex,
+          Types.decode(Types.fromNullable(IO.string, '')),
+          R.split(','),
+          R.map(R.trim),
+          R.reject((t) => !t),
+          R.uniq,
+          R.sortBy(R.identity),
+          (exts) =>
+            exts.length
+              ? (exts as FP.nonEmptyArray.NonEmptyArray<Types.NonEmptyString>)
+              : null,
+        ),
+  indexContentBytes: (values) =>
+    !values.enableDeepIndexing
+      ? 0
+      : FP.function.pipe(
+          values.indexContentBytes,
+          Types.decode(Types.fromNullable(IO.string, '')),
+          R.trim,
+          R.ifElse(R.equals(''), R.always(null), Types.decode(Types.IntFromString)),
+        ),
   scannerParallelShardsDepth: R.pipe(
     R.prop('scannerParallelShardsDepth'),
     Types.decode(Types.fromNullable(IO.string, '')),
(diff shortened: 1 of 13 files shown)


sir-sigurd:
Height of extensions field looks strange:
image

nl0:
> Height of extensions field looks strange:

indeed it does. looks fine in ff on mac. which browser are you using?

sir-sigurd:
> indeed it does. looks fine in ff on mac. which browser are you using?

Firefox 88.0.1. Also tried 90.0 and 78.12.0esr and it's the same.
Looks fine in Chrome.

nl0:
next iteration
Screenshot 2021-08-03 at 19 26 52

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -371,17 +452,125 @@ function BucketFields({ name, reindex }: BucketFieldsProps) {
...
+
+          <M.Box pt={2.5}>
+            <RF.Field
+              component={Form.Checkbox}
+              type="checkbox"
+              name="enableDeepIndexing"
+              label={
+                <>
+                  Enable deep indexing
+                  <Hint>Deep indexing help text TBD @akarve pls halp!</Hint>
```
nl0, line 464: @akarve wordsmith pls

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -371,17 +452,125 @@ function BucketFields({ name, reindex }: BucketFieldsProps) {
...
+                <Lab.Alert
+                  className={classes.warning}
+                  icon={
+                    <M.Icon fontSize="inherit" className={classes.warningIcon}>
+                      error
+                    </M.Icon>
+                  }
+                  severity="warning"
+                >
+                  New deep indexing settings will only affect newly indexed files. You can
```
nl0, line 506: @akarve wordsmith pls

fiskus:
It's a bit confusing, that empty file extensions field means ".csv, .txt, .json etc". If I want to add one more extension to the default list, I need manually add all of them first. I remember that we talk about it, but I don't remember why we didn't change this :)

catalog/app/containers/Admin/Form.js
```diff
@@ -16,8 +17,8 @@ export function Field({ input, meta, errors, helperText, ...rest }) {
 
 const useCheckboxStyles = M.makeStyles({
   root: {
-    paddingBottom: 0,
-    paddingTop: 0,
+    marginBottom: -9,
+    marginTop: -9,
```
fiskus, line 21: I think we use units everywhere:

```
marginBottom: '-9px'
```

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -98,16 +102,32 @@ const editFormSpec: FormSpec<Model.GQLTypes.BucketUpdateInput> = {
...
-    R.prop('fileExtensionsToIndex'),
-    Types.decode(Types.fromNullable(IO.string, '')),
-    R.split(','),
-    R.map(R.trim),
-    R.reject((t) => !t),
-    R.uniq,
-    (exts) =>
-      exts.length ? (exts as FP.nonEmptyArray.NonEmptyArray<Types.NonEmptyString>) : null,
-  ),
+  fileExtensionsToIndex: (values) =>
```
fiskus, line 105: `values` is untyped (it's just a Record<string, unknown>), but I think it should

In callback `onSubmit(values)` values are untyped too

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -203,6 +223,48 @@ function SnsField({
   )
 }
 
+const useHintStyles = M.makeStyles((t) => ({
+  icon: {
+    fontSize: '1.125em',
+    marginLeft: t.spacing(0.5),
+    marginTop: -1,
+    opacity: 0.5,
+    verticalAlign: -4,
```
fiskus, line 232: This looks a bit hacky. As I see, `vertical-align: middle` gives the same result as marginTop + verticalAlign

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -222,20 +284,45 @@ const useBucketFieldsStyles = M.makeStyles((t) => ({
...
 
-function BucketFields({ name, reindex }: BucketFieldsProps) {
+function BucketFields({ bucket, reindex }: BucketFieldsProps) {
   const classes = useBucketFieldsStyles()
+
+  const [{ error, data }] = urql.useQuery({ query: CONTENT_INDEXING_SETTINGS_QUERY })
+  if (!data && error) throw error
+
+  const sentry = Sentry.use()
+  React.useEffect(() => {
```
fiskus, line 309: If an error is thrown, doesn't React complain that it's a different number of hooks between renders?

fiskus: In other words, it seems that `useEffect` should be after `if (condition) throw`

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -367,21 +455,128 @@ function BucketFields({ name, reindex }: BucketFieldsProps) {
         </M.AccordionSummary>
         <M.Box className={classes.group} pt={1}>
           {!!reindex && (
-            <M.Button variant="outlined" fullWidth onClick={reindex}>
-              Re-index and repair
-            </M.Button>
+            <M.Box pb={2.5}>
```
fiskus, line 458: You are already using `classes = useStyles()` in this component. So it's better to place styles there

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -367,21 +455,128 @@ function BucketFields({ name, reindex }: BucketFieldsProps) {
...
+                    editFormSpec.fileExtensionsToIndex(values),
+                  ) &&
+                  R.equals(
+                    bucket.indexContentBytes,
+                    editFormSpec.indexContentBytes(values),
+                  )
+                )
+                  return null
+              } catch {
+                return null
```
fiskus, line 500: Maybe, at least, make `console.error`?

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -957,7 +1172,14 @@ interface CRUDProps {
 }
 
 function CRUD({ bucketName }: CRUDProps) {
-  const [{ data }] = urql.useQuery({ query: BUCKET_CONFIGS_QUERY })
+  const [{ error, data }] = urql.useQuery({ query: BUCKET_CONFIGS_QUERY })
+  if (!data && error) throw error
+
+  const sentry = Sentry.use()
+  React.useEffect(() => {
```
fiskus, line 1179: This, I think, violates Rules of Hooks too

fiskus (approved):

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -367,21 +455,128 @@ function BucketFields({ name, reindex }: BucketFieldsProps) {
...
-            label="File extensions to index (comma-separated)"
-            fullWidth
-            margin="normal"
+            component={Form.Checkbox}
+            type="checkbox"
+            name="enableDeepIndexing"
+            label={
+              <>
+                Enable deep indexing
+                <Hint>Deep indexing help text TBD @akarve pls halp!</Hint>
```
akarve, line 472:
```suggestion
                <Hint>Deep indexing adds the _contents_ of an object to your search index, while shallow indexing only covers object metadata. Deep indexing may require more disk in ElasticSearch. Enable deep index when you want your users to find files by their contents.</Hint>_
```

catalog/app/containers/Admin/Buckets.tsx
```diff
@@ -367,21 +455,128 @@ function BucketFields({ name, reindex }: BucketFieldsProps) {
...
+                <Lab.Alert
+                  className={classes.warning}
+                  icon={
+                    <M.Icon fontSize="inherit" className={classes.warningIcon}>
+                      error
+                    </M.Icon>
+                  }
+                  severity="warning"
+                >
+                  New deep indexing settings will only affect newly indexed files. You can
```
akarve, line 513: Changing these settings affects files that are indexed after the change. If you wish to deep index existing feels, click "Re-index and repair."

nl0:
> It's a bit confusing, that empty file extensions field means ".csv, .txt, .json etc". If I want to add one more extension to the default list, I need manually add all of them first. I remember that we talk about it, but I don't remember why we didn't change this :)

@fiskus i dont think it's confusing, but i agree that maybe it's not an optimal UX, and i don't remember this discussion taking place tbh. i'm open to implementing a better UX if you have ideas, but we should account for default values vs overrides, like, for example, the case when defaults change in the template.

catalog/app/containers/Admin/Form.js
nl0: well, our JSS setup uses pixels as default unit, so i think it's ok. the only place it may cause bugs is line-height.

catalog/app/containers/Admin/Buckets.tsx
nl0: it is untyped bc this is how final-form works. i agree we need a solution for typing (defining) forms properly, and i think it's even backlogged, but right now this set of functions (a "form spec") solves exactly this problem -- converts untyped values dict into a properly typed object or throws an error if that's impossible.

catalog/app/containers/Admin/Buckets.tsx
nl0: `vertical-align: middle` results in a different layout:
Screenshot 2021-08-09 at 10 57 31

vs layout w/ `marginTop: -1` and `verticalAlign: -4`:
Screenshot 2021-08-09 at 10 57 44

catalog/app/containers/Admin/Buckets.tsx
nl0: > If an error is thrown, doesn't React complain that it's a different number of hooks between renders?

it would, but it throws and exception and crashes anyways, and this case is more like an assertion -- this should never happen and if it happens it indicates an application error and should be handled by developers, it's not an expected behaviour

> in other words, it seems that useEffect should be after if (condition) throw

but it is

catalog/app/containers/Admin/Buckets.tsx
nl0: well, this seems a little too trivial and unspecific to make a class for it

catalog/app/containers/Admin/Buckets.tsx
nl0: it does, but when it happens it means an application error, not something we should recover from

catalog/app/containers/Admin/Buckets.tsx
nl0: when it throws, it means the input is invalid (cannot be converted to a proper output format), so it's just a validation error -- we don't want to log validation errors

PR #2291, opened 2021-08-02
Reset ErrorBoundary on route change

fiskus:
This adds ability to click on Quilt logo and show home page instead of error

### DIFF
--- a/catalog/app/utils/ErrorBoundary.js
+++ b/catalog/app/utils/ErrorBoundary.js
@@ -1,4 +1,3 @@
-import { boundMethod } from 'autobind-decorator'
 import PT from 'prop-types'
 import * as React from 'react'
 import * as RC from 'recompose'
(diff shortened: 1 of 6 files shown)


catalog/app/utils/ErrorBoundary.js
```diff
@@ -17,23 +16,22 @@ export const createBoundary = (handle, name = 'ErrorBoundary') =>
...
-      const handled = handle(this.props, { reset })(error, info)
+      const handled = handle(this.props, { reset: this.boundReset })(error, info)
       if (handled && handled !== error) {
         this.setState({ handled })
       } else {
         throw error
       }
     }
 
-    @boundMethod
```
fiskus, line 36: This doesn't work. Maybe, because of typescript. I didn't dig into it because it is used very rarely, so it's better to remove this dependency

nl0 (approved):

nl0:
@fiskus pls add a changelog entry for this

fiskus:
Ok, I didn't add a changelog entry for a reason. I think this fix is to small and almost doesn't matter

PR #2293, opened 2021-08-04
require content indexing settings in environ

sir-sigurd:


nl0 (approved):

lambdas/preview/index.py
```diff
@@ -138,7 +140,7 @@ def lambda_handler(request):
         elif input_type == 'ipynb':
             html, info = extract_ipynb(get_bytes(content_iter, compression), exclude_output)
         elif input_type == 'parquet':
-            html, info = extract_parquet(get_bytes(content_iter, compression))
+            html, info = extract_parquet(get_bytes(content_iter, compression), max_bytes=EXTRACT_PARQUET_MAX_BYTES)
```
sir-sigurd, line 143: Shouldn't we use `max_bytes` var instead?
https://github.com/quiltdata/quilt/blob/b379c2b4749f57dc50a44f29afa37fb7ffe5bdd6/lambdas/preview/index.py#L99

nl0: we probably should, i dont have enough context tho. @akarve wdyt?

akarve: in theory yes but we shouldn't touch that in this changeset and it might make a huge difference in the number of bytes used here sinc C_L_B is so much larger. i honestly don't remember why I did this. but we should change in a separate PR if desired. https://github.com/quiltdata/quilt/blob/b379c2b4749f57dc50a44f29afa37fb7ffe5bdd6/lambdas/shared/t4_lambda_shared/preview.py#L195

PR #2295, opened 2021-08-06
Update @types, lint and test dependencies

fiskus:
* `@types/fontfaceobserver`: 0.0.6 → 2.1.0
* `@types`: patch updates
* **`@typescript-eslint/eslint-plugin`: 4.28.2 → 4.29.0** "return-await" rule fix
* `@typescript-eslint/parser`: 4.28.2 → 4.29.0
* `eslint`: 7.30.0 → 7.32.0
* **`eslint-import-resolver-webpack`: 0.13.0 → 0.13.1**
* `lint-staged`: 11.0.0 → 11.1.2
* `ts-jest`: 27.0.3 → 27.0.4
* `ts-loader`: 9.2.3 → 9.2.5
* `util`: new dependency because `aws-sdk` requires it but `eslint-import-resolver-webpack` removed it


catalog/app/containers/Bucket/requests/requestsUntyped.js
```diff
@@ -1,6 +1,7 @@
 import { join as pathJoin } from 'path'
 
 import * as dateFns from 'date-fns'
+import * as FP from 'fp-ts'
```
nl0, line 4: why are these changes necessary?

nl0:
> util: new dependency because aws-sdk requires it but eslint-import-resolver-webpack removed it

but why it wasnt installed automatically if it's a dependency of aws-sdk? is it a peer dependency?

fiskus:
> > util: new dependency because aws-sdk requires it but eslint-import-resolver-webpack removed it
> 
> but why it wasnt installed automatically if it's a dependency of aws-sdk? is it a peer dependency?

I don't know why they didn't include it as dependency. `querystring`, `buffer`, `events` etc. already included. Issue here https://github.com/aws/aws-sdk-js/issues/3501

catalog/app/containers/Bucket/requests/requestsUntyped.js
fiskus: > Why I replaced R.pipe with FP.functions.pipe? Or why I removed `.then`? Or both?

I removed `return await func().then()` because it violated `return-await` rule. Eslint showed error.
I replaced `R.pipe` with `FP.functions.pipe` because I thought you want it.

nl0 (approved):

PR #2297, opened 2021-08-09
Remove react-intl

fiskus:
Since we don't use `react-intl`, it went too far from ver.2, and it mostly uses js `Intl` module now, I removed it completely:

* `FormattedMessage` is replaced by an inline message
* `FormattedPlural` is replaced by a simple function (`utils/plural`)
* `FormattedNumber` is replaced by Intl.NumberFormat
* `FormattedRelative` is replaced by a custom function similar to that in react-intl@v2 (`utils/date`). They use `Intl.RelativeTimeFormat` now and it behaves differently


catalog/app/embed/File.js
```diff
@@ -114,7 +114,7 @@ function VersionInfo({ bucket, path, version }) {
                   <M.ListItemText
                     primary={
                       <span>
-                        <FormattedRelative value={v.lastModified} />
+                        {dateUtils.formatRelative(v.lastModified)}
```
nl0, line 117: using format function directly here is a little too naive, bc it doesnt handle memoization and updates like the [original component](https://github.com/formatjs/formatjs/blob/main/packages/react-intl/src/components/relative.tsx) did, so we probably should create a component that schedules updates and optionally memoizes the result

catalog/app/containers/Bucket/PackageRevisions.js
```diff
@@ -338,7 +348,7 @@ function Revision({ bucket, name, hash, stats, message, modified, metadata, coun
           <M.Typography component="span" variant="body2">
             {readableQuantity(stats.files)}
             &nbsp;
-            <FormattedPlural value={stats.files} one="file" other="files" />
+            {filesNumberStr}
```
nl0, line 351: can we create a component wrapping the `plural.format` call to keep the api?

catalog/app/utils/plural.ts
```diff
@@ -0,0 +1,31 @@
...
+  // NOTE: we don't need it yet and, maybe, never
+  // two?: (v: number) => string
+  // few?: (v: number) => string
+  // many?: (v: number) => string
+  other: (v: number) => string
+}
+
+function numberToRule(value: number) {
+  if (value === 0) return 'zero'
+  if (value === 1) return 'one'
```
nl0, line 13: we should keep either `if`s or `switch`

nl0 (changes requested):
We should create components for formatting plurals and relative dates which will handle memoization and updates (for dates). They will also have more ergonomic API for writing them inline inside JSX (and we can make API the same as the removed components have, so we don't need to change the callsites). I propose to unify formatting functions and components under `utils/format` module and give them shorter names for convenience (maybe keep the old ones as aliases to avoid renaming), e.g.
```tsx
import * as Format from 'utils/format'

function MyComponent() {
  return (
    <>
      <Format.Plural value={1} one="one" />
      <Format.Relative value={date} />
    </>
  )
}
```

catalog/app/utils/format.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const [t, setT] = React.useState(Date.now())
+
+  React.useEffect(() => {
+    const timerId = window.setInterval(() => setT(Date.now()), 1000)
+    return () => clearInterval(timerId)
+  }, [value])
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const str = React.useMemo(() => relativify(value), [value, t])
+  return <>{str}</>
```
nl0, line 46: can we return the resulting string directly?

PR #2299, opened 2021-08-12
Use Suspense/lazy to defer JsonDisplay rendering nested parts

fiskus:
When a package has huge metadata page freezes and consumes 100% CPU even when idle.

Solution: don't render (not just hide) collapsed parts of JSON. This leads to the new problem: UI becomes unresponsive after click while the expanded part is rendering. Next solution: render this expanded part with lazy (setTimeout trick to render DOM on new tick).

---
Please, see details in Slack, if you need it: URL with huge metadata and screencasts

### DIFF
--- a/catalog/app/components/JsonDisplay/JsonDisplay.js
+++ b/catalog/app/components/JsonDisplay/JsonDisplay.js
@@ -257,10 +257,12 @@ export default function JsonDisplay({
 
   return (
     <M.Box className={cx(className, classes.root)} {...props}>
-      <JsonDisplayInner
-        {...{ name, value, topLevel, defaultExpanded, classes }}
-        showKeysWhenCollapsed={computedKeys}
-      />
+      <React.Suspense fallback={<WaitingJsonRender />}>
+        <JsonDisplayInner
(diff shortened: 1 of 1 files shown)


catalog/app/components/JsonDisplay/JsonDisplay.js
```diff
@@ -181,9 +206,23 @@ const isPrimitive = R.anyPass([
   R.equals(undefined),
 ])
 
+function useComponentOnNextTick(Component, props, optTimeout) {
+  return useMemoEq([Component, props], () =>
+    React.lazy(async () => {
+      await wait(optTimeout || 300)
```
nl0, line 212: 300 seems way more then a single tick, so probably we can reduce this a little

nl0 (changes requested):
this looks good, i have just some nitpicks on layout:
1. indent "rendering" placeholder one level, so its text is on the same level as the key labels rendered afterwards
2. render closing bracket beneath the "rendering" placeholder

btw there's new `startTransition` feature in upcoming React release which can help solve this problem with less hackery and boilerplate

nl0 (approved):

PR #2302, opened 2021-08-13
Unify package creation / update dialogs (add s3 file picker to package creation dialog)

nl0:
# Description

Use FilesInput in package creation dialog to support adding files from s3 + refactor package creation dialog to share most of the logic with package update dialog.

# TODO

- [x] use proper copy and visual treatment for creation vs update dialog
- [x] fix a bug with creation dialog state not being reset
- [x] [Changelog](CHANGELOG.md) entry

PR #2303, opened 2021-08-16
Update prettier

fiskus:
* prettier: 2.2.1 → 2.3.2
https://prettier.io/blog/2021/05/09/2.3.0.html#javascript-1

Also, I added `app/ipynb.css` to `.prettierignore`


catalog/.prettierignore
```diff
@@ -2,3 +2,5 @@ build/
 node_modules/
 internals/generators/
 package-lock.json
+*.generated.ts
```
nl0, line 5: that's not required, since generated files are run through prettier when running the generation script

nl0 (changes requested):
let's merge this right after #2302 to avoid conflicts

nl0 (approved):

Issue #1929, opened 2020-11-19
Accessing package causes ascii decoding error


saksham219:
While using quilt3 3.2.1 for browsing a package that was creating using 3.1.2, I get an ascii decoding error.

```
quilt_pkg = quilt3.Package.browse("test_package/data", quilt_s3_bucket)
Downloading manifest: 100%|##############################################################| 41.6k/41.6k [00:01<00:00, 23.6kB/s]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/quilt3/telemetry.py", line 132, in decorated
    results = func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/quilt3/packages.py", line 547, in browse
    return cls._browse(name=name, registry=registry, top_hash=top_hash)
  File "/usr/local/lib/python3.6/dist-packages/quilt3/packages.py", line 571, in _browse
    return cls._from_path(local_pkg_manifest)
  File "/usr/local/lib/python3.6/dist-packages/quilt3/packages.py", line 577, in _from_path
    pkg = cls._load(open_file)
  File "/usr/local/lib/python3.6/dist-packages/quilt3/packages.py", line 720, in _load
    for _ in readable_file:
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 4791: ordinal not in range(128)
```

sir-sigurd:
What's the output of this?
```python
import locale
print(locale.getpreferredencoding(False))
```

saksham219:
@sir-sigurd 
`
>>> print(locale.getpreferredencoding(False))
ANSI_X3.4-1968
`

sir-sigurd:
@saksham219 could you check if `quilt3==3.3.0rc1` works?

PR #2318, opened 2021-08-25
Upgrade pylint to 2.10.2

fiskus:


nl0 (approved):

fiskus:
@sir-sigurd can you check this? For some reason code coverage shows an error

---

Some background: pylint showed warning after update; I stuck version to last compatible; then in this PR I un-stuck pylint version and hide warnings instead

sir-sigurd (changes requested):
I think that we should specify encoding instead of ignoring this linter rule. We already had a [bug](https://github.com/quiltdata/quilt/issues/1929) that affected users. In some cases we might want to use ASCII and use UTF-8 in others, so I'll take other this PR.

fiskus:
> I think that we should specify encoding instead of ignoring this linter rule

Sure, but I won't do this. If you don't want to temporarily ignore warnings and un-stick pylint version, I'm going to close this PR

PR #2325, opened 2021-09-01
Fix JsonEditor breadcrumbs styles broken after sanitize.css upgrade

fiskus:
Sanitize.css adds `nav li:before {content: 'invisible-symbol-for-voice-something'}`


nl0 (approved):
i hate this shit w/ sanitize tbh, like sometimes we need to clear some stupid margins on some neutral elements like paragraphs (and those margins only work for some special cases e.g. block-start or smth like this, so it's not that trivial to spot it). maybe we should switch to some plain old style reset?

fiskus:
I agree, sanitize.css is too smart. I use https://github.com/necolas/normalize.css/blob/master/normalize.css for many years and didn't have problems with it

PR #2326, opened 2021-09-01
Less bright colorschema for charts

fiskus:
I changed the algorithm a little: instead of getting color from a predefined scheme I get color alias and set shade programmatically shading color every three items

### DIFF
--- a/catalog/app/containers/Bucket/Overview.js
+++ b/catalog/app/containers/Bucket/Overview.js
@@ -165,36 +165,33 @@ function ObjectsByExt({ data, colorPool, ...props }) {
             const maxBytes = capped.reduce((max, e) => Math.max(max, e.bytes), 0)
             const max = Math.log(maxBytes + 1)
             const scale = (x) => Math.log(x + 1) / max
-            return capped.map(({ ext, bytes, objects }, i) => {
-              const color = getFaidedColor(colorPool, ext, i)
-              return (
-                <React.Fragment key={`ext:${ext}`}>
-                  <div className={classes.ext} style={{ gridRow: i + 2 }}>
-                    {ext || 'other'}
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Bucket/Overview.js
```diff
@@ -161,36 +165,33 @@ function ObjectsByExt({ data, colorPool, ...props }) {
             const maxBytes = capped.reduce((max, e) => Math.max(max, e.bytes), 0)
             const max = Math.log(maxBytes + 1)
             const scale = (x) => Math.log(x + 1) / max
-            return capped.map(({ ext, bytes, objects }, i) => {
-              const color = colorPool.get(ext)
-              return (
```
fiskus, line 166: After this line it's indentation change only

nl0 (changes requested):
i think there was some misunderstanding and we should make something a little different:

1. the stacked areas shouldn't be progressively more transparent from bottom to top, but should have somewhat transparent fill plus opaque stroke instead, without variation in transparency between each other (we should also make sure this works nice with hover effect, which renders non-focused areas more transparent);

2. the colors themselves should be less saturated (from the list provided by Matt or some close equivalents from the MUI pattern; the reason there were some different shades in the list, e.g. A200 vs 300, is that not all colors from the MUI palette are perceived as having the same(ish) level of saturation and brightness, so we should pick the specific shades we need by hand);

3. we should make sure white text is readable on every bar (this is not the case e.g. for yellow background in both the previous implementation and this pr).

nl0:
hover effect can be changed to smth like making the hovered area fill opaque

fiskus:
https://github.com/quiltdata/quilt/pull/2329

PR #2327, opened 2021-09-02
Fix image render recursion state

fiskus:


fiskus:
https://github.com/quiltdata/quilt/pull/2328

PR #2328, opened 2021-09-02
Fix preview rendering inside packages, refactor logic related to view modes

nl0:
# Description

- fix condition causing infinite re-renders
- move setState out of the render phase into useEffect call
- refactor for more reuse and encapsulation

# TODO

- [x] Fix unit tests


fiskus (approved):

catalog/app/containers/Bucket/viewModes.spec.ts
```diff
@@ -1,136 +1,164 @@
 /**
  * @jest-environment jsdom
```
fiskus, line 2: I think this is obsolete since you don't use window.fetch anymore

nl0: hm maybe

PR #2329, opened 2021-09-02
Use warmer colors for bucket's overview chart

fiskus:
Before:
1 before

After:
Screenshot from 2021-09-02 15-37-14


nl0 (approved):
works for now, pls make sure to backlog adding the stroke

PR #2332, opened 2021-09-03
Show message describing whats happening during manifest download

fiskus:
We can't use progressbar for now because after manifest download UI freezes, so there is text message instead


nl0 (approved):
works for me

PR #2338, opened 2021-09-07
Warmer chart colors with stroke

fiskus:
Screenshot from 2021-09-08 08-45-25

### DIFF
--- a/catalog/app/components/StackedAreaChart/StackedAreaChart.js
+++ b/catalog/app/components/StackedAreaChart/StackedAreaChart.js
@@ -78,7 +78,11 @@ export default function MultiSparkline({
       R.pipe(
         R.aperture(2),
         R.map(([bottom, top]) => ({
-          area: R.concat(mkPoints(top), R.reverse(mkPoints(bottom))),
+          area: R.concat(
+            mkPoints(top),
+            // Leave one pixel room for stroke
+            R.reverse(mkPoints(bottom).map(({ x, y }) => ({ x, y: y - 1 }))),
+          ),
           stroke: mkPoints(top),
         })),
       )(stacked),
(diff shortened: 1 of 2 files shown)


nl0 (approved):
looks nice, tho on hover i'd make the whole area opaque and the same color as the stroke, but i guess this is just a matter of preference

catalog/app/components/StackedAreaChart/StackedAreaChart.js
```diff
@@ -133,22 +141,36 @@ export default function MultiSparkline({
...
+            // eslint-disable-next-line react/no-array-index-key
+            key={`area:${i}`}
+            onMouseEnter={handleAreaEnter(i)}
+            onMouseLeave={handleAreaLeave}
+          >
+            <polygon
+              points={SVG.pointsToSVG(area)}
+              fill={
+                // If color is #COLOR7
+                areaPaints[i].ref.length === 7
```
nl0, line 155: i dont get it. what's special about color 7?

fiskus:
```
#COLORX
1234567
```

fiskus: #ABCDEF - hex color is a string 7 characters long

fiskus: Another variant is `url(#key)`, and it throws error when passed to `lighten()` function

nl0: oh i see

nl0:
i've spotted a bug -- mouse tracking is a little broken: try hovering over the areas somewhere on the left side of the chart -- area highlighting is a little off (i've checked staging and it's ok there)

nl0:
it's because of how polylines work, here's the explanation (i set polylines' fill opacity to non-zero):
Screenshot 2021-09-08 at 18 48 55

PR #2341, opened 2021-09-08
chart tweaks

nl0:
- softer (grey) cursor lines
- fix cursor tracking
- use transparency instead of lightening for fills
- remove transparency for root element
- remove unused and unnecessary `fade` prop
- swap area rendering order so that bottom areas overlap top areas (this also aligns order of the areas with the order of the items in the legend tooltip)


nl0:
Screenshot 2021-09-08 at 19 30 04

PR #2351, opened 2021-09-15
Remove Immutable.js

fiskus:
I mostly replicated Immutable-specific code with plain javascript or Ramda. As I see frontend is working, but, maybe, I broke something


catalog/app/containers/Auth/SSOAzure.js
```diff
@@ -1,4 +1,4 @@
-import { push } from 'connected-react-router/esm/immutable'
+import { push } from 'connected-react-router'
```
nl0, line 1: please backlog ditching `connected-react-router` as well

catalog/app/utils/Store.js
```diff
@@ -25,21 +24,18 @@ export const Provider = function StoreProvider({ initialState = {}, history, chi
       process.env.NODE_ENV === 'development' &&
       process.env.LOGGER_REDUX === 'enabled'
     ) {
-      const stateTransformer = (state) =>
-        // pure JS is easier to read than Immutable objects
-        Iterable.isIterable(state) ? state.toJS() : state
       // eslint-disable-next-line global-require
       const { createLogger } = require('redux-logger')
-      middlewares.push(createLogger({ stateTransformer, collapsed: true }))
+      middlewares.push(createLogger({ stateTransformer: R.identity, collapsed: true }))
```
nl0, line 29: do we still need `stateTransformer` parameter here?

catalog/app/utils/ResourceCache.js
```diff
@@ -57,83 +56,71 @@ const Action = tagged([
   'CleanUp', // { time: Date }
 ])
 
-const keyFor = (resource, input) => [resource.id, I.fromJS(resource.key(input))]
+const keyFor = (resource, input) => [resource.id, resource.key(input)]
```
nl0, line 59: i think this is where things have been broken: `resource.key` doesn't always return string, in fact, it usually returns some object passed as an input, and this object gets used as a key for the entry, so we're basically doing smth like `cache[{ some: input }] = entry`, which turns `cache` var into `{ '[object Object]': entry }` (check redux logs and inspect the state to make sure), so, it kinda works when there's no conflicts for every resource (only single item is fetched) or every key is a string, but the logic is broken.
solution: use some kind of proper map implementation which can use arbitrary objects as keys (js native `Map` is okayish, but it's mutable, and i'm not sure about its performance when using it as immutable -- creating a new map on every change)

nl0 (changes requested):
ResourceCache is broken, we need to fix it

nl0:
another option would be to allow only string keys for resources and explicitly require key construction functions for every resource (grep for `createResource` -- there's not that many resources defined currently)

catalog/app/containers/Auth/SSOAzure.js
fiskus: I don't need task for this

catalog/app/utils/Store.js
fiskus: I don't know. And I don't refactor code in this PR, only remove Immutable.js

nl0: well, you can remove it to avoid future confusion as to why was it kept

nl0:
what's the status here? it seems ready, but abandoned

catalog/app/utils/Store.js
fiskus: I shouldn't but I can remove it if you would be more polite

fiskus:
It's not abandoned, I just have tasks with higher priority

PR #2356, opened 2021-09-16
Landing pages

nl0:


akarve:
Good start. Do I understand correctly that `__example` only shows up on dev mode and therefore needs no robots.txt etc?

There is a bug in the testimonial slider that it is blank to start with.
Screen Shot 2021-09-17 at 19 51 57

nl0:
> Do I understand correctly that __example only shows up on dev mode and therefore needs no robots.txt etc?

yep

> There is a bug in the testimonial slider that it is blank to start with.

yeah i noticed that, and seems like we have it on the deployed website version as well, so ima backlog a fix (bc it's not that trivial to fix -- some upstream library bug or just weird behavior). it's not easily noticed tho bc it fixes itself after first automatic slide change event (~7 seconds after page load), so by the time a user reaches this section it usually looks good.

nl0:
@akarve should we merge this?

PR #2364, opened 2021-09-28
Using two JSON Schema from workflows on frontend

fiskus:


shared/schemas/workflows.yml.json
```diff
@@ -29,24 +28,36 @@
...
             "minLength": 1
           },
           "description": {
-            "type": "string",
             "description": "The workflow description displayed by Quilt in the Python API or web UI.",
+            "type": "string",
             "minLength": 1
           },
+          "handle_pattern": {
+            "type": "string",
```
sir-sigurd, line 46:
```suggestion
            "type": "string",
            "format": "regex",
```

nl0:
> > [Bug] Package name validation against handle_pattern isn't triggered when switching
> > workflows -- it's only triggered when editing the name afterwards. This allows
> > to bypass name validation.
> 
> Validation is triggered, but we show error only when:
> 
>     * submit failed
> 
>     * value was modified

it's not


https://user-images.githubusercontent.com/122294/137279618-7cd56917-6445-49bd-ae16-4cc58718d3f4.mov

fiskus:
Reproduced with your example, thanks. Technically, validation is triggered but with the previous workflow, didn't figure out why yet

PR #2366, opened 2021-10-05
Add a video transcode lambda

dimaryaz:
Requires an ffmpeg layer


lambdas/transcode/transcode/index.py
```diff
@@ -0,0 +1,51 @@
...
+
+@api(cors_origins=get_default_origins())
+@validate(SCHEMA)
+def lambda_handler(request):
+    """
+    Generate thumbnails for images in S3
+    """
+    url = request.args['url']
+
+    with tempfile.NamedTemporaryFile(suffix='.mp4') as output_file:
```
akarve, line 38: Can we run out of disk here (over 512MB)? I guess the Lambda just fails at that point.

lambdas/transcode/index.py
```diff
@@ -0,0 +1 @@
+from transcode.index import lambda_handler
```
akarve, line 1: I had to read this twice then I realized what it was.
```suggestion
"""entrypoint for lambda at root"""


from transcode.index import lambda_handler
```

lambdas/transcode/transcode/index.py
```diff
@@ -0,0 +1,51 @@
...
+FFMPEG = pathlib.Path('/opt/bin/ffmpeg')
+FFMPEG_TIME_LIMIT = 20
+DURATION = 5
+WIDTH = 320
+HEIGHT = 240
+
+
+@api(cors_origins=get_default_origins())
+@validate(SCHEMA)
+def lambda_handler(request):
```
akarve, line 32: We should have one unit test that calls this handler (all other lambdas have one). It can be simple / use tiny files. That way it gets invoked under 3.6, 3.7, 3.8, etc.

akarve: And then we'll need to add this lambda to the CircleCI pass.

akarve:
Oh @dimaryaz needs a changelog entry: https://github.com/quiltdata/quilt/blob/master/docs/CHANGELOG.md

lambdas/transcode/transcode/index.py
akarve: Since this depends on a binary, for unit test you can shim out the subprocess call (still useful test imo; b/c the calling code runs), or (not necessary) do what I did for thumbnail and make some tests conditional on a flag.

lambdas/transcode/index.py
dimaryaz: Haha, I got confused by this myself just now. It's actually not needed if we're not packaging ffmpeg here, so I'll remove it.

docs/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 ## CLI
 
 ## Catalog, Lambdas
+* [Added] Add a "transcode" lambda for previewing video files
```
akarve, line 16:
```suggestion
* [Added] Add a "transcode" lambda for previewing video files ([#2366](https://github.com/quiltdata/quilt/pull/2366/)
```

akarve:
you can `isort myfile` to kill that CI error. i am mystified with the lambda-shared test failed though.

lambdas/transcode/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+def pytest_addoption(parser):
```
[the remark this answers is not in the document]
akarve: actually maybe this is just an unintentional file add because you don't actually hit the ffmpeg binary in your tests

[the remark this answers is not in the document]
dimaryaz: Oops, yes, will delete it

lambdas/transcode/index.py
```diff
@@ -0,0 +1,54 @@
...
+    },
+    'required': ['url'],
+    'additionalProperties': False
+}
+
+FFMPEG = '/opt/bin/ffmpeg'
+FFMPEG_TIME_LIMIT = 20
+DURATION = 5
+WIDTH = 320
+HEIGHT = 240
```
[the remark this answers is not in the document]
dimaryaz: Oh, sure. Actually, it could even be a query parameter - maybe the way the thumbnail lambda does it:
https://github.com/quiltdata/quilt/blob/master/lambdas/thumbnail/index.py#L36

lambdas/transcode/index.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    url = request.args['url']
+
+    with tempfile.NamedTemporaryFile(suffix='.mp4') as output_file:
+        p = subprocess.run([
+            FFMPEG,
+            "-t", str(DURATION),
+            "-i", url,
+            "-vf", f"scale=w={WIDTH}:h={HEIGHT}:force_original_aspect_ratio=decrease",
+            "-timelimit", str(FFMPEG_TIME_LIMIT),
```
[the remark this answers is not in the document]
sir-sigurd: https://ffmpeg.org/ffmpeg.html
```
-timelimit duration (global)
    Exit after ffmpeg has been running for duration seconds in CPU user time. 
```

lambdas/transcode/index.py
```diff
@@ -0,0 +1,54 @@
...
+def lambda_handler(request):
+    """
+    Generate thumbnails for images in S3
+    """
+    url = request.args['url']
+
+    with tempfile.NamedTemporaryFile(suffix='.mp4') as output_file:
+        p = subprocess.run([
+            FFMPEG,
+            "-t", str(DURATION),
```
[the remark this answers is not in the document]
sir-sigurd: > Is mp4 supported by every browser these days?

I'm not sure about that. As I understand support of both container formats and codecs varies from browser to browser, so probably we should get audio/video codecs and container format from querystring.

I think @nl0 could bring more details.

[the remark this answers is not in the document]
nl0: according to the [wikipedia article](https://en.wikipedia.org/wiki/HTML5_video) and [caniuse query](https://caniuse.com/?search=video%20format), MPEG-4/H.264 is our safest bet for now

nl0 (commented):
looks ok mostly, Sergey already noted most of the issues (all of them are quite minor)

dimaryaz:
Changing the memory limit does in fact fix the performance. webm is still slower, but at least it works.

akarve:
> It almost looks like the URL download takes too long - but that shouldn't be possible, since it's S3.

wait can FFMPEG work on local data? in that case we can do a much faster ranged request from S3 and just take like the first 100MB of the file (need to do some math here). as in i think we are waiting on FFMPEG to download the whole file which is bad.

dimaryaz:
I'm pretty sure it's smart enough to stream the input.

On my computer, it transcodes the first 1s of the video faster than I can download the whole video.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@
 ## Catalog, Lambdas
 !-->
 
+# unreleased - YYYY-MM-DD
+## Python API
+
+## CLI
+
+## Catalog, Lambdas
+* [Added] Add a "transcode" lambda for previewing video files ([#2366](https://github.com/quiltdata/quilt/pull/2366/)
```
akarve, line 16:
```suggestion
* [Added] Add a "transcode" lambda for previewing video files ([#2366](https://github.com/quiltdata/quilt/pull/2366/))
```

lambdas/transcode/index.py
```diff
@@ -0,0 +1,95 @@
...
+    },
+    'required': ['url'],
+    'additionalProperties': False
+}
+
+FFMPEG = '/opt/bin/ffmpeg'
+
+# Lambda has a 6MB limit for request and response, however, base64 adds 33% overhead.
+# Also, leave a few KB for the headers.
+MAX_VIDEO_SIZE = 6 * 1024 * 1024 * 3 // 4 - 4096
```
akarve, line 37: What is the logic behind 4.1 kB for headers?

lambdas/transcode/index.py
```diff
@@ -0,0 +1,95 @@
...
+    width_str = request.args.get('width', '320')
+    height_str = request.args.get('height', '240')
+    duration_str = request.args.get('duration', '5')
+
+    try:
+        width = int(width_str)
+        if not 10 <= width <= 640:
+            raise ValueError
+    except ValueError:
+        return make_json_response(400, {'error': "Invalid 'width'"})
```
akarve, line 57:
```suggestion
        return make_json_response(400, {'error': "Invalid 'width', not in [10, 640])"})
```

akarve: And might be nice for maintainers if we have MIN_WIDTH and MAX_WIDTH vars.

lambdas/transcode/index.py
```diff
@@ -0,0 +1,95 @@
...
+
+    try:
+        height = int(height_str)
+        if not 10 <= height <= 480:
+            raise ValueError
+    except ValueError:
+        return make_json_response(400, {'error': "Invalid 'height'"})
```
akarve, line 64: Same suggestions as above.

lambdas/transcode/tests/conftest.py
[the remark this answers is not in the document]
akarve: I think we still need to delete this file

lambdas/transcode/tests/test_transcode.py
```diff
@@ -0,0 +1,77 @@
...
+        return 42
+
+
+def test_403():
+    """test 403 cases, such as Glacier"""
+    url = "https://example.com/folder/file.ext"
+    event = _make_event({"url": url})
+
+    # Get the response
+    with patch.object(index, 'FFMPEG', '/bin/false'):
```
akarve, line 32: oh nice use of /bin/false

akarve (changes requested):
mostly looks good a few small things. only hard requirement is to clean out conftest.py

lambdas/transcode/index.py
dimaryaz: Kind of arbitrary - but it's pretty unlikely for the headers to take up more than 4KB. When I try it in a browser, it's just under 1KB.

nl0 (approved):
lgtm, don't see anything suspicious

akarve (approved):

PR #2381, opened 2021-10-19
Fix internal refs in GitBook

sir-sigurd:
Can be checked [here](https://quiltdocs.gitbook.io/test/SV66xPMJkhcfXUxMxyl0/).


fiskus:
> 2021
> don't use spaces in filenames, because something will break eventually

nl0:
haha classic

PR #2482, opened 2021-10-29
Declare compatibility with jsonschema==4.*

sir-sigurd:
Its breaking changes seem to not affect us: https://github.com/Julian/jsonschema/blob/main/CHANGELOG.rst#v400

### DIFF
--- a/api/python/setup.py
+++ b/api/python/setup.py
@@ -71,7 +71,7 @@ setup(
         'tenacity>=5.1.1',
         'tqdm>=4.32',
         'requests_futures==1.0.0',
-        'jsonschema==3.*',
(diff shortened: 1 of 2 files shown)


akarve (approved):

nl0:
> > if that's the only limitation, we can probably try using the new drafts
> 
> For what exactly?

To be a little more "modern" and up-to-date with the current state of the ecosystem, and also to be able to use official docs, since they cover the latest draft, and oftentimes the notes on previous drafts state smth like "docs TBD"

> Also we'd need to drop Python 3.6 support at first, since jsonschema==4 dropped it.

sounds like a plan :)

PR #2489, opened 2021-11-01
Add Voila renderer and set default height

fiskus:


catalog/app/components/Preview/renderers/Voila.tsx
```diff
@@ -0,0 +1,14 @@
...
+import Iframe from './IFrame'
+
+const DEFAULT_IFRAME_HEIGHT = '1024px'
+
+export default (
+  ifr: React.HTMLProps<HTMLIFrameElement>,
+  props: React.HTMLProps<HTMLIFrameElement>,
+) => {
+  const heightLens = R.lensPath(['style', 'height'])
+  const defaultHeight = R.defaultTo(DEFAULT_IFRAME_HEIGHT)
```
nl0, line 12: all these functions (`heightLens`, `defaultHeight`, `R.over(heightLens, defaultHeight)`) don't depend on anything from the component scope, so it's possible to move them out of here to avoid creating functions on every render (in practice it's probably a very minor optimization, but anyways)

nl0 (approved):

PR #2500, opened 2021-11-08
update eslintrc (imports-related rules stopped working)

nl0:


fiskus:
Oh, right
I noticed that too but didn't pay attention why

fiskus (approved):

PR #2502, opened 2021-11-08
utlils/parseSearch: handle arrays (multiple values) explicitly

nl0:


fiskus:
Where do we use multiple values in query params?

fiskus (approved):

fiskus:
> nowhere, but they are supported, and this case is not handled in our code

Maybe, better to remove support of multiple values then. So, instead of `Array.isArray(key)` in every component that uses this API, make one change in `parseSearch`

nl0:
> > nowhere, but they are supported, and this case is not handled in our code
> 
> Maybe, better to remove support of multiple values then.

this makes sense, but i decided to introduce a backwards-compatible change instead, so that the new behavior is opt-in (even tho it may be considered a fix). i'll consider doing it in a clean PR later.

>  So, instead of `Array.isArray(key)` in every component that uses this API, make one change in `parseSearch`

but it's not `Array.isArray(key)` at every callsite, it's just adding an extra boolean param.

PR #2510, opened 2021-11-08
FIle metadata editor and fullscreen package metadata editor

fiskus:
# Description

* Move `MetaInput` from `PackageDialog` to its own module
* Rename `containers/Bucket/PackageDialog/MetaInputErrorHelper.js` to `components/JsonValidationErrors.tsx`
* At Create/Revise dialogs replace `JsonEditor` with `JsonDislpay` + `Button` opening popup
* Add slightly different view for `JsonEditor` with multiple columns side-by-side. It's enabled by `multiColumned` switcher (probably, I need to remove it soon)
* Add `MetadataEditor` for switching between Excel-like `JsonEditor` and IDE-like `jsoneditor-react`
* Use `EditFileMeta` for editing file's meta and `MetaInput` for editing package's meta, both use `components/MetadataEditor`

Screenshot from 2021-11-09 18-06-16

Screenshot from 2021-11-09 18-07-34

# TODO

- [ ] <del>Unit tests</del>
- [ ] Documentation


nl0 (commented):
Looks good and feels far more convenient than inline editor. Some comments inline, plus a bug and some UX enhancement suggestions (we can discuss these in more detail in person if you'd like):

1. When changing metadata (both package- and object-level), then clicking "cancel" and then reopening the metadata editor, changes persist ("cancel" doesn't revert the changes). These changes, however, are not sent to the backend when submitting the form, so it seems like a local state of the editor isn't being reset on "cancel".
2. When opening object-level metadata editor and clicking "submit" without changing anything, "undo changes" button appears, even tho nothing was really changed -- probably some equality checks could be improved to track this properly.
3. It would be very handy to see the object-level metadata state (unchanged / modified / added / removed) -- similar to the entry state, we could use an overlay icon, maybe with color treatment.
4. When removing an entry, we could show a metadata icon similar to non-removed entries -- this way we can see if the deleted entry had any metadata associated with it.
5. Probably it makes sense to normalize metadata, so that empty object is converted to `null` and not being sent to the backend nor displayed in the entry list.
6. Do we want some sort of drag-and-drop support for object-level metadata?
7. It probably would be very useful to have a way to revert metadata to the initial state and to completely remove it from an entry (additional buttons in the modal are probably the easiest solution).

fiskus:
>     1. Reset editor's state on "Cancel/Discard" click
Done
>     2. Don't show "undo changes" when meta isn't changed
Can't reproduce
>     3. Object-level metadata state (unchanged / modified / added / removed)
I need to move state handling from one component to another at the upper level. I'll do this in the next clean PR
>     4. Show "meta" icon for deleted files
Done
>     5. Normalize metadata
I'll do in the next PR
>     6. Drag-and-drop for object-level metadata
I'd prefer to wait for the customer's request (developer who can frequently work with this metadata is a kinda customer too)
>     7. Button for reverting metadata to the initial state
Same reason as №3. Also, I need to think about where to put this button in UI.

nl0 (approved):

PR #2516, opened 2021-11-10
Add a FAQ on pushing from Jupyter

akarve:
<del>
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation
</del>

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -62,3 +62,18 @@ a local machine or foreign region)—I/O is much faster.
 
 1. Increase [`QUILT_TRANSFER_MAX_CONCURRENCY`](api-reference/cli.md#quilt_transfer_max_concurrency)
 above its default to match your available vCPUs.
+
+## I'm having trouble pushing packages from inside of a notebook
+Since `quilt3.Package.push` displays progress bars, we recommended one of the
+following:
+
+1. Push from outside your notebook with the CLI (`quilt3 push ....`). You can
+use Jupyter terminal for this.
+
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -62,3 +62,18 @@ a local machine or foreign region)—I/O is much faster.
...
+
+1. Push from outside your notebook with the CLI (`quilt3 push ....`). You can
+use Jupyter terminal for this.
+
+1. If you must push from within the notebook, save it first, and use the `%%capture`
+magic to prevent output from cells that change during push. This prevents a race
+condition between hashing the file and physically copying it to S3, during which
+time the file hash and size might change, due to progress bars and other text
+that `quilt3` writes to your notebook. Failing this precaution, `quilt3 install`
+may fail since the file in S3 contains may have a different hash than your local
```
sir-sigurd, line 78: I don't think `quilt3 install` verifies hashes of data.

sir-sigurd (commented):
The problem doesn't look specific to quilt3 progress bars inside notebooks, so instead of adding such FAQ entry I think we should address somehow the underlying problem of "eventual consistency" of hashes and uploaded data.

docs/FAQ.md
akarve: I think it check sizes then. The important thing: this race condition will cause `quilt3 install` to fail.

sir-sigurd: It doesn't look like so.

drernie:
Doesn't seem to still be an issue. Closing.

PR #2522, opened 2021-11-10
Use lambda for hashing in pkgpush lambda 

sir-sigurd:
# Description

# TODO

- [x] limit max file size
- [x] Unit tests


lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,51 @@
...
 
+def calculate_pkg_hashes(s3_client, pkg):
+    entries = []
+    for lk, entry in pkg.walk():
+        if entry.hash is not None:
+            continue
+        if entry.size > S3_HASH_LAMBDA_MAX_FILE_SIZE:
+            raise FileTooLargeForHashing(lk)
+
+        entries.append(entry)
```
akarve, line 190: where do we actually use `entries`? i see `data['entries']` below so maybe this is some context magic but i find it unclear.

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,51 @@
...
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=S3_HASH_LAMBDA_CONCURRENCY) as pool:
+        fs = [
+            pool.submit(calculate_pkg_entry_hash, s3_client, entry)
+            for lk, entry in pkg.walk()
```
akarve, line 195: prefer something like `for lk, entry in needs_hash:`

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,51 @@
...
+    pkg_entry.hash = {
+        'type': 'SHA256',
+        'value': invoke_hash_lambda(url),
+    }
+
+
+def invoke_hash_lambda(url):
+    resp = lambda_.invoke(FunctionName=S3_HASH_LAMBDA, Payload=json.dumps(url))
+    assert 'FunctionError' not in resp
+    return json.load(resp['Payload'])
```
akarve, line 224: i think we can tighten this up a bit. first the S3_HASH_LAMBDA should take a hash type ("SHA256" for now) in case we change that. and that brings me to my second point. it's somewhat opaque to return a string fro Payload and just assume it's a SHA256.

so this function should do two things differently:
1. Return a full hash value: `{'type': 'SHA256', 'value': 'somethingsomethingsomething'}`
2. Verify that the hash is the right length and composition relative to the type (JSON schema good for this; we already use it to validate query params)

in this way invoke_hash_lambda will have a clearer contract and better ownership of the inputs and outputs.

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,51 @@
 logger = get_quilt_logger()
 
 
+def calculate_pkg_hashes(s3_client, pkg):
+    entries = []
+    for lk, entry in pkg.walk():
+        if entry.hash is not None:
```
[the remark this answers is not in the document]
akarve: so we walk the same package twice with complementary logic. i think it'd be much clearer if you walk the package once and partition it into two lists: `needs_hash` and `dont_hash`, or something like that.

lambdas/pkgpush/index.py
```diff
@@ -229,6 +281,15 @@ def from_botocore_error(cls, boto_error: ClientError):
         return cls(status_code, message)
 
 
+class FileTooLargeForHashing(ApiException):
+    def __init__(self, logical_key):
+        super().__init__(
+            HTTPStatus.BAD_REQUEST,
+            f'Package entry {logical_key!r} is too large for hashing. '
+            f'Max size is {S3_HASH_LAMBDA_MAX_FILE_SIZE} bytes.'
```
akarve, line 289: This is a good example of why units are useful in variable names.

`S3_HASH_LAMBDA_MAX_FILE_SIZE_BYTES` would be clearer.

lambdas/pkgpush/index.py
```diff
@@ -29,6 +30,11 @@
 PROMOTE_PKG_MAX_FILES = int(os.environ['PROMOTE_PKG_MAX_FILES'])
 PKG_FROM_FOLDER_MAX_PKG_SIZE = int(os.environ['PKG_FROM_FOLDER_MAX_PKG_SIZE'])
 PKG_FROM_FOLDER_MAX_FILES = int(os.environ['PKG_FROM_FOLDER_MAX_FILES'])
+S3_HASH_LAMBDA = os.environ['S3_HASH_LAMBDA']
+S3_HASH_LAMBDA_CONCURRENCY = int(os.environ['S3_HASH_LAMBDA_CONCURRENCY'])
```
akarve, line 34: we should call this something else. because it's not the same as lambda reserved concurrency. and that also makes me wonder what happens if this number is greater than reserved concurrency and how we keep those numbers in sync?

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,51 @@
...
+    )
+    pkg_entry.hash = {
+        'type': 'SHA256',
+        'value': invoke_hash_lambda(url),
+    }
+
+
+def invoke_hash_lambda(url):
+    resp = lambda_.invoke(FunctionName=S3_HASH_LAMBDA, Payload=json.dumps(url))
+    assert 'FunctionError' not in resp
```
akarve, line 223: Shouldn't we be checking status code as well? What happens if this assert fails to concurrent hashes in progress?

lambdas/pkgpush/tests/test_index.py
```diff
@@ -900,29 +900,106 @@ def test_s3_error(self):
...
+                'VersionId': self.entry_without_hash.physical_key.version_id,
+            },
+        )
+
+        assert self.entry_without_hash.hash == {
+            'type': 'SHA256',
+            'value': invoke_hash_lambda_mock.return_value,
+        }
+
+    def test_invoke_hash_lambda(self):
```
akarve, line 986: Did I miss a test where the invocation fails? Good to have.

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,51 @@
...
+        if entry.size > S3_HASH_LAMBDA_MAX_FILE_SIZE:
+            raise FileTooLargeForHashing(lk)
+
+        entries.append(entry)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=S3_HASH_LAMBDA_CONCURRENCY) as pool:
+        fs = [
+            pool.submit(calculate_pkg_entry_hash, s3_client, entry)
+            for lk, entry in pkg.walk()
+            if entry.hash is None
```
sir-sigurd, line 196:
```suggestion
            for lk, entry in entries
```

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,50 @@
...
+            continue
+        if entry.size > S3_HASH_LAMBDA_MAX_FILE_SIZE:
+            raise FileTooLargeForHashing(lk)
+
+        entries.append(entry)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=S3_HASH_LAMBDA_CONCURRENCY) as pool:
+        fs = [
+            pool.submit(calculate_pkg_entry_hash, s3_client, entry)
+            for lk, entry in entries
```
sir-sigurd, line 195:
```suggestion
            for entry in entries
```

lambdas/pkgpush/index.py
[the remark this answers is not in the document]
sir-sigurd: It uses the same pattern as here:
https://github.com/quiltdata/quilt/blob/6e212cd309355c96662352115d557920ff8beaeb/api/python/quilt3/packages.py#L983

Changing this is out of scope.

lambdas/pkgpush/index.py
sir-sigurd: 1. I think making hash lambda take/return hash type is overengineering, since I don't expect this function will work with anything besides sha256 anytime soon (actually I think it will never be needed)
2. if we need a sort of sanity checks the right place for them seems quilt3 API

lambdas/pkgpush/index.py
sir-sigurd: > This is a good example of why units are useful in variable names.

I'm fine with including units in variable name, but don't quite get why this is a good example. :thinking:

akarve: well, reader can't know that bytes is the right unit for the f-string without more research into units of variable

lambdas/pkgpush/index.py
akarve: well returning a full hash object is a quick change and it's more modular. like even if we don't pass SHA256 in, we should return it `type` so that the lambda is clear about what it's returning.

i still think it's important to check `len(hash.value)` at the least.

if for some reason there is a bug here we will be making make manifests that are impossible to reproduce b/c hash is unattainable.

lambdas/pkgpush/index.py
```diff
@@ -172,6 +179,50 @@
...
+    if pk.version_id is not None:
+        params['VersionId'] = pk.version_id
+    url = s3_client.generate_presigned_url(
+        ClientMethod='get_object',
+        ExpiresIn=S3_HASH_LAMBDA_SIGNED_URL_EXPIRES_IN,
+        Params=params,
+    )
+    pkg_entry.hash = {
+        'type': 'SHA256',
+        'value': invoke_hash_lambda(url),
```
akarve, line 216: Are we sure that one file per lambda is the fastest? What about 2 or 3 or N?

docs/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 ## Python API
 * [Added] Declared compatibility with `jsonschema==4.*`.
 
+## Catalog, Lambdas
+* [Changed] `pkgpush` lambda now uses dedicated lambda for hashing files.
```
sir-sigurd, line 15:
```suggestion
* [Changed] `pkgpush` lambda now uses dedicated lambda for hashing files to push larger packages faster from catalog.
```

lambdas/pkgpush/index.py
sir-sigurd: > well returning a full hash object is a quick change and it's more modular. like even if we don't pass SHA256 in, we should return type so that the lambda is clear about what it's doing.

Lambda gets URL and returns sha256, that's its API. It knows nothing about `PackageEntry.hash`. `invoke_hash_lambda()` is a simple wrapper for hash lambda. I could replace `S3_HASH` with `S3_SHA256` if it would make someone's life easier.

> i don't really understand why this couldn't be SHA-3 at some point.

It could be but I don't think it would be anytime soon so I don't see a point in this generalization for hash type.

sir-sigurd: > i still think it's important to check len(hash.value) at the least.

Still not understand why it should be here and not in quilt3 API.

lambdas/pkgpush/index.py
sir-sigurd: N is what we do now without dedicated hashing lambda.

There is no "optimal" predefined number for this as it depends on number of files, their sizes and lambda's remaining time. I don't want to add complexity with implementing such scheduler and making hash lambda work with many files in parallel because current implementation seem to work OK.

lambdas/pkgpush/index.py
sir-sigurd: > Shouldn't we be checking status code as well?

No, I think we get boto3 exceptions in this case.

lambdas/pkgpush/index.py
```diff
@@ -172,6 +186,55 @@
...
+    entries = []
+    for lk, entry in pkg.walk():
+        if entry.hash is not None:
+            continue
+        if entry.size > S3_HASH_LAMBDA_MAX_FILE_SIZE_BYTES:
+            raise FileTooLargeForHashing(lk)
+
+        entries.append(entry)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=S3_HASH_LAMBDA_CONCURRENCY) as pool:
```
dimaryaz, line 199: If I understand all this correctly, the only purpose of the thread pool is to wait for multiple lambas - and no actual work is being done there, right?

I see that the lamba API supports async calls using `InvocationType='Event'` - would that be useful? Though if the async API makes everything unnecessarily complicated, then thread pool is fine. (I can't tell how to wait for the results from the async call...)

sir-sigurd: > I see that the lamba API supports async calls using `InvocationType='Event'` - would that be useful? Though if the async API makes everything unnecessarily complicated, then thread pool is fine. (I can't tell how to wait for the results from the async call...)

It's not Python async and I don't think it could be useful.

lambdas/pkgpush/index.py
akarve: Because defensive programming. In many cases (e.g. pure catalog) `quilt3` is not even in the picture.

akarve (approved):

lambdas/pkgpush/index.py
sir-sigurd: > What happens if this assert fails to concurrent hashes in progress?

Actually that's a good question, unfortunately we can't "cancel" already invoked lambdas.

PR #2528, opened 2021-11-12
Metadata editor inline again

fiskus:
UI changes based on Aneesh feedback:

* Return table-view JsonEditor because small objects can be edited quickly and customers are used to it (with next iteration I make JsonEditor smaller)
* Remove "Main" header
* Edit metadata button uses "fullscreen" icon
* Consistent wording: package-level metadata, object-level metadata
* Don't use `<Code>` for filename

Screenshot from 2021-11-12 17-28-29


catalog/app/containers/Bucket/PackageDialog/MetaInput.tsx
```diff
@@ -246,7 +257,8 @@ export const MetaInput = React.forwardRef<HTMLDivElement, MetaInputProps>(
     const [locked, setLocked] = React.useState(false)
 
     // used to force json editor re-initialization
-    const [jsonEditorKey, setJsonEditorKey] = React.useState(1)
+    const [jsonInlineEditorKey, setJsonInlineEditorKey] = React.useState(1)
```
nl0, line 260: do we really need two different keys here? seems like they are updated simultaneously anyways

nl0 (approved):

catalog/app/containers/Bucket/PackageDialog/MetaInput.tsx
fiskus: There are two JsonEditors and they update each other respectively with different react keys.
I can't use one key: inline JsonEditor shouldn't re-render on inline change, because otherwise it loses focus

fiskus: Maybe, I'll add one layer of abstraction (one more component) with value/onChange interface, and remove these keys

PR #2532, opened 2021-11-15
Tune upload perf, better cleanup logic

nl0:
# Description

1. Limit hashing concurrency to prevent file access errors
2. Optimize upload code to reduce memory pressure (avoid creating and storing a closure for every upload)
3. Fall back to server-side hashing when it can't be done client-side

# TODO

### DIFF
--- a/catalog/app/utils/ResourceCache.js
+++ b/catalog/app/utils/ResourceCache.js
@@ -148,6 +148,8 @@ function* handleInit({ resource, input, resolver }) {
 }
 
 function* cleanup() {
+  // TODO: refactor cleanup logic, so that the cleanup action is only dispatched
+  // when there's anything to cleanup (to avoid re-renders every 5 sec)
   while (true) {
     yield effects.delay(RELEASE_TIME)
     yield effects.put(Action.CleanUp({ time: new Date() }))
--- a/catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
+++ b/catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
@@ -1,4 +1,5 @@
 import cx from 'classnames'
+import pLimit from 'p-limit'
 import * as R from 'ramda'
 import * as React from 'react'
 import { useDropzone, FileWithPath } from 'react-dropzone'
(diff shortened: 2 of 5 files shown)


akarve (commented):
I will test tomorrow and circle back

catalog/app/utils/ResourceCache.js
```diff
@@ -148,6 +148,8 @@ function* handleInit({ resource, input, resolver }) {
 }
 
 function* cleanup() {
+  // TODO: refactor cleanup logic, so that the cleanup action is only dispatched
+  // when there's anything to cleanup (to avoid re-renders every 5 sec)
   while (true) {
```
akarve, line 153: 😬

akarve (approved):
I am currently testing this with a 6.9GB upload on Chrome / Mac OS / 2016 MacBook Pro.

1. OK for separate PR: We should bump the warning size to idk 10GB (I got a warning for being over 1GB)
2. The UI thread stalled for me as I was typing in the name field (after dragging and dropping my files) — about a 2.5 sec pause but then it was OK.
3. My uplink is 21.5 Mbps and I seem to be ~ about saturating it; which is good. I'd say I'll finish 6.9GB in about 35 min total wall time.
4. OK for separate PR: The progress optically appears to stall when the upload goes over 1GB but it's only because the number changes more slowly :) We should either always show more decimal places of precision above 1GB and/or we should show indeterminate progress
5. OK for separate PR; it would be really nice to have some concurrency knobs (if meaningful) in the catalog settings so that different machines/users have some control and we can learn from the field how to optimize performance

nl0:
> The UI thread stalled for me as I was typing in the name field (after dragging and dropping my files) — about a 2.5 sec pause but then it was OK.

it was processing the dropped files (synchronously), so it's kinda expected, tho not very nice, i guess

nl0:
> OK for separate PR; it would be really nice to have some concurrency knobs (if meaningful) in the catalog settings so that different machines/users have some control and we can learn from the field how to optimize performance

@akarve i'm not really sure how to approach this, ping me in slack for a discussion if you feel this is important

PR #2542, opened 2021-11-19
Add audio transcoding: mp3 and ogg

dimaryaz:


lambdas/transcode/index.py
```diff
@@ -7,14 +7,24 @@
 from t4_lambda_shared.decorator import api, validate
 from t4_lambda_shared.utils import get_default_origins, make_json_response
 
+VIDEO_FORMATS = ['mp4', 'webm']
+AUDIO_FORMATS = ['mp3', 'ogg']
```
akarve, line 11: Can we support FLAC? WAV? (Customer requests)
Same question for MPEG-TS and M3U8 (might just work?). Some .TS examples here https://open.quiltdata.com/b/streaming-orcasound-net/tree/rpi_orcasound_lab/hls/1541027406/

OK to do these new formats in a separate PR or postpone until after re:Invent.

lambdas/transcode/index.py
```diff
@@ -75,16 +85,30 @@ def lambda_handler(request):
     except ValueError:
         return make_json_response(400, {'error': "Invalid 'duration'"})
 
+    format_params = []
+
+    if format in AUDIO_FORMATS:
+        format_params.extend([
+            '-b:a', '128k',  # 128KB audio bitrate
```
akarve, line 92: Should the requester be able to override this with a query param?

akarve (approved):

lambdas/transcode/index.py
dimaryaz: Sure.

Though there's also a question: do we care about audio bitrate for videos? And video bitrate?

akarve: > do we care about audio bitrate for videos?
No for simplicity

> And video bitrate?
Yes if it's simple to parameterize, otherwise No for simplicity

lambdas/transcode/index.py
sir-sigurd: > Can we support FLAC? WAV? (Customer requests)

For output?

lambdas/transcode/index.py
```diff
@@ -7,14 +7,24 @@
...
 
+VIDEO_FORMATS = ['mp4', 'webm']
+AUDIO_FORMATS = ['mp3', 'ogg']
+
+CONTENT_TYPES = {
+    'mp4': 'video/mp4',
+    'webm': 'video/webm',
+    'mp3': 'audio/mpeg',
+    'ogg': 'audio/ogg',
+}
```
sir-sigurd, line 18: We could do something like this:

```suggestion
VIDEO_FORMATS =  {
    'mp4': 'video/mp4',
    'webm': 'video/webm',
}
AUDIO_FORMATS = {
    'mp3': 'audio/mpeg',
    'ogg': 'audio/ogg',
}

CONTENT_TYPES = {
    **VIDEO_FORMATS,
    **AUDIO_FORMATS,
}
```

Or we could just accept output format as content type so wouldn't need these mappings at all.

lambdas/transcode/index.py
akarve: No for input

lambdas/transcode/index.py
```diff
@@ -75,16 +85,30 @@ def lambda_handler(request):
...
             "-t", str(duration),
             "-i", url,
             "-f", format,
-            "-vf", ','.join([
-                f"scale=w={width}:h={height}:force_original_aspect_ratio=decrease",
-                "crop='iw-mod(iw\\,2)':'ih-mod(ih\\,2)'",
-            ]),
+            *format_params,
             "-timelimit", str(request.context.get_remaining_time_in_millis() // 1000 - 2),  # 2 seconds for padding
             "-fs", str(MAX_VIDEO_SIZE),
```
fiskus, line 113: would be useful to parametrize this, so we could use this **instead** of duration.

Use case: show list of videos each no more than 6Mb, no matter how long it is

fiskus (approved):

lambdas/transcode/index.py
dimaryaz: Oh, this is all about output formats. It'll support whatever ffmpeg supports as input - not sure if there's even a way to specify it.

lambdas/transcode/index.py
dimaryaz: Hm, content type is an interesting idea. It would also solve the video/audio ambiguity - e.g., could have `video/webm` and `audio/webm` if we wanted to.

lambdas/transcode/index.py
sir-sigurd: From man:

> Set the file size limit, expressed in bytes. No further chunk of bytes is written after the limit is exceeded. **The size of the output file is slightly more than the requested file size**.

so I don't think we could rely on it.

fiskus: It's more predictable than relying on duration+bitrate for setting expected filesize :)

akarve: ☝️ I agree with Max

lambdas/transcode/index.py
akarve: OK resolving this

lambdas/transcode/index.py
dimaryaz: So what does everyone think about using mime types here?

fiskus: Do I understand correctly, that by using mime type you're meaning that API instead of `format=mp4` will have something like `mime=video/mp4`, and will return mp4 video according to to this mime type? Yes, it's probably better because - you're already pointing this out - we can set audio or video for the same extension.

dimaryaz: Yeah, exactly. (Could still call it `format`, though.)

The only argument against it I can think of is: we will be relying on someone else's standard, which might not be perfect and might not exactly fit what we want. E.g., video/mp4 might refer to multiple codecs, and we'll have no way to specify that. If we wanted to be perfect, we'd specify the container format, the audio codec, and the video codec - but that seems a bit too much.

But I think it's a pretty minor thing right now.

sir-sigurd: > E.g., video/mp4 might refer to multiple codecs, and we'll have no way to specify that.

https://developer.mozilla.org/en-US/docs/Web/Media/Formats/codecs_parameter

lambdas/transcode/index.py
```diff
@@ -3,18 +3,30 @@
...
+
 SCHEMA = {
     'type': 'object',
     'properties': {
         'url': {
             'type': 'string'
         },
         'format': {
-            'enum': ['mp4', 'webm']
+            'enum': list(FORMATS) + list(LEGACY_FORMATS)
```
akarve, line 29: not a bug as written but it could become a bug if keys and values overlap so might wanna check that e.g. with `len(set(SCHEMA['format']['enum']))`

lambdas/transcode/index.py
```diff
@@ -49,44 +62,56 @@ def lambda_handler(request):
...
-            raise ValueError
-    except ValueError:
-        return make_json_response(400, {'error': f"Invalid 'height'; must be between {MIN_HEIGHT} and {MAX_HEIGHT}"})
+        try:
+            value = type(default_value)(value_str)
+            if not min_value <= value <= max_value:
+                raise ValueError
+            return value
+        except ValueError:
+            raise ValueError(f"Invalid {name!r}; must be between {min_value} and {max_value}")
```
akarve, line 78:
```suggestion
            raise ValueError(f"Invalid {name!r}; must be between {min_value} and {max_value} inclusive")
```

lambdas/transcode/index.py
```diff
@@ -49,44 +62,56 @@ def lambda_handler(request):
...
+            '-b:a', f'{audio_bitrate}k',
+            '-vn',  # Drop the video stream
+        ])
+    elif category == 'video':
+        format_params.extend([
+            "-vf", ','.join([
+                f"scale=w={width}:h={height}:force_original_aspect_ratio=decrease",
+                "crop='iw-mod(iw\\,2)':'ih-mod(ih\\,2)'",
+            ]),
+        ])
```
akarve, line 104: Do you want `else throw` here?

lambdas/transcode/index.py
```diff
@@ -97,4 +122,13 @@ def lambda_handler(request):
 
         data = output_file.read()
 
-    return 200, data, {'Content-Type': f'video/{format}'}
+    parsed = urlparse(url)
+    filename = parsed.path.rsplit('/', 1)[-1]
```
akarve, line 126: very minor and sorry if I'm being pedantic, but this feels cleaner
```suggestion
    filename = parsed.path.rpartition('/')[-1]
```

akarve (approved):

lambdas/transcode/index.py
dimaryaz: LOL. I used to have `else: assert False`, and got a pycov error cause that line wasn't covered. But, obviously, it can't be covered since it can't happen. What's the solution here?

akarve: You _could_ patch `FORMATS`... up to you though

lambdas/transcode/index.py
dimaryaz: I'd just remove the legacy formats as soon as nl0 updates his code

PR #2545, opened 2021-11-19
Object-level metadata validation (frontend)

fiskus:


fiskus:
Too trivial to keep it

PR #2547, opened 2021-11-22
Audio preview

fiskus:
Screenshot from 2021-11-22 13-46-30

# TODO

- [ ] <del>Unit tests</del>
- [ ] <del>Automated tests (e.g. Preflight)</del>
- [x] Documentation

### DIFF
--- a/catalog/app/components/Preview/load.js
+++ b/catalog/app/components/Preview/load.js
@@ -31,10 +31,10 @@ const loaderChain = [
   Pdf,
   Vcf,
   Html,
-  Text,
   Image,
   Video,
   Audio,
+  Text,
   fallback,
 ]
 
--- a/catalog/app/components/Preview/load.js
+++ b/catalog/app/components/Preview/load.js
@@ -1,5 +1,6 @@
 import * as React from 'react'
 
+import * as Audio from './loaders/Audio'
 import * as Csv from './loaders/Csv'
 import * as Echarts from './loaders/Echarts'
 import * as Excel from './loaders/Excel'
@@ -33,6 +34,7 @@ const loaderChain = [
   Text,
   Image,
   Video,
+  Audio,
   fallback,
 ]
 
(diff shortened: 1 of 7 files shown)


catalog/app/components/Preview/load.js
```diff
@@ -30,9 +31,10 @@ const loaderChain = [
   Pdf,
   Vcf,
   Html,
-  Text,
```
fiskus, line 33: `Text` handles audio format `.ts`. And, I think Text should be at the end of list, because it's the last resort to render file :)

I think, we need a switcher for `.ts`, the same as for Voila/Notebook

nl0: not sure about switch, probably we can do just fine without typescript previews)

nl0:
the ui is quite ugly imo, but that's totally fine for the initial implementation. probably we can do some fancy stuff like rendering waveforms in some not so distant future)

catalog/app/components/Preview/loaders/Audio.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const detect = utils.extIn([
+  '.flac',
+  '.m2t',
+  '.mp3',
+  '.mp4',
+  '.ogg',
+  '.ts',
+  '.tsa',
+  '.tsv',
```
nl0, line 20: this conflicts w/ tab-separated values

nl0 (approved):

fiskus:
I think, we can use some library like howler.js, though howler.js doesn't support playlists yet https://github.com/goldfire/howler.js/issues/357

catalog/app/components/Preview/loaders/Audio.tsx
fiskus: Thanks
And, it's also a video format

catalog/app/components/Preview/loaders/Audio.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { PreviewData } from '../types'
+
+import * as utils from './utils'
+
+export const detect = utils.extIn([
+  '.flac',
+  '.m2t',
+  '.mp3',
+  '.mp4',
```
nl0, line 16: also, wont this conflict with video (since it's a multipurpose container)?

fiskus: Yes, thanks
`.m2t` too

PR #2549, opened 2021-11-22
Add an s3proxy lambda

dimaryaz:
(Only tested using quilt-lite, not as an actual AWS lambda)

### DIFF
--- a/lambdas/s3proxy/index.py
+++ b/lambdas/s3proxy/index.py
@@ -40,7 +40,11 @@ def lambda_handler(request):
     elif len(parts) == 3:
         s3_region, s3_bucket, s3_path = parts
     else:
-        return requests.codes.bad_request, 'Expected region/bucket/path', {'content-type': 'text/plain', **cors_headers}
+        return (
+            requests.codes.bad_request,
+            'Expected region/bucket/path',
+            {'content-type': 'text/plain', **cors_headers}
+        )
 
     if request.method == 'OPTIONS':
         return requests.codes.ok, '', cors_headers
--- a/lambdas/s3proxy/index.py
+++ b/lambdas/s3proxy/index.py
@@ -1,0 +1,77 @@
+"""
+Python implementation of the s3 proxy from quilt/s3-proxy/nginx.conf
+"""
+import urllib.parse
+
+import requests
+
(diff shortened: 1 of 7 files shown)


nl0 (commented):
this looks ok, but since it's only used in `quilt3_lite`, maybe it makes sense to implement it as an ASGI app/middleware, so it can be mounted directly, not via lambda harness (and if we need to run it in lambda, we can use smth like `magnum` for that)

lambdas/s3proxy/index.py
```diff
@@ -0,0 +1,81 @@
...
+    cors_headers = {
+        'access-control-allow-headers': request.headers.get('access-control-request-headers', ''),
+        'access-control-allow-methods': request.headers.get('access-control-request-method', ''),
+        'access-control-allow-origin': '*',
+        'access-control-max-age': '3000',
+        'access-control-expose-headers': ', '.join([
+            'Content-Length',
+            'Content-Range',
+            'ETag',
+            'x-amz-meta-helium',
```
sir-sigurd, line 27: Do we really need this one?

lambdas/s3proxy/index.py
```diff
@@ -0,0 +1,81 @@
...
+
+    request_headers = dict(request.headers)
+    request_headers.pop('host', None)  # Correct host header will come from the URL.
+    request_headers.pop('connection', None)  # Let requests handle keep-alive, etc.
+
+    response = session.request(
+        method=request.method,
+        url=url,
+        headers=request_headers,
+    )
```
sir-sigurd, line 69: This seems to not work for upload. Probably we don't need it right now, but might make sense to support it from the beginning.

sir-sigurd (commented):
I think its not useful as lambda because of lambda's restrictions on size of request/response.

lambdas/s3proxy/index.py
dimaryaz: Oh right, need the body. I didn't actually test uploads.

dimaryaz:
I'll move this into quilt-lite directly, instead of making it a lambda.

sir-sigurd:
Should we close this?

PR #2550, opened 2021-11-22
Add a lambda endpoint

dimaryaz:


quilt_lite/quilt_lite/lambdas.py
```diff
@@ -0,0 +1,57 @@
...
+        },
+        'queryStringParameters': dict(request.query_params) or None,
+        'headers': request.headers or None,  # FastAPI makes headers lower-case, just like AWS.
+        'body': b64encode(req_body),
+        'isBase64Encoded': True
+    }
+
+    module = importlib.import_module(f'{name}.index')
+
+    result = module.lambda_handler(args, None)
```
dimaryaz, line 45: I _just_ realized that FastAPI uses asyncio. So technically, this should run in a thread-pool... Don't know if that matters just yet.

akarve (approved):

nl0 (approved):
looks good, tho you need to adjust `def config` in `main.py` to pass the lambdas routes (relevant keys: "apiGatewayEndpoint", "binaryApiGatewayEndpoint", "s3Proxy").
and also it seems like we should merge s3proxy lambda PR first for this one to work

quilt_lite/quilt_lite/lambdas.py
nl0: i think the only async / non-blocking call in this function is `await request.body`, but the processing is sync / blocking (as far as i understand python async stuff -- it's quite different from node.js in some details) -- we can try using smth like [asyncio.loop.run_in_executor](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) to make lambda handler invocations truly async, but i'm not sure we'll gain anything from it

nl0:
@dimaryaz i've rebased base branch and cherry-picked your commit onto it

PR #2556, opened 2021-11-24
Add an s3proxy endpoint

dimaryaz:
Pretty much the same code as #2549, but as a FastAPI endpoint instead of a lambda.

### DIFF
--- a/quilt_lite/quilt_lite/s3proxy.py
+++ b/quilt_lite/quilt_lite/s3proxy.py
@@ -1,0 +1,74 @@
+"""
+Python implementation of the s3 proxy from quilt/s3-proxy/nginx.conf
+"""
+import urllib.parse
+
+import aiohttp
+from fastapi import FastAPI, Request, Response
+
+
+s3proxy = FastAPI()
+
+session = aiohttp.ClientSession()
+
+
+@s3proxy.api_route("/{s3_region}/{s3_bucket}", methods=['GET', 'HEAD', 'POST', 'PUT', 'OPTIONS'])
+@s3proxy.api_route("/{s3_region}/{s3_bucket}/{s3_path:path}", methods=['GET', 'HEAD', 'POST', 'PUT', 'OPTIONS'])
+async def s3proxy_request(request: Request, s3_region: str, s3_bucket:str, s3_path: str = ''):
+    """
+    Forward the request to S3.
+    """
+    cors_headers = {
+        'access-control-allow-headers': request.headers.get('access-control-request-headers', ''),
+        'access-control-allow-methods': request.headers.get('access-control-request-method', ''),
+        'access-control-allow-origin': '*',
+        'access-control-max-age': '3000',
+        'access-control-expose-headers': ', '.join([
+            'Content-Length',
+            'Content-Range',
+            'ETag',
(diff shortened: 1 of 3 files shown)


quilt_lite/quilt_lite/s3proxy.py
```diff
@@ -0,0 +1,74 @@
...
+            'x-amz-storage-class',
+        ])
+    }
+
+    req_body = await request.body()
+
+    if request.method == 'OPTIONS':
+        return Response(content='', status_code=200, headers=cors_headers)
+
+    if s3_region == '-':
```
akarve, line 43: Is `-` an AWS thing or just a convention we use?

akarve (approved):
This looks reasonable. Main question is do we want to find some way to share logic between quilt_lite and the NGINX proxy? Headers for instance. The routing logic and syntax in NGINX might be weird enough not to be shareable.

quilt_lite/quilt_lite/s3proxy.py
dimaryaz: Our convention. It was added in #1888; I forgot exactly why - it might have been only for Zip downloads - unless the catalog uses it too somewhere (@nl0 ?)

But, regardless, S3 has two URL formats, with and without a region, so it'll come up sooner or later.

nl0: i don't remember exactly, but i think we use regionless requests a lot

quilt_lite/quilt_lite/s3proxy.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+import urllib.parse
+
+import aiohttp
+from fastapi import FastAPI, Request, Response
+
+
+s3proxy = FastAPI()
+
+session = aiohttp.ClientSession()
```
nl0, line 12: don't we want a new session for every request?

nl0 (approved):
looks good

nl0:
> Main question is do we want to find some way to share logic between quilt_lite and the NGINX proxy? Headers for instance. The routing logic and syntax in NGINX might be weird enough not to be shareable.

@akarve i'd say we ditch nginx and keep just this one if it's stable and performant enough

quilt_lite/quilt_lite/s3proxy.py
dimaryaz: Possibly... Not sure. In general, re-using a session is good for performance - but it might not be correct in this case. I guess I'll change it for now, until we know what's safe.

dimaryaz:
Re nginx: I'd consider this one good enough for local mode (i.e., since user) - but definitely not something we'd want to run for e.g. opendata. I doubt a Python version can ever be as performant as Nginx.

nl0:
> Re nginx: I'd consider this one good enough for local mode (i.e., since user) - but definitely not something we'd want to run for e.g. opendata. I doubt a Python version can ever be as performant as Nginx.

well, we don't have that much requests going thru the proxy anyways (mostly zip downloads, some head requests and config fetching; requests for previews and stuff go thru lambdas; also, we plan to reduce this number even more by migrating most of the calls to GraphQL), so i'd say we don't need that much performance and python/fastapi would probably be good enough

PR #2560, opened 2021-11-24
pkgselect lambda: make directly invocable, always use execution role

nl0:
# TODO

- [ ] ~~adjust tests~~ (defer)

### DIFF
--- a/lambdas/pkgselect/index.py
+++ b/lambdas/pkgselect/index.py
@@ -2,62 +2,52 @@
 Provide a virtual-file-system view of a package's logical keys.
 """
 
+import asyncio
+import dataclasses
+import functools
 import json
-import os
+import typing as T
 
 import boto3
-import botocore
 import pandas as pd
 
-from t4_lambda_shared.decorator import api, validate
-from t4_lambda_shared.utils import (
-    get_default_origins,
-    make_json_response,
-    query_manifest_content,
-    sql_escape,
-)
+from t4_lambda_shared.utils import query_manifest_content, sql_escape
 
-SCHEMA = {
-    'type': 'object',
-    'properties': {
-        'bucket': {
-            'type': 'string'
-        },
-        'manifest': {
-            'type': 'string'
-        },
-        'access_key': {
-            'type': 'string'
-        },
-        'secret_key': {
-            'type': 'string'
-        },
-        'session_token': {
-            'type': 'string'
-        },
-        'prefix': {
-            'type': 'string'
-        },
-        'logical_key': {
-            'type': 'string'
-        },
-        'offset': {
-            'type': 'integer'
-        },
-        'limit': {
-            'type': 'integer'
-        }
-    },
-    'required': ['bucket', 'manifest'],
-    'additionalProperties': False
-}
+
+async def run_async(fn, executor=None, loop=None):
+    if loop is None:
+        loop = asyncio.get_running_loop()
+    return await loop.run_in_executor(executor, fn)
+
+
+class PkgselectException(Exception):
+    def __str__(self):
+        s = self.__class__.__name__
+        if self.args:
+            s = f"{s}: {self.args[0]}"
+        return s
+
+
+class BadInputParameters(PkgselectException):
+    pass
+
+
+class AccessDenied(PkgselectException):
+    pass
+
+
+class NotFound(PkgselectException):
+    pass
+
+
+def validate(condition: T.Any, message: str):
+    if not condition:
+        raise BadInputParameters(message)
 
 
 def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
     """
-    Post process a set of logical keys to return only the
-    top-level folder view (a special case of the s3-select
-    lambda).
+    Post process a set of logical keys to return only the top-level folder view.
     """
     if {'physical_key', 'logical_key', 'size'}.issubset(df.columns):
         groups = df.groupby(df.logical_key.str.extract('([^/]+/?).*')[0], dropna=True)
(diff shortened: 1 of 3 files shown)


lambdas/pkgselect/index.py
```diff
@@ -59,203 +36,220 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
     top-level folder view (a special case of the s3-select
     lambda).
     """
-    if {'physical_key', 'logical_key', 'size'}.issubset(df.columns):
-        groups = df.groupby(df.logical_key.str.extract('([^/]+/?).*')[0], dropna=True)
+    if {"physical_key", "logical_key", "size"}.issubset(df.columns):
```
sir-sigurd, line 39: could you reduce patch to substantial changes?

nl0: i will, but it won't make much difference

lambdas/pkgselect/index.py
```diff
@@ -90,172 +86,227 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
-    aws_access_key_id: str,
-    aws_secret_access_key: str,
-    aws_session_token: str
-):
+def object_is_publicly_readable(bucket: str, key: str):
+    try:
+        boto3.client(
+            "s3",
+            config=botocore.client.Config(signature_version=botocore.UNSIGNED),
+        ).head_object(Bucket=bucket, Key=key)
```
sir-sigurd, line 101: I'd have a separate [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache)d function for creation of client, because at least locally it's quite expensive (2 ms).

lambdas/pkgselect/index.py
```diff
@@ -90,172 +86,227 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
+
+    if (credentials is not None and not all((
+        credentials.get("aws_access_key_id"),
+        credentials.get("aws_secret_access_key"),
+        credentials.get("aws_session_token"),
+    ))):
+        raise IncompleteCredentials
+
+    if object_is_publicly_readable(bucket, key):
+        return boto3.client("s3")
```
sir-sigurd, line 130: This could be `lru_cache`d as well.

lambdas/pkgselect/index.py
```diff
@@ -90,172 +86,227 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
+
+    if credentials is None:
+        raise AccessDenied
+
+    return boto3.Session(**credentials).client("s3")
```
sir-sigurd, line 135: I think this also works:
```suggestion
    return boto3.client("s3", **credentials)
```

lambdas/pkgselect/index.py
```diff
@@ -2,55 +2,51 @@
...
+
+class PkgselectException(Exception):
+    def __str__(self):
+        s = self.__class__.__name__
+        if self.args:
+            s = f"{s}: {self.args[0]}"
+        return s
+
+
+class InvocationError(PkgselectException):
```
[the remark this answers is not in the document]
nl0: well, i think it says exactly that. any ideas on better options?

lambdas/pkgselect/index.py
```diff
@@ -2,55 +2,51 @@
...
+from t4_lambda_shared.utils import query_manifest_content, sql_escape
+
+
+async def run_async(fn, executor=None, loop=None):
+    if loop is None:
+        loop = asyncio.get_running_loop()
+    return await loop.run_in_executor(executor, fn)
+
+
+class PkgselectException(Exception):
```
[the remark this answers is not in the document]
nl0: in this case `pkgselect` is kinda one word, and so it is capitalized as one word (it's not `pkg_select`), so doing this consistently may help avoid confusion and mistypings. in general, `some words` == `SomeWords` == `some_words`

lambdas/pkgselect/index.py
nl0: it's only called once, so we won't benefit from caching it (well, maybe in localmode case only)

lambdas/pkgselect/index.py
nl0: same here, it's called only once

lambdas/pkgselect/index.py
sir-sigurd: Cache will be used when lambda's environment is reused between invocations, you can read some details at the end of this [page](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html).

lambdas/pkgselect/index.py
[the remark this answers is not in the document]
sir-sigurd: `BadInputParamsError`?

lambdas/pkgselect/index.py
```diff
@@ -82,180 +72,199 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
+async def dir_view(
+    bucket: str,
+    manifest: str,
+    path: T.Optional[str],
+    limit: T.Optional[int] = None,
+    offset: T.Optional[int] = None,
+) -> DirView:
+    validate(
+        isinstance(bucket, str) and bool(bucket),
+        "dir_view: bucket must be a non-empty string (given: {repr(bucket)})",
```
sir-sigurd, line 181:
```suggestion
        f"dir_view: bucket must be a non-empty string (given: {bucket!r})",
```

lambdas/pkgselect/index.py
```diff
@@ -82,180 +72,199 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
+
+async def dir_view(
+    bucket: str,
+    manifest: str,
+    path: T.Optional[str],
+    limit: T.Optional[int] = None,
+    offset: T.Optional[int] = None,
+) -> DirView:
+    validate(
+        isinstance(bucket, str) and bool(bucket),
```
sir-sigurd, line 180:
```suggestion
        isinstance(bucket, str) and bucket,
```

lambdas/pkgselect/index.py
```diff
@@ -82,180 +72,199 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
+        "dir_view: path must be a string if provided (given: {repr(path)})",
+    )
+    validate(
+        limit is None or isinstance(limit, int),
+        "dir_view: limit must be an int if provided (given: {repr(limit)})",
+    )
+    validate(
+        offset is None or isinstance(offset, int),
+        "dir_view: offset must be an int if provided (given: {repr(offset)})",
+    )
```
sir-sigurd, line 198: Might make sense to check that these values are positive/non-negative.

lambdas/pkgselect/index.py
```diff
@@ -82,180 +72,199 @@ def file_list_to_folder(df: pd.DataFrame, limit: int, offset: int) -> dict:
...
-                'title': 'Incomplete credentials',
-                'detail': "access_key, secret_key and session_token are required"
-            }
-        )
-    assert s3_client
+    Get details of a single file in the package.
+    """
+    validate(
+        isinstance(bucket, str) and bucket,
+        "file_view: bucket must be a non-empty string (given: {bucket!r})",
```
sir-sigurd, line 140: Here and below:
```suggestion
        f"file_view: bucket must be a non-empty string (given: {bucket!r})",
```

sir-sigurd (approved):

lambdas/pkgselect/index.py
nl0: oh good catch thx

PR #2564, opened 2021-11-29
Replace blackthorn with neumora

fiskus:


fiskus:
https://github.com/quiltdata/quilt/pull/2565

PR #2566, opened 2021-11-29
Add AWS Partner logo

fiskus:


fiskus:
We have AwsPartner logo at #2567

PR #2571, opened 2021-11-30
Bring back calendly events logs

fiskus:


catalog/app/components/TalkToUs/TalkToUs.js
```diff
@@ -36,42 +36,50 @@ function getCalendlyPromise() {
   return window[CALENDLY_PROMISE]
 }
 
-// const getCalendlyEvent = (e) =>
-//   e.data.event &&
-//   e.data.event.startsWith('calendly.') &&
-//   e.data.event.substring('calendly.'.length)
+function useSingletonListener() {
+  const ref = React.useRef(null)
+  return (eventName, callback) => {
```
nl0, line 41: i'd add smth like `if (ref.current === callback) return` to avoid unnecessary re-subscription of the same listener

catalog/app/components/TalkToUs/TalkToUs.js
```diff
@@ -36,42 +36,50 @@ function getCalendlyPromise() {
...
-        // const handleClose = () => {
-        //   window.removeEventListener('message', handleCalendlyEvent)
-        // }
-        // window.addEventListener('message', handleCalendlyEvent)
+        listen('message', handleCalendlyEvent)
         C.initPopupWidget({ url: cfg.calendlyLink })
       })
     },
-    [t, cfg.calendlyLink, calendlyP],
+    [t, cfg.calendlyLink, calendlyP, listen],
```
nl0, line 82: since you're using `listen` in the dependency list, it should have a stable identity (`useSingletonListener`'s return value should be memoized via `useCallback` or similar), otherwise memoizing `showPopup` doesn't make sense

catalog/app/components/TalkToUs/TalkToUs.js
nl0: return value should be memoized via `useCallback` or similar, so that it can be used as a dependency for other memoized functions/values

nl0 (approved):

PR #2572, opened 2021-11-30
Move close button inside calendly popup

fiskus:
It's still not good, but, in my opinion, better. If it doesn't improve UX in your opinion, I'll remove PR

Before:
1  before

After:
2  after


nl0 (commented):
well, this seems a little bit better, but quite fragile, like, what if the button gets re-rendered?
can we achieve similar effect with css?

PR #2576, opened 2021-12-06
Perspective viewer

fiskus:
* Added third argument to `Preview.load` with `contenxt` property - where file is rendered - listing or file view
* Re-use this argument for passing previously added `mode` options forcing file type (JSON, Jupyter etc.)
* Added Tabular loader and Perspective renderer
* For Tabular loader I re-used `Csv.detect`, `Excel.detect` etc., I didn't remove these loaders yet but they aren't used now

Screenshot from 2022-02-18 18-34-47


nl0:
Screenshot 2021-12-07 at 20 59 47
noice

sir-sigurd:
Might make sense to update perspective to the latest version ([1.2.0](https://github.com/finos/perspective/releases/tag/v1.2.0)).

nl0:
@fiskus could you please resolve merge conflicts here?

catalog/app/components/Preview/loaders/Tabular.tsx
```diff
@@ -0,0 +1,191 @@
+import * as R from 'ramda'
+import * as React from 'react'
+
+import { HTTPError } from 'utils/APIConnector'
+import * as AWS from 'utils/AWS'
+import * as Config from 'utils/Config'
+import * as Data from 'utils/Data'
+import { mkSearch } from 'utils/NamedRoutes'
```
nl0, line 8: this function can now be imported directly from [`utils/mkSearch`](https://github.com/quiltdata/quilt/blob/master/catalog/app/utils/mkSearch.ts)

catalog/app/components/Preview/loaders/Tabular.tsx
```diff
@@ -0,0 +1,191 @@
...
+import * as Csv from './Csv'
+import * as Excel from './Excel'
+import * as Parquet from './Parquet'
+import * as utils from './utils'
+
+const isJsonl = R.pipe(utils.stripCompression, utils.extIs('.jsonl'))
+
+export const detect = R.anyPass([Csv.detect, Excel.detect, Parquet.detect, isJsonl])
+
+// TODO: ['bed', 'fcs', 'ipynb', 'vcf']
```
nl0, line 22: why are these in TODO? they don't seem like a general-purpose tabular data

catalog/app/components/Preview/loaders/Tabular.tsx
```diff
@@ -0,0 +1,191 @@
...
+  switch (context) {
+    case CONTEXT.FILE:
+      return gated ? 'medium' : 'large'
+    case CONTEXT.LISTING:
+      return gated ? 'small' : 'large'
+    // no default
+  }
+}
+
+// function useContentLength(handle: S3HandleBase): number | null {
```
nl0, line 127: why are you keeping this?

catalog/app/components/Preview/loaders/Tabular.tsx
```diff
@@ -0,0 +1,191 @@
...
+  }
+}
+
+async function getCsvFromResponse(r: Response): Promise<ArrayBuffer | string> {
+  const isArrow = r.headers.get('content-type') === 'application/vnd.apache.arrow.file'
+  return isArrow ? r.arrayBuffer() : r.text()
+}
+
+interface LoadTabularDataArgs {
+  compression?: '.gz' | '.bz2'
```
nl0, line 62: i believe these shouldnt have leading dots

catalog/app/components/Preview/renderers/Perspective.js
```diff
@@ -0,0 +1,102 @@
+import cx from 'classnames'
```
nl0, line 1: why is this file JS and not TS?

catalog/app/components/Preview/types.js
```diff
@@ -34,11 +35,11 @@ export const PreviewData = tagged([
   'Notebook', // { preview: string, ...PreviewStatus }
   'Parquet', // { preview: string, ...ParquetMeta, ...PreviewStatus }
   'Pdf', // { handle: object, pages: number, firstPageBlob: Blob, type: 'pdf' | 'pptx' }
+  'Perspective', // { preview: string, ...PreviewStatus }
```
nl0, line 38: seems like type description is incorrect, should be smth like `{ context: ..., data: ..., handle: ..., ... }`

catalog/app/utils/perspective.js
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
```
nl0, line 1: why not use TS for all new files?

catalog/app/components/Preview/renderers/Perspective.js
```diff
@@ -0,0 +1,102 @@
...
+
+function TruncatedWarning({ className, onLoadMore }) {
+  const classes = useTruncatedWarningStyles()
+  return (
+    <div className={cx(classes.root, className)}>
+      <span className={classes.message}>
+        <M.Icon fontSize="small" color="inherit" className={classes.icon}>
+          info_outlined
+        </M.Icon>
+        Partial preview <SizeEstimation />
```
nl0, line 44: props aren't passed to `SizeEstimation`, therefore it always returns null and renderes nothing. is this intentional?

nl0 (commented):
looks good mostly, questions and comments inline

nl0:
> * Added third argument to Preview.load with contenxt property - where file is rendered - listing or file view
> * Re-use this argument for passing previously added mode options forcing file type (JSON, Jupyter etc.)

👍 

> * Added Tabular loader and Perspective renderer

👍 

> * For Tabular loader I re-used Csv.detect, Excel.detect etc., I didn't remove these loaders yet but they aren't used now

🤔 i wonder why. i believe we should remove the unused code, bc its presence and partial imports from there complicate the codebase. ofc it's ok to do it in a separate pr, but we shouldn't keep those leftovers indefinitely

fiskus:
I agree that unused code should be removed. Even if it will be used in the future, it can be restored from git (that's why I want to remove it in separate PR, because it's easier to restore). I just wasn't sure that we will not use it in the nearest future for something, and I was accidentally right: we just decided to use .vcf and .fcs with old renderers :)

So, let's remove old renderers in the very next PR after this one

nl0:
> I agree that unused code should be removed. Even if it will be used in the future, it can be restored from git (that's why I want to remove it in separate PR, because it's easier to restore).

makes sense

> I just wasn't sure that we will not use it in the nearest future for something, and I was accidentally right: we just decided to use .vcf and .fcs with old renderers :)

i'm actually a little confused bc those formats don't seem to me like general-purpose tabular data at all, but that's a topic for another discussion

> So, let's remove old renderers in the very next PR after this one

👍

catalog/app/components/Preview/loaders/Tabular.tsx
fiskus: I copied them from API stub, but you are right, I'll remove them

catalog/app/components/Preview/renderers/Perspective.js
fiskus: I started with perspective@1.0.3, and now it's 1.2.0 :)
Type definitions work now, but earlier they were buggy

catalog/internals/webpack/webpack.base.js
```diff
@@ -4,6 +4,7 @@
 
 const path = require('path')
 
+const PerspectivePlugin = require("@finos/perspective-webpack-plugin");
```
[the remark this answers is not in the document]
fiskus: Yes, and default VS Code prettier config sets double quotes

nl0 (approved):

PR #2594, opened 2021-12-27
Use WebpackDevServer

fiskus:


catalog/internals/webpack/webpack.prod.js
```diff
@@ -6,7 +6,6 @@ const TerserPlugin = require('terser-webpack-plugin')
 module.exports = require('./webpack.base')({
   mode: 'production',
 
-  // In production, we skip all hot-reloading stuff
   entry: {
```
nl0, line 9: if this is the same in both prod and dev, we can move it to the base

catalog/package.json
```diff
@@ -20,12 +20,9 @@
     "prebuild": "npm run build:clean",
     "build": "cross-env NODE_ENV=production webpack --config internals/webpack/webpack.prod.js --color --progress",
     "build:clean": "rimraf ./build",
-    "start": "cross-env NODE_ENV=development node server",
-    "start:tunnel": "cross-env NODE_ENV=development ENABLE_TUNNEL=true node server",
-    "start:production": "npm run test && npm run build && npm run start:prod",
-    "start:prod": "cross-env NODE_ENV=production node server",
```
nl0, line 26: so, we don't have a way to start the node server in production mode now? i know it's not used in production deployments, but i was using it to test the production build

catalog/package.json
```diff
@@ -20,12 +20,9 @@
...
-    "presetup": "npm i chalk shelljs",
-    "setup": "node ./internals/scripts/setup.js",
+    "build:desktop": "cross-env NODE_ENV=production webpack --config internals/webpack/webpack.desktop-prod.js --color --progress",
+    "desktop": "cross-env NODE_ENV=development webpack serve --config internals/webpack/webpack.desktop-dev.js",
```
nl0, line 24: these files (desktop-prod and desktop-dev) are missing from the PR

catalog/package.json
```diff
@@ -67,16 +64,14 @@
...
     "classnames": "^2.3.1",
-    "compression": "^1.7.4",
     "connected-react-router": "^6.9.1",
     "cross-env": "^7.0.3",
     "date-fns": "^2.25.0",
     "dedent": "^0.7.0",
     "dompurify": "^2.3.3",
     "echarts": "^5.2.2",
-    "express": "^4.17.1",
+    "electron": "^13.1.7",
```
nl0, line 74: why do we need electron in the webapp repo?

catalog/server/middlewares/frontendMiddleware.js
```diff
@@ -1,66 +0,0 @@
...
-    fs.readFile(path.join(compiler.outputPath, name), (err, file) => {
-      if (err) {
-        res.sendStatus(404)
-      } else {
-        res.send(file.toString())
-      }
-    })
-  }
-
-  app.get('/__embed-debug', sendFile('embed-debug-harness.html'))
```
nl0, line 31: does devserver support these routes somehow?

nl0 (commented):
well, the first question is why do we need this?
the second is are we dropping our custom routes, e.g. `/__embed` etc?

catalog/internals/webpack/webpack.prod.js
fiskus: It's different for desktop

fiskus:
> why do we need this?
1. It's easier to maintain
2. Live reload is fixed

fiskus:
* Made `/__embed-debug` route works using `devServer.historyApiFallback.rewrites`
* Reverted `./server` and `npm run start:prod` removal
* Removed extra dependencies, that will be introduced later but not now

nl0 (approved):
looking good now

PR #2597, opened 2021-12-30
JSON editor Ajv errors

fiskus:
* Pass errors from Ajv validation to JSON editor
* Change path to value syntax inside JSON similar to Ajv (`{"root": {"nested": {"value": null}}})` → `/root/nested/value`)
* Show errors from Ajv near JSON value and highlight row with red border
* Add ReactTools#join to join list of components using separator component


catalog/app/utils/reactTools.tsx
```diff
@@ -30,3 +30,25 @@ export const mkLazy = (
...
+ * `RT.join([<A1 />, <A2 />, <A3 />], <S />) -> <A1 /><S /><A2 /><S /><A3 />`
+ */
+export const join = (
+  list: React.ReactNode[],
+  separator: React.ReactNode,
+): React.ReactNode => {
+  if (list.length < 2) return <>{list}</>
+  return (
+    <>
+      {list.reduce((acc, item) => (
```
nl0, line 45: this is probably fine for smallish lists, but it creates a lot of fragments nested within fragments nested within fragments nested within one fragment.
the only way i see we can do it properly and not having to mess with the keys is to use `React.createElement` API (https://reactjs.org/docs/react-without-jsx.html) -- you can pass children directly to that call, e.g., first create a list of all the children with separators, and then pass this list to `createElement`, smth. like this: `React.createElement(React.Fragment, null, ...children)`

nl0: this can be backlogged ofc, since it's more like an optimization, not an actual bug

nl0 (approved):
i haven't tested this, but the code looks good

PR #2598, opened 2021-12-30
Powerpoint preview

sir-sigurd:
# Description

# TODO
- [x] move to existing preview lambda
- [x] polish Dockerfile to minimize image size
- [x] Unit tests


lambdas/thumbnail/tests/test_thumbnail.py
```diff
@@ -109,6 +107,18 @@ def test_403():
...
+            {"size": "w1024h768", "input": "pptx", "page": "1", "countPages": "true"},
+            "pptx/out-page1-1024w.jpeg", None, [1024, 1449], 2, 200,
+            marks=(pytest.mark.poppler, pytest.mark.loffice),
+        ),
+        pytest.param(
+            "pptx/in.pptx",
+            {"size": "w1024h768", "input": "pptx", "page": "2", "countPages": "true"},
+            "pptx/out-page2-1024w.jpeg", None, [1024, 1449], 2, 200,
+            marks=(pytest.mark.poppler, pytest.mark.loffice),
+        ),
```
sir-sigurd, line 121: @akarve 
I used thumbs generated on the test stack as expected thumbs, but these tests fail for me locally, probably because of fonts. `MUMmer.pdf` tests also fail for me. Not sure if I could do something with it without running tests in the real lambda's Docker image, but I think that would be out of scope of this PR.

lambdas/thumbnail/Dockerfile
```diff
@@ -1,27 +1,47 @@
...
-
-# Update pip
-RUN pip3 install --upgrade pip
-
-# Get Cython for numpy
-RUN pip3 install Cython
-
-# Install package
-RUN pip3 install -r requirements.txt
+FROM amazon/aws-lambda-python:3.8 as base
```
nl0, line 1: or maybe we should set target platform here in dockerfile? https://docs.docker.com/engine/reference/builder/#from

nl0 (approved):

lambdas/thumbnail/Dockerfile
```diff
@@ -1,27 +1,47 @@
...
-RUN ln -sf /usr/bin/pip-3.6 /usr/bin/pip3
-
-# Get requirements file
-COPY requirements.txt requirements.txt
-
-# Update pip
-RUN pip3 install --upgrade pip
-
-# Get Cython for numpy
-RUN pip3 install Cython
```
akarve, line 24: I'm nervous that this will break something. Probably we should try building/deploying all lambdas before merging this.

lambdas/thumbnail/tests/test_thumbnail.py
akarve: But Lambda itself is linux; why can't you install the fonts locally? There are also some extended unit tests hidden behind a CLI switch that we should run.

akarve (approved):

PR #2625, opened 2022-01-19
Storybook/Playground page

fiskus:
Special page dedicated to showing separate components. There is a JSON editor only now.
It's useful to develop components in isolation without. Also, they can be used as live docs and test specs.


nl0 (approved):
good idea

PR #2631, opened 2022-01-25
downolad vega data files from s3 via proxy

nl0:
# Description

Vega fails downloading data files from s3 buckets that dont have CORS properly set up. We can work around that by routing such requests via s3 proxy.

# TODO

### DIFF
--- a/catalog/app/utils/AWS/S3.js
+++ b/catalog/app/utils/AWS/S3.js
@@ -19,6 +19,7 @@ const DEFAULT_OPTS = {
 
 const PROXIED = Symbol('proxied')
 const PRESIGN = Symbol('presign')
+const FORCE_PROXY = Symbol('forceProxy')
 
 const Ctx = React.createContext()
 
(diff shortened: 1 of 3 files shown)


catalog/app/utils/AWS/S3.js
```diff
@@ -124,7 +125,15 @@ function useSmartS3() {
           params = null
         }
 
+        let forceProxy = false
+        if (params && 'forceProxy' in params) {
```
fiskus, line 129: `'forceProxy' in params` instead of `params?.forceProxy` because you need to check the prototype chain?

catalog/app/utils/AWS/S3.js
```diff
@@ -124,7 +125,15 @@ function useSmartS3() {
...
+          forceProxy = params.forceProxy
+          delete params.forceProxy
+        }
         const req = super.makeRequest(operation, params)
+        if (forceProxy) {
+          req.httpRequest[FORCE_PROXY] = true
+        }
```
fiskus, line 136: I think you can always try to `delete params.forceProxy` and replace forceProxy as boolean flag:

```ts
const isForceProxy = params && 'forceProxy' in params
delete params.forceProxy
const req = super.makeRequest(operation, params)
if (isForceProxy) {
  req.httpRequest[FORCE_PROXY] = true
}
```

This way you don't use `let` and reassign variable, that I intuitively avoid

fiskus (approved):

fiskus (approved):

PR #2637, opened 2022-01-26
Add non-interactive rows to JSON editor to fill empty space and get more Excel-like look

fiskus:
Screenshot from 2022-01-26 16-12-46


catalog/app/components/JsonEditor/EmptyRow.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react'
+import * as M from '@material-ui/core'
+
+const useStyles = M.makeStyles((t) => ({
+  inputCell: {
+    // border: `1px solid ${t.palette.divider}`,
```
nl0, line 6: forget to clean-up comments?

nl0 (approved):
looking good

catalog/app/components/JsonEditor/EmptyRow.tsx
fiskus: Yes, thanks!

PR #2647, opened 2022-01-31
quilt3-local: version bump

nl0:
update pillow

### DIFF
--- a/quilt3_local/pyproject.toml
+++ b/quilt3_local/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "quilt3_local"
-version = "1.0.0"
+version = "1.0.1"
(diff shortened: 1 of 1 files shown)


sir-sigurd (approved):
changelog?

nl0:
> changelog?

we haven't started it yet for quilt3-local (i believe it should be separate)

sir-sigurd:
> we haven't started it yet for quilt3-local (i believe it should be separate)

I think we also discussed putting quilt3-local into the separate repo...

nl0:
> > we haven't started it yet for quilt3-local (i believe it should be separate)
> 
> I think we also discussed putting quilt3-local into the separate repo...

yeah, after the initial release -- so now is the time (when i get some capacity)

PR #2656, opened 2022-02-02
Desktop flag

fiskus:
* Added config flag `{ "desktop": "ENABLED" }`
* If there is the `desktop` flag, then config is already provided and isn't downloading
* Disabled "Revise package" and "Download directory" for desktop


nl0 (approved):
looks ok, tho i think it would make more sense to set desired value of `noDownload` in your custom config, istead of adding an extra check

PR #2660, opened 2022-02-03
Desktop download package

fiskus:
* Added IPC React context for electron events. Actual electron implementation is not in this PR
* Wrapped "Download package" button in separate component switching web and electron implementation
* Added LocalFolderInput to set local directory destination and ConfirmDialog (it is also progressabar with fake progress)

Screenshot from 2022-02-03 19-03-53

Screenshot from 2022-02-03 19-04-09

Screenshot from 2022-02-03 19-04-19

### DIFF
--- a/catalog/app/utils/electron/ipc-provider.tsx
+++ b/catalog/app/utils/electron/ipc-provider.tsx
@@ -1,0 +1,66 @@
+import * as React from 'react'
+
+import * as AWS from 'utils/AWS'
+
+export * as EVENTS from './events'
+
+interface Credentials {
+  accessKeyId: string
+  secretAccessKey: string
+  sessionToken: string
+}
+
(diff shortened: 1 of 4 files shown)


catalog/app/utils/electron/ipc-provider.tsx
```diff
@@ -0,0 +1,66 @@
...
+  value: IPC
+}
+
+const serializeCredentials = (credentials: Credentials) => ({
+  accessKeyId: credentials.accessKeyId,
+  secretAccessKey: credentials.secretAccessKey,
+  sessionToken: credentials.sessionToken,
+})
+
+export const Provider = function SentryProvider({
```
nl0, line 44: Sentry?

nl0 (approved):
some typos, but otherwise seems ok

PR #2661, opened 2022-02-04
Fixed JSON editor help tooltip

fiskus:
* Show description even when no Schema type for this value
* Also, refactored component. Now I create an array of type helps separated into three groups: errors, type, description. And create tooltip help from this array
* Removed RT.join, I don't use it, and there is no reason to leave it since it has optimization issues

Screenshot from 2022-02-04 16-55-57
Screenshot from 2022-02-04 16-56-22


nl0 (approved):
looks cool

PR #2662, opened 2022-02-04
Increase maximum media duration to 30 sec, and use audio 30sec by default

fiskus:


dimaryaz:
Won't this also change video previews to 30s?

dimaryaz (approved):
Oh nevermind, it's correct.

PR #2682, opened 2022-02-15
Password strength

fiskus:
Screenshot from 2022-02-15 18-24-04

### DIFF
--- a/catalog/app/containers/Auth/SignUp.js
+++ b/catalog/app/containers/Auth/SignUp.js
@@ -15,7 +15,7 @@ import parseSearch from 'utils/parseSearch'
 import useMutex from 'utils/useMutex'
 import validate, * as validators from 'utils/validators'
 
-import PasswordStrength from './PasswordStrength'
+import * as PasswordStrength from './PasswordStrength'
 import * as Layout from './Layout'
 import SSOAzure from './SSOAzure'
 import SSOGoogle from './SSOGoogle'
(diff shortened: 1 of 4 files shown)


catalog/app/containers/Auth/SignUp.js
```diff
@@ -28,6 +29,48 @@ const Container = Layout.mkLayout('Complete sign-up')
...
+    <>
+      <Layout.Field
+        InputProps={{
+          endAdornment: isWeak && (
+            <M.InputAdornment position="end">
+              <WeakPasswordIcon />
+            </M.InputAdornment>
+          ),
+        }}
+        helperText="Hint: use combination of two-three words, mix uppercase with lowercase letters, and use special characters"
```
nl0, line 63: shouldnt we use hints from `zxcvbn`?

```
result.feedback   # verbal feedback to help choose better passwords. set when score <= 2.

  result.feedback.warning     # explains what's wrong, eg. 'this is a top-10 common password'.
                              # not always set -- sometimes an empty string

  result.feedback.suggestions # a possibly-empty list of suggestions to help choose a less
                              # guessable password. eg. 'Add another word or two'
```

nl0 (approved):
looks really nice and useful, tho i'd try showing `zxcvbn`'s feedback since it seems quite relevant and based on the actual user input (not some generic advice).
also, we should implement the same constraints on the backend probably

catalog/app/containers/Auth/SignUp.js
nl0: > I use result.feedback.suggestions with the new changeset. But warning almost always is empty for me and shows some strange warnings

ok, you can revert if you think it's not useful (i trust your judgement)

fiskus: Suggestions are useful, warning for some reason - not useful :man_shrugging:

PR #2689, opened 2022-02-18
Implement semi-atomic push

dimaryaz:
Make Package.push fail if the destination already contains a package with a top hash that's different from the current package's parent.

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -41,6 +41,7 @@ from .util import TEMPFILE_DIR_PATH as APP_DIR_TEMPFILE_DIR
 from .util import (
     PhysicalKey,
     QuiltException,
+    QuiltConflictException,
     catalog_package_url,
     extract_file_extension,
     fix_url,
@@ -1409,9 +1410,10 @@ class Package:
                 raise
 
             if self._origin is None or latest_hash != self._origin.top_hash:
-                raise QuiltException(
-                    f"Package with an unexpected hash {latest_hash} already exists at the destination; "
-                    "use force=True to overwrite"
+                raise QuiltConflictException(
+                    f"Package with hash {latest_hash} already exists at the destination; "
+                    f"expected {None if self._origin is None else self._origin.top_hash}. "
+                    "Use force=True to overwrite."
                 )
 
         # Check the top hash and fail early if it's unexpected.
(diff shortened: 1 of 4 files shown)


akarve (approved):

api/python/quilt3/packages.py
```diff
@@ -1391,12 +1396,34 @@ def dest_fn(lk, *args, **kwargs):
...
+            try:
+                latest_hash = get_bytes(registry.pointer_latest_pk(name)).decode()
+            except botocore.exceptions.ClientError as ex:
+                if ex.response['Error']['Code'] == 'NoSuchKey':
+                    # Expected
+                    return
+                raise
+
+            if self._origin is None or latest_hash != self._origin.top_hash:
+                raise QuiltException(
```
sir-sigurd, line 1412: Might make sense to add specific subclass of `QuiltException` for that.

api/python/quilt3/packages.py
```diff
@@ -1391,12 +1396,34 @@ def dest_fn(lk, *args, **kwargs):
...
+                    f"Package with an unexpected hash {latest_hash} already exists at the destination; "
```
sir-sigurd, line 1413: Might make sense to specify both expected and actual hashes.

api/python/tests/integration/test_packages.py
```diff
@@ -462,7 +474,7 @@ def add_pkg_file(pkg, lk, filename, data, *, version):
         )
         with patch('time.time', return_value=timestamp1), \
              patch('quilt3.data_transfer.MAX_CONCURRENCY', 1):
-            remote_pkg = new_pkg.push(pkg_name, registry)
+            remote_pkg = new_pkg.push(pkg_name, registry, force=True)
```
sir-sigurd, line 477: Why do we need `force=True` here?

dimaryaz: Because otherwise, it makes extra S3 calls, so I'd need to mock those. I guess `setup_s3_stubber_push_manifest` could be updated to mock them by default, but `force=True` plus a new unit test for push conflicts seemed simpler.

api/python/quilt3/packages.py
dimaryaz: Ok

api/python/tests/integration/test_packages.py
sir-sigurd: > but force=True plus a new unit test for push conflicts seemed simpler

Well, yes, it's simpler, but are we going to use `force=True` in every new test?

PR #2693, opened 2022-02-21
Avoid leaking pass reset token to trackers

nl0:
Fix for "Password reset token leaked via Referer and via Websockets - Low - Password reset token is leaked to the third parties (analytics and such)"

### DIFF
--- a/catalog/app/utils/tracking.js
+++ b/catalog/app/utils/tracking.js
@@ -1,9 +1,11 @@
 import * as R from 'ramda'
 import * as React from 'react'
 import * as redux from 'react-redux'
+import { matchPath } from 'react-router-dom'
 
 import { useExperiments } from 'components/Experiments'
 import * as Config from 'utils/Config'
+import * as NamedRoutes from 'utils/NamedRoutes'
 import usePrevious from 'utils/usePrevious'
 
 const NAV_TIMEOUT = 500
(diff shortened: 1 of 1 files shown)


catalog/app/utils/tracking.js
```diff
@@ -21,7 +23,22 @@ const consoleTracker = Promise.resolve({
...
 
-const mkLocation = (l) => `${l.pathname}${l.search}${l.hash}`
+function useMkLocation() {
+  const {
+    paths: { passChange: passChangePath },
+    urls: { passChange: passChangeUrl },
+  } = NamedRoutes.use()
+  return React.useCallback(
+    (l) => {
+      const pathname = matchPath(l.pathname, { path: passChangePath, exact: true })
```
fiskus, line 33: I'm curious why don't use `paths.passChange` and `urls.passChange`?

catalog/app/utils/tracking.js
```diff
@@ -21,7 +23,22 @@ const consoleTracker = Promise.resolve({
...
+        ? passChangeUrl('REDACTED')
+        : l.pathname
+
+      return `${pathname}${l.search}${l.hash}`
```
fiskus, line 37: I think it doesn't really matter in this place but you can create an output string with URL constructor because `urls.passChange` can occasionally contain `search`. And, it seems that search param can leak here before you reset it in PassChange (but it's from my naive perspective, I didn't dig into it)

fiskus (approved):

catalog/app/utils/tracking.js
nl0: to avoid passing whole `urls` and `paths` to memo arguments, to avoid thinking whether or not they are memoized (you can't pass just `urls.passReset`, you must pass whole `urls`, bc react eslint plugin assumes the method call uses `this`)

catalog/app/utils/tracking.js
nl0: in this specific case it's safe, bc token is passed in pathname, not as a querystring parameter

PR #2698, opened 2022-02-22
Remove Csv, Excel, Parquet loaders

fiskus:


catalog/app/components/Preview/types.js
```diff
@@ -1,28 +1,5 @@
 import tagged from 'utils/tagged'
 
-/*
-ParquetMeta: {
```
nl0, line 4: these should be preserved probably (it's not related to this PR, but anyways, it looks like a regression introduced by the previous one)

nl0 (approved):
lgtm, tho we probably should restore displaying parquet metadata

catalog/app/components/Preview/types.js
fiskus: Started to work on resolving regression

PR #2718, opened 2022-03-02
Fix pkgpush for some regions

sir-sigurd:
# Description
Fix pkgpush lambda failing because it was generating presigned URLs for s3hash lambda not considering region.

# TODO


nl0 (approved):
looks legit

PR #2722, opened 2022-03-04
Add --force command line. Use an existing local package as a parent

dimaryaz:


api/python/quilt3/main.py
```diff
@@ -445,6 +449,13 @@ def create_parser():
         default=...,
         type=lambda v: None if v == '' else v
     )
+    optional_args.add_argument(
+        "--force",
+        action="store_true",
+        help="""
+            Skip the top hash check and overwrite any existing package.
```
akarve, line 456:
```suggestion
            Skip the parent top hash check and create a new revision even if your local state is behind the remote registry.
```

api/python/tests/test_cli.py
```diff
@@ -88,12 +129,15 @@ def test_push_with_meta_data_error(
     with patch_package_class as mocked_package_class, \
          mock.patch('quilt3.main.parse_arg_json', wraps=main.parse_arg_json) as mocked_parse_json_arg:
```
akarve, line 130: Not for this PR, but there's an extra leading space here. IDK why pylint doesn't complain.

akarve: nm, i'm wrong

akarve (approved):

PR #2723, opened 2022-03-09
Lazy password strength

fiskus:
Made PasswordField standalone component and load it with `import()` and `React.lazy/React.Suspense`


nl0 (approved):
lazy-loading the whole field doesn't seem like optimal UX to me -- it delays user's ability to interact with the form. maybe it would make sense to lazy-load `zxcvbn` and make the strength-checking code async instead? just an idea, can be merged as is

fiskus:
I understand that it's better to lazy load zxcvbn, not input component itself, but it requires rewriting this implementation

PR #2724, opened 2022-03-09
Load perspective lazily

fiskus:
I checked the bucket overview page: the size of js files decreased in 650Kb

### DIFF
--- a/catalog/app/components/Preview/renderers/Perspective/types.ts
+++ b/catalog/app/components/Preview/renderers/Perspective/types.ts
@@ -1,0 +1,14 @@
+import type * as React from 'react'
+
+import type { S3HandleBase } from 'utils/s3paths'
+
+import { ParquetMetadata } from '../../loaders/Tabular'
+
+export interface PerspectiveProps extends React.HTMLAttributes<HTMLDivElement> {
+  context: 'file' | 'listing'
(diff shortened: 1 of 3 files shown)


catalog/app/components/Preview/renderers/Perspective/types.ts
```diff
@@ -0,0 +1,14 @@
+import type * as React from 'react'
+
+import type { S3HandleBase } from 'utils/s3paths'
+
+import { ParquetMetadata } from '../../loaders/Tabular'
+
+export interface PerspectiveProps extends React.HTMLAttributes<HTMLDivElement> {
```
nl0, line 7: *probably* this can be kept in `Perspective.tsx` and imported as `import type ...` in `index.ts` (type import *should not* affect code-splitting) to avoid creating an extra file and moving definition away from the code using it, but it's not a big deal

nl0 (approved):

catalog/app/components/Preview/renderers/Perspective/types.ts
fiskus: I thought about that but didn't try :)
Just a minute

fiskus: Seems like it works

PR #2730, opened 2022-03-11
File size and timestamp

fiskus:
Created FileProperties module with Container (accepting size and lastModified) and Wrapper (accepting AsyncData containing size/lastModified) components

Screenshot from 2022-03-14 13-22-31
Screenshot from 2022-03-14 13-23-02


catalog/app/containers/Bucket/FileProperties.tsx
```diff
@@ -0,0 +1,121 @@
...
+    () =>
+      lastModified
+        ? dateFns.format(
+            lastModified,
+            today.getFullYear() === lastModified.getFullYear() ? 'd MMM' : 'd MMM yyyy',
+          )
+        : null,
+    [lastModified, today],
+  )
+  const formattedSize = React.useMemo(() => (size ? readableBytes(size) : null), [size])
```
nl0, line 87: we should probably support (display) zero size as well

catalog/app/containers/Bucket/FileProperties.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <div className={cx(classes.root, className)}>
+      {data.case({
+        Ok: requests.ObjectExistence.case({
+          Exists: ({ lastModified, size }: { lastModified?: Date; size?: number }) => (
+            <FileProperties {...{ className, lastModified, size }} />
+          ),
+          _: () => <FilePropertiesSkeleton className={className} />,
+        }),
+        Err: (e: Error) => e.message,
```
nl0, line 116: this may look a little awkward in the ui depending on the message (and it also does not have any styling). maybe show an error icon instead and display error message as a tooltip?

catalog/app/containers/Bucket/FileProperties.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {formattedSize && (
+        <FileProperty className={classes.property}>{formattedSize}</FileProperty>
+      )}
+    </div>
+  )
+}
+
+export const Container = FileProperties
+
+export function Wrapper({ className, data }: FilePropertiesProps) {
```
nl0, line 105: this component is used more often and it doesn't require any extra imports when used, so we can make it a default export to save several characters

nl0 (approved):
overall looks good, tho we should probably support zero size

catalog/app/containers/Bucket/FileProperties.tsx
fiskus: Thanks, I didn't mean it, I forgot to change

PR #2732, opened 2022-03-11
Use Perspective for JSONL in .quilt

fiskus:


fiskus:
https://github.com/quiltdata/quilt/pull/2971 + https://github.com/quiltdata/quilt/pull/2977

PR #2750, opened 2022-03-21
pkgselect: flat view

nl0:
Support returning flat list of package entries (similar to reading the manifest directly, but paginated via limit/offset).
Required for upcoming gql changes.

### DIFF
--- a/lambdas/pkgselect/index.py
+++ b/lambdas/pkgselect/index.py
@@ -236,9 +236,78 @@ async def dir_view(
     )
 
 
+@dataclasses.dataclass
+class FlatView:
+    entries: T.List[dict]  # {logical_key: str, physical_key: str, size: float, hash: str, meta: dict}
+    meta: dict
+
+
+async def flat_view(
+    bucket: str,
+    manifest: str,
+    limit: T.Optional[int] = None,
+    offset: T.Optional[int] = None,
+) -> FlatView:
+    validate(
+        isinstance(bucket, str) and bucket,
+        f"flat_view: bucket must be a non-empty string (given: {bucket!r})",
+    )
+    validate(
+        isinstance(manifest, str) and manifest,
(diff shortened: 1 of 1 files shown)


lambdas/pkgselect/index.py
```diff
@@ -236,9 +236,78 @@ async def dir_view(
...
+            s.meta
+        FROM s3object s
+        LIMIT {limit}
+        OFFSET {offset + 1}
+        """
+
+    result = await select(bucket, manifest, sql_stmt)
+
+    df = (
+        pd.read_json(
```
nl0, line 292: @sir-sigurd probably pandas is not required here, but i'm not sure how to read json lines properly (also, maybe pyarrow would be faster?)

sir-sigurd: If you need to read json lines, using pandas is OK, pyarrow should be faster though it might work worse for not well-defined fields like meta.

sir-sigurd (commented):
I'm not sure if it really makes sense to put this code in lambda because it doesn't look like it offloads something from registry.

lambdas/pkgselect/index.py
```diff
@@ -236,9 +236,78 @@ async def dir_view(
...
+    )
+
+    if limit is None:
+        limit = 1000
+    if offset is None:
+        offset = 0
+
+    meta = asyncio.create_task(select_meta(bucket, manifest))
+
+    # skip first row (header) by increasing offset by 1
```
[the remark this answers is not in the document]
nl0: > Also do you really want to select entries with sub-packages meta here?

yes, i need them to carry over when revising a package

nl0:
> I'm not sure if it really makes sense to put this code in lambda because it doesn't look like it offloads something from registry.

yeah you're right. i moved it to registry

PR #2755, opened 2022-03-24
Hide parquet meta by default

fiskus:
Screenshot from 2022-03-24 17-09-48
Screenshot from 2022-03-24 17-09-56

### DIFF
--- a/catalog/app/components/Preview/renderers/Perspective/Perspective.tsx
+++ b/catalog/app/components/Preview/renderers/Perspective/Perspective.tsx
@@ -1,6 +1,7 @@
 import cx from 'classnames'
 import * as R from 'ramda'
 import * as React from 'react'
+import { TransitionGroup } from 'react-transition-group'
 import * as M from '@material-ui/core'
 
 import JsonDisplay from 'components/JsonDisplay'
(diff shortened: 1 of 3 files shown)


catalog/app/components/Preview/renderers/Perspective/Perspective.tsx
```diff
@@ -57,28 +73,42 @@ function ParquetMeta({
...
+      <M.Typography className={classes.header} onClick={toggleShow}>
+        <M.Icon
+          className={cx(classes.headerIcon, { [classes.headerIconExpanded]: show })}
+        >
+          expand_more
+        </M.Icon>
+        Parquet metadata
+      </M.Typography>
+      <TransitionGroup>
+        {show && (
```
nl0, line 86: do we actually need this check? `<M.Collapse in={false}>` will hide it contents anyways, no? and it also seems that `TransitionGroup` is not required in this case, or am i missing smth? (probably we should keep one of these -- either transition group with `show && ...` stuff or `<M.Collapse in={show}>`)

nl0: > M.Collapse reserved space for invisible table.

oh that makes sense actually

nl0 (approved):

PR #2767, opened 2022-03-30
Perspective Safari bugs

fiskus:
1. CSS variables weren't applied. Adding className besides `perspective-viewer` tag name fixed that. CSS code looks like this:
```css
perspective-viewer, .perspective-viewer-material {
  --variableA: 'one'
  --variableB: 'two'
}
```
2. `perspecttive-viewer`'s shadow HTML was covering resize handle in Safari, so I moved resize handle to outer wrapper and added small bottom padding

### DIFF
--- a/catalog/app/utils/perspective.ts
+++ b/catalog/app/utils/perspective.ts
@@ -1,3 +1,4 @@
+import cx from 'classnames'
 import * as React from 'react'
 
 import 'utils/perspective-pollution'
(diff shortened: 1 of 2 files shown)


catalog/app/utils/perspective.ts
```diff
@@ -18,7 +19,8 @@ export function renderViewer(
 ): HTMLPerspectiveViewerElement {
   const element = document.createElement('perspective-viewer')
   if (className) {
```
nl0, line 21: seems like this check is not needed anymore, bc 'perspective-viewer-material' class should be always set

nl0 (approved):

catalog/app/utils/perspective.ts
fiskus: Right, thanks

PR #2769, opened 2022-03-30
Format URLs as links

fiskus:
You can test with `/search?q=metadata%3Ahttps`

Screenshot from 2022-03-30 18-17-58


fiskus:
@nl0 did I miss something? I think it's ready to merge

nl0 (approved):

nl0:
> @nl0 did I miss something? I think it's ready to merge

yeah seems good

PR #2770, opened 2022-03-30
Wording and correct Intercom *chat* logo

fiskus:


nl0 (approved):

nl0:
btw media link layout in the footer should be adjusted for removing intercom from there

PR #2771, opened 2022-03-31
Show `quilt_summarize.json` failed state

fiskus:
Screenshot from 2022-03-31 19-14-22

### DIFF
--- a/catalog/app/components/Collaborators/Avatars.tsx
+++ b/catalog/app/components/Collaborators/Avatars.tsx
@@ -115,6 +115,7 @@ export default function Avatars({
     >
       {avatars.map(({ collaborator: { email } }, index) => (
         <Avatar
+          key={`${email}_${index}`}
           className={classes.userpic}
           avatarsLength={avatarsLength}
           email={email}
--- a/catalog/app/containers/Bucket/Summarize.tsx
+++ b/catalog/app/containers/Bucket/Summarize.tsx
@@ -3,12 +3,14 @@ import cx from 'classnames'
 import type { LocationDescriptor } from 'history'
 import * as R from 'ramda'
 import * as React from 'react'
+import * as Lab from '@material-ui/lab'
 import * as M from '@material-ui/core'
 
 import { copyWithoutSpaces } from 'components/BreadCrumbs'
 import Markdown from 'components/Markdown'
 import * as Preview from 'components/Preview'
 import Skeleton, { SkeletonProps } from 'components/Skeleton'
+import { docs } from 'constants/urls'
 import * as APIConnector from 'utils/APIConnector'
 import * as AWS from 'utils/AWS'
 import { useData } from 'utils/Data'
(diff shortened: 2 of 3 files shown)


akarve (changes requested):
Let's udpate the text a bit:
*Oops*
Check your quilt_summarize.json file for errors. See the [summarize docs]() for more.

catalog/app/components/Collaborators/Avatars.tsx
```diff
@@ -115,6 +115,7 @@ export default function Avatars({
     >
       {avatars.map(({ collaborator: { email } }, index) => (
         <Avatar
+          key={`${email}_${index}`}
```
nl0, line 118: doesn't seem like a relevant change, but whatever

catalog/app/containers/Bucket/Summarize.tsx
```diff
@@ -83,13 +84,11 @@ const FileThemeContext = React.createContext(FileThemes.Overview)
 
 const useSectionStyles = M.makeStyles((t) => ({
   root: {
+    marginTop: t.spacing(2),
```
nl0, line 87: does this look right on xs screens? bc it was specifically used on sm+ screens where we have rounded boxes and disabled on xs screens where we have straight boxes which collapse and are separated just by a thin divider line

catalog/app/containers/Bucket/requests/requestsUntyped.js
```diff
@@ -628,8 +628,6 @@ export const summarize = async ({ s3, handle: inputHandle, resolveLogicalKey })
     const manifest = JSON.parse(json)
     const configErrors = isValidManifest(manifest)
     if (configErrors.length) {
-      // eslint-disable-next-line no-console
-      console.error(configErrors[0])
```
nl0, line 632: probably it would still be helpful to see actual errors given by the validator to not guess what's wrong with your config

catalog/app/containers/Bucket/Summarize.tsx
fiskus: It doesn't look rignt on small screen, because there rounded border and no margin between them.

But intention is now clear, I'll fix this in another PR, and revert this change here

fiskus: Border radius on previous block, files l
Screenshot from 2022-04-01 10-23-32
isting

catalog/app/containers/Bucket/requests/requestsUntyped.js
fiskus: Yes, I should stop sending PR at night

nl0 (approved):

catalog/app/containers/Bucket/Summarize.tsx
fiskus: Fixed here https://github.com/quiltdata/quilt/pull/2773

PR #2775, opened 2022-04-01
Intercom launcher, fix UI bugs

fiskus:
* Replaced conditions `cfg.mode === 'PRODUCT'` with intercom.isCustom
* Removed intercom.dummy, because it's not in use anymore
* Reorganized styles a little: classNames instead of M.Box attributes, set margins outside components
* Fixed margins: in NavBar when no custom launcher and for default intercom Launcher

### DIFF
--- a/catalog/app/app.tsx
+++ b/catalog/app/app.tsx
@@ -150,7 +150,7 @@ const render = () => {
           userSelector: intercomUserSelector,
           horizontal_padding:
             // align the launcher with the right side of the container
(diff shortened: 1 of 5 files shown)


nl0 (changes requested):
dummy mode and custom launcher layout are two absolutely different concepts. dummy mode is used when no intercom credentials provided -- in this mode no launcher is shown and all the intercom calls are logged to the console

fiskus:
I think you mean that I need `if (!intercom.dummy)` check everywhere I use Intercom

nl0 (approved):

PR #2776, opened 2022-04-02
pkgpush: make directly invocable, add exceptions

nl0:
# Description

# TODO

- [x] Unit tests

### DIFF
--- a/lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
+++ b/lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
@@ -24,12 +24,9 @@ import quilt3
 from quilt3.backends import get_package_registry
 from quilt3.backends.s3 import S3PackageRegistryV1
 from quilt3.util import PhysicalKey
-from t4_lambda_shared.decorator import ELBRequest, api
 from t4_lambda_shared.utils import (
     LAMBDA_TMP_SPACE,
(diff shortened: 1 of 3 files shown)


lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -285,50 +272,176 @@ def setup_user_boto_session(session):
...
+    def __str__(self):
+        return f"{self.__class__.__name__}: {self.message}"
+
+
+class S3HashLambdaUnhandledError(PkgpushException):
+    def __init__(self):
+        super().__init__("An error occured while calling S3 Hash Lambda.")
+
+
+class InvalidCredentials(PkgpushException):
```
nl0, line 303: @sir-sigurd wdyt about having distinct exception classes for everything (vs having one class with assignable attributes e.g. name/code, message, data)?

sir-sigurd: Having distinct exception classes seem to make more sense because it allows to catch the exact exception reason without further analysis of exception.

sir-sigurd (commented):
Looks OK.

sir-sigurd (commented):
Looks OK.

sir-sigurd (approved):

PR #2778, opened 2022-04-03
Button for toggling config

fiskus:
Screenshot from 2022-04-04 14-55-33
Screenshot from 2022-04-04 14-55-41

### DIFF
--- a/catalog/app/components/Preview/renderers/Perspective/Perspective.tsx
+++ b/catalog/app/components/Preview/renderers/Perspective/Perspective.tsx
@@ -131,7 +131,7 @@ const useTruncatedWarningStyles = M.makeStyles((t) => ({
 interface ToolbarProps {
   className: string
   onLoadMore: () => void
-  state: perspective.State
+  state: perspective.State | null
   truncated: boolean
 }
 
@@ -144,7 +144,7 @@ function Toolbar({ className, onLoadMore, state, truncated }: ToolbarProps) {
           <M.Icon fontSize="small" color="inherit" className={classes.icon}>
             info_outlined
           </M.Icon>
-          {state.size ? `Showing only ${state.size} rows` : `Partial preview`}
+          {state?.size ? `Showing only ${state?.size} rows` : `Partial preview`}
         </span>
       )}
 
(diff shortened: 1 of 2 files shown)


catalog/app/utils/perspective.ts
```diff
@@ -39,40 +41,43 @@ function usePerspective(
...
   settings?: boolean,
 ) {
-  const [tableData, setTableData] = React.useState<TableData | null>(null)
+  const table = React.useRef<Table | null>(null)
+  const viewer = React.useRef<HTMLPerspectiveViewerElement | null>(null)
 
-  React.useEffect(() => {
-    let table: Table, viewer: HTMLPerspectiveViewerElement
+  const [state, setState] = React.useState<State>({
+    toggleConfig: () => viewer.current?.toggleConfig(),
```
nl0, line 48: i don't see how this is a part of the state. why not just make it a memoized function of `viewer`?

catalog/app/utils/perspective.ts
```diff
@@ -39,40 +41,43 @@ function usePerspective(
   attrs: React.HTMLAttributes<HTMLDivElement>,
   settings?: boolean,
 ) {
-  const [tableData, setTableData] = React.useState<TableData | null>(null)
+  const table = React.useRef<Table | null>(null)
+  const viewer = React.useRef<HTMLPerspectiveViewerElement | null>(null)
```
nl0, line 45: all these shared mutable references are very likely going to cause some concurrency-related issues that may be quite hard to debug. we should try to avoid them as much as possible.
before this pr you were mostly using local variables, and it was safer. why can't we keep using them and expose the `toggleConfig` fn via state, same as `size`?

fiskus: Reverted back to local variables

catalog/app/utils/perspective.ts
fiskus: I didn't understand this
You don't like `state` name? Or, do you mean I can pass the `viewer`, not just function?

catalog/app/components/Preview/renderers/Perspective/Perspective.tsx
```diff
@@ -124,28 +128,47 @@ const useTruncatedWarningStyles = M.makeStyles((t) => ({
...
           Load more
         </M.Button>
       )}
+
+      {state?.toggleConfig && (
+        <M.Button
+          className={classes.item}
+          startIcon={<M.Icon>tune</M.Icon>}
+          size="small"
+          onClick={() => state?.toggleConfig()}
```
nl0, line 167: why not just `onClick={state.toggleConfig}`?

catalog/app/utils/perspective.ts
nl0: since you had `viewer` in scope (in the ref-based implementation), you could've made `toggleConfig` a memoized function of viewer, e.g. `const toggleConfig = React.useCallback(() => viewer.current?.toggleConfig(), [viewer])`, but doesn't seem relevant anymore

PR #2782, opened 2022-04-06
Use chunked SHA256 hashes calculated server-side by S3

dimaryaz:
- Use S3's `ChecksumAlgorithm='SHA256'` whenever possible, instead of calculating hashes manually
- Breaking change: switch from normal SHA256 to chunked SHA256: hash file chunks, then hash the hashes


dimaryaz:
@sir-sigurd: I removed a lot of the existing hashing code and made everything much simpler, for two reasons:
- Hashing parts instead of whole files is better for parallelism, so makes everything easier (and I'm sure that's why AWS does it that way)
- Since most of the time, hashing will be done by AWS, our own code's performance is less critical.

But, I didn't fully understand the existing code, so it's possible I missed something.

api/python/quilt3/data_transfer.py
```diff
@@ -269,14 +266,17 @@ def _upload_file(ctx, size, src_path, dest_bucket, dest_key):
                 Body=fd,
                 Bucket=dest_bucket,
                 Key=dest_key,
+                ChecksumAlgorithm='SHA256',
             )
 
         version_id = resp.get('VersionId')  # Absent in unversioned buckets.
-        ctx.done(PhysicalKey(dest_bucket, dest_key, version_id))
+        sha256 = resp['ChecksumSHA256']
+        ctx.done(PhysicalKey(dest_bucket, dest_key, version_id, sha256))
```
akarve, line 274: Does the now additional passing of `sha256` guarantee that this serializes with the new hash type?

api/python/quilt3/data_transfer.py
```diff
@@ -298,10 +298,15 @@ def upload_part(i, start, end):
...
-                    PartNumber=part_id
+                    PartNumber=part_id,
+                    ChecksumAlgorithm='SHA256',
                 )
             with lock:
-                parts[i] = {"PartNumber": part_id, "ETag": part["ETag"]}
+                parts[i] = dict(
+                    PartNumber=part_id,
+                    ETag=part["ETag"],
+                    ChecksumSHA256=part['ChecksumSHA256'],
```
akarve, line 308: Why single quotes here and double quotes on the line above?

akarve: Same comment for a few other places.

api/python/quilt3/packages.py
```diff
@@ -66,6 +66,9 @@
     MANIFEST_MAX_RECORD_SIZE = DEFAULT_MANIFEST_MAX_RECORD_SIZE
 
 
+HASH_NAME = 'AWS-SHA256'
```
akarve, line 69:
```suggestion
HASH_NAME = 'AWSChecksumSHA256'
```

api/python/quilt3/data_transfer.py
```diff
@@ -483,16 +502,17 @@ def _upload_or_copy_file(ctx, size, src_path, dest_bucket, dest_path):
...
-            dest_size = resp['ContentLength']
+            dest_size = resp['ObjectSize']
             dest_etag = resp['ETag']
             dest_version_id = resp.get('VersionId')
             if size == dest_size:
                 src_etag = _calculate_etag(src_path)
                 if src_etag == dest_etag:
                     # Nothing more to do. We should not attempt to copy the object because
                     # that would cause the "copy object to itself" error.
+                    sha256 = resp['Checksum'].get('ChecksumSHA256')
```
akarve, line 513: It would be easier to read if we didn't call this `sha256` but `checksum`. Would also help to have a comment below for each case:
1) object in S3 and has checksum
2) object in S3 and does not have checksum
3) object local and we will request checksum
Or similar.

api/python/quilt3/data_transfer.py
```diff
@@ -298,10 +298,15 @@ def upload_part(i, start, end):
                     Bucket=dest_bucket,
                     Key=dest_key,
                     UploadId=upload_id,
-                    PartNumber=part_id
+                    PartNumber=part_id,
+                    ChecksumAlgorithm='SHA256',
                 )
             with lock:
-                parts[i] = {"PartNumber": part_id, "ETag": part["ETag"]}
+                parts[i] = dict(
```
akarve, line 305: I was expecting at least one case in which we need to ListParts or know the part size from S3? E.g. what about `quilt3 verify` or any case where we only have the checksum and the file and need to verify it.

akarve: We _could_ drop support for quilt3 verify if this is a pain, but it's nice. And actually checksums should make it *much* faster.

akarve (commented):
Looks reasonable. A few suggestions. I'm going to come back and reread the

api/python/quilt3/data_transfer.py
dimaryaz: Oh, didn't notice. Don't think there's an official style, but I like to follow [this](https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python):

> double quotes around strings that are used for interpolation or that are natural language messages, and single quotes for small symbol-like strings

api/python/quilt3/data_transfer.py
dimaryaz: `ListParts` requires an upload ID - and I'm not sure it'll even work for completed uploads (though I didn't check).

In any case, I got `verify` working - but, it relies on the assumption that the files were uploaded using the default values for `multipart_threshold` and `multipart_chunksize` (both 8MB). We're always using the defaults - but, if somehow the user uploads the files using other tools _and_ changes the defaults _and_ enables the sha256 checksum, then they won't match. But that seems like a pretty extreme edge case.

api/python/quilt3/data_transfer.py
akarve: When this physical key serializes in a manifest it will have a different type, right (not `"type": "sha256"`)? I'm just wondering how the manifest serializer knows that.

dimaryaz: Oh, serialization only looks at `PackageEntry`, not at `PhysicalKey`. The type gets set here: https://github.com/quiltdata/quilt/pull/2782/files#diff-652f7ab61c8420f5d72f57c54f12337d64a53343502a81d912026e0831bd53e0R1464

sir-sigurd (commented):
I left some comments, but I'll need to have another look.

api/python/quilt3/data_transfer.py
sir-sigurd: A list of parts along with their checksums cant be obtained like this:
```
aws s3api get-object-attributes --bucket quilt-sergey-test-promote --key 32MB-file --object-attributes ObjectParts,Checksum
```
```
{
    "LastModified": "Thu, 07 Apr 2022 14:50:33 GMT",
    "VersionId": "fR2lZdKdWx5YF4C6hbpN6Hx7MzB6svcv",
    "Checksum": {
        "ChecksumSHA256": "fMJJ08ZgZU9Z8aMET6oXADfcmluSqcXXtR07Xy3DCoQ="
    },
    "ObjectParts": {
        "TotalPartsCount": 2,
        "PartNumberMarker": 0,
        "NextPartNumberMarker": 2,
        "MaxParts": 1000,
        "IsTruncated": false,
        "Parts": [
            {
                "PartNumber": 1,
                "Size": 17179870,
                "ChecksumSHA256": "DNRfQslATbmfAaQNE15vP9kXaATe/NoGJ5Di2pkDIOc="
            },
            {
                "PartNumber": 2,
                "Size": 16374562,
                "ChecksumSHA256": "ZekGMfeF9+gphL8RrAL5ppwFWfRdNLO73mx6i1qph2M="
            }
        ]
    }
}
```

api/python/quilt3/util.py
```diff
@@ -142,6 +142,7 @@ def __init__(self, bucket, path, version_id):
         self.bucket = bucket
         self.path = path
         self.version_id = version_id
+        self.checksum = checksum
```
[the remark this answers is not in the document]
akarve: > IMO this breaks `PhysicalKey` abstraction.

Why? it already has version_id.

api/python/quilt3/packages.py
```diff
@@ -1440,7 +1443,7 @@ def check_latest_hash():
             # Copy the datafiles in the package.
             physical_key = entry.physical_key
 
-            new_physical_key = dest_fn(logical_key, entry, top_hash)
+            new_physical_key = dest_fn(logical_key, entry, None)
```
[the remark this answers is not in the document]
akarve: > This is breaking.

What do you mean, not backward compatible?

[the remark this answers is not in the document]
sir-sigurd: This breaks using top_hash in `dest` callable ([docs](https://docs.quiltdata.com/api-reference/package#package.push)), doesn't it?

api/python/quilt3/util.py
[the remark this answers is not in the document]
sir-sigurd: version_id is required to specify exact object, it's not some additional object property such as size, otherwise why we don't also put size here?

api/python/quilt3/packages.py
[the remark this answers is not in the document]
dimaryaz: Yes; there's no way around it. I guess I should update the docs, and just remove the last parameter completely.

api/python/setup.py
```diff
@@ -58,7 +58,7 @@ def run(self):
     install_requires=[
         'appdirs>=1.4.0',
         'aws-requests-auth>=0.4.2',
-        'boto3>=1.10.0',
+        'boto3>=1.21.10',
```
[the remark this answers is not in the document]
dimaryaz: Weird... 1.21.10 seems to support the checksums - but I can make it 1.27.7 if that's where they were "officially" supported.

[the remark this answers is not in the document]
sir-sigurd: Oops, I meant, 1.21.7 is enough.

sir-sigurd:
1. it looks like we need to modify policies in registry since `s3:GetObjectVersion` and `s3:GetObjectVersionAttributes` are required
2. should we also modify catalog to enable checksums for uploads?
3. does this work for public buckets?

api/python/quilt3/data_transfer.py
```diff
@@ -473,7 +469,10 @@ def _upload_or_copy_file(ctx, size, src_path, dest_bucket, dest_path):
         try:
             params = dict(Bucket=dest_bucket, Key=dest_path)
             s3_client = ctx.s3_client_provider.find_correct_client(S3Api.HEAD_OBJECT, dest_bucket, params)
-            resp = s3_client.head_object(**params)
+            resp = s3_client.get_object_attributes(
```
sir-sigurd, line 472: I think you can use `head_object(**params, ChecksumMode='ENABLED')` instead. That way we don't need new permissions. Also `get_object_attributes()` returns checksum without a number of parts at the end and as I understand we want it, BTW do we really want it?

api/python/quilt3/data_transfer.py
```diff
@@ -991,81 +918,114 @@ def _calculate_sha256_internal(src_list, sizes, results):
...
-                future, idx = future_to_idx.popitem()
-                future.cancel()
+            for _, future_list in futures:
+                for future in future_list:
+                    future.cancel()
 
     return results
 
 
+def calculate_sha256_bytes(data: bytes):
```
sir-sigurd, line 1010: It's weird we have `calculate_sha256()` and `calculate_sha256_bytes()` and they calculate something different.

sir-sigurd (commented):
For something like `quilt3.Package().set_dir('.', 's3://allencell/test').build('test/test', registry='s3://quilt-dev-null')` will it still use the old algorithm?

## TODO
- [ ]  add changelog entry
- [ ] fix doc tests

api/python/quilt3/data_transfer.py
dimaryaz: Will try `head_bucket`; didn't know about that parameter.

`get_object_attributes` returns parts if the object was uploaded using a multipart upload.

sir-sigurd: > `get_object_attributes` returns parts if the object was uploaded using a multipart upload.

```
In [82]: resp = s3.get_object_attributes(
    ...:     Bucket='quilt-sergey-test-promote',
    ...:     Key='32MB-file',
    ...:     ObjectAttributes=['ObjectParts', 'Checksum'],
    ...: )
    ...: print(resp['Checksum']['ChecksumSHA256'])
    ...: print(resp['ObjectParts']['TotalPartsCount'])
fMJJ08ZgZU9Z8aMET6oXADfcmluSqcXXtR07Xy3DCoQ=
2
```

dimaryaz: Oh wow, you're right. I somehow missed it, and it looks like a bug in S3 - it's literally the same checksum, just without the number of parts.

But `head_object(..., ChecksumMode='ENABLED')` works correctly - so I'll do that.

dimaryaz:
@akarve: not sure I understand what you mean here:
 
>     2. Object in S3 does not have checksum, will not be copied - use old SHA256 hash (requires streaming object)

In cases where you're e.g. building a package out of S3 URLs, it'll need to stream the object and calculate the hash - but it'll use the new algorithm, not the old one.

dimaryaz:
Any idea what's going on with docs? I ran `build.py`, but it didn't do anything. Also, it's just blank lines.

sir-sigurd:
@dimaryaz 
does this PR supports `quilt3 verify` for packages with old hashes?
https://github.com/quiltdata/quilt/blob/d88df8c826685f04d04aef0df4191f794ef2e824/api/python/quilt3/packages.py#L1625-L1660

dimaryaz:
> @dimaryaz does this PR supports `quilt3 verify` for packages with old hashes?

It does not. It's doable - we'd need to keep the old hashing code for that.

akarve:
> @dimaryaz does this PR supports `quilt3 verify` for packages with old hashes?
> It does not. It's doable - we'd need to keep the old hashing code for that.

I am OK dropping support for `quilt3 verify` but in that case we should delete the code so users don't call it.

dimaryaz:
> I am OK dropping support for `quilt3 verify` but in that case we should delete the code so users don't call it.

You mean, we don't need it for new packages, either?

akarve:
> In cases where you're e.g. building a package out of S3 URLs, it'll need to stream the object and calculate the hash - but it'll use the new algorithm, not the old one.

That can work. If the object is ever copied (from S3 to S3) in the future, it may or may not have the same new hash, though, or do we also control part size in that case? Even if the hash breaks this is OK, I'm just trying to understand all the cases.

akarve:
> > I am OK dropping support for `quilt3 verify` but in that case we should delete the code so users don't call it.
> You mean, we don't need it for new packages, either?

Sorry, I didn't realize that it was an option. As long as we have an error message for the "old package" case (we can use manifest version) then the new client can fail `quilt3 verify` on old manifests and should still work on new manifests.

That reminds me, do we need to bump the manifest version?

akarve:
> @akarve: not sure I understand what you mean here:
> 
> > ```
> > 2. Object in S3 does not have checksum, will not be copied - use old SHA256 hash (requires streaming object)
> > ```
> 
> In cases where you're e.g. building a package out of S3 URLs, it'll need to stream the object and calculate the hash - but it'll use the new algorithm, not the old one.

I didn't realize that we could use the new algorithm on an old object that isn't being copied. That sounds good as long as if the client copies the object in the future we enforce part-size and get the same checksum.

dimaryaz:
> As long as we have an error message for the "old package" case (we can use manifest version) then the new client can fail `quilt3 verify` on old manifests and should still work on new manifests.

The `verify` API just returns `True` or `False` - so unless we throw an exception instead, there isn't really a way to report an "old package". Also... right now, it doesn't appear to check the hash type at all - should definitely fail if it's unexpected.

> That reminds me, do we need to bump the manifest version?

Not really. We're just changing the hash type - and the hash type is already part of the manifest format.

akarve:
> > As long as we have an error message for the "old package" case (we can use manifest version) then the new client can fail `quilt3 verify` on old manifests and should still work on new manifests.
> 
> The `verify` API just returns `True` or `False` - so unless we throw an exception instead, there isn't really a way to report an "old package". Also... right now, it doesn't appear to check the hash type at all - should definitely fail if it's unexpected.
> 
> > That reminds me, do we need to bump the manifest version?
> 
> Not really. We're just changing the hash type - and the hash type is already part of the manifest format.

OK do you agree that old clients (which from the code don't even seem to check `hash.type` (right?) that verify and install both will just fail with a new manifest whenever the physical key is multi-part. That's fine.

I can live with this as a minor update or a major one, but it shouldn't be a patch imo. Lmk what you think.

akarve (dismissed):

dimaryaz:
> OK do you agree that old clients (which from the code don't even seem to check `hash.type` (right?)) that verify and install both will just fail with a new manifest whenever the physical key is multi-part? That's fine, btw, just confirming.

Yes - though they'll fail even for non-multipart objects cause the new checksum is base64 encoded instead of base16.

sir-sigurd (changes requested):
* [CHANGELOG](https://github.com/quiltdata/quilt/blob/master/docs/CHANGELOG.md) entry is missing
* as I understand to calculate these new hashes you need to use a certain part sizing policy that is stable (i.e. works the same way for every client and every version), if so do we really need to include number of parts in hash?
* if we need stable part sizing policy, do you think it's safe enough to rely on defaults of `TransferConfig` and implementation of `ChunksizeAdjuster`?
* I think we shouldn't release this until we make sure that catalog properly works with the new hash type, though it might be OK to merge this PR before

api/python/quilt3/packages.py
```diff
@@ -66,6 +67,9 @@
     MANIFEST_MAX_RECORD_SIZE = DEFAULT_MANIFEST_MAX_RECORD_SIZE
 
 
+HASH_NAME = 'AWSChecksumSHA256'
```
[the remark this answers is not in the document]
dimaryaz: What would be a good name for it? (Though if we rely on `TransferConfig`'s chunk sizes, then it's still somewhat AWS-specific)

dimaryaz:
> as I understand to calculate these new hashes you need to use a certain part sizing policy that is stable (i.e. works the same way for every client and every version), if so do we really need to include number of parts in hash?

I would still include it:
- Just to be consistent with the way AWS does it
- To catch common mismatches like multi-part upload vs normal upload (e.g., in the browser, S3 console does not seem to do multipart uploads at all)

> if we need stable part sizing policy, do you think it's safe enough to rely on defaults of TransferConfig and implementation of ChunksizeAdjuster?

Yeah, I think that's the best we can do.

sir-sigurd:
> Just to be consistent with the way AWS does it

Well, `get_object_attributes()` returns checksum without number of parts, so either option is <del>eventually</del> somewhat consistent with AWS :wink: 

> To catch common mismatches like multi-part upload vs normal upload (e.g., in the browser, S3 console does not seem to do multipart uploads at all)

S3 console does multipart uploads (for [example](https://github.com/quiltdata/quilt/pull/2782#discussion_r845265459) this file was uploaded via S3 console), anyway I doesn't understand how it's helpful, could you explain further or maybe provide a full example?

> Yeah, I think that's the best we can do.

Can't we have our own implementation?

I think that we use s3transfer could be a problem by itself, from https://github.com/boto/s3transfer/blob/develop/README.rst:
> This project is not currently GA. If you are planning to use this code in production, make sure to lock to a minor version as interfaces may break from minor version to minor version. For a basic, stable interface of s3transfer, try the interfaces exposed in [boto3](https://boto3.readthedocs.io/en/latest/guide/s3.html#using-the-transfer-manager)

api/python/quilt3/packages.py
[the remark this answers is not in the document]
akarve: I think we have the right name and prefix in "AWS" because they either calculate it for us or we try to do it their way and that's what matters. If they changed their algorithm we would too. I suppose you could call it "hash of hashes" but they doesn't help future developers who will want to Google "AWS checksum" so I say we leave it as is.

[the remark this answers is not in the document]
sir-sigurd: > I think we have the right name and prefix in "AWS" because they either calculate it for us or we try to do it their way 

I'd rather say that we define a hash function in a way that it can be efficiently calculated using S3 features in most of our use cases, but in cases when it can't (existing S3 object without calculated `ChecksumSHA256` or with part sizes that are not compliant with our part sizing policy or in-memory data) we calculate hash ourselves.

> If they changed their algorithm we would too.

What this means? That we would produce a different hash for the same data and hash type?

dimaryaz:
> S3 console does multipart uploads (for [example](https://github.com/quiltdata/quilt/pull/2782#discussion_r845265459) this file was uploaded via S3 console), anyway I doesn't understand how it's helpful, could you explain further or maybe provide a full example?

Ok, maybe it has different config or something. I uploaded a 15MB file - but it didn't use a multi-part upload, even though `TransferConfig`'s maximum is 8MB:

```
$ aws s3api get-object-attributes --bucket quilt-dima --key 'Blaise Car.mp4' --object-attributes ObjectSize,ObjectParts,Checksum
{
    "LastModified": "Thu, 21 Apr 2022 18:58:47 GMT",
    "VersionId": "id.G2YemnvIXL4H758csbGUEH86JY.2l",
    "Checksum": {
        "ChecksumSHA256": "84bMq5fjSOF5rBQUSK75wKD2vgM7wGgjj9S9fvD/Pu8="
    },
    "ObjectSize": 15814262
}
```

> Can't we have our own implementation?

Yes - what I meant is, we should _try_ to make our own implementation compatible with the defaults used by S3, but we can't always succeed. Anyways, we should really decide what our goals are here. I don't disagree with anything - just not sure what else I can do.

sir-sigurd:
> Ok, maybe it has different config or something. I uploaded a 15MB file - but it didn't use a multi-part upload

I think it has 16 MB as multipart threshold:
image

So in my [example](https://github.com/quiltdata/quilt/pull/2782#discussion_r845265459) I uploaded 32 MiB file and its first part is 16.384 MiB and the second one is 15.616 MiB :confused: 

> what I meant is, we should try to make our own implementation compatible with the defaults used by S3

That makes sense, but I don't think such defaults exist: see my comment about multipart upload in S3 console, [JS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3/ManagedUpload.html) has `partSize (Number) — default: 5mb`, in boto3 default part size is 8 MiB.

So I think goals are:

1. Make our sizing policy stable (e.g. we don't want it to change because some defaults of some dependency changed with version update), because hash calculation relies on that
2. This policy should be well-defined so it can be consistently implemented in all clients (e.g. we can't use s3transfer's `ChunksizeAdjuster` from catalog)

dimaryaz:
Ping?

api/python/quilt3/data_transfer.py
```diff
@@ -249,6 +228,34 @@ def read_file_chunks(file, chunksize=s3_transfer_config.io_chunksize):
...
+# 8MB - same as TransferConfig().multipart_threshold - but hard-coded to guarantee it won't change.
+CHECKSUM_MULTIPART_THRESHOLD = 8 * 1024 * 1024
+
+# Maximum number of parts supported by S3
+CHECKSUM_MAX_PARTS = 10_000
+
+
+def adjust_checksum_chunksize(file_size):
+    """
+    Calculate the chunk size to be used for the checksum. It is normally 8MB,
```
sir-sigurd, line 240:
```suggestion
    Calculate the chunk size to be used for the checksum. It is normally 8 MiB,
```

api/python/quilt3/data_transfer.py
```diff
@@ -249,6 +228,34 @@ def read_file_chunks(file, chunksize=s3_transfer_config.io_chunksize):
...
+    but gets doubled as long as the number of parts exceeds the maximum of 10,000.
+
+    It is the same as
+    `ChunksizeAdjuster().adjust_chunksize(s3_transfer_config.multipart_chunksize, file_size)`,
+    but hard-coded to guarantee it won't change and make the current behavior a part of the API.
+    """
+    assert file_size >= CHECKSUM_MULTIPART_THRESHOLD, "file_size too small for multi-part"
+
+    chunksize = 8 * 1024 * 1024
+    num_parts = int(math.ceil(file_size / float(chunksize)))
```
sir-sigurd, line 250: I think it can be simplified:
```suggestion
    num_parts = math.ceil(file_size / chunksize)
```

api/python/quilt3/data_transfer.py
```diff
@@ -249,6 +228,34 @@ def read_file_chunks(file, chunksize=s3_transfer_config.io_chunksize):
 UPLOAD_ETAG_OPTIMIZATION_THRESHOLD = 1024
 
 
+# 8MB - same as TransferConfig().multipart_threshold - but hard-coded to guarantee it won't change.
```
sir-sigurd, line 231:
```suggestion
# 8 MiB - same as TransferConfig().multipart_threshold - but hard-coded to guarantee it won't change.
```

api/python/quilt3/data_transfer.py
```diff
@@ -249,6 +228,34 @@ def read_file_chunks(file, chunksize=s3_transfer_config.io_chunksize):
...
+CHECKSUM_MULTIPART_THRESHOLD = 8 * 1024 * 1024
+
+# Maximum number of parts supported by S3
+CHECKSUM_MAX_PARTS = 10_000
+
+
+def adjust_checksum_chunksize(file_size):
```
sir-sigurd, line 238:
```suggestion
def adjust_checksum_chunksize(file_size: int) -> int:
```

sir-sigurd, line 238: This might be a better name:
```suggestion
def get_checksum_chunksize(file_size):
```

docs/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 ## Catalog, Lambdas
 !-->
+
+## Python API
+* [Changed] Switch from a regular SHA256 checksum to a hash list to match S3's built-in checksums ([#2782](https://github.com/quiltdata/quilt/pull/2782))
+* [Changed] Delay object hashing until package push to take advantage of S3's hashing ([2782](https://github.com/quiltdata/quilt/pull/2782))
```
sir-sigurd, line 12: I think breaking changes should be listed more explicitly.

api/python/quilt3/data_transfer.py
```diff
@@ -30,12 +27,7 @@
     HTTPClientError,
     ReadTimeoutError,
 )
-from s3transfer.utils import (
-    ChunksizeAdjuster,
-    OSUtils,
-    signal_not_transferring,
-    signal_transferring,
-)
+from s3transfer.utils import ChunksizeAdjuster, ReadFileChunk
```
sir-sigurd, line 30: It's not clear to me why we still use `ChunksizeAdjuster`.

sir-sigurd (changes requested):
I don't quite understand why we break `quilt3 verify` for old packages, but if we decide to do that it at least should be explicitly list in changelog.

docs/CHANGELOG.md
dimaryaz: @akarve: what should we say here?

api/python/quilt3/data_transfer.py
```diff
@@ -249,6 +232,34 @@ def read_file_chunks(file, chunksize=s3_transfer_config.io_chunksize):
...
+    but hard-coded to guarantee it won't change and make the current behavior a part of the API.
+    """
+    assert file_size >= CHECKSUM_MULTIPART_THRESHOLD, "file_size too small for multi-part"
+
+    chunksize = 8 * 1024 * 1024
+    num_parts = math.ceil(file_size / chunksize)
+
+    while num_parts > CHECKSUM_MAX_PARTS:
+        chunksize *= 2
+        num_parts = int(math.ceil(file_size / float(chunksize)))
```
sir-sigurd, line 258:
```suggestion
        num_parts = math.ceil(file_size / chunksize)
```

dimaryaz: Oh, sorry, missed this one

docs/CHANGELOG.md
sir-sigurd: Well, I think that `top_hash` is not passed to `dest` function anymore.

sir-sigurd (dismissed):

PelleUllberg:
Hi, we really need this feature. What's the current status? Seems to have been stuck for a while. Is anybody actively working on it?

api/python/quilt3/data_transfer.py
```diff
@@ -897,7 +925,124 @@ def calculate_sha256(src_list: List[PhysicalKey], sizes: List[int]):
...
+                    Key=src.path,
+                    Range=f'bytes={offset}-{end}',
+                )
+                if src.version_id is not None:
+                    params.update(VersionId=src.version_id)
+
+                s3_client = find_correct_client(S3Api.GET_OBJECT, src.bucket, params)
+
+                try:
+                    body = s3_client.get_object(**params)['Body']
```
nl0, line 985: @dimaryaz we can use MPU hack here to compute checksums in the cloud w/o streaming the bytes, tho we need some scratch location to host that MPU

robnewman:
> Hi, we really need this feature. What's the current status? Seems to have been stuck for a while. Is anybody actively working on it?

@PelleUllberg Thanks for your message - this is actively being worked on outside of this PR.

dimaryaz:
Just rebased everything on top of master.

We now have a bit of a problem with the "dedupe" parameter: it is supposed to skip the package push if the remote top hash already matches the local top hash - but we might not know the top hash until we actually push.

dimaryaz:
The code now matches the spec as far as I can tell - however, I'm pretty sure we cannot support `--dedupe` anymore because we might not know the hash until we're done pushing. @akarve - what do you think? Do we drop `--dedupe`?

akarve:
> Hi, we really need this feature. What's the current status? Seems to have been stuck for a while. Is anybody actively working on it?

This is back towards the top of our queue. The issue is mostly on the catalog side where more work was needed.

api/python/quilt3/data_transfer.py
sir-sigurd: we could configure service bucket during `quilt3 login`
but in that case it might be a better idea to just use backend
I wonder if we even move (at least part of) pkgpush logic to quilt3 🤔🤔🤔

api/python/quilt3/data_transfer.py
dimaryaz: (Just going through existing comments - want to either fix it or punt it.)

`get-object-attributes` does appear to work - but only for objects in S3, obviously. I think the normal use-case for `verify` is to check if a local directory has correct checksums - but, once files are downloaded locally, we lose all the object parts info. (We _could_ special case it for a remote package, with S3 physical keys, and use those to get object parts, but then use local files to calculate the hash... but this seems a bit crazy.)

sir-sigurd: > In any case, I got verify working - but, it relies on the assumption that the files were uploaded using the default values for multipart_threshold and multipart_chunksize (both 8MB). We're always using the defaults - but, if somehow the user uploads the files using other tools and changes the defaults and enables the sha256 checksum, then they won't match. But that seems like a pretty extreme edge case.

I haven't re-checked PR yet, so I'm not sure how that comment relates to the existing code, but if we get existing checksum from S3 object we can't *assume* it's compliant, we **MUST** check object has compliant part sizes.

dimaryaz: Right now, the code doesn't rely on existing checksums at all, compliant or not - and only uses them when uploading to S3. When an existing S3 file gets added to a package, it's downloaded and hashed locally.

docs/PARALLEL_CHECKSUMS.md
```diff
@@ -0,0 +1,32 @@
...
+The algorithm has an upper limit of 10,000 hashes;
+if the file is larger than 80,000 MiB,\
+it will double the block size until the number of blocks
+is under that limit.
+
+* If the file is less than 8 MiB,
+  the result will be identical to "serial" sha2-256.
+
+* If the file is exactly 8 MiB
+  the result will be the hash of the file hashed with a zero-size block.
```
dimaryaz, line 21: @drernie: I don't think I understand "hashed with a zero-size block". It's good to call out this corner case, but I'd describe it as a hash of one individual hash, or something along those words. I.e., the multi-hash algorithm is used for files ≥8MB, even though >8MB would have been more logical.

docs/PARALLEL_CHECKSUMS.md
```diff
@@ -0,0 +1,32 @@
+# PARALLEL CHECKSUMS
+
+## sha2-256-parallel-8
+
+### 0xb518
+
+This variant of sha2-256 is designed to enable very large files
+to be uploaded and hashed in parallel using fixed size blocks
+(8MiB to start with), with the final result being a "top_hash"
```
dimaryaz, line 9:
```suggestion
(8MiB to start with), with the final result being a "top hash"
```

(so it doesn't look like a Python identifier)

docs/PARALLEL_CHECKSUMS.md
```diff
@@ -0,0 +1,32 @@
...
+that is an order-independent hash of the individual hashes.
```
dimaryaz, line 10: "order-independent" sounds wrong to me. It's true that it doesn't depend on the upload order, but it certainly depends on the order of chunks within the file, so I'd make that more clear.

docs/sha2-256-chunked.md
```diff
@@ -0,0 +1,42 @@
...
+
+The algorithm has an upper limit of 10,000 hashes.
+If the file is larger than 80,000 MiB,
+it will double the block size until the number of blocks
+is under that limit.
+
+```pseudocode
+n = floor(source_size / (10_000 * 8 MiB))
+chunk_size = 8*2^n MiB
+```
```
dimaryaz, line 20: @drernie: this is actually wrong - we need `log` to calculate `n`, since we're doing `2^n` later. Though in the Python code, we're just using a loop cause it's simpler (and that's what boto3 does).

Simple example of the failure:

Let's say the file size would result in 32 * 10,000 parts using 8MiB chunks. The correct solution is to double the chunk size 5 times, resulting in 32*8MiB chunks. However, what we end up is:

```python
>>> MiB = 2**20
>>> source_size = 10_000 * 8 * MiB * 32
>>> n = math.floor(source_size / (10_000 * 8 * MiB))
>>> chunk_size = 8 * 2**n * MiB
>>> 
>>> source_size
2684354560000
>>> chunk_size
36028797018963968
```

Chunk size is bigger than the source size...

drernie: Doh! Good catch.  Is this better?
```python
> Hash of concatenated SHA2-256 digests of 8*2^n MiB source chunks
> where n = log(source_size / (10^4 * 8MiB))

n = floor(log(source_size / (10_000 * 8 MiB)))
```

docs/sha2-256-chunked.md
```diff
@@ -0,0 +1,43 @@
...
+
+The algorithm has an upper limit of 10,000 hashes.
+If the file is larger than 80,000 MiB,
+it will double the block size until the number of blocks
+is under that limit.
+
+```pseudocode
+n = floor(log(source_size / (10_000 * 8 MiB)))
+chunk_size = 8*2^n MiB
+```
```
sir-sigurd, line 21: that doesn't actually work
I suggest to use code that we actually use and that is tested
if for some reason we don't want to use the existing iterative algorithm these are my best guesses:

```
MIN_CHUNK_SIZE = 8 * 2 ** 20
MAX_NUM_CHUNKS = 10_000

from math import ceil, log2

def get_chunk_size1(source_size: int):
    if source_size == 0:
        return MIN_CHUNK_SIZE
    return MIN_CHUNK_SIZE * 2 ** ceil(log2(ceil(source_size / (MAX_NUM_CHUNKS * MIN_CHUNK_SIZE))))
def get_chunk_size2(source_size: int):
    if source_size == 0:
        return MIN_CHUNK_SIZE
    return MIN_CHUNK_SIZE * 2 ** ((source_size - 1) // (MAX_NUM_CHUNKS * MIN_CHUNK_SIZE)).bit_length()
```

docs/sha2-256-chunked.md
```diff
@@ -0,0 +1,43 @@
...
+## Inspiration
+
+This algorithm is inspired by Amazon's 
+[S3 Checksums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/)
+implementation.
+
+The key differences are:
+
+* Fixing the chunk size as (starting form) 8MiB.
+* Always hashing the result (even if source_size < 8MiB).
```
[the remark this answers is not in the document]
nl0: "starting f**or**m" or "f**ro**m"?

docs/sha2-256-chunked.md
dimaryaz: Just commented above, but I think ceil(log2(...)) is best.

api/python/quilt3/data_transfer.py
drernie: Is this resolved?

api/python/quilt3/data_transfer.py
drernie: Okay, that's fine. I will document that this differs from the Catalog that way.

docs/sha2-256-chunked.md
drernie:
```suggestion
n = ceil(log2(source_size / (10_000 * 8 MiB)))
```

docs/sha2-256-chunked.md
drernie: Updated with Dima's comment.
```
ceil(log2(ceil(source_size...)))
``
@sir-sigurd Why do we need the double `ceil`? Is that actually necessary for correctness, or just to match the way we do doubling?

Comment (sir-sigurd) [reply docs/sha2-256-chunked.md]: @drernie 

seems wrong, doesn't it?

```python
In [55]: MIN_CHUNK_SIZE = 8 * 2 ** 20
    ...: MAX_NUM_CHUNKS = 10_000
    ...:
    ...: from math import ceil, log2
    ...:
    ...: def get_chunk_size(source_size: int):
    ...:     return MIN_CHUNK_SIZE * 2 ** ceil(log2(source_size / (MAX_NUM_CHUNKS * MIN_CHUNK_SIZE)))
    ...:

In [56]: get_chunk_size(0)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[56], line 1
----> 1 get_chunk_size(0)

Cell In[55], line 7, in get_chunk_size(source_size)
      6 def get_chunk_size(source_size: int):
----> 7     return MIN_CHUNK_SIZE * 2 ** ceil(log2(source_size / (MAX_NUM_CHUNKS * MIN_CHUNK_SIZE)))

ValueError: math domain error

In [57]: get_chunk_size(1)
Out[57]: 0.0001220703125
```

sir-sigurd:
@dimaryaz could you update this (recently added) type?
https://github.com/quiltdata/quilt/blob/7a82dfcd869035c5e71a5cc7cd912af35d72515c/api/python/quilt3/packages.py#L74C7-L81

docs/sha2-256-chunked.md
```diff
@@ -0,0 +1,43 @@
+# sha2-256-chunked
```
nl0, line 1: will this page end up on the docs website somewhere? or will it be accessible only in the repo?

nl0 (approved):
i can't say i've fully digested the entire changeset, but it looks good to me so far -- i don't see anything suspicious. please 1) update the PR title and description to match the actual changes (it got somewhat out of sync thru the years lol); and 2) add a changelog entry (breaking) referencing this PR

sir-sigurd (dismissed):
I haven't fully digested the entire changeset as well, but LGTM

1. please use 'sha256' and 'sha256_chunked' vs 'legacy' and 'modern' if that makes sense to you (I thought it does)
2. there a few lines in PR uncovered by tests, might make sense finish them, but I won't insist

api/python/quilt3/data_transfer.py
```diff
@@ -890,15 +951,123 @@ def get_size_and_version(src: PhysicalKey):
...
+
+            if src.is_local():
+                bytes_remaining = length
+                with open(src.path, 'rb') as fd:
+                    fd.seek(offset)
+                    while bytes_remaining > 0:
+                        chunk = fd.read(min(s3_transfer_config.io_chunksize, bytes_remaining))
+                        if not chunk:
+                            # Should not happen, but let's not get stuck in an infinite loop.
+                            break
```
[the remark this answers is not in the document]
dimaryaz: Well, this won't catch the case where file has changed, but became longer. We also have the same problem for S3 files, but it's less obvious. In any case, we'll need to come up with a "real" solution.

dimaryaz:
@nl0 - I already had the changelog entries, unless there's something else I missed?

docs/sha2-256-chunked.md
drernie: In the short term, just in the repo.  I'm not sure what the right long-term doc strategy is.

drernie (dismissed):

api/python/quilt3/data_transfer.py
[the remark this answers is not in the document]
dimaryaz: Another solution is to say that we don't care what the user does to the file, but whatever ends up in S3 is what we consider to be "correct", and use its hash in the manifest.

In that case, we should only use `bytes_remaining` for the progress bar, but keep reading the file to the end.

sir-sigurd (dismissed):

api/python/quilt3/data_transfer.py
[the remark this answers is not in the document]
dimaryaz: I did some more digging to figure out the correct solution.

Because we now rely on server-side checksums, this code only runs for `Package.verify` (in which case, it'll fail later anyway, so that's fine), and when the user explicitly runs `Package.build` (and I'm not sure when that's actually useful?)

If the user _does_ run `build` before `push`, then we should also worry about files changing _between_ `build` and `push`, not just during `build`. This can result all kinds of problems:
- `push` fails with a weird exception because the number of chunks decreased, and we tried doing a 0-byte part upload
- `push` succeeds, and returns a package with correct checksums but incorrect size (if the file got smaller)
- `push` succeeds, and returns a package with truncated files (if the file got larger)

We _could_ detect the case of files changing between `build` and `push` (rather than during) and handle that more gracefully. But we should decide what the expected behavior is, or even what the purpose of `build` is. E.g., suppose the file contents changes but the size stays the same; we detect that, but only _after_ the push is done - now what?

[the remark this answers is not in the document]
dimaryaz: (Anyways, I changed it to raise an exception.)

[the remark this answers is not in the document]
sir-sigurd: > Because we now rely on server-side checksums, this code only runs for Package.verify (in which case, it'll fail later anyway, so that's fine), and when the user explicitly runs Package.build (and I'm not sure when that's actually useful?)

I'd also expect us to hash local file when ETag check succeeds and remote file doesn't have ChecksumSHA256, but IIUC now we end up hashing from S3. If my understanding is correct that looks like regression and it should be fixed.

docs/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Entries inside each section should be ordered by type:
 ## Python API
 * [Removed] Drop Python 3.7 support ([#3841](https://github.com/quiltdata/quilt/pull/3841))
 * [Changed] Set S3 client `max_pool_connections` to `QUILT_TRANSFER_MAX_CONCURRENCY` ([#3867](https://github.com/quiltdata/quilt/pull/3867))
+* [Changed] Switch from a regular SHA256 checksum to a hash list to match S3's built-in checksums ([#2782](https://github.com/quiltdata/quilt/pull/2782))
+* [Changed] Delay object hashing until package push to take advantage of S3's hashing; as a result, `dest` functions no longer receive a `top_hash` ([2782](https://github.com/quiltdata/quilt/pull/2782))
```
nl0, line 23:
```suggestion
* [Changed] **BREAKING:** Delay object hashing until package push to take advantage of S3's hashing; as a result, `dest` functions no longer receive a `top_hash` ([#2782](https://github.com/quiltdata/quilt/pull/2782))
```

api/python/quilt3/data_transfer.py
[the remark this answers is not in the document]
nl0: > I'd also expect us to hash local file when ETag check succeeds and remote file doesn't have ChecksumSHA256, but IIUC now we end up hashing from S3. If my understanding is correct that looks like regression and it should be fixed.

i'd say this is important enough to be fixed soon (in a follow-up (pre?)release), but not important enough to block the current PR and (pre)release

nl0 (approved):

api/python/quilt3/data_transfer.py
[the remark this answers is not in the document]
dimaryaz: Hmm interesting. I guess we could calculate the local SHA256 right there, and also use it to check SHA256 before we check the ETag, as @sir-sigurd suggested elsewhere. But yes, that's a big enough change that I'd rather do separately.

docs/sha2-256-chunked.md
dimaryaz: Technically, we need a max(0, log2(...)) to avoid <8MB chunks. And something similar to avoid log2(0). @drernie - is it worth the complexity here? We'd need something like:

```
n = floor(max(0, log(max(1, source_size / (10_000 * 8 MiB)))))
```

dimaryaz: Anyways, we can fix it later.

PR #2787, opened 2022-04-07
Fix scroll on homepage when there are few buckets

fiskus:
Instead of hardcoding height, I stuck the background element to parent's bottom. It's possible because for product mode there are fewer blocks.


fiskus:
It seems, `test-catalog` is failing because of some codecov issues

nl0:
> It seems, `test-catalog` is failing because of some codecov issues

yeah no problem

nl0 (approved):

PR #2791, opened 2022-04-12
Sshow/hide blocks according to `catalog/config.yaml`

fiskus:
Spec: https://github.com/quiltdata/specs/pull/40

All blocks are hidden in b/fiskus-sandbox-dev

### DIFF
--- a/shared/schemas/bucketConfig.yml.json
+++ b/shared/schemas/bucketConfig.yml.json
@@ -54,6 +54,31 @@
             }
           }
         },
+        "blocks": {
+          "type": "object",
+          "properties": {
+            "analytics": {
+              "default": true,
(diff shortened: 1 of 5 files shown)


shared/schemas/bucketConfig.yml.json
```diff
@@ -54,6 +54,31 @@
             }
           }
         },
+        "blocks": {
+          "type": "object",
+          "properties": {
+            "analytics": {
+              "default": true,
+              "description": "Shows/hides analytics block",
```
nl0, line 62:
```suggestion
              "description": "Show/hide analytics block",
```

nl0 (approved):
looks ok. shouldn't we add default (true) values to `defaultPreferences` tho?

PR #2793, opened 2022-04-12
Theming logo and background

fiskus:
* Added ThemeEditor: custom logo and custom background color for header/footer
* Refactored Logo:
  - optional src
  - height and width set from outside component
  - one image for quilt logo, just cut square for short one
 * Removed footer for internal pages, and restore "Docs" link in header

Screenshot from 2022-04-18 10-44-34


catalog/app/containers/Admin/Settings/ThemeEditor.tsx
```diff
@@ -0,0 +1,449 @@
...
+                )}
+
+                <M.Button
+                  type="submit"
+                  onClick={handleSubmit}
+                  variant="contained"
+                  color="primary"
+                  disabled={submitting || (submitFailed && hasValidationErrors)}
+                >
+                  Save
```
nl0, line 437: as per convention we use (based on mui guidelines, probably), "save" should be the rightmost button

nl0 (approved):
seems ok

PR #2768, opened 2022-03-30
Migrate package push API to GraphQL

nl0:
# Description

# TODO

- [x] resolve merge conflicts
- [x] [Changelog](CHANGELOG.md) entry

PR #2797, opened 2022-04-14
Disable package name validation, because it destroys copy package dialog

fiskus:


fiskus:
Fixed here https://github.com/quiltdata/quilt/pull/2768

PR #2804, opened 2022-04-18
Search help two columns

fiskus:
Screenshot from 2022-04-18 11-53-11


nl0 (approved):
the part with zeroes and ones is a little confusing (maybe i'd do smth with lambdas instead), but whatever

fiskus:
Implementation was influenced by regexp substitutions. But since there is no Regexp here it can be even more frustrating :)

I refactored it to functions+components

PR #2811, opened 2022-04-19
Refactor collaborators fading background

fiskus:


nl0 (commented):
what's the purpose of this change? seems like the only change is notation, but effect stays the same

fiskus:
Instead of manually converting `hex` to `rgba`, I use mui function. It's more readable, because now it's obvious these colors are the same. And it's easier to change color
The effect stays the same, right.

nl0:
> Instead of manually converting `hex` to `rgba`, I use mui function. It's more readable, because now it's obvious these colors are the same. And it's easier to change color The effect stays the same, right.

ah ok

nl0 (approved):

PR #2812, opened 2022-04-20
Theme improvements

fiskus:
* Footer is not customized again
* Custom Intercom launcher is a button with "marketing" colors
* Custom company logo can be any size, square is not required
* There is a curved borderline between custom background and quilt background: it's made with two colored blocks and border-radius 

Screenshot from 2022-04-20 15-54-04
Screenshot from 2022-04-20 15-56-11


fiskus:
Thanks, just two more fixes:
1. quilt logo is not aligned to the left (probably, this bug was from the very beginning
2. no padding if custom logo, but no custom background

fiskus:
Screenshot from 2022-04-21 17-59-20
Screenshot from 2022-04-21 17-59-48

PR #2816, opened 2022-04-22
Update bundlewatch

fiskus:
Also updates axios


nl0:
#2608

fiskus:
This PR is clean, manually created, and manually tested. I'd prefer to merge it rather than another one

nl0:
> This PR is clean

that PR is as clean as this one -- they are identical (just check the diffs)

> manually created

i don't see this as an advantage because it takes time and effort and is a subject to human error (renovate's PRs also have nice useful changelogs in the description to see what actually changed between versions)

> and manually tested.

no-one stops you from checking out renovate's branch and manually testing it (i'd even say it's encouraged)

PR #2819, opened 2022-04-25
Policies UI

nl0:
## TODO


catalog/app/containers/Admin/RolesAndPolicies/BucketsPermissions.tsx
```diff
@@ -0,0 +1,308 @@
...
+            <M.ListItemSecondaryAction style={{ right: 0 }}>
+              <M.Tooltip
+                title={`Read-${perm.level === Level.READ ? 'only' : 'write'} access`}
+              >
+                <M.IconButton
+                  onClick={(event) => openPermissionMenu(event, perm)}
+                  edge="end"
+                  size="small"
+                >
+                  <M.Icon>{perm.level === Level.READ ? 'visibility' : 'edit'}</M.Icon>
```
fiskus, line 256: I spent 10 minutes debugging this. Realizing it's not a bug :)
It's confusing, that the edit button does the same as the visibility icon. I'd suggest splitting it into two elements:
1. show status. Words are better than icons. It's not obvious that the eye icon means read-only
2. button to trigger menu

catalog/app/containers/Admin/RolesAndPolicies/Policies.tsx
```diff
@@ -0,0 +1,735 @@
...
+    getValue: R.prop('title'),
+    props: { component: 'th', scope: 'row' },
+  },
+  {
+    id: 'source',
+    label: 'Source',
+    getValue: (p: Policy) => p.managed,
+    getDisplay: (value: boolean) =>
+      value ? (
+        <abbr title="This IAM policy is created and managed by Quilt">Quilt</abbr>
```
fiskus, line 52: The more right way to add a dotted line is to use CSS. `<abbr>` has semantic meaning

catalog/app/containers/Admin/RolesAndPolicies/Policies.tsx
```diff
@@ -0,0 +1,735 @@
...
+  {
+    id: 'buckets',
+    label: 'Buckets',
+    getValue: (p: Policy) => (p.managed ? p.permissions.length : null),
+    getDisplay: (_buckets: any, p: Policy) =>
+      p.managed ? (
+        <M.Tooltip
+          arrow
+          title={
+            p.permissions.length ? (
```
fiskus, line 68: I think it deserves to be a standalone component

fiskus (approved):
Read a code without diving deep (I need a couple of days for this probably)
And checked the UI, it's good besides new popups after popups and issue with visibility/edit icons/buttons I mentioned in comments

catalog/app/containers/Admin/RolesAndPolicies/BucketsPermissions.tsx
```diff
@@ -0,0 +1,308 @@
...
+import defaultBucketIcon from 'components/BucketIcon/bucket.svg'
+import * as Model from 'model'
+import StyledLink from 'utils/StyledLink'
+import useQuery from 'utils/useQuery'
+
+import BUCKETS_QUERY from './gql/Buckets.generated'
+import { BucketPermissionSelectionFragment as BucketPermission } from './gql/BucketPermissionSelection.generated'
+
+const Level = Model.GQLTypes.BucketPermissionLevel
+type Level = Model.GQLTypes.BucketPermissionLevel
```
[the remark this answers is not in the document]
fiskus: Also, it seems that this branch merged with master long time ago

catalog/app/containers/Admin/RolesAndPolicies/AssociatedRoles.tsx
```diff
@@ -0,0 +1,200 @@
...
+                <M.Checkbox
+                  checked={selected.includes(role)}
+                  onChange={() => toggle(role)}
+                  color="primary"
+                />
+              }
+              label={
+                <>
+                  {role.name}{' '}
+                  <M.Box component="span" color="text.secondary">
```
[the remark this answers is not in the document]
nl0: i'm not sure about that, i think Box (or even inline styles) works well for elements with small number of styles -- it's easier to maintain than having a distinct class for every little thing (and forgetting to remove it once that little thing is not used anymore)

catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const attachPolicy = React.useCallback(
+    (policy: Policy) => {
+      onChange(value.concat([policy]))
+    },
+    [onChange, value],
+  )
+
+  const detachPolicy = (policy: Policy) => {
+    onChange(value.filter((p) => p.id !== policy.id))
+  }
```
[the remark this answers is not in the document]
nl0: it's used inside a closure which is regenerated on every render anyways (`onClick={() => detachPolicy(policy)}`), so there's no point in memoizing it

catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
```diff
@@ -0,0 +1,215 @@
...
+    setPolicySelectionOpen(true)
+  }, [setPolicySelectionOpen])
+
+  const closePolicySelection = React.useCallback(() => {
+    setPolicySelectionOpen(false)
+  }, [setPolicySelectionOpen])
+
+  const attachPolicy = React.useCallback(
+    (policy: Policy) => {
+      onChange(value.concat([policy]))
```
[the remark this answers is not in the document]
nl0: it's strange, but typescript was giving me errors when i was triyng to concat with a single item -- now it doesnt

catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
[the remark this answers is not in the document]
fiskus: As well as `attachPolicy` where you're using `useCallback`

catalog/app/containers/Admin/RolesAndPolicies/BucketsPermissions.tsx
nl0: i'm not sure how to do this in MUI, bc ListItem only has secondary action which is an IconButton (well, i guess you can put smth like a select there, but it doesn't look natural)

fiskus: Probably, you can use table instead of a list

catalog/app/containers/Admin/RolesAndPolicies/Policies.tsx
nl0: abbr here isnt used primarily for a dotted line, but more importantly for the tooltip that explains the meaning of the type of a role, so i think it's kinda correct semantically (to some extent)

catalog/app/containers/Admin/RolesAndPolicies/Policies.tsx
nl0: seems like more trouble than gain, since it's quite simple and not very reusable

catalog/app/containers/Admin/RolesAndPolicies/BucketsPermissions.tsx
nl0: well, table was the original layout, but list seems to suite this data better (apart from the icon/select problem), since there's not that many "fields" to represent this as a table (only bucket name and access level)

catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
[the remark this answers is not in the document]
nl0: `attachPolicy` is passed directly as a prop: `attachPolicy={attachPolicy}`

PR #2835, opened 2022-04-29
Add .ent for NGL loader

fiskus:
http://localhost:3000/b/pdbsnapshots/search?q=ext%3A.ent
http://localhost:3000/b/pdbsnapshots/search?q=ext%3A.ent.gz%20AND%20size%3A%3E100000

Screenshot from 2022-04-29 17-06-23


fiskus:
I didn't add `.Z` extension (no, it has nothing to do with politics).
We have a few README.Z files only. It works if I use gzip on them, but probably it needs its own uncompressor

nl0 (approved):

PR #2838, opened 2022-05-10
"Something went wrong" → 'Secure token expires"

fiskus:
* Move ErrorBoundar'ies to app/containers/Errors
* Add ErrorsCredentials in case `Credentials.use()` returns nothing
* Add refresh button to FinalBoundary

Screenshot from 2022-05-10 10-24-22

### DIFF
--- a/catalog/app/utils/error.ts
+++ b/catalog/app/utils/error.ts
@@ -40,16 +40,3 @@ export class ErrorDisplay extends BaseError {
     super(headline, { headline, detail, object })
   }
 }
-
-export class ErrorCredentials extends BaseError {
-  /**
-   * @param headline
-   *
-   * @param detail
-   *
-   * @param object
-   */
-  constructor(headline: string, detail?: string, object?: {}) {
-    super(headline, { headline, detail, object })
-  }
-}
--- a/catalog/app/utils/error.ts
+++ b/catalog/app/utils/error.ts
@@ -41,9 +41,6 @@ export class ErrorDisplay extends BaseError {
   }
 }
 
-/**
- * Error class with fields designed to feed <Error /> via object rest spread.
- */
 export class ErrorCredentials extends BaseError {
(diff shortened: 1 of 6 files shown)


catalog/app/utils/error.ts
```diff
@@ -40,3 +40,16 @@ export class ErrorDisplay extends BaseError {
     super(headline, { headline, detail, object })
   }
 }
+
+export class ErrorCredentials extends BaseError {
```
nl0, line 44: i don't think this util module is a proper place for this error class, probably `utils/AWS` or `utils/AWS/Credentials` is better. also, `CredentialsError` seems like a better name.

nl0 (approved):
looks good in general, but i feel like it's a stopgap and the actual solution would be to investigate why the tokens stop working and re-request them (page reload fixing the issue means the tokens are ok and the problem should be fixable)

fiskus:
> it's a stopgap and the actual solution would be to investigate
100% agree

catalog/app/utils/error.ts
fiskus: It seemed to me as a module to rule all errors. If it's not then I can move it to AWS/Credentials

fiskus: I understand that CredentialsError is nicer but naming things from general to specific is more convenient from my experience: listing is better, search is faster. Vue styleguide section is related https://v2.vuejs.org/v2/style-guide/?redirect=true#Order-of-words-in-component-names-strongly-recommended

nl0: > I understand that CredentialsError is nicer but naming things from general to specific is more convenient

yeah i see what you mean and i do that myself most of the times (i guess) as well

> It seemed to me as a module to rule all errors

it's more like a place for reusable general stuff, and more specific things should be kept closer to their domain

PR #2840, opened 2022-05-12
Improve Okta docs (auth servers), add default role guidance

akarve:


docs/technical-reference.md
```diff
@@ -345,6 +348,11 @@ Use current template > Next > Specify stack details), set the following paramete
 * *AuthClientId*: *Client ID*
 * *AuthBaseUrl*: *Issuer URL V2*
 
+#### Setting the default role
+**The Quilt admin must log in and set the default role** in order for new users to be able to use Quilt.
+Without this, new users will have no permissions.
+![](imgs/default-role.png)
+
```
sir-sigurd, line 355: What's wrong with this?
https://github.com/quiltdata/quilt/blob/49bbe650aa7e6c97ae82ea24b8a79b1861175d4d/docs/technical-reference.md?plain=1#L253-L260

akarve: I have removed the duplication in favor of the SSO section, where it makes the most sense and can be addressed sequentially after deployment.

sir-sigurd: Well, I'm not sure that's right, because it also affects password signup like on open.

nl0 (approved):
looks ok, but i feel that it's just a stopgap and we should automatically create an empty managed role on stack bring-up (or on migration) and make it default

docs/technical-reference.md
```diff
@@ -250,14 +250,13 @@ User email addresses are stored by the Identity Service in RDS Postgres (part of
...
 
-### Setting default role
+### Setting the default role
 
-If your stack settings allow users signing up by themselves, you must set the
-default role which will be assigned to all new users.
-Users won't be able to sign up by themselves until the default role is set,
-though they still can be invited by an admin via user management UI.
+**The Quilt admin must log in and set the default role** in order for new 
+users to be able to log in. Without this, new users will have no permissions.
```
nl0, line 256: @akarve this is not quite correct -- users won't be able to sign up unless default role is set up

akarve, line 256:
```suggestion
users to be able to sign up.
```

akarve: Fixed.

PR #2842, opened 2022-05-16
Files input checkboxes and menus

fiskus:
Create package:
Screenshot from 2022-05-16 09-28-38

Revise package:
Screenshot from 2022-05-16 09-29-50

Push from directory:
Screenshot from 2022-05-16 09-30-32


nl0 (commented):
I think hiding multiple actions under a single icon is a good idea, but in general I feel like we're trying to stretch one UI to handle distinct UXs and it looks and feels quite contrived IMO. Checkbox behavior on package create/revise list feels odd, bc when unchecking a box I would expect it to become unchecked rather than an item disappearing altogether. OTOH having menu icons with "no actions available" in the package from directory UI feels like a useless visual noise.

What's the main purpose and principles behind this refactoring? Like, what were you trying to achieve with this?

Some nitpicking:

1. Icon buttons are practically unusable with the scrollbar on the right:
Screenshot 2022-06-28 at 10 22 43

2. We may want tighter spacing between icon and text here: 
Screenshot 2022-06-28 at 10 24 25

Some suggestions:

1. Maybe render an action icon button directly in place of the menu icon when's there's only one action in the list?

2. Maybe don't render menu icon altogether when there's no actions available?

fiskus:
> 1. Icon buttons are practically unusable with the scrollbar on the right:

Can you clarify which OS and browser did you get this screenshot on? I can't reproduce on Linux/Firefox and macOS/Safari/Firefox: scrollbar always auto hides, and even when it is shown there is enough space to click on "More" button

fiskus:
I think I can add additional padding (4px). It looks natural.

I was able to accidentally click on the scrollbar only when tried to do so :) But when I try to click on a specific file when I'm targeting it scrollbar already disappears

---

One more solution is replacing "More" button with some word like Github does in container at the bottom of this page with link "Details"

nl0:
> I think I can add additional padding (4px). It looks natural.

that's ok i guess

fiskus:
As I see in Gmail: they disobey the rule to have only one action per row if it's even the rule :)
I think I can do the same: inline menu on hover over the file size

fiskus:
Like this:
Screenshot from 2022-06-28 15-19-42

fiskus:
https://user-images.githubusercontent.com/533229/176185394-08522f63-e645-47c7-ade3-3919e591d35a.mp4

nl0:
> As I see in Gmail: they disobey the rule to have only one action per row if it's even the rule :)
> I think I can do the same: inline menu on hover over the file size

maybe we should copy their approach a little more closely, e.g. use smaller nicer icon buttons instead of huge buttons with text?

nl0:
also we should consider usability on touch devices with no proper "hover" mechanics

fiskus:
> smaller nicer icon buttons instead of huge buttons with text?

I'd prefer to have explanatory text on buttons. Simplified buttons a little:

https://user-images.githubusercontent.com/533229/177179045-33407694-c830-4eb7-9471-f0524dd9445c.mp4
https://user-images.githubusercontent.com/533229/177179051-e65bfbaf-b445-47f6-a9d5-3802e58ba542.mp4

fiskus:
Mobile tap works as hover, but I removed opening OS's file browser on click on files

fiskus:
I would like to approach this with standalone web-components and import them as a separate UI library/component.

PR #2846, opened 2022-05-18
Temporary workaround to trigger CI

fiskus:


nl0 (commented):
if we go this route, probably it would make more sense to create this file under .circle dir and have some simple content in it (e.g. a single character would be enough)

fiskus:
It was supposed as a workaround for one day before Aneesh change my permissions. But we have another workaround: "Trigger pipeline" button

PR #2873, opened 2022-05-26
Bug: adding date causes infinite reload

fiskus:
1. Go to /b/fiskus-sandbox-dev/packages/fiskus/sandbox/tree/3bb7131572d3e4b1cfb6429cfdaad2316c3690609201994d6548db8a9a56092d/
Result: page will never finish rendering


fiskus:
It's just a repro for bug

PR #2879, opened 2022-05-31
Bucket preferences iteration

fiskus:
* Splited BucketPrefences module into React specific and data-focused
* Use one way to extend preferences for both yaml config and local stack
* Added tests for data

### DIFF
--- a/catalog/app/utils/BucketPreferences/BucketPreferences.spec.ts
+++ b/catalog/app/utils/BucketPreferences/BucketPreferences.spec.ts
@@ -1,3 +1,5 @@
+import dedent from 'dedent'
+
 import { parse } from './BucketPreferences'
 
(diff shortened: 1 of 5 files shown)


catalog/app/utils/BucketPreferences/BucketPreferences.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      const config = dedent`
+            ui:
+                nav:
+                    queries: QUERY
+      `
+      expect(() => parse(config, sentryMock)).toThrowError()
+    })
+  })
+
+  describe('extendUiDefaults', () => {
```
nl0, line 93:
```suggestion
  describe('extendDefaults', () => {
```

nl0 (approved):

PR #2880, opened 2022-06-01
Always show bucket browser on bucket tab

fiskus:
`blocks.browser` will hide files browser for packages. But on the Bucket tab the files browser is the main element


fiskus:
Yes, he told me to do this :)

PR #2883, opened 2022-06-02
thumbnail lambda: remove unused json output

sir-sigurd:


lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
```diff
@@ -73,7 +72,7 @@
             'enum': ['pdf', 'pptx']
         },
         'output': {
-            'enum': ['json', 'raw']
+            'enum': ['raw']
         },
```
sir-sigurd, line 76: This could be removed once catalog stop send `output` parameter in querystring. Does it make sense to do it in this PR?

nl0: i think we can remove it here altogether and i'll send a PR removing that from the catalog -- it's trivial

nl0 (approved):

JacksonMaxfield (approved):
This is some byte passing that I don't really know about / am missing context but from my read of things you are no longer encoding / decoding with base64 and I am assuming are just sending the raw bytes? 

The aicsimageio tests are passing so I think it's probably safe.

PR #2888, opened 2022-06-03
Thumbnail lambda png fix

eode:
# Description
Bugfix for issue: PNG file preview fails

### Problem
PIL 'modes' are the organization of the fundamental image data into color (or alpha) bands, bits-per-band, and byte order.
Each element of RGBA, for example, is a band.  Each band might have 1bpp, 8bpp, 16bpp, 32bpp, etc -- and the byte order might be different, as well. 

Each mode is an implementation of a unique collection of these facets -- and each implementation may or may not have the default resample filter used when downscaling an image.  Conversion also isn't effective for every mode<->mode pair.

### Solution
* convert 16-bit 'I' modes to 32 bits, which supports all of the available resampling filters.
* if a mode fails, try all of the filters (from greatest to least quality) to see if any are supported.  This results in a thumbnail of the best available quality for that image without conversion, so long as there is at least one resampling filter available to that mode.

### Sidenote
Pytest was warning about missing custom markers.  Although they can still be used without registering them, they were straightforward to register, and that has been done (reduces test noise).  If need be, I can pull this into a separate PR.

# TODO
- [x] Unit tests
- [n/a] Documentation
- [n/a] [Changelog](CHANGELOG.md) entry


lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
```diff
@@ -368,8 +368,38 @@ def lambda_handler(request):
...
+                success = False
+                for resampler in fallback_resampler_order:
+                    try:
+                        img.thumbnail(size, resample=resampler)
+                        success = True
+                        break
+                    except ValueError:
+                        continue
+                if not success:
+                    raise ValueError(
```
akarve, line 397: can we reraise the original here instead? (hopefully with added information that we tried to handle this, but failed).

lambdas/thumbnail/tests/conftest.py
```diff
@@ -16,6 +16,12 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "poppler: mark a test to run only if --poppler is indicated (poppler tools is installed)"
+    )
+    config.addinivalue_line(
+        "markers", "loffice: mark a test to run only if --loffice is indicated (LibreOffice is installed)"
+    )
```
akarve, line 24: Nice, please submit as separate PR.

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
akarve: reason being: all PIL errors should look like PIL errors and maybe caller wants to handle so we shouldn't change exception type, etc.

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
```diff
@@ -368,8 +368,38 @@ def lambda_handler(request):
...
+        try:
+            # 'I;16' modes have limited resamplers for scaling (poor quality)
+            # convert to 'I' to use a better resampler.
+            # remove the following two lines to prefer fast (but much lower quality) conversion
+            if img.mode.startswith('I;16'):
+                img = img.convert('I')  # 32-bit (more completely implemented by PIL)
+            # attempt to use the default resampler - we have test images using this.
+            img.thumbnail(size)
+        except ValueError as err:
+            if err.args[0] == 'image has wrong mode':
```
akarve, line 382:
```suggestion
            if 'image has wrong mode' in str(err):
```

lambdas/thumbnail/tests/conftest.py
eode: Separated out into another branch.  I'll make a PR for it as soon as this goes through.

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
eode: > can we reraise the original here instead?

Sure.

>  (hopefully with added information that we tried to handle this, but failed).

As a comment in that area of code where it's re-raised?

eode: ..or I could attach a message to it, like

```python
except ValuError as err:
    [...]
    err.note = "Re-raised: Tried all known resamplers, but none worked for this image mode."
```
I still think creating a new exception is a more informative path, though, as it'll have both the original traceback and the newly-created one, whereas re-raising only shows the original code path (even though it also was caught, handled, and re-raised).

eode: Note that I've peeled this segment of code into a separate function, `generate_thumbnail()` -- the one it was a part of is getting pretty long (70 lines still, after removal of that segment).

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
eode: Done.  This change makes the check more general in case they change or add things to the error.

akarve (approved):

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
```diff
@@ -90,6 +90,9 @@
 }
 
 
+_TEST_FALLBACK = False
```
eode, line 93: If this is True, the mode-specific fix I implemented is skipped, and the general fallback fix (likely poor quality) is used.  This is necessary for testing.  I don't like mixing testing logic into my code, but I don't really see an alternative here.

lambdas/thumbnail/tests/test_thumbnail.py
```diff
@@ -144,7 +146,19 @@ def test_generate_thumbnail(
     # Create the lambda request event
     event = _make_event({"url": url, **params})
     # Get the response
-    response = t4_lambda_thumbnail.lambda_handler(event, None)
+    if expected_thumb == "I16-mode-128-fallback.png":
```
eode, line 149: I don't like piggybacking parameter meanings -- but in this case, it might be the least disruptive thing.  When we're testing for the result png `I16-mode-128-fallback.png`, we do the fallback test.

The only alternative I see is to add a new parameter to the test, `True` for the fallback test, and `False` for every other test.  It seems a bit overkill, which is why I went the way I did.  I don't like either option, tbh, and am open to either -- or any ideas you have.

eode:
@akarve : The above two comments are my two concerns -- if you're still good with it, I'll start working on deploy.

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
akarve: Go ahead and create a fresh exception. I understand your point of view.

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
```diff
@@ -322,6 +325,38 @@ def handle_pptx(*, src: bytes, page: int, size: int, count_pages: bool):
...
+        # attempt to use the default resampler - we have test images using this.
+        img.thumbnail(size)
+        return img
+    except ValueError as err:
+        if 'image has wrong mode' in str(err):
+            # The default resampler doesn't work with this image mode.
+            # PIL does not support all resamplers with all modes.
+            # These are all of the resamplers available, Ordered highest to lowest quality.
+            fallback_resampler_order = [Image.LANCZOS, Image.BICUBIC, Image.HAMMING,
+                                        Image.BILINEAR, Image.BOX, Image.NEAREST]
```
akarve, line 348:
```suggestion
            fallback_resampler_order = [
                Image.LANCZOS,
                Image.BICUBIC,
                Image.HAMMING,
                Image.BILINEAR,
                Image.BOX,
                Image.NEAREST
            ]
```

lambdas/thumbnail/tests/test_thumbnail.py
akarve: This feels wrong. Can't we just have different image format string to hit the fallback case? I'd rather see us patch something out in the tested function than add a test-only boolean like this.

eode: Yeah, I don't like it either.. ..but..

The image mode string is generated when the image is read in (or created from an array).  Just changing the mode doesn't change the bytes, it just changes how PIL treats them.  We would still need valid original and thumbnail images to test against.  PIL/PILLOW has a [plethora of testing images](https://github.com/python-pillow/Pillow/tree/main/Tests/images), as you may imagine -- and I'm sure an image of each mode is accounted for in there, so we could use those (if the licensing is OK) for testing exhaustively -- but it'd be time consuming.

I would ideally like to keep all testing-related code separate from the actual code -- but the basic issue is that if there's a mode that we're certain is a failure case, we may as well fix it and match the quality of our existing thumbnails.  But then that mode is no longer a failure case, and the fallback wouldn't be triggered. 

The current fix fixes two things:
* The mode "I;16", which doesn't have the normal scaling filters, now generates a normal-quality thumbnail, where it didn't generate any thumbnail before
* any undiscovered mode which doesn't have the normal scaling filters will still generate a thumbnail, so long as that mode has a scaling filter of any quality.  The bool is to test this segment of code.

Some alternate options I can think of:
* Ditch the global bool, and use a parameter like `fallback_only=False` to the `generate_thumbnail` function, and call it with `True` when testing the fallback.  Substantially similar to what we're doing now, minus the global.
* remove the bool, and either
  * remove the fallback handling, and just let any new failures fail to generate a thumbnail, and handle each failing mode when/if it happens, resulting in some potential missing thumbnails, but a known fix if they pop up
  * or switch to fallback-only on failure of the normal scaling, so that the thumbnail quality on `I;16` mode is reduced, but the code path is tested by the current tests and needs no bool for testing (neither as a parameter nor as global)
  * or just don't test that segment of code (but that seems wrong, too)

lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py
```diff
@@ -322,6 +325,44 @@ def handle_pptx(*, src: bytes, page: int, size: int, count_pages: bool):
...
 
 
+def generate_thumbnail(arr, size):
+    # Send to Image object for thumbnail generation and saving to bytes
+    img = Image.fromarray(arr)
+
+    # The mode I;16 has limited resamplers for scaling, and throws an error.
+    # Rather than use a non-default poor-quality resampler, convert to a better-handled mode.
+    if img.mode == 'I;16' and not _TEST_FALLBACK:
+        img = Image.fromarray((arr // 256).astype('uint8'))
```
akarve, line 335: If it's hard to find/generate an image that actually hits the fallback case then you can patch (or wrap) Image.fromarray to return`fromarray(arr)` and that effectively skips the above line.

akarve: Another thing you could do (maybe even cleaner) is make a sub function that takes the image mode as a string and just patch that value on a specific test run to skip the above line.

eode: > patch (or wrap) Image.fromarray

true....

> make a sub function

I'll do that.

eode: Alright.  My only remaining reservation is that after handling the PIL `ValueError("image has wrong mode")` and failing, I still think we should raise our own error.  But overall, I'm ok with what's here now.

akarve: Feel free to do error handling as you prefer: raise our own.

PR #2892, opened 2022-06-09
Move CSS files from app/ to app/assets/

fiskus:
* I think we can move favicon.ico and index.html there too
* `ipynb-styling/Makefile` doesn't work, maybe we can remove it, I don't know if it worths fixing


nl0 (approved):
what's the reasoning behind this?

fiskus:
The reason is a clean root directory :) 
I want to add another theme for Perspective

nl0:
> The reason is a clean root directory :) I want to add another theme for Perspective

mkay

fiskus:
> it's needed for building (updating) ipynb styles, which we still rely on
I understand, but if it's broken, I think we don't depend on it really :)

This script uses `master` which went ahead, I suppose. Imported files are not there, no bower.json. We can `checkout` some tag or commit instead of `master to fix it

PR #2895, opened 2022-06-10
Support .tab as .tsv

fiskus:
Wikipedia mentions both .tsv and .tab as tab separated values https://en.wikipedia.org/wiki/Tab-separated_values
Gnumeric too https://help.gnome.org/users/gnumeric/stable/gnumeric.html#sect-file-open-formats

I tested with two files found in https://open.quiltdata.com/search?q=ext%3A.tab


nl0 (approved):
thanks for taking this

PR #2896, opened 2022-06-13
Metadata → Details

fiskus:
Screenshot from 2022-06-13 13-35-09
Screenshot from 2022-06-13 13-35-29


akarve:
This looks good but needs some wordsmithing:

* METADATA (same as before)
    * Message
    * User metadata
    * Workflow

akarve (commented):
This looks good but needs some wordsmithing:

* METADATA (same as before)
    * Commit message
    * User metadata
    * Workflow

Nice-to-have: on hover show JSON path to values displayed

catalog/app/containers/Bucket/Meta.tsx
```diff
@@ -0,0 +1,104 @@
...
+interface HeadCellProps {
+  children: React.ReactNode
+  className?: string
+  title: string
+}
+
+const HeadCell = ({ className, children, title }: HeadCellProps) => {
+  const classes = useHeadCellStyles()
+  return (
+    <M.TableCell className={cx(classes.root, className)} component="th" scope="row">
```
nl0, line 27: maybe vertically align this to the top? seems like it would look better that way

nl0 (approved):
im not 100% convinced with the "table" design, maybe it would look better without borders idk, but anyways it feels like a step into the right direction

fiskus:
* Removed vertical borders
* Vertically aligned table row titles
* Added quotes
* Renamed "Commit 
Screenshot from 2022-06-23 12-29-03
Screenshot from 2022-06-23 12-29-16
message" to "Message"

nl0:
> it would be nicer to not have quotes around message

> Added quotes

=) sorry for misleading comment, but seems like there were NO quotes at the moment of review, but i was commenting based on the screenshot which had quotes

also you could probably make the spacing between labels and values a little smaller

fiskus:
Yeah, missed "not" in "to not have quotes" :)

PR #2900, opened 2022-06-15
Test quilt3 on Win on all Python versions

eode:
# Description
Cleaner redo of #2861 after incorrect rebase.

# TODO

- [n/a] Unit tests
- [n/a] Documentation
- [n/a] [Changelog](CHANGELOG.md) entry

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1462,7 +1462,7 @@ class Package:
         def physical_key_is_temp_file(pk):
             if not pk.is_local():
                 return False
-            return pathlib.Path(pk.path).parent == APP_DIR_TEMPFILE_DIR
+            return pathlib.Path(pk.path).parent.resolve() == APP_DIR_TEMPFILE_DIR.resolve()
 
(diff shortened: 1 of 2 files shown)


sir-sigurd (changes requested):
It looks like a real bug fix so changelog entry is needed.

api/python/quilt3/packages.py
```diff
@@ -1462,7 +1462,7 @@ def check_latest_hash():
         def physical_key_is_temp_file(pk):
             if not pk.is_local():
                 return False
-            return pathlib.Path(pk.path).parent == APP_DIR_TEMPFILE_DIR
+            return pathlib.Path(pk.path).parent.resolve() == APP_DIR_TEMPFILE_DIR.resolve()
```
sir-sigurd, line 1465: Does it make sense to do this normalization when we create `PhysicalKey` instances?

eode: I'd agree with both of those points.  In theory, the resolution should not have a problematic effect -- but knowing definitively is another matter.

akarve (approved):

sir-sigurd (approved):

PR #2919, opened 2022-06-23
Do not send useless data to ES

sir-sigurd:
`comment` and `target` were always empty for object index,
`size` was always 0 for package index.


akarve (approved):
I think it's just as easy to check for an empty string as it is for null, but whatever. My general suggestion with these kinds of changes is they don't fix anything, but they might break something. So I'm not convinced we should fix things that aren't broken. But leaving up to you.

nl0 (approved):

PR #2922, opened 2022-06-27
Beta switch and enabling "Open in Desktop"

fiskus:
* Add "beta" switch to enable "Beta" features
* Treat "Open in Desktop" as beta feature

### DIFF
--- a/catalog/app/containers/Admin/Settings.tsx
+++ b/catalog/app/containers/Admin/Settings.tsx
@@ -13,6 +13,25 @@ import * as validators from 'utils/validators'
 import * as Form from './Form'
 import ThemeEditor from './Settings/ThemeEditor'
 
+const useBetaEditorStyles = M.makeStyles((t) => ({}))
+
+function BetaEditor() {
+  const settings = CatalogSettings.use()
+  const [beta, setBeta] = React.useState(settings?.beta || false)
+  const [disabled, setDisabled] = React.useState(false)
+  const writeSettings = CatalogSettings.useWriteSettings()
+  const onChange = React.useCallback(async () => {
+    setDisabled(true)
+    setBeta(!beta)
+    await writeSettings({
+      ...settings,
+      beta: !beta,
+    })
+    setDisabled(false)
+  }, [writeSettings])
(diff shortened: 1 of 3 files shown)


catalog/app/containers/Admin/Settings.tsx
```diff
@@ -278,6 +299,12 @@ export default function Settings() {
             <ThemeEditor />
           </React.Suspense>
         </M.Paper>
+        <M.Paper className={classes.group}>
+          <M.Typography variant="h5" className={classes.sectionHeading}>
+            Enable beta-features
```
nl0, line 304:
```suggestion
            Enable beta features
```

nl0 (commented):
this looks good for now, tho going forward i'd prefer to have fine-grained feature flags rather than a single umbrella option for everything

fiskus:
So, "Enable Open in Desktop beta" rather then "Enable beta features"?

nl0:
> So, "Enable Open in Desktop beta" rather then "Enable beta features"?

for now (since we only have one beta feature) it's ok to keep it as is, but going forward i think it makes sense to have a top-level switch e.g. "Enable beta features" which, when switched on, opens up a list of individual feature flags e.g. "open in desktop"

fiskus:
I think, we already have these "flags" in `.quilt/catalog/config.yaml`. But they are hard to edit. We can add `brace` (or any other) editor to change text files inline.

fiskus:
> "Enable beta features" which, when switched on, opens up a list of individual feature flags e.g. "open in desktop"

I like this, but it's a nightmare to test :)

nl0:
> I think, we already have these "flags" in `.quilt/catalog/config.yaml`. But they are hard to edit. We can add `brace` (or any other) editor to change text files inline.

yeah, I'm thinking in more or less same direction, but I'd prefer a stricter API (GraphQL) and a friendlier UI (labeled switches)

fiskus:
Because, you need to test your feature once. Then with feature1 and feature2 enabled, then with feature 1 and feature 2 disabled, then feature1 enabled and feature2 disabled, then with feature1 disabled and feature2 enabled

nl0 (approved):

PR #2923, opened 2022-06-29
Handle s3:// URLs in JsonDisplay

fiskus:
Sample page:
/b/quilt-bio-staging/packages/aneesh/fcsbeads/tree/395103f0d513fa26e35b9518d92e77d787cdf6509c572ea6f485db9db97b8b8a/

Screenshot from 2022-06-29 15-19-00


fiskus:
Yes, the Markdown renderer handles some URLs already, but not such URLs as .quilt/workflows/config.yaml

catalog/app/components/JsonDisplay/JsonDisplay.js
```diff
@@ -75,18 +78,38 @@ function NonStringValue({ value }) {
...
 
+function S3UrlValue({ href, children }) {
+  const { urls } = NamedRoutes.use()
+  const isInStack = BucketConfig.useIsInStack()
+  const props = React.useMemo(() => {
+    const s3Handle = s3paths.parseS3Url(href)
+    const inStack = isInStack(s3Handle.bucket)
+    return {
+      href: inStack
+        ? urls.bucketFile(s3Handle.bucket, s3Handle.key, s3Handle.version)
```
nl0, line 89: internal catalog url should be passed as `to` prop to be handled by react-router

nl0 (approved):
looks ok, tho maybe i wouldn't make distinction btw in-stack and out-of-stack buckets and just navigate to the object via catalog (it's probably ok to see "access denied" if it's inaccessible via catalog)

fiskus:
I suppose, if we show "Access Denied" in Catalog for the file, then the same result would be with `https;//bucket.s3.amazon.com/key?credentials`, but with ugly XML :) So, yes, I'll open out-of-stack URLs on catalog too

catalog/app/components/JsonDisplay/JsonDisplay.js
fiskus: Thanks, I didn't notice reload

PR #2929, opened 2022-07-02
Auth screen: add test ids

nl0:


nl0:
cc @fiskus

nl0:
merging in, it's quite trivial

fiskus:
Also, it will help me `grep` code for components I inspect with DevTools

PR #2934, opened 2022-07-07
Fix quilt3 install path examples

fiskus:
Aneesh provided samples with trailing slash in Python code and without in CLI code. I don't know if it was intentional, maybe you know


nl0 (approved):

nl0:
> samples with trailing slash in Python code and without in CLI code
> I don't know if it was intentional, maybe you know

i don't remember tbh, probably @sir-sigurd is more familiar with the api conventions re:slashes

sir-sigurd:
Both should not have trailing slash, otherwise they fail.

PR #2935, opened 2022-07-07
Admin: Status UI + related refactorings

nl0:
I can't upload a screenshot for some reason, so I'm posting it in slack

# TODO

### DIFF
--- a/catalog/app/components/EChartsChart.tsx
+++ b/catalog/app/components/EChartsChart.tsx
@@ -1,0 +1,96 @@
+import * as echarts from 'echarts'
+import * as React from 'react'
+import * as R from 'ramda'
+import useResizeObserver from 'use-resize-observer'
+import * as M from '@material-ui/core'
+
+import { createBoundary } from 'utils/ErrorBoundary'
+import useMemoEq from 'utils/useMemoEq'
+
+export type DisposeHook = () => void
+export type InitHook = (chart: echarts.ECharts) => DisposeHook | void
+
+interface ChartInnerProps {
+  option: echarts.EChartsOption
+  resize?: boolean
+  theme?: Parameters<typeof echarts.init>[1]
+  initOptions?: Parameters<typeof echarts.init>[2]
+  onInit?: InitHook
+}
+
+function ChartInner({
+  option,
+  resize = true,
+  theme = undefined,
+  initOptions = undefined,
+  onInit = undefined,
+  ...props
+}: ChartInnerProps & M.BoxProps) {
+  const [error, setError] = React.useState<unknown>(null)
+  if (error) throw error
+
+  const containerRef = React.useRef<HTMLDivElement | null>(null)
+  const chartRef = React.useRef<echarts.ECharts | null>(null)
+  const initArgs = useMemoEq([theme, initOptions] as const, R.identity)
+  const optionMemo = useMemoEq(option, R.identity)
+  const onInitRef = React.useRef<InitHook | undefined>()
+  onInitRef.current = onInit
+
+  React.useEffect(() => {
+    if (!containerRef.current) return
+    try {
+      const chart = echarts.init(containerRef.current, ...initArgs)
+      // XXX: consider accepting setOption opts via props (notMerge, replaceMerge, lazyUpdate)
+      chart.setOption(optionMemo)
+      const { current: onInitCur } = onInitRef
+      let dispose: DisposeHook | void = undefined
+      if (onInitCur) dispose = onInitCur(chart)
+      chartRef.current = chart
+      return () => {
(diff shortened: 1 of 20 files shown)


catalog/app/components/EChartsChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+        chart.dispose()
+      }
+    } catch (e) {
+      setError(e)
+    }
+  }, [containerRef, initArgs, optionMemo, onInitRef])
+
+  const size = useResizeObserver({ ref: containerRef })
+  const sizeStr =
+    size.width != null && size.height != null ? `${size.width}x${size.height}` : ''
```
fiskus, line 61: It seems that you want `onResize` callback that `use-resize-observer` has

catalog/app/containers/Admin/Status/gql/Status.graphql
```diff
@@ -0,0 +1,24 @@
+query ($window: Int) {
```
fiskus, line 1: It's confusing to see the name "window" in the frontend environment meaning something different from UI windows. Maybe, "timespan" or "time frame" fits better?

fiskus (commented):
Couldn't test it manually. WIth both staging and searchminimal environment, I'm getting Error screen

catalog/app/components/EChartsChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+function ChartInner({
+  option,
+  resize = true,
+  theme = undefined,
+  initOptions = undefined,
+  onInit = undefined,
+  ...props
+}: ChartInnerProps & M.BoxProps) {
+  const [error, setError] = React.useState<unknown>(null)
+  if (error) throw error
```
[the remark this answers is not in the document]
nl0: > I think it will also produce another error: rendered less hooks than in the previous render

nope, it will just throw an error which will be propagated to a nearest error boundary

catalog/app/components/EChartsChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }, [containerRef, initArgs, optionMemo, onInitRef])
+
+  const size = useResizeObserver({ ref: containerRef })
+  const sizeStr =
+    size.width != null && size.height != null ? `${size.width}x${size.height}` : ''
+
+  React.useEffect(() => {
+    if (resize && sizeStr) chartRef.current?.resize()
+  }, [chartRef, sizeStr])
```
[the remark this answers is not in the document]
nl0: yep, thanks, eslint doesnt work in pre-commit hooks =(
we should fix it sooner

catalog/app/components/EChartsChart.tsx
nl0: i was thinking about that, but the current approach with computed values and effects seems cleaner bc it handles all the required state and references, without needing to pass chart ref etc to the callback and handling equality checks and stuff 🤷

catalog/app/containers/Admin/Status/gql/Status.graphql
fiskus: As I see, we use it only for ANALYICS_WINDOW constant. It think, it's better to rename it, because the naming is confusing

fiskus:
Please, let me know if and when I can test it manually

catalog/app/containers/Admin/Status/gql/Status.graphql
nl0: there's two more instances of `window` query parameter in our gql schema, and by analogy with ANALYTICS_WINDOW* js variable im introducing a new one -- STATS_WINDOW. i dont feel its confusing when seen in context. we can refactor it later in one take if we come up with a better language.

catalog/app/containers/Admin/Status/Status.tsx
```diff
@@ -0,0 +1,487 @@
...
+    marginLeft: t.spacing(1),
+  },
+}))
+
+function CountsByState() {
+  const classes = useCountsByStateStyles()
+  const apiRef = React.useContext(MDG.GridApiContext)
+  const counts = MDG.useGridSelector(apiRef, countsByStateSelector)
+
+  const renderItem = (count: number, label: string, ok: boolean | null) =>
```
fiskus, line 63: Probably, it can be standalone component or standalone function, outside `CountsByState`

fiskus (approved):

catalog/app/containers/Admin/Status/Status.tsx
nl0: yeah, it could, but this code is so small and it doesn't make sense on its own

PR #2936, opened 2022-07-09
Raise a TypeError if wrong type given to fix_url

eode:
# Description
This is in relation to "Package.set_dir() should be more restrictive to path".

Note that I've modified `fix_url` instead of `set_dir()`.  `fix_url()` is the centralized place for checking local-style paths.  All tests passed with this small change, but it could cause people to need to adapt, if they have code dependent on the previous behavior.

If changing this here is considered too broad, I can shift it into `set_dir`.

# TODO

- [x] Unit tests
- [x] Documentation
- [?] [Changelog](CHANGELOG.md) entry


eode:
It could be worth looking up functions that use `fix_url` and updating their docstrings as well.  Or, moving the change out of `fix_url` and into `set_dir()`

sir-sigurd (commented):
I'm not yet sure whether we should touch `fix_url()`, but there are some thoughts:

* supporting `PathLike` is a good idea
* do we have some tests with `PathLike`?
* When we know that a value is a path, but not a URL, I *think* it's preferred to use `PhysicalKey.from_path()` rather than `PhysicalKey.from_url()` like here:
https://github.com/quiltdata/quilt/blob/5f2ec40396ad1e95c9aa8974f56dc39900d8d008/api/python/quilt3/packages.py#L844-L847
So it might make sense to check whether `path` is instance of `PathLike` and use `PhysicalKey.from_path()` in that case.

eode:
@sir-sigurd Ran across something.

(master) This is a weird behavior:
```python
>>> p = Package()
>>> p.set_dir('example')
>>> p.set_dir('')
# Result: Package is set to PWD.

>>> p = Package()
>>> p.set_dir('')
>>> p.set_dir('example')
# Result: Package is set to PWD.
```
That seems like buggy behavior.  Is it actually intentional?

Extended example:
```python
>>> p = Package()
>>> p.set_dir('')
(local Package)
 └─bar
 └─example/
   └─baz
   └─bing
 └─foo
>>> p.set_dir('example')
(local Package)
 └─bar
 └─example/
   └─baz
   └─bing
 └─foo
>>> p = Package()
>>> p.set_dir('example')
(local Package)
 └─example/
   └─baz
   └─bing
>>> p.set_dir('')
(local Package)
 └─bar
 └─example/
   └─baz
   └─bing
 └─foo
```

I'd imagine that it should 
a: always set the package dir, or 
b: set the package dir once and create an error in other circumstances.

Can you specify what the behavior should be?  I'm working on that chunk of code, and may as well fix this if it's a bug.

sir-sigurd:
I don't understand what's unexpected. From [docs](https://docs.quiltdata.com/api-reference/package#package.set_dir):

> Package.set_dir(self, lkey, path=None, meta=None, update_policy='incoming')
...
path(string): path to scan for files to add to package. If None, lkey will be substituted in as the path.

Does this make sense?

eode:
Nevermind, I was thinking of `set_dir` as setting the package contents to the specified tree, rather than adding the specified tree to the package.  Still an RTFM issue, though -- my mistake.

QuiltSimon:
Closing as abandoned: opened by an external contributor, last touched 2025-05-16, no maintainer engagement since. Raising a TypeError on a wrong-typed fix_url argument is defensible, but a year-plus-stale one-liner against a client library that has moved on needs re-verification rather than a merge. If the underlying looseness in fix_url still matters, it is a fresh change with a test.

PR #2939, opened 2022-07-14
Make S3 proxy add ELB certificate at startup

sir-sigurd:


s3-proxy/run-nginx.sh
```diff
@@ -12,6 +12,11 @@ then
     exit 1
 fi
 
+if [ $ELB_CERTIFICATE_ARN ]
+then
+    aws --region $(echo $ELB_CERTIFICATE_ARN | cut -d ":" -f 4) acm get-certificate --certificate-arn $ELB_CERTIFICATE_ARN --output text --query Certificate >> /etc/nginx/certs.pem
```
sir-sigurd, line 17: I'm not sure if it should add `Certificate` or `CertificateChain` or both.

dimaryaz: From [Nginx's docs](https://nginx.org/en/docs/http/configuring_https_servers.html):

> Some browsers may complain about a certificate signed by a well-known certificate authority, while other browsers may accept the certificate without issues. This occurs because the issuing authority has signed the server certificate using an intermediate certificate that is not present in the certificate base of well-known trusted certificate authorities which is distributed with a particular browser. In this case the authority provides a bundle of chained certificates which should be concatenated to the signed server certificate.

I've definitely seen cases where e.g. Firefox accepts the cert, but curl does not because it doesn't know the intermediate certs. So including the chain is the safest thing to do.

So try concatenating the cert and the chain in a single .pem, and see if it works?

dimaryaz (approved):

PR #2946, opened 2022-07-21
Fix access counts position when there is metadata

fiskus:
Added wrapper "anchored" access count chart instead of positioning it with pixels

Before:

Screenshot from 2022-07-21 16-13-05

After;
Screenshot from 2022-07-21 16-13-15


nl0 (approved):
nice

PR #2949, opened 2022-07-25
More test ids

nl0:


catalog/app/components/SearchResults/SearchResults.js
```diff
@@ -457,7 +457,12 @@ function FileHit({ showBucket, hit: { path, versions, bucket } }) {
     })
 
   return (
-    <Section>
+    <Section
+      data-testid="search-hit"
+      data-search-hit-type="file"
+      data-search-hit-bucket={bucket}
+      data-search-hit-path={path}
```
fiskus, line 464: Probably, we can use some decorators for components, destructuring `data-search-hit={handle}` to this

fiskus (approved):

catalog/app/components/SearchResults/SearchResults.js
nl0: that's probably worth exploring

PR #2952, opened 2022-07-26
File editor redirect

fiskus:
* Added redirect to the next version, and remove `wait` (I can't reproduce stale content now)
* Added redirect to the previous page provided in the `next` query string
* Added `bucketFileEdit` route (I'm not sure about this)

### DIFF
--- a/catalog/app/components/FileEditor/FileEditor.tsx
+++ b/catalog/app/components/FileEditor/FileEditor.tsx
@@ -7,12 +7,23 @@ import AsyncResult from 'utils/AsyncResult'
 import * as NamedRoutes from 'utils/NamedRoutes'
 import parseSearch from 'utils/parseSearch'
 import type { S3HandleBase } from 'utils/s3paths'
-import wait from 'utils/wait'
 
 import Skeleton from './Skeleton'
 import TextEditor from './TextEditor'
 import { detect, loadMode, useWriteData } from './loader'
 import { EditorInputType } from './types'
+
+function useRedirect() {
+  const history = RRDom.useHistory()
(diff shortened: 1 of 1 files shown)


nl0 (approved):
looks ok overall, but i'd adjust route handling

> Added bucketFileEdit route (I'm not sure about this)

it would be ok probably to reuse the `bucketFile` route, e.g.
```ts
export const bucketFile = {
  path: '/b/:bucket/tree/:path(.*[^/])',
  url: (bucket, path, version = undefined, mode = undefined, { edit, next } = {}) =>
    `/b/${bucket}/tree/${encode(path)}${mkSearch({ mode, version, edit, next })}`,
}
```

ideally we should move all the options into the last options argument, e.g.
```ts
export const bucketFile = {
  path: '/b/:bucket/tree/:path(.*[^/])',
  url: (bucket, path, { version, mode, edit, next } = {}) =>
    `/b/${bucket}/tree/${encode(path)}${mkSearch({ mode, version, edit, next })}`,
}
```
but that would require adjusting this function calls all across the app (not that many, probably, but anyways)

fiskus:
```
export const bucketFile = {
  path: '/b/:bucket/tree/:path(.*[^/])',
  url: (bucket, path, { version, mode, edit, next } = {}) =>
    `/b/${bucket}/tree/${encode(path)}${mkSearch({ mode, version, edit, next })}`,
```
I'd prefer to call `bucketFile(handle, options)` because I handle is the small building block of data used across the app. Maybe it worth changing all routes that way at once at some point

fiskus:
Follow up PR with addressed changes https://github.com/quiltdata/quilt/pull/2953

PR #2955, opened 2022-07-28
Add to Package aka Shopping Cart

fiskus:
# Description

This PR doesn't change anything visually, because I need one last feature to implement: show Revise dialog https://github.com/quiltdata/quilt/pull/2957. Then I'll turn "Add README" on.

AddToPackage is basically a simple map with files in the same format as files added from S3FilesPicker to Package Dialog and accessible anywhere in the app to read and write. I suspect we'll need to sync this map to localStorage later.

Notable additions/changes:
  * `containers/AddToPackage` stores map of files
  * `S3FilesPicker.S3File` becomes `Model.S3File`
  * `urls.bucketFile` URL accepts `?add=true` to automatically append edited file to `AddToPackage` map

# TODO

- [ ] Unit tests
- [ ] Documentation


catalog/app/containers/AddToPackage/Provider.tsx
```diff
@@ -0,0 +1,34 @@
...
+import * as R from 'ramda'
+import * as React from 'react'
+
+import type * as Model from 'model'
+
+const Ctx = React.createContext<{
+  append: (file: Model.S3File) => void
+  entries: Record<string, Model.S3File>
+}>({
+  append: () => {},
```
nl0, line 12: you probably don't want this dummy context, bc it will just silently swallow data when there's no provider in tree. maybe just throw an error (e.g. 'Please initialize the provider') in `append()` fn

catalog/app/containers/AddToPackage/Provider.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export function Provider({ children }: ProviderProps) {
+  const [entries, setEntries] = React.useState({})
+  const append = React.useCallback((s3File) => {
+    setEntries(
+      R.mergeLeft({
+        [basename(s3File.key)]: s3File,
+      }),
+    )
+  }, [])
```
nl0, line 28: we probably need a way to clear the state once we consume the data -- it could be either a combined method (e.g. `consume()`) which would return the entries and clear the state, or a dedicated method (e.g. `clear()`) which can be called manually after we consume the entries

catalog/app/containers/Bucket/Summary.js
```diff
@@ -187,16 +209,18 @@ export default function BucketSummary({
   )
   const { readme, images, summarize } = extractSummary(files)
 
+  const canAddReadme = false // !readme && !!packageHandle AND can open RevisePackage with url
```
nl0, line 212: this is some stub i guess

nl0 (approved):
looks ok mostly

catalog/app/containers/Bucket/Summary.js
fiskus: Yes, I'm going to uncomment it after https://github.com/quiltdata/quilt/pull/2957 merge

catalog/app/containers/AddToPackage/Provider.tsx
fiskus: You're right, thanks

catalog/app/containers/AddToPackage/Provider.tsx
fiskus: Makes sense, I'll introduce `clear` method called after Package's onSubmit

catalog/app/containers/AddToPackage/Provider.tsx
fiskus: I'll add in next PR, where I turn this feature on

catalog/app/containers/AddToPackage/Provider.tsx
fiskus: I'll add in the next PR too

PR #2957, opened 2022-07-28
Open Revise Package dialog on `?action=revisePackage`

fiskus:
# Description

# TODO

- [ ] Unit tests
- [ ] Automated tests (e.g. Preflight)
- [ ] Documentation
    - [ ] [Python: Run `build.py`](../gendocs/build.py) for new docstrings
    - [ ] JavaScript: basic explanation and screenshot of new features
- [ ] [Changelog](CHANGELOG.md) entry

PR #2960, opened 2022-07-29
Add README.md

fiskus:
This is a finale of
* https://github.com/quiltdata/quilt/pull/2953
* https://github.com/quiltdata/quilt/pull/2955
* https://github.com/quiltdata/quilt/pull/2957

Addressed feedback from prev PRs:
* instead of throwing an error on the dummy context function, I just provide null as the default value
* `clear` method
* actions.filter → actions.every

Screenshot from 2022-07-29 14-26-12
Screenshot from 2022-07-29 14-26-38
Screenshot from 2022-07-29 14-26-51


nl0 (dismissed):

fiskus:
I removed "Add Readme" button from nested directories inside the package. We can, potentially, add Readme there too, but it needs more development and thoughts

nl0 (approved):

PR #2962, opened 2022-07-29
Bucket/Meta: expose test ids and some other data for tests

nl0:


catalog/app/containers/Bucket/Meta.tsx
```diff
@@ -98,7 +98,11 @@ function PackageMetaSection({ meta, ...props }: PackageMetaProps) {
             </M.TableRow>
           )}
           {userMeta && (
-            <M.TableRow className={classes.row}>
+            <M.TableRow
+              className={classes.row}
+              data-key="user_meta"
+              data-value={JSON.stringify(userMeta)}
```
fiskus, line 104: Probably worth memoize this if `userMeta` is big

fiskus (approved):

catalog/app/containers/Bucket/Meta.tsx
nl0: maybe, but seems like premature optimization at the moment

PR #2967, opened 2022-08-04
Use colored diff when checking doc changes

sir-sigurd:


sir-sigurd:
Seems to not work because https://github.com/community/community/discussions/18492.

PR #2969, opened 2022-08-04
Overwrite AppBar default orange color

fiskus:
We use `navTheme.palette.seconday.dark` for color, but AppBar extracts `.main` (so, `<AppBar color="secondary"` doesn't help)
Sometimes, this `primary.main` orange flashes during first load (maybe even once after deploy)


fiskus:
Still not fixed. I think we need a more in-depth refactoring of layout

PR #2975, opened 2022-08-09
Athena refactor

fiskus:
# Description

Goals:
* reduce callback hell of components inside State and then inside Athena
* reduce dependencies of components each from the other, or at least make them more predictable

Main changes:
* selected workflow is stored in URL rather than in React state
* split one large State into local (inside components) pieces
* redirect to Execution page after submit with `history.push` rather than <Redirect />
* move more or less independent parts to modules under `/Athena`: Workgroups, QueryEditor (queries select was moved here), Results, History.

Fixed bugs:
* if query run is failed, then query editor is not blocked now. Earlier it was error message *instead* of query editor
* if there is idempotency error, then just call the same query without idempotency token

# TODO

- [ ] Unit tests
- [ ] Documentation


catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
```diff
@@ -1,458 +1,212 @@
...
+                  queryResults.next ? () => results.loadMore(queryResults) : undefined
+                }
+              />
+            )
+          }
+          if (queryResults.queryExecution) {
+            return (
+              <History
+                bucket={bucket}
+                executions={[queryResults.queryExecution!]}
```
nl0, line 168: `!` doesn't seem necessary here, bc `if (queryResults.queryExecution)` check asserts it's truthy

catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
```diff
@@ -1,458 +1,212 @@
...
-  children: (props: QueryResultsFetcherRenderProps) => React.ReactElement
-  queryExecutionId: string | null
+interface QueryResults {
+  data: requests.AsyncData<requests.athena.QueryResultsResponse>
+  loadMore: (prev: requests.athena.QueryResultsResponse) => void
 }
 
-function QueryResultsFetcher({ children, queryExecutionId }: QueryResultsFetcherProps) {
+function useQueryResults(queryExecutionId?: string): QueryResults {
   const [prev, usePrev] = React.useState<requests.athena.QueryResultsResponse | null>(
```
nl0, line 205:
```suggestion
  const [prev, setPrev] = React.useState<requests.athena.QueryResultsResponse | null>(
```

nl0 (dismissed):
i'm not very familiar with this code, but the changes look reasonable

nl0 (approved):

PR #2987, opened 2022-08-16
Markdownlint

fiskus:
Installed `markdownlint-cli` and set up `npm run lint:docs` CLI command to lint Markdown docs.
To make changes more manageable and incremental, I disabled lint on almost all files, except `docs/Catalog/Preferences.md` for a sample.


nl0 (changes requested):
looks kinda useful, but breaks some actions

docs/Catalog/Preferences.md
```diff
@@ -48,16 +48,21 @@ drag-and-drop or from folders in S3
...
+that map to an empty object reserved for future enhancements;
 buckets in this dictionary are the ones offered when the user clicks
-Revise Package > Add files from Bucket; if the dictionary is not set or is empty the feature "Add files from Bucket" is disabled
-* `ui.defaultSourceBucket` - source bucket from `ui.sourceBuckets` that is selected by default; if it doesn't match any bucket then it's ignored
-* `ui.package_description` - a dictionary that maps package handle regular expressions or literals to JSONPath expressions of fields to show from package metadata in the package list view.
+Revise Package > Add files from Bucket; if the dictionary is not set
+or is empty the feature "Add files from Bucket" is disabled
+* `ui.defaultSourceBucket` - source bucket from `ui.sourceBuckets`
+that is selected by default; if it doesn't match any bucket then it's ignored
+* `ui.package_description` - a dictionary
```
sir-sigurd, line 58: I think GitBook doesn't work well with multi-line list items: see 16aad503cffcc57d86b1145fb8df2ce897415773 and f152a237f85988713cfe6ade77c2ae93dd6a8783.
I'm not sure what useful markdownlint does besides that.

fiskus: Already existed multiline list handled https://docs.quiltdata.com/catalog/preferences#properties
Maybe it was an old issue, that was fixed in GitBook?

fiskus: linter is useful to not think about formatting anymore, and to not have unnecessary changes introduced by formatting

docs/api-reference/Bucket.md
```diff
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable -->
```
sir-sigurd, line 1: docs/api-reference/*.md are generated from Python sources. Maybe these files could be ignored through markdownlint's config?

catalog/package.json
```diff
@@ -210,6 +211,7 @@
     "jest": "^27.5.1",
     "jest-localstorage-mock": "^2.4.21",
     "lint-staged": "^12.4.1",
+    "markdownlint-cli": "^0.32.1",
```
sir-sigurd, line 214: Shouldn't we put it to the different place, since it lints docs for the whole project?

fiskus: I think we should, but don't know where. Probably, in `./testdocs` directory?

nl0: or you can just use `npx` and avoid having a `package.json` entirely

catalog/package-lock.json
```diff
@@ -21237,13 +21237,13 @@
       }
     },
     "node_modules/micromatch": {
-      "version": "4.0.4",
-      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.4.tgz",
-      "integrity": "sha512-pRmzw/XUcwXGpD9aI9q/0XOwLNygjETJ8y0ao0wdqprrzDa4YnxLcz7fQRZr8voh8V10kGhABbNcHVk5wHgWwg==",
+      "version": "4.0.5",
```
sir-sigurd, line 21240: Are changes in this file related to PR?

fiskus: No, reverted, thanks

sir-sigurd (dismissed):
I'd prefer to list ignored files in config, but that's up to you.

.github/workflows/js-ci.yml
```diff
@@ -30,3 +30,16 @@ jobs:
...
+      run:
+        working-directory: docs
+    env:
+      NODE_OPTIONS: --max-old-space-size=4096
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.11'
+      - run: npx --package=markdownlint-cli markdownlint --ignore node_modules **/*.md
```
nl0, line 45: i'd prefer to have this command accessible somewhere outside GH actions config, e.g. in a makefile or smth (even a standalone shell script would do), so it's easy to run it locally with the same options as in CI

nl0 (approved):

sir-sigurd:
@fiskus 
is there anything to be done here?

fiskus:
Yes. Something, according to your and Alexei feedback :)
I didn't decide what actually

drernie:
I think these two suggestions (from above) are worth considering, to make it easier for people to incrementally adopt:

* list ignored files in config
* have this command accessible somewhere outside GH actions config, e.g. in a makefile or smth (even a standalone shell script would do), so it's easy to run it locally with the same options as in CI

Are those easy to do?

fiskus:
Merged with master, resolved conflicts and fixed lint errors in new file

nl0 (approved):

.github/workflows/js-ci.yml
fiskus: Added task in Asana

fiskus:
I ignored files in-place rather than in single config from my experience. It was convenient for me to chop one file by another:
1. remove first line
2. exec linter and see warnings
3. fix warnings, commit and merge (single file)

PR #2989, opened 2022-08-16
status_reports lambda

nl0:
# Description

# TODO


nl0:
@sir-sigurd i'll probably adjust html layout and add some inline styling, but the rest of the lambda code seems complete

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
```diff
@@ -0,0 +1,296 @@
...
+
+AWS_REGION = os.getenv("AWS_REGION")
+assert AWS_REGION
+
+
+session = aiobotocore.session.get_session()
+
+
+def create_cfn():
+    return session.create_client("cloudformation", region_name=AWS_REGION)
```
sir-sigurd, line 22: I think it gets region from `AWS_DEFAULT_REGION` env var set by lambda runtime, so it doesn't make sense to pass region explicitly.

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
```diff
@@ -0,0 +1,296 @@
...
+async def drain(syn, method: str, key: str, names: T.List[str]) -> T.List[dict]:
+    chunks = [
+        names[i:i + CANARIES_PER_REQUEST]
+        for i in range(0, len(names), CANARIES_PER_REQUEST)
+    ]
+    pages = await asyncio.gather(*[
+        getattr(syn, method)(Names=chunk)
+        for chunk in chunks
+    ])
+    return list(itertools.chain(*[p[key] for p in pages]))
```
sir-sigurd, line 47: Consider this:
```suggestion
    return list(itertools.chain.from_iterable(p[key] for p in pages))
```

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
```diff
@@ -0,0 +1,296 @@
...
+            c.update(ok=state == "PASSED", lastRun=prev["Timeline"]["Completed"])
+
+    return canaries
+
+
+async def get_resources(cfn, stack_name: str) -> T.List[dict]:
+    result = []
+    async for page in cfn.get_paginator("list_stack_resources").paginate(StackName=stack_name):
+        for r in page["StackResourceSummaries"]:
+            result.append(r)
```
sir-sigurd, line 108:
```suggestion
        result.extend(page["StackResourceSummaries"])
```

lambdas/status_reports/requirements.txt
```diff
@@ -0,0 +1,3 @@
+Jinja2==3.1.2
+aiobotocore==2.3.4
+botocore==1.24.21
```
sir-sigurd, line 3: I'd put these ones to setup.py as dependencies and generate requirements.txt with [pip-compile](https://pip-tools.readthedocs.io/en/latest/).

nl0: what for?

sir-sigurd: To have indirect dependencies pinned.

nl0: ok i'll have a look

nl0: done (tho i used py3.8-based venv to compile requirements, bc it's a PITA to install 3.7 on my machine -- it requires x64 arch, i hope it'll work)

sir-sigurd: Well, it seems to work, but I think it really should be generated with the same Python version (and actually the same arch). 
It doesn't look like installing Python is really hard (https://diewland.medium.com/how-to-install-python-3-7-on-macbook-m1-87c5b0fcb3b5) or maybe you could you Docker. If you don't want to spend time on it, I could regenerate it by myself.

nl0: > If you don't want to spend time on it, I could regenerate it by myself.

well, yeah, you can, but going forward i think it's time we move to a newer runtime (3.9 or even 3.10 would be nice)

lambdas/status_reports/setup.py
```diff
@@ -0,0 +1,13 @@
...
+
+setup(
+    name="t4_lambda_status_reports",
+    version="0.0.1",
+    packages=find_packages(where="src"),
+    package_dir={"": "src"},
+    install_requires=[
+        "Jinja2 ==3.1.2",
+        "aiobotocore ==2.3.4",
+        "botocore ==1.24.21",
```
sir-sigurd, line 11: I think either of these are far more common:
```suggestion
        "Jinja2==3.1.2",
        "aiobotocore==2.3.4",
        "botocore==1.24.21",
```
```suggestion
        "Jinja2 == 3.1.2",
        "aiobotocore == 2.3.4",
        "botocore == 1.24.21",
```

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+def async_handler(f):
+    @functools.wraps(f)
+    def wrapper(event, context):
+        return asyncio.run(f(event, context))
+
+    return wrapper
+
+
+@async_handler
+async def lambda_handler(*_):
```
sir-sigurd, line 268: I'd use this for uniformity, but that's up to you.

```suggestion
async def lambda_handler(event, context):
```

sir-sigurd (commented):
What about tests? :slightly_smiling_face:

lambdas/status_reports/requirements.txt
sir-sigurd: Well I think we already discussed that we use 3.7 so we can easily move code between components. 3.10 is not even available as lambda runtime.

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+    return result
+
+
+async def get_stack_data(cfn, stack_name: str) -> dict:
+    resp = await cfn.describe_stacks(StackName=stack_name)
+    return resp["Stacks"][0]
+
+
+jenv = jinja2.Environment(autoescape=jinja2.select_autoescape())
+# TODO: styling
```
sir-sigurd, line 116: What about styling?

lambdas/status_reports/setup.py
nl0: ok, i just saw this style on some docs page and considered it "official" 🤷

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
nl0: i'll do it in a subsequent PR

lambdas/status_reports/setup.py
sir-sigurd: > ok, i just saw this style on some docs page and considered it "official" shrug

I think I've never seen this before :man_shrugging:.

nl0:
> What about tests? 🙂

well...

lambdas/status_reports/src/t4_lambda_status_reports/__init__.py
nl0: but these args are unused, so my local linter(ish) barfs

lambdas/status_reports/requirements.txt
nl0: > Well I think we already discussed that we use 3.7 so we can easily move code between components

i dont remember that discussion, but well, this makes *some* sense, tho idk about moving anything from lambdas to the client, for example, and getting stuck with some really old runtime is quite frustrating

sir-sigurd:
> > What about tests? slightly_smiling_face
> 
> well...

at least you could add a stub test and add it to CI so at least we can be sure it's installable

nl0:
> > > What about tests? slightly_smiling_face
> > 
> > 
> > well...
> 
> at least you could add a stub test and add it to CI so at least we can be sure it's installable

ok i'll add a stub

sir-sigurd (approved):

PR #2992, opened 2022-08-18
Remove idempotency token for Athena queries

fiskus:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation

### DIFF
--- a/catalog/app/containers/Bucket/Queries/requests/athena.ts
+++ b/catalog/app/containers/Bucket/Queries/requests/athena.ts
@@ -316,15 +316,6 @@ export function useQueryResults(
   )
 }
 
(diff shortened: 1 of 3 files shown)


nl0 (approved):
why do we need this change? is idempotency useless in this case?

fiskus:
Documentation is a little confusing: it says that two requests with the same token will return the same query. It doesn't say about results. So, probably it means two simultaneously executed queries will execute only one query if they share the same token. But it seems it's already done internally by SDK, and this token is not required (they articulate it in docs, that it's not required)
And preserving this token in code can lead to bugs, and it's just unnecessary to maintain something we don't need.
---
TLDR:
1. It's not required
2. It can produce errors and additional work to maintain

PR #2994, opened 2022-08-19
Get rid of pkg_resources

flying-sheep:
# Description

`pkg_resources` is a deprecated, extremely slow-to-import package that only exists when `setuptools` is installed.

Instead of adding a runtime dependency on `setuptools`, one should just use `importlib.metadata` and `importlib.resources` from the stdlib.

See here for performance benefits: https://github.com/mtkennerly/dunamai/issues/28

# TODO


api/python/quilt3/session.py
```diff
@@ -23,33 +26,33 @@
...
 AUTH_PATH = BASE_PATH / 'auth.json'
 CREDENTIALS_PATH = BASE_PATH / 'credentials.json'
-VERSION = pkg_resources.require('quilt3')[0].version
+VERSION = metadata.version('quilt3')
 
 
 def _load_auth():
     if AUTH_PATH.exists():
-        with open(AUTH_PATH, encoding='utf-8') as fd:
+        with AUTH_PATH.open(encoding='utf-8') as fd:
```
dimaryaz, line 34: Is this the preferred way? I can't find anything about which `open` is better.

In any case, if this way is better, could you do it in a separate PR, just cause it's not related to `pkg_resources`?

gendocs/build.py
```diff
@@ -34,10 +36,9 @@ def gen_walkthrough_doc():
 
 def install_pydocmd():
     try:
-        pydocmd_dist = pkg_resources.get_distribution('pydoc-markdown')  # install name, not module name
-        version = pydocmd_dist.version
-    except pkg_resources.DistributionNotFound:
-        version = ''
+        version = metadata.version('pydoc-markdown')  # install name, not module name
+    except metadata.PackageNotFoundError:
+        version = None
```
dimaryaz, line 41: This will break the `version.endswith` below. Change it back to `None`, or change the check?

dimaryaz (changes requested):
Thanks for the fix! A couple comments below...

gendocs/build.py
flying-sheep: done, check without whitespace changes: https://github.com/quiltdata/quilt/pull/2994/files?w=1

dimaryaz (approved):

docs/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ## Python API
 * [Fixed] Fix check to determine if a file is a tempfile in Windows with Python 3.8+ ([#2900](https://github.com/quiltdata/quilt/pull/2900))
 * [Changed] Disable upload optimization for objects with SSE-KMS ([#2790](https://github.com/quiltdata/quilt/pull/2790))
+* [Fixed] Sped up import and got rid of undeclared runtime dependency on `setuptools` ([#2994](https://github.com/quiltdata/quilt/pull/2994))
```
nl0, line 12:
```suggestion
* [Fixed] Speed up import and get rid of undeclared runtime dependency on `setuptools` ([#2994](https://github.com/quiltdata/quilt/pull/2994))
```

nl0 (approved):

sir-sigurd (approved):

sir-sigurd:
Thanks!

PR #3017, opened 2022-08-24
Create text file

fiskus:
* Refactored component for showing menu (`RevisionMenu`) and re-used it on bucket directory page
*  Added custom prompt component using `M.Dialog`
* Prompt for file name and redirect to file editor

Screenshot from 2022-08-25 20-01-17
Screenshot from 2022-08-25 20-01-48
Screenshot from 2022-08-25 20-02-03

# TODO

- [ ] Documentation (in the next clean PR, with Ernie help, probably)


catalog/app/components/Dialog/Prompt.tsx
```diff
@@ -0,0 +1,115 @@
...
+function Dialog({
+  initialValue,
+  open,
+  onCancel,
+  onSubmit,
+  title,
+  validate,
+}: DialogProps) {
+  const [value, setValue] = React.useState(initialValue || '')
+  const [submited, setSubmited] = React.useState(false)
```
nl0, line 24:
```suggestion
  const [submitted, setSubmitted] = React.useState(false)
```

catalog/app/components/Dialog/Prompt.tsx
```diff
@@ -0,0 +1,115 @@
...
+  validate: (value: string) => Error | undefined
+}
+
+export function usePrompt({ initialValue, title, onSubmit, validate }: PromptProps) {
+  const [key, setKey] = React.useState(0)
+  const [opened, setOpened] = React.useState(false)
+  const open = React.useCallback(() => setOpened(true), [])
+  const close = React.useCallback(() => {
+    setOpened(false)
+    setKey(R.inc)
```
nl0, line 82: this would probably cause flash of the dialog contents while it's in closing transition.
solution to this would be doing key change either after the transition is complete (which would complicate the code with additional state and callbacks) or before the dialog is shown (before `setOpened(true)`)

nl0 (approved):
looks good, some suggestions inline.
also, does it work for creating nested files (i.e. if i type smth like `prefix/file.md` into the prompt)?

catalog/app/components/Dialog/Prompt.tsx
fiskus: > before the dialog is shown

:thinking:

fiskus:
> also, does it work for creating nested files (i.e. if i type smth like prefix/file.md into the prompt)?

Yes, it works, but I don't want to advertise this "feature" yet. I want to come up with some nicer way of creating folders.

nl0 (approved):

PR #3052, opened 2022-08-29
Stretch Athena page to full widhth, remove vertical menu

fiskus:
* Open Athena queries in Queries tab
* Added transient ElasticSearch tab. It's now deprecated in favor of already existing search bar

Screenshot from 2022-08-29 15-29-02
Screenshot from 2022-08-29 15-29-09


catalog/app/containers/Bucket/BucketNav.tsx
```diff
@@ -79,6 +79,9 @@ function Tabs({ bucket, preferences, section = false }: TabsProps) {
       {section === 'search' && (
         <NavTab label="Search" value="search" to={urls.bucketSearch(bucket)} />
       )}
+      {(section === 'queries' || section === 'es') && (
```
nl0, line 82: this condition should be probably changed to `{preferences.queries && (`, since otherwise ES tab is only shown while it's already active (or while the queries tab is active)

nl0 (approved):

catalog/app/containers/Bucket/BucketNav.tsx
fiskus: RIght, thanks!

fiskus: Oh, I didn't get your point.

The goal was to show ES tab only when ES or Queries tab is active. Similar to Search tab

fiskus: I added `preferences.queries` condition but in addition to existing condition

nl0: well ok if that's intentional, i just i though it makes little sense showing extra ES tab after switching to Queries tab

fiskus: I'd better move ES link to somewhere on Queries tab, but didn't find a proper place for it.

Maybe, we can move it to dropdown help in search bar :thinking:

fiskus:
Merged it, because layout changes block further changes. We can move link to ElasticSearch page lsomewhere else ater

PR #3071, opened 2022-09-05
Markdownlint: ignore GitHub template

fiskus:


sir-sigurd:
Why do we want this?

fiskus:
Result remains the same: this file is ignored. But useless extra line will be disappeared when you create new PR

sir-sigurd (approved):

PR #3072, opened 2022-09-05
Hotfix: Athena localStorage key fix

fiskus:
Fixes incorrect use of `utils/storage` module


sir-sigurd (approved):

drernie:
Great. Will we need new checks to prevent this failure mode from happening again?

fiskus:
1. Always run your code, even if you refactor something trivial (it's obvious, and I always run code before requesting PR, but as you see - not always :) )
2. We need to convert this module to typescript

PR #3074, opened 2022-09-06
Footer revision hash

fiskus:
Screenshot from 2022-09-12 12-57-14


nl0 (dismissed):
the code looks ok, but im not sure about the design. i think this plain hash string may look kinda random / suspicious to some not very technical customer. maybe we can add some description to it and show several first characters of the hash, e.g. make it look like `Revision: 1d54afcb`, and remove the copy button, just copy the full hash on click

nl0:
and maybe use "product revision" language instead of "version" in tooltips and notifications

fiskus:
Agree

Changed layout

nl0 (approved):

PR #3075, opened 2022-09-06
Don't throw error when no access to workgroup

fiskus:


catalog/app/containers/Bucket/Queries/requests/athena.ts
```diff
@@ -103,12 +103,19 @@ async function fetchWorkgroup({
...
+    const workgroupOutput = await athena.getWorkGroup({ WorkGroup: workgroup }).promise()
+    if (
+      workgroupOutput?.WorkGroup?.Configuration?.ResultConfiguration?.OutputLocation &&
+      workgroupOutput?.WorkGroup?.State === 'ENABLED' &&
+      workgroupOutput?.WorkGroup?.Name
+    ) {
+      return workgroupOutput.WorkGroup.Name
+    }
+    return null
+  } catch (error) {
```
nl0, line 116: it would be nice to log this error at least, or better yet, catch only specific type of errors (check on `.code`?) and re-throw unexpected ones

nl0 (approved):

catalog/app/containers/Bucket/Queries/requests/athena.ts
fiskus: Check for HTTP `code` is a good one. I'll try to do it next. I want to merge it as is just to make staging work sooner

nl0: i meant AWS error code, but you get the idea

PR #3076, opened 2022-09-07
Fix "`name.replaceAll` is not a function" error

fiskus:
I don't know how to reproduce it, but I saw these errors in Sentry


nl0 (approved):

fiskus:
I think, this "fix" just hides the problem. And according to code there is no `row.id`
I can't reproduce this error on https://open.quiltdata.com/b/tcga-2-open/tree/68287677-25f3-4f95-9614-e8edde299812/MSK_0168-1_251469329153_S01_CGH-v4_91.txt

It would be good to not crash frontend with exception (thought I don't know if it was crashed because I couldn't reproduce), but this fix is useless for this.
Probably, it would be better to leave this line of code intact

PR #3088, opened 2022-09-14
Refine status reports design

nl0:
# TODO


nl0:
@sir-sigurd do i need to change anything for template files to be included into the package?

sir-sigurd:
> @sir-sigurd do i need to change anything (in setup.py?) for template files to be included into the package?

I think you need to add it to MANIFEST.in and enable `include_package_data`, see 
https://github.com/quiltdata/quilt/blob/b62757f1ed0f6ae8f6ddbd1250d6e4543914e833/api/python/MANIFEST.in#L1-L3
and 
https://github.com/quiltdata/quilt/blob/b62757f1ed0f6ae8f6ddbd1250d6e4543914e833/api/python/setup.py#L94
for example.

You should also try to deploy lambda with our deploy script because it tries to remove some useless stuff, but probably can remove useful stuff as well.

nl0:
> > @sir-sigurd do i need to change anything (in setup.py?) for template files to be included into the package?
> 
> I think you need to add it to MANIFEST.in and enable `include_package_data`, see
> 
> https://github.com/quiltdata/quilt/blob/b62757f1ed0f6ae8f6ddbd1250d6e4543914e833/api/python/MANIFEST.in#L1-L3
> 
> and
> 
> https://github.com/quiltdata/quilt/blob/b62757f1ed0f6ae8f6ddbd1250d6e4543914e833/api/python/setup.py#L94
> 
> for example.
> 
> You should also try to deploy lambda with our deploy script because it tries to remove some useless stuff, but probably can remove useful stuff as well.

that worked, thanks

nl0:
@sir-sigurd pls have a look

sir-sigurd (approved):

PR #3089, opened 2022-09-14
Documentation faqs

robnewman:
# Description
Add overview of using Quilt with R

# TODO


docs/FAQ.md
```diff
@@ -64,3 +64,31 @@ a local machine or foreign region)—I/O is much faster.
...
+
+### Using the Quilt CLI API with R
+You can script the Quilt CLI directly from your shell environment and chain it
+with your R scripts to create a unified workflow:
+
+```bash
+quilt3 install
+[Run R commands or scripts]
+quilt3 push
+```
```
drernie, line 90: I find it hard to understand follow what  is happening without context. Perhaps:
```
$ quilt3 install # download package
[Run R commands or scripts] # modify or add files
$ quilt3 push # upload package
```
Unless there's a different convention that works better?

drernie (approved):

docs/FAQ.md
robnewman: I'm not sure you can isolate this to just a CLI (i.e. the $ CLI prefix). You can have a bash script that does all this in one executable

robnewman: I followed the convention used earlier in the page

robnewman: I've added more context and resubmitted

drernie (dismissed):

robnewman:
@drernie Need another review please

drernie (dismissed):

drernie:
FYI, in an ideal world I would want that bash script to be an actual runnable code example, so we could test that it was valid.  However, that requires adding write-capable credentials to our CI environment, which is not worth burdening this PR with. I'll create a new Asana for that...

robnewman:
@drernie Unfortunately adding that failed (perhaps because I didn't add it to the first line) so I added just a `<!--pytest.mark.skip-->` for the code block in question. Hopefully that resolves the issue

drernie (approved):
Ah, sorry. Looks good now.

PR #3097, opened 2022-09-19
Add object deletion faq

robnewman:
# Description
Add object deletion example with python snippet

# TODO

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -120,3 +120,6 @@ for (k, e) in p.walk():
     pk = e.physical_key
     s3.delete_object(Bucket=pk.bucket, Key=pk.path, VersionId=pk.version_id)
 ```
+
+You can then follow the above with `q3.Package.delete(pname, registry=reg, top_hash=p.top_hash)`.
+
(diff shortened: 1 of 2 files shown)


docs/FAQ.md
```diff
@@ -93,3 +93,33 @@ quilt3 push --dir path/to/remote-registry my-package # upload Quilt data package
 ### Using Quilt with Reticulate
 The [Reticulate](https://rstudio.github.io/reticulate/) package provides a set of tools
 for interoperability between Python and R by embedding a Python session within your R session.
+
+## How do I delete all of the objects in a data package?
+
+*Very carefully!* In favor of immutability Quilt makes deletion a
+bit tricky. First, note that `quilt3.Package.delete` only deletes the
+_package manifest_, not the *underlying objects*. If you wish to delete
+a package revision *and* its objects, _delete the objects first_.
```
drernie, line 102: I'm confused. Is there a difference between "deleting a manifest" and "deleting a revision"?
If so, what? And which is shown in the example below?

drernie (changes requested):

docs/FAQ.md
drernie: Okay, then may I suggest replacing:
> delete a package revision
with
> delete the entire package

robnewman: @drernie Good feedback. Incorporated. Thanks!

drernie (approved):

PR #3098, opened 2022-09-20
Docs: incorrect parsing of definitions list

fiskus:
I think, parser treat `s3:` as definition term here https://docs.quiltdata.com/api-reference/package#package.push
Solution: I replaced colon to a visually similar character (https://en.wikipedia.org/wiki/Colon_(letter)).

- [x] Documentation


robnewman:
@fiskus This conflicts with a very recent change made by @drernie: https://github.com/quiltdata/quilt/pull/3424

Are you both aligned on what needs fixing and why?

fiskus:
I guess, braces do the trick in Ernie's PR

PR #3114, opened 2022-09-29
Suggest creating a stub readme file when pushing an empty package

nl0:
# Description

# TODO


catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
```diff
@@ -62,6 +66,66 @@ export interface PackageCreationSuccess {
...
+    `# ${name}`,
+    '\n\n',
+    `Stub README for the **${name}** package generated by Quilt Catalog`,
+    '\n',
+  ]
+  const f = new File(contents, README_PATH, { type: 'text/markdown' })
+  return FI.computeHash(f) as FI.LocalFile
+}
+
+// XXX: move to dialogs module
```
fiskus, line 80: Would be good if you mean this module app/components/Dialog/Confirm.tsx

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
```diff
@@ -62,6 +66,66 @@ export interface PackageCreationSuccess {
...
+interface DialogsOpenProps {
+  close: (reason?: string) => void
+}
+
+function ConfirmReadme({ close }: DialogsOpenProps) {
+  return (
+    <>
+      <M.DialogTitle>Add a README file?</M.DialogTitle>
+      <M.DialogContent dividers>
```
fiskus, line 89: I think,  it's nicer without dividers

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
```diff
@@ -235,6 +301,18 @@ function PackageCreationForm({
       })),
     )
 
+    if (!entries.length) {
```
fiskus, line 304: I think, it makes more sense to check if package has README rather than if package has no files

fiskus (commented):
I think there are two mistakes:
1. now users can push empty packages
2. confirmation Dialog shows when no files, but should be shown when no README

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
```diff
@@ -62,6 +66,66 @@ export interface PackageCreationSuccess {
...
+          You are about to push an empty package.
+          <br />
+          Would you like to add a stub <b>README.md</b> file?
+        </M.DialogContentText>
+      </M.DialogContent>
+      <M.DialogActions>
+        <M.Button onClick={() => close('cancel')} color="primary">
+          Cancel
+        </M.Button>
+        <M.Button onClick={() => close('empty')} color="primary" variant="outlined">
```
fiskus, line 100: So, as I see this feature, this third option is redundant:

Add a README file?
* Yes -> create empty (or not empty) package with empty README
* No/Cancel -> continue creating package

nl0:
> I think there are two mistakes:

both these things are intentional.

>     1. now users can push empty packages

yeah, and i believe it has its use cases (like, metadata-only packages, or stub packages to add files to later)

>     2. confirmation Dialog shows when no files, but should be shown when no README

i assumed that when you push a package with some files, you know what you're doing and you don't want that readme file (you can add it later via catalog or the client)

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
nl0: i mean `app/utils/Dialogs.js`

fiskus (approved):

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
nl0: in its current state it makes sense to keep this option

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
nl0: addressed in another comment

nl0:
we can revisit this feature later based on feedback from the leadership and the team (and maybe even customers?)

fiskus (approved):

PR #3115, opened 2022-09-29
Fix Vega reloading

fiskus:
The main problem was with two unmemoized objects. Handle is used as dependency for signing URLs in JSON for Vega.
```
const withPreview = ({ handle }) => {
  const previewHandle = { ...handle, packageHandle }
  return Preview.load(previewHandle, ...)
}
return fileQuery.case({
  data: () => {
    const handle: PackageEntryHandle = {
      ...s3paths.parseS3Url(file.physicalKey),
      logicalKey: file.path,
    }
    return <Section>{withPreview({ handle })}</Section>
})
```

* Convert `renderProgresss` and `renderError` functions to components
* Create wrapper component `FileDisplayQuery` to move `fileQuery` in it
* Memoize objects described above


nl0 (approved):
i haven't digested the whole diff, but the description of the changes looks reasonable

PR #3120, opened 2022-10-03
adjust xlsx import

nl0:
webpack dev server started giving me some warnings after xlsx update


fiskus (approved):
I checked that drag'n'drop .xlsx file works

nl0:
> I checked that drag'n'drop .xlsx file works

thanks for testing

PR #3121, opened 2022-10-04
Quilt config editor

fiskus:
* Make `utils/yaml.ts` API similar to JSON: YAML.stringify, YAML.parse
* Move config paths to `constants/quiltConfigs.ts`
* Add additional internal type for FileEditor: "__quiltConfig" (it's temporary, I'll find better solution)
* Use JsonEditor for editing quilt configs: config.yaml is converted to JSON + add JSON Schema for validation and showing placeholders
* Minor JsonEditor improvements

Screenshot from 2022-10-05 14-41-06


catalog/app/utils/yaml.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+// eslint-disable-next-line consistent-return
+export function parse(inputStr?: string) {
+  if (!inputStr) return undefined
+  try {
+    return yaml.load(inputStr)
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error)
```
nl0, line 16: shouldnt we re-throw the error here?

nl0 (approved):

nl0:
@fiskus is there still an option to edit these files as plain text? bc i'd very much prefer that option

fiskus:
You can edit them locally:
```
aws s3 cp s3://config.yaml .
vi config.yaml
aws s3 cp config.yaml s3://config.yaml
```

But I agree, we need to add a way to use text editor in UI too. I'll add it later, after some feedback. Maybe, we'll throw this solution at all

catalog/app/utils/yaml.ts
fiskus: You're right, we should. At least, for matching with `JSON.parse`
But current design doesn't expect this.

I'll refactor this later

fiskus:
I think, this UI is not very intuitive and can be hard to use, potentially.
Let's see in the wild, is it better to remove it (it's easy to do) or improve

PR #3122, opened 2022-10-04
Hide actions buttons (push, revise etc.) on open.quiltdata.com

fiskus:
Package creation on open.quiltdata.com doesn't work. It's implied users push packages using `quilt3`


nl0 (approved):

fiskus:
We decided to use GraphQL to send permission level for bucket, and show/hide controls according to that level

PR #3124, opened 2022-10-06
Forbid to save file if it's not on latest revision

fiskus:


catalog/app/containers/Bucket/File.js
```diff
@@ -382,15 +382,20 @@ export default function File({
...
+      downloadable: false,
+    }),
+    Ok: requests.ObjectExistence.case({
+      _: () => ({
+        downloadable: false,
       }),
-    })
+      Exists: ({ deleted, archived, version: versionHash }) => ({
+        downloadable: !noDownload && !deleted && !archived,
+        hash: versionHash,
```
nl0, line 395: i wouldnt call this hash -- it's a version id, it's not linked to the contents of the file and isn't related to any hash functions

nl0 (approved):

catalog/app/containers/Bucket/File.js
fiskus: Ok, got it.
`hash` for package version is valid name, because we use hash function
`hash` for file version is invalid name, because it's some S3 ID

fiskus: I'm renaming this `hash` to `fileVersionId`

fiskus:
@nl0 I renamed `hash` to `fileVersionId`. Please, approve once again if you don't mind

nl0 (approved):

PR #3128, opened 2022-10-06
Use value if array has single value

fiskus:
Added the way we parse spreadsheet values:
* if there are multiple values - it's an array for sure
* if there is single value, then check user wants it as array using Schema
* otherwise (single value, and/or schema doesn't force as array) - it's a single value

Changed and added unit-tests


nl0 (dismissed):

nl0:
@fiskus tests are red tho

fiskus:
Good! I didn't notice it. It means it's definitely not a regression :)

nl0 (approved):

PR #3131, opened 2022-10-11
Disable creating files in packages alognside with disabling revising packages

fiskus:


catalog/app/containers/Bucket/Summary.js
```diff
@@ -261,9 +263,12 @@ export default function BucketSummary({ files, mkUrl: mkUrlProp, packageHandle,
           mkUrl={mkUrl}
         />
       )}
-      {!readme && !path && !!packageHandle && (
-        <AddReadmeSection packageHandle={packageHandle} />
-      )}
+      {!readme &&
+        !path &&
+        !!packageHandle &&
+        !!preferences?.ui?.actions?.revisePackage && (
```
nl0, line 269: i'm not sure this check is required here, bc it's a summary for a bucket, not a package, or am i missing something?

fiskus: This is specific to packages.

Bucket/Summary.js is used for Bucket/Dir and for PackageTree, but `!path` hides button for nested directories

And, on Overview page this component is not used

nl0: ok makes sense

nl0 (approved):

PR #3133, opened 2022-10-11
Reload file preview after file creation

fiskus:


nl0 (approved):

fiskus:
Something doesn't work for versioned bucket

fiskus:
@nl0 the same fix now works for both versioned and unversioned buckets

nl0 (approved):

nl0:
@fiskus merging in?

fiskus:
Yes, I didn't get notification over email

PR #3140, opened 2022-10-13
Switch to platformdirs

flying-sheep:
# Description
See https://github.com/platformdirs/platformdirs#why-this-fork:

> ## Why this Fork?
> This repository is a friendly fork of the wonderful work started by [ActiveState](https://github.com/ActiveState/appdirs) who created appdirs, this package's ancestor.
> 
> Maintaining an open source project is no easy task, particularly from within an organization, and the Python community is indebted to appdirs (and to Trent Mick and Jeff Rouse in particular) for creating an incredibly useful simple module, as evidenced by the wide number of users it has attracted over the years.
> 
> Nonetheless, given the number of long-standing open issues and pull requests, and no clear path towards https://github.com/ActiveState/appdirs/issues/79, this fork was created.
> 
> Contributions are most welcome.

# TODO


api/python/setup.py
```diff
@@ -56,7 +56,7 @@ def run(self):
     url='https://github.com/quiltdata/quilt',
     keywords='',
     install_requires=[
-        'appdirs>=1.4.0',
+        'platformdirs>=2.5.2',
```
sir-sigurd, line 59: Makes sense?
 ```suggestion
        'platformdirs>=2,<3',
```

File: api/python/tests/conftest.py [hunk 2]
```diff
@@ -22,23 +23,26 @@ def pytest_sessionstart(session):
...
 
+    def get_dir(*args, d):
+        return str(d / args[0] if args else d)
+
     # Mockers that need to be loaded before any of our code
     Vars.extrasession_mockers.extend([
-        mock.patch('appdirs.user_data_dir', lambda *x: str(Vars.tmpdir_data / x[0] if x else Vars.tmpdir_data)),
-        mock.patch('appdirs.user_cache_dir', lambda *x: str(Vars.tmpdir_cache / x[0] if x else Vars.tmpdir_cache)),
+        mock.patch('platformdirs.user_data_dir', partial(get_dir, d=Vars.tmpdir_data)),
+        mock.patch('platformdirs.user_cache_dir', partial(get_dir, d=Vars.tmpdir_cache)),
```
sir-sigurd, line 45: This change seems unrelated, so I'd revert it.

api/python/setup.py
flying-sheep: I don’t think so. If you’d add a comment here that explains why you add an upper limit, maybe, but you shouldn’t do it without a reason: https://iscinumpy.dev/post/bound-version-constraints/

api/python/tests/conftest.py
flying-sheep: I did it because the slightly longer name caused black to do an ugly line break.

api/python/setup.py
sir-sigurd: OK, I don't insist on setting upper limit, but any reason for setting lower limit to the newest version? Doesn't 'platformdirs==2.0.0' really works for us?

flying-sheep: Definitely not, you don’t want to pin any dependencies in a library.

You might want to read the “Python is not JavaScript” section of the linked article.

nl0: "`>=` everywhere" doesn't seem safe tbh, bc it means potentially accepting some version with breaking changes in the future, especially considering that python community starts to embrace SemVer (well, yeah, somewhat).
but i think what @sir-sigurd meant in this particular case is actually `platformdirs>=2.0.0` (more relaxed lower bound) vs having a lower bound of `2.5.2`

flying-sheep: `platformdirs>=2` would work, yes. I often set pins to the current version because

1. I’m not testing with lower versions, so I don’t know if I’m relying on a bug fix or a feature of the current version, and can therefore only say that that one works. If someone files an issue in my project wanting to use some lower version for some reason I happily lower it. Another reason is that
2. Some libraries *do* pin dependencies or specify upper bounds (mostly because they don’t know better, sometimes for good reasons). In that case `pip` can sometimes end up veeery slooowly downloading a bunch of versions of many packages to maybe find a compatible combination. Setting your lower bounds relatively high means the space it searches is smaller resulting in a quicker resolution (or notification of resolution failing) when that happens.

I saw that you’re doing some pinning in your dependencies. We should fix that so you don’t cause the problem I described for people. See also https://iscinumpy.dev/post/bound-version-constraints/#python-is-not-javascript

flying-sheep: > ">= everywhere" doesn't seem safe tbh, bc it means potentially accepting some version with breaking changes in the future

Please read the blog post if you think that. Henry and I know the Python packaging landscape quite well, and this is an informed decision that takes your concerns into account.

I know it’s long, but I’m sure you can Ctrl-F your way to the parts that address your questions.

sir-sigurd (approved):

nl0 (approved):

dimaryaz (approved):

sir-sigurd:
@flying-sheep, thanks!

PR #3144, opened 2022-10-14
Handle `additionalProperties` and `items` in JSONEditor

fiskus:
One of the main source of data for JSONEditor is dictionary `{ [JSON Pointer]: data }`, where `JSON Pointer` is a pointer to every object value or JSON Schema type, and `data` is everything I need about Schema and object value for this Pointer.
Before this PR I stored only explicitly described properties. Here, I introduced placeholder `__*` that substitutes `additionalProperties` or `items`.

So, for JSON Schema

```js
const schema = {
  type: 'object',
  properties: {
    a: { anyOf: [{ type: 'number', minimum: 1024 }, { type: 'string' }] },
    b: { type: 'number' },
    c: {
      type: 'object',
      additionalProperties: {
        type: 'object',
        properties: {
          a: { anyOf: [{ type: 'number', minimum: 1024 }, { type: 'string' }] },
          b: {
            type: 'array',
            items: {
              type: 'number',
            },
          },
        },
      },
    },
  },
}
```

I create such dictionary:
```js
const dict = {         
  '/a': ...,           
  '/b': ...,           
  '/c': ...,           
  '/c/__*': ...,       
  '/c/__*/a': ...,     
  '/c/__*/b': ...,     
  '/c/__*/b/__*': ..., 
}                      
```

Screenshot from 2022-10-14 22-12-49

- [x] Javascript unit-tests


catalog/app/components/JsonEditor/State.js
```diff
@@ -142,9 +165,38 @@ function collectErrors(allErrors, itemAddress, value) {
...
+  let weight = 0
+  let placeholderItem = undefined
+  Object.entries(jsonDict).forEach(([path, value]) => {
+    if (doesPlaceholderPathMatch(deserializeAddress(path), addressPath)) {
+      if (weight < addressPath.length) {
+        weight = addressPath.length
+        placeholderItem = value
+      }
+    }
+  }, {})
```
nl0, line 193: the last argument (`thisArgs`) doesn't seem necessary here

nl0 (approved):

catalog/app/components/JsonEditor/State.js
fiskus: It's a leftover `.reduce` instead of `.forEach` implementation, thanks

nl0 (approved):

PR #3145, opened 2022-10-14
S3 proxy: Use path-style requests for buckets containing dots

dimaryaz:
Virtual host style does not support dots because sub-subdomains are not allowed by wildcard SSL certificates. The AWS console uses path-style in those cases, so let's do the same thing in the S3 proxy.


akarve (approved):
Thanks. I googled NGINX unit tests and apparently they're written in Perl. So there's that.

PR #3147, opened 2022-10-18
Catalog: AWS/S3: support dots in bucket names while proxying

nl0:
# TODO

### DIFF
--- a/catalog/app/utils/AWS/S3.js
+++ b/catalog/app/utils/AWS/S3.js
@@ -90,6 +90,14 @@ function useSmartS3() {
               path: req.httpRequest.path,
             }
             const basePath = endpoint.path.replace(/\/$/, '')
+            // handle buckets with dots in their names
+            if (
+              req.httpRequest.path.startsWith(`/${b}`) &&
+              !req.httpRequest.endpoint.host.startsWith(`${b}.`)
+            ) {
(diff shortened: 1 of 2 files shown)


catalog/app/utils/AWS/S3.js
```diff
@@ -90,6 +90,14 @@ function useSmartS3() {
               path: req.httpRequest.path,
             }
             const basePath = endpoint.path.replace(/\/$/, '')
+            // handle buckets with dots in their names
+            if (
+              req.httpRequest.path.startsWith(`/${b}`) &&
+              !req.httpRequest.endpoint.host.startsWith(`${b}.`)
+            ) {
+              req.httpRequest.path = req.httpRequest.path.replace(`/${b}`, '')
```
fiskus, line 98: Just a note, that `req.httpRequest[PROXIED].path` stays unchanged. It's by design, probably

And, as I see, on "retry" `req.httpRequest.path` will be reset to original value, without workaround for dots

fiskus (approved):

catalog/app/utils/AWS/S3.js
nl0: > Just a note, that req.httpRequest[PROXIED].path stays unchanged. It's by design, probably

yeah, it is -- we're keeping what comes from the sdk as it is

PR #3148, opened 2022-10-18
S3 Proxy: fix support for buckets with dots

nl0:
Add slash after the bucket name conditionally (only when there's more path segments after it), so that the operations on the bucket itself (head bucket, list objects in the root) don't lead to signature mismatch errors.

### DIFF
--- a/s3-proxy/nginx.conf
+++ b/s3-proxy/nginx.conf
@@ -60,13 +60,29 @@ http {
             set $s3_path_prefix '';
             if ($s3_bucket ~ "\.") {
                 set $s3_host '$s3_host_suffix';
-                set $s3_path_prefix '$s3_bucket/';
+                set $s3_path_prefix '$s3_bucket';
+            }
+
+            set $s3_path_suffix '';
+            # Use $request_uri rather than $s3_path because it needs to stay encoded.
+            if ($request_uri ~ "^/[^/?]+/[^/?]+/?(.*)") {
+                set $s3_path_suffix "$1";
+            }
+
+            set $slash_needed '';
+            if ($s3_path_prefix) {
+                set $slash_needed '${slash_needed}1';
+            }
+            if ($s3_path_suffix ~ '^[^?]+') {
+                set $slash_needed '${slash_needed}1';
+            }
+            if ($slash_needed = '11') {
+                set $s3_path_prefix '$s3_path_prefix/';
             }
 
             # Proxy the request to S3.
-            # Use $request_uri rather than $s3_path because it needs to stay encoded.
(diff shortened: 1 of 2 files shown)


nl0:
> I haven't gotten a chance to test the signatures myself, but can you try if this logic works? I think it's slightly less hacky than "11":

@dimaryaz i've tried and it works, thanks! i've also added some changes to support empty path suffix (needed e.g. for head bucket requests)

s3-proxy/nginx.conf
```diff
@@ -58,17 +58,29 @@ http {
...
-            # Proxy the request to S3.
+            set $s3_path_suffix '';
             # Use $request_uri rather than $s3_path because it needs to stay encoded.
             if ($request_uri ~ "^/[^/?]+/[^/?]+/?(.*)") {
-                proxy_pass 'https://$s3_host/$s3_path_prefix$1';
+                set $s3_path_suffix "$1";
+            }
+
+            # Don't add a slash if there's no more path segments
+            if ($s3_path_suffix !~ '^[^?]+') {
```
dimaryaz, line 75: Don't need the `+`, right?

dimaryaz (approved):

dimaryaz:
(Sorry, looks like my first "+" comment didn't submitted, and then it ended up with two.)

s3-proxy/nginx.conf
nl0: yeah, seems so

nl0:
```suggestion
            if ($s3_path_suffix !~ '^[^?]') {
```

s3-proxy/nginx.conf
```diff
@@ -58,17 +58,29 @@ http {
...
+                set $s3_path_prefix '$s3_bucket';
+                set $s3_path_sep '/';
             }
 
-            # Proxy the request to S3.
+            set $s3_path_suffix '';
             # Use $request_uri rather than $s3_path because it needs to stay encoded.
             if ($request_uri ~ "^/[^/?]+/[^/?]+/?(.*)") {
-                proxy_pass 'https://$s3_host/$s3_path_prefix$1';
+                set $s3_path_suffix "$1";
```
dimaryaz, line 71: I've run into this before... It's likely an Nginx bug.

This will keep URL encoding:
```
  proxy_pass 'https://$s3_host/$s3_path_prefix$s3_path_sep$1;
```

But this will break it:

```
  set $s3_path_suffix "$1";
  proxy_pass 'https://$s3_host/$s3_path_prefix$s3_path_sep$s3_path_suffix';
```

Even though they are supposed to be exactly the same thing.

Will need to work around it somehow. Worst case, have two `if`s, for requests with and without path segments.

PR #3155, opened 2022-10-20
Fix URL encoding in s3 proxy

dimaryaz:
Using "set" causes the URL to be decoded - likely an Nginx bug. We have to use "$1" immediately, without storing it in a variable.

We already have an "s3_path" variable we can use for the "/" logic, though, so that makes it simpler.


nl0 (approved):
ok, this is working. thanks!

PR #3158, opened 2022-10-25
Extend JsonEditor with additional controls and forms and add workflows in particular

fiskus:
* Added React context provider where we can set elements for using inside JSONEditor Column
* Added WorkflowsToolbar in particular, that can create additional fields for new workflow using form
* Added components/Form, that can be used for elements for react-final-form. We already use more or less similar but not identical fields in Admin forms and PackageCreate forms, it's easy to unify them

Screenshot from 2022-10-26 11-56-49
Screenshot from 2022-10-26 11-56-59
Screenshot from 2022-10-26 11-57-29


catalog/app/components/FileEditor/QuiltConfigEditor/WorkflowsToolbar.tsx
```diff
@@ -0,0 +1,320 @@
...
+      setOpen(false)
+    },
+    [onChange],
+  )
+
+  return (
+    <>
+      <Button onClick={() => setOpen(true)} />
+      <Popup
+        bucket={bucket || 'fiskus-sandbox-dev'}
```
nl0, line 246: are you hardcoding `fiskus-sandbox-dev` bucket intentionally?

catalog/app/components/Form/Field.tsx
```diff
@@ -0,0 +1,40 @@
...
+})
+
+interface FieldProps {
+  error?: string
+  helperText?: React.ReactNode
+  input: RF.FieldInputProps<string>
+  validating?: boolean
+}
+
+export default function Field({
```
nl0, line 19: "Field" seems too generic imo, especially considering that in fact it's a TextField under the hood

nl0 (approved):

catalog/app/components/FileEditor/QuiltConfigEditor/WorkflowsToolbar.tsx
fiskus: Unintentional, while I was debugging in `/storybook` where is no `bucket`

catalog/app/components/Form/Field.tsx
fiskus: Renamed to `TextField`

nl0 (approved):

PR #3165, opened 2022-10-31
Adopt `loglevel` lib

nl0:


fiskus (approved):
What is a plan to use it?

I really want to replace all console.log's with it as I proposed here https://github.com/quiltdata/quilt/pull/1882

nl0:
> What is a plan to use it?

just use it everywhere you need any logging done =)

> I really want to replace all console.log's with it as I proposed here #1882

yeah that's the point

fiskus:
I usually use https://www.npmjs.com/package/winston btw

nl0:
> I usually use https://www.npmjs.com/package/winston btw

looks quite cool, but seems like an overkill for the catalog (tho would fit nicely into a node server if we ever need one)

PR #3166, opened 2022-10-31
Synchronous catalog config from a global var + types

nl0:
In order to simplify access to the config data throughout the app (to a simple `import cfg from 'utils/Config'`) and setup of some tools (such as sentry),
I've decided to refactor the config system to use a synchronous JS file which sets a global variable that later can be accessed by the app code.

This requires a change in deployment which will come soon (later today).

**UPD**: Actually, deployment change is only relevant for OPEN, so we can merge this PR without waiting for that change.

In order for local catalog instance to continue functioning, one should rename `config.json` file in `static-dev` folder to `config.js` and prepend the contents with `window.QUILT_CATALOG_CONFIG = `

## TODO

- [x] changelog


catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,116 @@
...
+  ssoProviders: cfg.ssoProviders.length ? cfg.ssoProviders.split(' ') : [],
+  enableMarketingPages: cfg.mode === 'PRODUCT' || cfg.mode === 'MARKETING',
+  disableNavigator: cfg.mode === 'MARKETING',
+  s3Proxy: startWithOrigin(cfg.s3Proxy),
+  apiGatewayEndpoint: startWithOrigin(cfg.apiGatewayEndpoint),
+  binaryApiGatewayEndpoint: startWithOrigin(cfg.binaryApiGatewayEndpoint),
+  noDownload: !!cfg.noDownload,
+  noOverviewImages: !!cfg.noOverviewImages,
+  // XXX: there's no such field in the schema, so it will be stripped from the config when present
+  desktop: !!(cfg as any).desktop,
```
nl0, line 85: @fiskus do we need this? if yes, we can add an untyped optional `desktop` field to the config schema and you can store anything there

catalog/test.js
```diff
@@ -1 +0,0 @@
-console.log(a === undefined)
```
nl0, line 1: i'm not sure where this came from 🤔 
@fiskus any idea?

nl0:
Actually, deployment change is only relevant for OPEN, so we can merge this PR without waiting for that change.

catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const ajv = new Ajv({ allErrors: true, removeAdditional: true })
+
+ajv.addSchema(configSchema, 'Config')
+
+export class ConfigError extends BaseError {
+  static displayName = 'ConfigError'
+}
+
+const pprint = (x: unknown) => JSON.stringify(x, null, 2)
```
fiskus, line 49: There is an printObject in utils/string

catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,115 @@
...
+
+function validateConfig(input: unknown): asserts input is ConfigJson {
+  if (ajv.validate('Config', input)) return
+  const printErrors = `Errors array:\n${pprint(ajv.errors)}`
+  const printInput = `Input:\n${pprint(input)}`
+  throw new ConfigError(
+    `invalid config format:\n${ajv.errorsText()}\n${printErrors}\n${printInput}`,
```
fiskus, line 56: I think you can use visible line endings instead, if you use "``"

fiskus, line 56:
```suggestion
    dedent`invalid config format:
      ${ajv.errorsText()}
      ${printErrors}
      ${printInput}`,
```

catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,115 @@
...
+  ssoAuth: AUTH_MAP[cfg.ssoAuth],
+  ssoProviders: cfg.ssoProviders.length ? cfg.ssoProviders.split(' ') : [],
+  enableMarketingPages: cfg.mode === 'PRODUCT' || cfg.mode === 'MARKETING',
+  disableNavigator: cfg.mode === 'MARKETING',
+  s3Proxy: startWithOrigin(cfg.s3Proxy),
+  apiGatewayEndpoint: startWithOrigin(cfg.apiGatewayEndpoint),
+  binaryApiGatewayEndpoint: startWithOrigin(cfg.binaryApiGatewayEndpoint),
+  noDownload: !!cfg.noDownload,
+  noOverviewImages: !!cfg.noOverviewImages,
+  // XXX: there's no such field in the schema, so it will be stripped from the config when present
```
fiskus, line 84: Can you add it to Schema, so this PR won't be a breaking change for desktop?

catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,115 @@
...
+  } catch (e) {
+    if (e instanceof ConfigError) throw e
+    throw new ConfigError('unexpected error', { originalError: e })
+  }
+}
+
+export type Config = ReturnType<typeof getConfig>
+
+const configKey = 'QUILT_CATALOG_CONFIG'
+const rawConfig = (window as any)[configKey]
```
fiskus, line 101: Maybe, it's worth to describe `window` at `app/@types/window.d.ts`

catalog/test.js
fiskus: Probably, I tested eslint or husky, and missed that it was commited

fiskus (approved):

catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,115 @@
...
+invariant(rawConfig, `window.${configKey} must be defined`)
+const CONFIG = getConfig(rawConfig)
+
+export default CONFIG
+
+export function useConfig(opts: { suspend: false }): { promise: Promise<Config> }
+export function useConfig(): Config
+/** @deprecated Config is now synchronous -- just import Config module directly */
```
fiskus, line 109: I would prefer to use getter function to get synchronous config

fiskus: It's quite important to easier refactoring

fiskus: And extending

catalog/Dockerfile
```diff
@@ -11,6 +11,8 @@ COPY build /usr/share/nginx/html
 # Copy config file
 COPY config.json.tmpl config.json.tmpl
 
-# Substitute environment variables into config.js before starting nginx.
+# Substitute environment variables into config.json and generate config.js based on that before starting nginx.
+CMD envsubst < config.json.tmpl > /usr/share/nginx/html/config.json
+CMD echo "window.QUILT_CATALOG_CONFIG = `cat /usr/share/nginx/html/config.json`" > /usr/share/nginx/html/config.js
```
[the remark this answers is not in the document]
dimaryaz: Just do `RUN cmd1 && cmd2 && cmd3` to run multiple commands.

If it doesn't work or gets annoying, you can create a script like `run_nging.sh` that does all the work. (I was trying to use `$(<...)` instead of `cat` - but either it's not supported by `/bin/sh`, or somehow `$` doesn't work - but it might be better to move everything into a helper script.)

catalog/app/utils/Config.ts
nl0: well, window has its type defined in typescript lib, and i'm not sure it's worth it ATM to explicitly add our "extensions"

catalog/app/utils/Config.ts
nl0: well, `useConfig` is a synchronous getter function, but it's a deprecated API anyways, bc doing `import cfg from 'utils/Config'` is more ergonomic

fiskus:
Sorry, I wasn't clear about getter

I thought about something like this:
```ts
// utilsConfig.ts

const config = readAndValidateConfigOnStart()
export default function getConfig() {
  return config
}
```

```ts
// some-module.ts
import getConfig from 'utils/Config'

const { noDownload } = getConfig()
```

It's almost the same as in the previous iteration (`import { cfg } from 'utils/Config'`), but instead of importing object you can import getter function.
Using getter function instead of object is more future-proof, because you can provide some options (ex. environment, context) to that getter, and it involves less rewriting.

fiskus (commented):
If your last iteration (https://github.com/quiltdata/quilt/pull/3166/commits/590ea3a3a0a9a9cfa7eafaa05216b6c05ab1b41f) is a trade-off for my feedback, then it's better to revert it, because I like previous iteration more, and, probably, you too

nl0:
> Sorry, I wasn't clear about getter
> 
> I thought about something like this:
> 
> ```ts
> // utilsConfig.ts
> 
> const config = readAndValidateConfigOnStart()
> export default function getConfig() {
>   return config
> }
> ```

right now the code looks almost exactly like this

> ```ts
> // some-module.ts
> import getConfig from 'utils/Config'
> 
> const { noDownload } = getConfig()
> ```
> 
> It's almost the same as in the previous iteration (`import { cfg } from 'utils/Config'`), but instead of importing object you can import getter function.

you can do that with the current implementation if you need (via `getGlobalConfig()` or `useConfig()` functions)

> Using getter function instead of object is more future-proof, because you can provide some options (ex. environment, context) to that getter, and it involves less rewriting.

i don't quite get that argument, please elaborate on this. what would some actual examples of this look like?
using a callable every time "just because" makes the code lengthier for no reason (you need to declare a new variable to assign the result of the call every time, and you also need to cache the result)

catalog/app/utils/Config.ts
fiskus: `useConfig` is a hook, I don't like using it for config at all :)

---

```ts
import cfg from 'utils/Config'
const { noDownload } = cfg
```
has almost the same ergonomic as
```ts
import getConfig from `utils/Config`
const { noDownload } = getConfig()
```

But with `getConfig` you are free to extend it with additional functionality, like providing context or environment without much rewriting:
```ts
const { noDownload } = getConfig(user)
```

catalog/app/utils/Config.ts
```diff
@@ -0,0 +1,123 @@
...
+  disableNavigator: cfg.mode === 'MARKETING',
+  s3Proxy: startWithOrigin(cfg.s3Proxy),
+  apiGatewayEndpoint: startWithOrigin(cfg.apiGatewayEndpoint),
+  binaryApiGatewayEndpoint: startWithOrigin(cfg.binaryApiGatewayEndpoint),
+  noDownload: !!cfg.noDownload,
+  noOverviewImages: !!cfg.noOverviewImages,
+  desktop: !!cfg.desktop,
+})
+
+export function getConfig(input: unknown) {
```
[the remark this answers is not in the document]
fiskus: Both `getConfig` and `getGlobalConfig` do additional work inside besides getting memoized config. I think such base functionality as getting config should be without additional work as much as possible, so some `export` (preferably `export default`) should provide already validated and transformed config

catalog/app/constants/config.ts
```diff
@@ -0,0 +1,3 @@
+import { getGlobalConfig } from 'utils/Config'
+
+export default getGlobalConfig()
```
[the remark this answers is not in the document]
nl0: well, i wasn't thinking in such terms, i was thinking a constant is just a value which is shared globally across the system and can be imported wherever you need it. like, i can put the config into some module in `utils/`, but i think it makes quite a lot of sense to have it in `constants/`

fiskus:
> right now the code looks almost exactly like this

The main difference. In my pseudocode config validated and created only once, and then you can use the same object over and over

catalog/app/utils/Config.ts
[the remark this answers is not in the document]
nl0: > export (preferably export default) should provide already validated and transformed config

i agree with that, and that's why there's `constants/config` module which has this transformed and validated config as a default export

fiskus:
Ok, I see.
I talked about getter, and then I saw you made `export getConfig` and `export getClobalConfig`, and thought it's addressing to my feedback. But it's not what I meant.

Now I see, that you actually ignored my feedback on getter, and that's kinda ok :)

fiskus:
If `utils/Config` will be used only for `constants/config` - I agree with new changes

I care more about API, then implementation, because implementation is a subject to change, but API is what should be more or less stable in collaborative code. API wasn't clear for me

If you want to replace all:
```ts
import * as Config from 'utils/Config'
const { noDownload } = useConfig()
```
to
```ts
import cfg from `constants/config`
const { noDownload } = cfg
```
or
```ts
import { noDownload } from `constants/config`
```
then I OK with that, though I would prefer
```ts
import getConfig from `constants/config`
const { noDownload } = getConfig()
```

catalog/app/utils/Config.ts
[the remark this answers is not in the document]
nl0: > I think that config is an important base module and should have strict limited API

and it does (see `constants/config` module, which is a new entrypoint for getting the config).
`utils/Config` module is a utility module which contains a set of utilities for working with config data.

nl0:
> > right now the code looks almost exactly like this
> 
> The main difference. In my pseudocode config validated and created only once, and then you can use the same object over and over

`useConfig` does exactly that

nl0:
> I talked about getter, and then I saw you made export getConfig and export getClobalConfig, and thought it's addressing to my feedback. But it's not what I meant.
>
> Now I see, that you actually ignored my feedback on getter, and that's kinda ok :)

i haven't ignored it and implemented that logic in `useConfig` function (tho i didnt make it the primary api)

> If `utils/Config` will be used only for `constants/config` - I agree with new changes

yeah it will (after we replace all the `useConfig` calls)

> I care more about API, then implementation, because implementation is a subject to change, but API is what should be more or less stable in collaborative code. API wasn't clear for me

ok makes sense

> If you want to replace all:
> 
> ```ts
> import * as Config from 'utils/Config'
> const { noDownload } = useConfig()
> ```
> 
> to
> 
> ```ts
> import cfg from `constants/config`
> const { noDownload } = cfg
> ```

yeah more or less

> or
> 
> ```ts
> import { noDownload } from `constants/config`
> ```

that's not quite feasible due to how webpack and ES modules work

> then I OK with that, though I would prefer
> 
> ```ts
> import getConfig from `config`
> const { noDownload } = getConfig()
> ```

ok i'm glad we can *more or less* agree on the api surface, but i still don't see a compelling reason to keep the extra unnecessary step of calling the function instead of just accessing the result directly (tho i'm still open to being convinced otherwise)

fiskus:
Future needs for refactoring will convince you :)

fiskus (approved):

catalog/app/utils/Config.ts
fiskus: I didn't get why 
```
import getConfig from `utils/Config`
const { noDownload } = getConfig()
```
should be like this:
```
import { getConfig } from 'utils/Config'
const cfg = getConfig()
const { noDownload } = cfg
```
but it's almost out of scope of this PR, because using getter it's not my hard requirement, more like my guts feeling

catalog/app/utils/Config.ts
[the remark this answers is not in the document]
fiskus: We use import 'utils/Config' a lot, and I thought you will continue to use it
`constants/config` is OK

catalog/app/utils/Config.ts
nl0: > I didn't get why

because of this part:

> const { noDownload } = cfg

it seems quite meaningless to use destructuring here and it feels more convenient to access `cfg.*` members directly without assigning them to variables

catalog/app/constants/config.ts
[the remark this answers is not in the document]
nl0: new top-level namespace to do `import cfg from 'config'`? well, it kinda makes sense, but it's more work because we need to adjust webpack and typescript setup

fiskus (approved):

PR #3167, opened 2022-10-31
Send catalog release hash to mixpanel

nl0:


kevinemoore (commented):
I recommend we close this PR in favor of one that sends the template hash. I assume that's more difficult?

nl0:
> I recommend we close this PR in favor of one that sends the template hash. I assume that's more difficult?

yeah it is, we need to pass the template hash to the catalog container and then expose it

akarve:
Well we can still merge this and backlog the deployment hash. There's no reason not to have this info imo. What about sentry?

nl0:
> What about sentry?

it has this data if you're asking that

PR #3171, opened 2022-11-03
s3-proxy: return the original Content-Length in a custom header

dimaryaz:
When compression is used, the Content-Length header gets removed. The original file size can still be useful for e.g. the catalog, so return it in a custom header.

### DIFF
--- a/s3-proxy/nginx.conf
+++ b/s3-proxy/nginx.conf
@@ -42,7 +42,11 @@ http {
             add_header 'Access-Control-Allow-Methods' $http_access_control_request_method always;
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Max-Age' '3000' always;
-            add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class' always;
+            add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class, x-quilt-size' always;
+
+            # When compression is used, the original Content-Length header gets removed.
+            # The original file size may still be useful, so return it in a custom header.
+            add_header 'x-quilt-size' $upstream_http_content_length;
 
             # Return success on OPTIONS.
             if ($request_method = 'OPTIONS') {


s3-proxy/nginx.conf
```diff
@@ -42,7 +42,11 @@ http {
             add_header 'Access-Control-Allow-Methods' $http_access_control_request_method always;
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Max-Age' '3000' always;
-            add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class' always;
+            add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class, x-quilt-size' always;
+
+            # When compression is used, the original Content-Length header gets removed.
+            # The original file size may still be useful, so return it in a custom header.
+            add_header 'x-quilt-size' $upstream_http_content_length;
```
dimaryaz, line 49: Don't know what's a good name for it. I was trying to copy the `x-amz-...` headers.

s3-proxy/nginx.conf
```diff
@@ -42,7 +42,11 @@ http {
             add_header 'Access-Control-Allow-Methods' $http_access_control_request_method always;
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Max-Age' '3000' always;
-            add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class' always;
+            add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class, x-quilt-size' always;
```
dimaryaz, line 45: Could probably remove the `Content-Length` now (unless it'll make things difficult because catalog still needs to work with un-proxied S3?)

s3-proxy/nginx.conf
```diff
@@ -42,7 +42,11 @@ http {
...
+
+            # When compression is used, the original Content-Length header gets removed.
+            # The original file size may still be useful, so return it in a custom header.
```
sir-sigurd, line 48: I think it's more like "required for catalog to properly work", rather than "may still be useful".

sir-sigurd (commented):
I understand why nginx compression could be useful, but obviously it requires more work on frontend, I'm not sure if this work worth benefits of compression.

ETag header (others?) is also modified for compressed responses, I'm not sure if catalog needs this header.

nl0:
i tend to agree w/ @sir-sigurd that this will create extra work on the front-end for unclear benefit.
i know that it may be useful especially for larger text files, but it just seems quite low priority rn.
we may want to get back to that later

dimaryaz:
Ok

PR #3181, opened 2022-11-10
Add a dedupe parameter to push

dimaryaz:
Remaining question: how do `force` and `dedupe` interact with each other?

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1437,7 +1437,7 @@ class Package:
             latest_hash = get_latest_hash()
             if top_hash == latest_hash:
                 if print_info:
-                    print(f"Package with hash {latest_hash} already exists at the destination; skipping.")
+                    print(f"Skipping since package with hash {latest_hash} already exists at the destination and dedupe parameter is true.")
                 return
 
         # Since all that is modified is physical keys, pkg will have the same top hash
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -1292,7 +1292,10 @@ class Package:
 
     @ApiTelemetry("package.push")
     @_fix_docstring(workflow=_WORKFLOW_PARAM_DOCSTRING)
-    def push(self, name, registry=None, dest=None, message=None, selector_fn=None, *, workflow=..., force=False, dedupe=False):
+    def push(
+        self, name, registry=None, dest=None, message=None, selector_fn=None, *,
+        workflow=..., force=False, dedupe=False
+    ):
         """
         Copies objects to path, then creates a new package that points to those objects.
         Copies each object in this package to path according to logical key structure,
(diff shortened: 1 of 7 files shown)


api/python/quilt3/packages.py
```diff
@@ -1428,6 +1433,13 @@ def check_latest_hash():
         top_hash = self._calculate_top_hash(pkg._meta, self.walk())
         pkg._origin = PackageRevInfo(str(registry.base), name, top_hash)
 
+        if dedupe:
+            latest_hash = get_latest_hash()
+            if top_hash == latest_hash:
+                if print_info:
+                    print(f"Package with hash {latest_hash} already exists at the destination; skipping.")
```
akarve, line 1440:
```suggestion
                    print(f"Skipping since package with hash {latest_hash} already exists at the destination and dedupe is true.")
```

akarve (changes requested):
Can you add CLI bindings for --dedupe just like for --force? Also, a unit test would be 💯.

api/python/quilt3/packages.py
```diff
@@ -1395,30 +1404,42 @@ def dest_fn(lk, *args, **kwargs):
                         f"in the {registry!r} package registry specified by 'registry'."
                     )
 
+        if dedupe and not force:
+            raise QuiltException("dedupe=True requires force=True")
```
sir-sigurd, line 1408: Haven't thought much about it but why `dedupe` requires `force`?

sir-sigurd (commented):
Please add the new parameter to docstring.

api/python/quilt3/packages.py
akarve: I'm also not sure that dedupe requires force.

sir-sigurd: I'm not sure that force=True means "always push". I thought it means to skip only the specific check.
https://docs.quiltdata.com/api-reference/cli#push
```
  --force              Skip the parent top hash check and create a new
                       revision even if your local state is behind the remote
                       registry.
```

akarve: You're right.

akarve: That's how we use it, but "force" means "no matter what"... anway as long as we are consistent with how the app works.

dimaryaz: I only did it because dedupe without force seemed like a very obscure scenario: the parent hash has to match the latest hash and _also_ match the current package hash. Not sure how you end up in that scenario by accident (and if it's not by accident, then just don't call push unnecessarily?)

But anyways, I'll remove it.

sir-sigurd (commented):
LGTM, but please add changelog entry.

sir-sigurd (approved):

dimaryaz:
Btw, @akarve: still blocked on your review

api/python/quilt3/packages.py
```diff
@@ -1333,13 +1336,20 @@ def push(self, name, registry=None, dest=None, message=None, selector_fn=None, *
                 push the local file to s3 (instead of pushing all data to the destination bucket).
             %(workflow)s
             force: skip the top hash check and overwrite any existing package
+            dedupe: don't push if the top hash matches the existing package top hash; return the current package
```
akarve, line 1339: I guess I'm OK with this if you feel it's Pythonic but is it wise to change the return type based on a param? Shouldn't we always return `None`? Or is the idea that the return value is how users know that they didn't push? I guess that's OK but it seems weird to me.

akarve (approved):

PR #3187, opened 2022-11-16
Add new section with example for array keyword

robnewman:
# Description
Add new section to workflows describing support for drop-down strict lists in metadata validation with the `array` keyword

# TODO

### DIFF
--- a/docs/advanced-features/workflows.md
+++ b/docs/advanced-features/workflows.md
@@ -225,6 +225,33 @@ keyword `dateformat` in your schemas. For example:
 The `dateformat` template follows
 [Unicode Technical Standard #35](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table).
 
+### Drop-down list of predefined values
+Quilt supports the [`array` data type](https://json-schema.org/understanding-json-schema/reference/array.html). 
+Arrays are used for ordered elements. If you need your users to
+select one or more metadata entries from a strict list of values, then use
+the `array` data type. This is important for consistency across
+your data package metadata.
+
+Define your list items with the `items` keyword,
+then in a list use the `enum` data type to strictly define the
+available options in a tuple. You also need to define a minimum number
+of entries using the `minItems` keyword and a maximum number of
+entries using the `maxItems` keyword. For example:
+
+```json
+{
+    "type": "array",
+    "items": [
+        {"enum": ["Fixed Value 1","Fixed Value 2","Fixed Value 3","Fixed Value 4"]}
+    ],
+    "minItems": 2,
+    "maxItems": 4
+}
+```
+
+Note that we use the Draft 4 Json Schema where tuples are validated with `items`, and not `prefixItems`.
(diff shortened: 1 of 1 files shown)


QuiltSimon (approved):
👌

fiskus:
This section is OK for me, and mabe it's enough

But I think it can be confusing for someone who is not familiar with JSON and JSON Schema when we talk about lists:
* array is a list (of unordered elements)
* tuple is a list (of ordered elements)
* enum is a list (of predefined values as single element or single value)

So, maybe it's worth to make this section more tutorial-ish with strict distinction between these terms

fiskus:
My attempt:

### Arrays, tuples and enums
You can use `array` keyword if you need to define list of elements. These elements can be of any type.
If the order in this list is not significant for you, use "arrays":
```json
{
  "type": "array",
  "items": {
    "anyOf": [
      {
        "type": "string"
      },
      {
        "type": "number"
      }
    ]
  }
}
```
With this Schema you can create lists like: `["Any string A", 123, "Any string B"]` or `[123, "Any string", 456]`

If order is important, then use "tuples":
```json
{
  "type": "array",
  "items": [
    {
      "type": "string"
    },
    {
      "type": "number"
    }
  ],
  "minItems": 2,
  "maxItems": 2
}
```
With this Schema you can create lists like `["Any string", 123]`. Incorrect order will give an error `[123, "Any string"] // invalid`
Consider that you should define `"minItems"` and `"maxItems"` or `"minItems"` and `"additonalItems": false`, because tuples should have fixed size.

Instead of letting users set any value, you can define list of available options with `enum`:
```json
{
  "type": "array",
  "items": {
    "type": "string",
    "enum": ["Fixed 1", "Fixed 2"]
  }
}
```
With this Schema users are allowed to create list of any length with predefined values only like `["Fixed 1", "Fixed 2", "Fixed 1"]`.
Or for ordered list:
```json
{
  "type": "array",
  "items": [
    "type": "string",
    "enum": ["Fixed 1", "Fixed 2"]
  ],
  "minItems": 1,
  "additionalItems": false,
}
```
With this Schema users are allowed to create tuples like `["Fixed 1"]` or `["Fixed 2"]`.

If you want to provide users with a list of predefined values but let them fill any values out of this list, you can use `anyOf` keyword:
```json
{
  "type": "array",
  "items": {
    "anyOf": [
      {
        "type": "string"
        "enum": ["Fixed 1", "Fixed 2"]
      },
      {
        "type": "string"
      }
    ]
  }
}
```
Then lists `["Fixed 1", "Fixed 2"]`, `["Fixed 1", "Any string"]` or `["Any string 1", "Any string 2"]` are all valid.

In some cases you could want to create lists that have first ordered items of predefined values and the rest are any other. Then you can can create tuples with `"additionalItems": true`:
```json
{
  "type": "array",
  "items": [
    "type": "string",
    "enum": ["Fixed 1", "Fixed 2"]
  ],
  "minItems": 1,
  "additionalItems": true,
}
```
With this Schema lists like this are valid `["Fixed 1", "Any string", 123]` and this is invalid `["Any string", 123]`

fiskus (approved):

PR #3197, opened 2022-11-23
Scrub any unnecessary references to Package.build()

robnewman:
# Description
Remove any (unnecessary) references to `Package.build()` to avoid user confusion. Everyone should be using `Package.push()` _only_, unless there are very clear scenarios in which `Package.build()` are **uniquely applicable** . Those scenarios have been left in the documentation untouched.


docs/advanced-features/materialization.md
```diff
@@ -1,34 +1,35 @@
...
+referencing the remote files in the remote registry.
+
+This is known as **materialization**,
+and it means that `push` creates **materialized packages**: packages
+which point solely to files located in an Amazon S3 bucket.
 
-A materialized package is stronger than a unmaterialized (or "local") package because S3 guarantees that individual object versions are never lost or destroyed (assuming object versioning is enabled). Furthermore, only materialized packages may be browsed in the online Quilt catalog.
+A materialized package is powerful because Amazon S3 guarantees that
+individual object versions are never lost or destroyed (assuming
+object versioning is enabled).  Furthermore, only materialized
```
akarve, line 16: > S3 guarantees that individual object versions are never lost or destroyed
This is not true. More like: "with appropriate IAM controls S3 object versions can be protected against accidental deletion"

docs/advanced-features/materialization.md
```diff
@@ -1,34 +1,35 @@
...
+packages may be browsed in the online Quilt catalog.
 
 ## Pushing unmaterialized packages
```
akarve, line 19: We should drop this whole section.

akarve (changes requested):
See comments inline.

akarve (approved):

PR #3233, opened 2022-12-16
Docs page for writing custom queries against elastic search cluster indices

robnewman:
# Description

New docs page in similar vein to `Working with the Catalog` and `Working with a Bucket` specifically for writing custom search queries against ES cluster indices


drernie (approved):
Looks good. Might be worth explaining that we currently use traditional ElasticSearch, rather than AWS' custom OpenSearch. That terminology always confused me...

PR #3237, opened 2022-12-22
Add help message to `tagged#case` to find out which variants are redundant

fiskus:


catalog/app/utils/tagged.js
```diff
@@ -70,7 +70,10 @@ export default (variants) => {
     )
     invariant(
       R.all(R.contains(R.__, ['_', '__', ...variants]), R.keys(cases)),
-      `${scope}/case: cases may only include type variants and placeholders (_ and __)`,
+      `${scope}/case: cases may only include type variants and placeholders (_ and __). These variants are redundant ${R.without(
```
nl0, line 73:
```suggestion
      `${scope}/case: cases may only include type variants and placeholders (_ and __). Unrecognized variants: ${R.without(
```

nl0 (approved):

nl0 (approved):

PR #3242, opened 2022-12-27
Athena perspective link

fiskus:
Listen re-renders of table and change table cell content on-the-fly.
`addStyleListener` calls only on re-render, so we have to call it first time manually.


nl0 (approved):
looks a little bit hacky, but, well, we work with what we have 🤷 
tho isn't there a cleaner way to do this, e.g. customize a cell rendering function or smth?

fiskus:
Perspective library author referenced his own example of customizing cells by listening `addStyleListener` and rewriting `td.innerHTML`:
https://github.com/finos/perspective/discussions/1510#discussioncomment-1146608

https://github.com/texodus/mtg-perspective/blob/main/src/index.js#L153
https://github.com/texodus/mtg-perspective/blob/main/src/mana_cost_utils.js#L12

`addStyleListener` is undocumented in @finos/perspective, but it is from regular-table package from the same author:
https://www.npmjs.com/package/regular-table#addstylelistener-and-getmeta-styling

PR #3247, opened 2023-01-04
Rename metadata to user metadata and expand JSON to 100% width

fiskus:
Before:
Screenshot from 2023-01-04 13-01-40

After:
Screenshot from 2023-01-04 13-01-25

* Added `onToggle` callback to `JsonDisplay`
* Changed table layout to more flexible `div`/`dl`
* (Irrelevant here but) Added link to package revisions from package list


catalog/app/components/JsonDisplay/JsonDisplay.js
```diff
@@ -157,11 +157,19 @@ function CompoundEntry({
...
+  onToggle,
 }) {
   const braces = Array.isArray(value) ? '[]' : '{}'
   const entries = React.useMemo(() => Object.entries(value), [value])
   const [stateExpanded, setExpanded] = React.useState(!!defaultExpanded)
-  const toggle = React.useCallback(() => setExpanded((e) => !e), [])
+  const toggle = React.useCallback(() => {
+    const newStateExpanded = !stateExpanded
+    if (onToggle) {
+      onToggle(newStateExpanded)
```
nl0, line 168: i don't quite get how this works: if you provide the onToggle prop, how the state is changed?

nl0: if you plan to rely on `defaultExpanded`, this won't work afaict, bc it's only used once to initialize the state, and all the updates to this prop are ignored

catalog/app/containers/Bucket/Meta.tsx
```diff
@@ -85,58 +73,78 @@ interface PackageMetaProps extends Partial<SectionProps> {
 function PackageMetaSection({ meta, preferences, ...props }: PackageMetaProps) {
   const classes = usePackageMetaStyles()
   const { message, user_meta: userMeta, workflow } = meta
+  const [expanded, setExpanded] = React.useState({
```
nl0, line 76: combining these states seem like an unnecessary complication -- it would be simpler to just have two separate states

fiskus: Yeah, since I still have two callbacks anyway `onMetaToggle` and `onWorkflowToggle`
I made one callback, but the code looked too complicated for such trivial operation

I'll revert to two states

catalog/app/components/JsonDisplay/JsonDisplay.js
fiskus: It works accidentally :)
But I realized my mistake, thank you

I planned to use `onToggle` as a callback without changing the state, so I'll return to that decision

fiskus:
I modified state manipulation and no longer change `defaultExpanded` prop

catalog/app/containers/Bucket/Meta.tsx
```diff
@@ -24,54 +24,42 @@ const errorHandler = (error: Error) => (
...
-  )
-}
-
-const usePackageMetaStyles = M.makeStyles({
+const usePackageMetaStyles = M.makeStyles((t) => ({
   headCell: {
+    paddingTop: t.spacing(0.25),
     width: '137px',
   },
+  cellExpanded: {
```
nl0, line 32: does this mean that the json block is displayed inline while not expanded, and after expansion it jumps to the new line?

fiskus: Yes, that was the main idea. Does it bother you?

It doesn't look nice at this moment, but after https://github.com/quiltdata/quilt/pull/3249 it will look nicer

nl0: yeah, this kind of jump seems like a bad UX to me. i'd rather always display it on its own line or keep it inline, it doesn't matter that much imo, but the jump bothers me

fiskus: I agree that it doesn't look good. The idea was to fill all width, but it's not too important to have 100% width, so we can make such UX tradeoff

Since this PR consists mostly from changes for that expanding, I'll close it and create new one for renaming metadata -> user metadata

fiskus:
UX of the main change is not optimal and doesn't create significant benefits
Other changes moved to separate PRs:
* https://github.com/quiltdata/quilt/pull/3255
* https://github.com/quiltdata/quilt/pull/3256

PR #3248, opened 2023-01-06
Cross-account SNS topic notifications & subscription

robnewman:
# Description
* Documentation to describe how to get notifications for cross-account S3 bucket changes
* All PNG screenshot images have private data blurred and have been crushed with ImageOptim

---
  - https://app.asana.com/0/0/1203130258649777


docs/CrossAccount.md
```diff
@@ -62,3 +63,109 @@ For security, auditing, and user-facing analytics, it is recommended that all
...
+
+This is done by adding a new IAM policy in the
+relevant *DataAccountTwo* Amazon S3 bucket that provides permissions
+to write to an existing SNS topic in the *ControlAccount*.
+
+1. Create a S3 bucket policy in the *DataAccountTwo* account that
+allows the *ControlAccount* access to all S3 events:
+
+<!--pytest.mark.skip-->
+```json
```
sir-sigurd, line 82: This looks the same as the policy above, let's avoid duplication so we won't have to update it multiple places.

robnewman: Done, although I think it's now less readable and more error prone.

docs/CrossAccount.md
```diff
@@ -62,3 +62,83 @@ For security, auditing, and user-facing analytics, it is recommended that all
...
+![](imgs/cross-account-s3-event-setup.png)
+
+In the Destination section, choose SNS topic and specify the relevant SNS
+topic either from your list of SNS topics or by entering it manually:
+
+![](imgs/cross-account-s3-event-destination.png)
+
+Once you hit `Save changes`, Amazon S3 will send a test message to the event notification destination.
+
+[AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-event-notifications.html)
```
sir-sigurd, line 144: It's not clear what this link refers to.

sir-sigurd (commented):
I don't quite understand what scenario this doc covers.

I did some testing:
1. create S3 bucket in data account
2. added it to Quilt 
3. put object to the new bucket
4. I see the object indexed (which relies on SNS notifications)

So everything seems to work just fine.

The additional setup from customer seems to be needed in the case when S3 bucket from data account is already configured to publish events to SNS topic from data account. In that case SNS topic should allow SQS queues from control account to subscribe to it (might be not quite correct).

sir-sigurd (changes requested):
See my previous comments.

akarve (changes requested):
Refer to Slack notes and convo; needs simplification and correction

akarve (changes requested):
Please simplify and correct per Slack and 1:1 convo

drernie:
Doesn't seem necessary, or at least worth the effort.

PR #3260, opened 2023-01-13
Unify all OIDC code and remove client IDs

dimaryaz:
Use the new /oidc-authorize registry endpoint and remove all provider-specific details.


nl0 (dismissed):
this looks very nice

PR #3262, opened 2023-01-13
[PR cleanup] add kms docs

drernie:
Redoing old PR from September, as part of January cleanup

# Description

Added minimal documentation on adding KMS-SSE keys.
Per Aneesh's request on original PR, only works if the customer has already defined Source=Quilt Roles


robnewman (commented):
Is the PNG screenshot crushed with ImageOptim or equivalent utility?

docs/technical-reference.md
```diff
@@ -437,6 +437,97 @@ similar to the following to the KMS key policy:
 }
 ```
 
+### S3 buckets with SSE-KMS
```
robnewman, line 440: Unless we have already defined the acronym SSE above, we should first use the full string (`Server-Side Encryption`) and _then_ use the abbreviation elsewhere. Same for `Key Management Service (KMS)`

drernie:
FYI, the largest file in doc/imgs is quilt.gif, which I doubt is even used anymore.

robnewman:
> @robnewman I didn't do any explicit compression, but it is only 119KB so I don't think that matters.

It does matter when everything (for now) sits in the `quilt` repo. Screenshot file size add up. I recommend crushing

docs/technical-reference.md
```diff
@@ -437,6 +437,97 @@ similar to the following to the KMS key policy:
...
+
+1. Add KMS Key Usage to Quilt Permission Boundary
+2. Add Quilt Principals to KMS Key Policy
+3. Add KMS Key Access to a Scoure=Quilt Role
+
+NOTE: This will not work with the default Source=Custom Roles.
+
+#### 1. Add KMS Key Usage to Quilt Permission Boundary
+
+By default, AWS does not allow anything in your account to access KMS. If you haven't done so already,  create an IAM policy that explicitly enables KMS access.
```
robnewman, line 452: We need to ensure long lines are wrapped as per PEP-8 (max chars 79)

docs/technical-reference.md
```diff
@@ -437,6 +437,97 @@ similar to the following to the KMS key policy:
...
+    "Resource": "arn:aws:kms:us-west-2:111122223333:key/*"
+  }
+}
+```
+
+Go to CloudFormation > Your Quilt Stack -> Update -> Parameters and add the ARN of that IAM policy to  `ManagedUserRoleExtraPolicies` at the bottom of the page:
+
+![](../imgs/ManagedUserRoleExtraPolicies.png)
+
+If other policies are already in that field, you will need to add a comma before appending the ARN.
```
robnewman, line 472: This should be highlighted/add emphasis with a `>` at the beginning

docs/technical-reference.md
```diff
@@ -437,6 +437,97 @@ similar to the following to the KMS key policy:
...
+        "kms:Decrypt",
+        "kms:GenerateDataKey"
+    ],
+    "Resource": "*"
+}
+```
+
+#### 3. Add KMS Key Access to Quilt Role
+
+Finally, you need create a restricted policy that gives a Quilt role access to the keys for specific buckets, e.g:
```
robnewman, line 509: Nit-picky, but line length again

robnewman (changes requested):

docs/technical-reference.md
```diff
@@ -437,6 +437,107 @@ similar to the following to the KMS key policy:
 }
 ```
 
+### S3 buckets with SSE-KMS
```
drernie, line 440:
```suggestion
### S3 buckets with Server-Side Encryption via Key Management Service (SSE-KMS)
```

docs/technical-reference.md
```diff
@@ -437,6 +437,97 @@ similar to the following to the KMS key policy:
...
+  }
+}
+```
+
+You can now create a Quilt Policy from this policy using the Catalog's admin panel.
+Afterwards, you can attach that Policy to a user-defined Quilt Role (which has Source=Quilt in the Roles panel, as opposed to system-defined Source=Custom Roles).
+
 ## Backup and Recovery
 
 All data and metadata in Quilt is stored in S3. S3 data is automatically backed up (replicated across multiple available zones). To protect against accidental deletion or overwriting of data, we strongly recommend enabling object versioning for all S3 buckets connected to Quilt.
```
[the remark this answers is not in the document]
drernie: I have fixed the long-lines in MY section.  I would like to get that new content approved before rewrapping the  entire document (if necessary).

docs/technical-reference.md
drernie: Propose we ignore so we can get this merged.

robnewman (approved):

PR #3265, opened 2023-01-16
Update key and key_text samples

fiskus:
Screenshot from 2023-01-16 16-27-41


catalog/app/containers/NavBar/Help.tsx
```diff
@@ -79,7 +79,7 @@ const syntaxHelpRows: SyntaxHelpNamespace[] = [
         title: 'Object key',
       },
       {
-        example: (s) => <>{s}"research"</>,
+        example: (s) => <>{s}"phase*"</>,
```
robnewman, line 82: Shouldn't it be `{s}phase*` without the quotes? Without the quotes works for me on open. With quotes returns no results.

robnewman (changes requested):

robnewman (approved):

PR #3277, opened 2023-01-18
Fix collaborators render warning

fiskus:


catalog/app/components/Collaborators/Table.tsx
```diff
@@ -77,7 +77,7 @@ export default function Table({ collaborators }: TableProps) {
       <M.TableBody>
         {sorted.map(
           ({ collaborator: { email, username }, permissionLevel = 'UNKNOWN' }) => (
-            <M.TableRow>
+            <M.TableRow key={email + username}>
```
nl0, line 80: username is unique, so it should be enough, when email + username *may be* non-unique, e.g. `user@example.com` + `user` vs `user@example.co` + `muser` (or we should at least use a separator)

nl0 (approved):

PR #3280, opened 2023-01-18
Clean up auth code and use the new /oidc-authorize endpoint

dimaryaz:
Okta, OneLogin, and Azure now use a registry endpoint that handles base URLs and client IDs, so the catalog does not need their configs anymore.

OneLogin and Azure auth is merged into a single implementation. Okta still uses custom code because changing it would require the customer to update the redirect URI.

Google still uses old code and requires a client ID because changing it would also require the customer to update the redirect URI.


catalog/app/utils/Okta.js
```diff
@@ -1,40 +1,40 @@
...
 
 export class OktaError extends BaseError {
   constructor(code, details) {
-    super('Okta login failure', { code, details })
+    super('Login failure', { code, details })
   }
 }
 
-export function useOkta({ clientId, baseUrl }) {
+export function useOkta({ provider, popupParams }) {
```
nl0, line 13: it seems a little odd to pass a provider name here, since it's always "okta"

nl0 (dismissed):
looks good, tho we probably should hardcode the `provider` variable in okta code

nl0 (approved):

PR #3289, opened 2023-01-23
More explicit word choice about catalog hash in the footer

fiskus:


nl0 (approved):

sir-sigurd:
Maybe we should really use the revision of CFN template here?

fiskus:
The use case for this is to be confident, does the customer have a code with discussed features.
So, when they say that feature doesn't work, we can tell from that hash that they have outdated product and don't have that feature.

Maybe, CFN revision is more correct. But this git hash is useful as well

sir-sigurd (approved):

PR #3290, opened 2023-01-24
Toggle button for file previews

fiskus:
Screenshot


nl0 (approved):
look fine, couple thoughts:
- search results display is not affected by this change, is it?
- i'd put the expand controls to the right, and hidden all the other stuff (e.g. download) into a "more" menu next to it -- this would imo work better for search results, bc search results show an entity type icon on the left (package vs object)

fiskus:
Hmmm... 
Screenshot from 2023-01-24 23-15-42

fiskus:
Does it look ok for you? I'm not sure about my opinion yet

fiskus:
Screenshot from 2023-01-25 11-10-22

fiskus:
I put the expand/collapse button to the right, and shrank the download button into a 3-dots menu.
I moved these buttons to app/components/Preview directory and re-used them for SearchResults.

Also, I removed "auto-expand" functionality because it requires more work: some previews obtain height after a while (Jupyter, JsonDisplay). So, we need to listen resize etc., I don't want to pollute this PR with this.

nl0:
> Does it look ok for you? I'm not sure about my opinion yet

that looked ok to me, tho the spacing btw the icon buttons might be a little tighter

nl0 (approved):
the code seems fine, tho im not sold on the design -- full-fledged button seems too bulky for the place imo. but it's your call if you think it's better this way

fiskus:
I agree, that design looks nicer without EXPAND text and bordered button, but I'm afraid user couldn't find that button.

Let's let users get used to the new button position, then we can optimize the space and shrink this button.

PR #3292, opened 2023-01-25
Reduce backtracking during `pip install quilt3[catalog]`

sir-sigurd:


api/python/setup.py
```diff
@@ -89,6 +89,7 @@ def run(self):
         ],
         'catalog': [
             'quilt3_local>=1,<2',
+            'aiobotocore[boto3]>=2',
```
nl0, line 92: i'd leave a note that this is a workaround for a specific issue

nl0 (approved):

PR #3296, opened 2023-01-26
Use code authorization flow instead of ID tokens.

dimaryaz:


catalog/app/containers/Auth/SSOAzure.js
```diff
@@ -36,16 +36,16 @@ export default function SSOAzure({ mutex, next, ...props }) {
...
       mutex.claim(MUTEX_REQUEST)
       try {
-        dispatch(actions.signIn({ provider, token }, result.resolver))
+        dispatch(actions.signIn({ provider, code }, result.resolver))
         await result.promise
       } catch (e) {
         if (e instanceof errors.SSOUserNotFound) {
           if (cfg.ssoAuth === true) {
-            dispatch(push(urls.ssoSignUp({ provider, token, next })))
+            dispatch(push(urls.ssoSignUp({ provider, code, next })))
```
nl0, line 48: you also have to adjust the url function signature in `app/constants/routes.js:65`

nl0 (commented):
looks sane so far

catalog/app/containers/Auth/SSOAzure.js
dimaryaz: I was actually going to ask you about this one... Unlike the ID token, the code only works once - so we need to either prompt the user to authenticate again and get a new code, or just ignore this case and return an error. I can't remember anymore - when does this come up? I couldn't even test it.

nl0: i believe this happens when a user first tries to sign in via sso having no account yet, so basically this is how sso sign-up works (but only when password authentication is enabled as well: https://github.com/quiltdata/enterprise/pull/455/files#diff-a8a37d67caff3550aa0298bd5f9022038c317ff2e20381cbaf3b93246bc17a29R484).

> Unlike the ID token, the code only works once - so we need to either prompt the user to authenticate again and get a new code

well, this complicates things for sure

> or just ignore this case and return an error.

in this case users won't be able to sign up via sso with password-based authentication enabled

i see these potential approaches here:

1. create the user anyways, but set the unusable password, so that they are required to reset it afterwards (seems sane)
2. store some state in the db in such a case so that the user can continue signing up not requiring to re-request the code (or maybe just return an access token in the error response so that the front-end can use it to complete the registration?)

dimaryaz: Unusable password is probably good enough.

Another idea: authorization code flow gives us both a code and an ID token, so the login endpoint could use the ID token for everything, and avoid using the code until the very end, so it stays valid if there's an error. That said... I don't think passing the code as a URL parameter is a good practice cause it's pretty sensitive, and this whole flow seems unnecessarily complicated. Also, it asks for a username, while other flows don't - that's a whole different story.

@akarve, @kevinemoore: are you ok with letting users sign up without a password and doing a password reset later if they want to? This seems like a very obscure case - a customer who wants SSO, but also allows passwords (which defeats the point of SSO) - so I'd rather not spend more effort making it work.

dimaryaz: Going with unusable passwords

nl0: > This seems like a very obscure case - a customer who wants SSO, but also allows passwords (which defeats the point of SSO)

as mentioned in slack, this case is mostly for social sign-in, which is enabled on OPEN

nl0 (approved):

nl0 (approved):
don't see anything suspicious

PR #3297, opened 2023-01-26
Add a note about secure search flag

sir-sigurd:


akarve:
See #3312

sir-sigurd:
> See #3312

It looks information from this PR, so why this PR is closed?

PR #3300, opened 2023-01-26
Buttons refactoring

fiskus:
Changed:
* Buttons "Create package from folder" and "Push to bucket" have icon in wide layouts too. I think it would be consistent (though chosen icon is not optimal)
* Some small flex/margins changes like moving buttons to the next line on small screens (`flexWrap: 'wrap'`, see screenshot)

Refactored:
* Added `components/Buttons/ButtonIconized`, and replace `FileView.AdaptiveLayoutButton`
* Refactor and unify margins for these buttons
* Remove `components/ButtonIcon`. It was used only once, and it's trivial
Screenshot from 2023-01-30 18-27-08


nl0 (approved):
looks ok, tho i'm a little reluctant about adding a whole new namespace (`components/Button`) for only one export. to me, `components/ButtonIconized` makes more sense at the moment, tho this whole arguments smells of bike-shedding, so whatever

fiskus:
Hmm..., Ok, I'll rename it to ButtonIconized, and move to Buttons again only when there will be one more button reused across app

nl0:
> Hmm..., Ok, I'll rename it to ButtonIconized, and move to Buttons again only when there will be one more button reused across app

As i said, it's not a requirement and you can keep it as is if you feel like it would be better in perspective

fiskus:
I mean, I agree, namespace Buttons will be reasonable when it contains at least two buttons

nl0 (approved):

PR #3304, opened 2023-01-27
Remove the SSO signup page

dimaryaz:
It's no longer needed: the registry now allows SSO users to sign up without a password, even if passwords are allowed.


nl0 (approved):
nice. removing the code is always fun

PR #3305, opened 2023-01-29
Iframe browsing front-end PoC

nl0:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation


fiskus (commented):
I agree with concept
The code needs some linting, probably it was intentional to skip it. And I would move creation of the session to the hook:

```jsx
// loaders/Html.js
function useSessionRoot() { // or SessionId, or SessionUrl
  const [sessionRoot] = useState()
  useEffect(() => {...}, [])
}

function IFrameLoader({ handle }) {
  const sesionRoot = useSessionRoot()
  const src = `${sessionRoot}${handle.key}`
  return <iframe src={src} />
}
```

It's easier to read for me, because you deal with a limited amount of information at a time. And it tells what code does with variables names: so, I can skip reading `useEffect` thoroughly, if it does what I expect from it

fiskus:
> i remember you mentioning implementing this in some way

~I don't remember all details, but I think I didn't implement it with graphql~
Found it in `file-server-handler-renderer` branch

PR #3306, opened 2023-01-30
File edit as text

fiskus:
Main change is in FileEditor/loader.ts - detected type changed from single value to array 
* If there are multiple detected types to edit, then there is a menu to select which type to use
* Refactored state, `types` stores detected type, `editing` converted from boolean (is editing or not) to file type (if editing then with what type)
* Move state to separate module and re-used state type instead of creating type individually for every component


nl0 (approved):
looks good

PR #3310, opened 2023-02-01
Excel/Perspective editor

fiskus:
* Fixed and made explicit types for TypeScript
* Added Excel file editor using Perspective and `{ plugin_config: editable }`
* Made every heavy module in `components/FileEditor` lazy loaded
* Added option to set Schema for Perspective columns, and edit Schema using JSON text editor  

- [ ] Unit tests
- [ ] Documentation


fiskus:
I think we decreased priority over QuiltSync. Because we can't compete with Excel.

PR #3311, opened 2023-02-01
S3 Proxy: browse endpoint

nl0:
# Description

The new `browse` endpoint is supposed to stream the objects from s3 resolved using the `browse` endpoint in the registry providing a virtual hierarchy based on package manifests (or any other potential providers like raw s3 buckets or whatever).

Summary of how it is supposed to work:

1. A request is being issued to this endpoint by the catalog
2. nginx requests the same url from the registry (`/browse/<session_id>/*<path>`)
3. registry resolves this url to a presigned s3 url and returns an http redirect to it
4. nginx intercepts the redirect response and proxies (streams) the contents of that url

# TODO

- [x] polish the config


nl0:
@dimaryaz @sir-sigurd i need your help on this one, since i have a very vague idea how a proper nginx config should look 🥲

dimaryaz:
I've never even seen `location @handle_redirect`; interesting.

I'm wondering, though, if there are easier solutions? E.g.:
- Have the registry do a redirect to the s3-proxy location instead of the original S3 signed URL
- Make an API call to the registry that returns the signed URL, then rewrite it to point to the s3-proxy

Or is this solution supposed to be more efficient?

Also, I think in some cases, catalog doesn't proxy S3 requests - would this change cause problems there?

dimaryaz:
Can you also test special characters in the path - spaces, pluses, etc.?

nl0:
> Can you also test special characters in the path - spaces, pluses, etc.?

well, yeah, i probably should

> I'm wondering, though, if there are easier solutions? E.g.:
>
> - Have the registry do a redirect to the s3-proxy location instead of the original S3 signed URL
> - Make an API call to the registry that returns the signed URL, then rewrite it to point to the s3-proxy
>
> Or is this solution supposed to be more efficient?

well, this solution is supposed to be more "transparent" to the end user, so that it works like a "regular" web-server with relative paths, without redirects and without exposing underlying s3 urls.

the problems with redirects are:
1. relative urls would not work as desired (they will be resolved relative to the real s3 location, but we want them to be resolved relative to the virtual/logical package structure)
2. there's a possibility that some js code won't be following the redirects properly

> Also, I think in some cases, catalog doesn't proxy S3 requests - would this change cause problems there?

i don't think these things are related. in theory, we could make this into a standalone service not tied to the s3 proxy at all, bc it's completely unrelated conceptually, but since there's already an unrelated `zip` endpoint here, it seems okayish to have more utility stuff here instead of going thru all the complexities of weaving another service thru CFT

s3-proxy/nginx.conf
```diff
@@ -112,6 +112,60 @@ http {
...
+            #     proxy_pass $upstream_http_location;
+            # }
+        }
+
+        location @handle_redirect {
+            internal;
+            # TODO: this must be forced only for html files or, better yet,
+            # not forced at all -- we need to provide proper content-type headers
+            # when signing the urls in the registry
+            add_header 'Content-Type' 'text/html';
```
dimaryaz, line 151: I'd expect .html files to already have the correct header - e.g., if you use `aws s3 cp foo.html s3://whatever/`, it'll set the content type. But if not, then yeah, doing it through signed URLs should work,

s3-proxy/nginx.conf
```diff
@@ -112,6 +112,60 @@ http {
...
+            add_header 'Access-Control-Allow-Headers' $http_access_control_request_headers always;
+            add_header 'Access-Control-Allow-Methods' $http_access_control_request_method always;
+            add_header 'Access-Control-Allow-Origin' '*' always;
+            add_header 'Access-Control-Max-Age' '3000' always;
+            # add_header 'Access-Control-Expose-Headers' 'Content-Length, Content-Range, ETag, x-amz-meta-helium, x-amz-bucket-region, x-amz-delete-marker, x-amz-request-id, x-amz-version-id, x-amz-storage-class' always;
+
+            # Return success on OPTIONS.
+            if ($request_method = 'OPTIONS') {
+                return 200;
+            }
```
dimaryaz, line 163: This is only needed if you use XHR from JavaScript. If it's only used in an iframe, you don't need this at all - just the `proxy_pass`.

dimaryaz:
Ok, makes sense.

Just tried it myself. It seems to work correctly, including special characters.

Delete the commented out / unneeded code, and it should be good.

s3-proxy/nginx.conf
nl0: well, js in iframe cold use xhr to fetch the adjacent data files, but according to the docs, preflight requests arent made in most cases we care about, so ima drop this

dimaryaz (approved):

PR #3313, opened 2023-02-02
Make JsonDisplay nested keys lighter

fiskus:
Before:
1 before

After:
2 after


nl0 (approved):
looks ok, but i'm not quite sure why we're doing this

fiskus:
I think it was hard to read the one-liner, when it's not obvious which key are from the first level, and which from the second

fiskus:
Maybe, I'll wait feedback from customer who wanted these key/values (visually separated) in packages list

fiskus:
Never heard since that anyone expressed interest in this.

PR #3314, opened 2023-02-02
pkgpush lambda: fix metadata handling for entries without hash or size

sir-sigurd:
# Description

# TODO

- [x] Unit tests

---
  - https://app.asana.com/0/0/1203335357356363


lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -606,7 +609,8 @@ def create_package(req_file):
                     )
                 )
             else:
-                pkg.set(logical_key, str(physical_key), meta)
+                pkg.set(logical_key, str(physical_key))
+                pkg[logical_key]._meta = meta or {}
```
sir-sigurd, line 613: I know that smells but I don't think I could do better without major refactoring.

nl0: ok 🤷

nl0 (approved):

nl0 (approved):

PR #3251, opened 2023-01-09
Landing for AWS Marketplace

fiskus:
image

TODO:
* Links for "Read more"
* Links for One Liner (PDF) and Marketplace Listing
* Add more padding space for Case Studies block?
* Move Mailchimp endpoint and Calendly link somewhere?

- [ ] ~[Changelog](CHANGELOG.md) entry~ Probably, we should miss changelog entry here

PR #3319, opened 2023-02-03
Prettify `AwsMarketplace/Features.tsx` and everything

fiskus:
* Prettified `AwsMarketplace/Features.tsx` that slipped through in https://github.com/quiltdata/quilt/pull/3251
* And why not prettify a few other files in this PR. So, I did `cd catalog && npm run prettify .`


nl0 (approved):

catalog/static/oauth-callback.html
```diff
@@ -1,8 +1,11 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <body>
     <script>
-      window.opener.postMessage({type: "callback", fragment: window.location.hash}, window.location.origin);
+      window.opener.postMessage(
+        { type: 'callback', fragment: window.location.hash },
+        window.location.origin,
+      )
```
dimaryaz, line 8: FYI, this file is served as is, without being compiled/processed, so this change could theoretically break things: the trailing comma was not supported before 2017: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#browser_compatibility

(Maybe it should be compiled? I'm not sure how to set it up properly.)

fiskus: Thanks for this find!
I'll redefine prettier config for static directory, so it will use ES5 rules for commas

PR #3322, opened 2023-02-07
Add exception for prettifying static directory

fiskus:


catalog/static/.prettierrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "trailingComma": "es5"
```
nl0, line 2: won't this file be added to the build?

fiskus: Yes
I added `ignore` rule to CopyWebpackPlugin config

nl0 (approved):

PR #3325, opened 2023-02-11
Add provider-specific OAuth2 overrides to ask for refresh tokens

dimaryaz:
Currently, we're getting ID tokens and access tokens, but not refresh tokens (which aren't used yet, but will be soon).

It appears that requesting a refresh token is not standardized, and different providers do it in different, incompatible ways:
- Google needs `access_type=offline`, and will actively reject anything unexpected
- Okta and Azure need `scope=offline_access`, and will also reject anything unexpected
- OneLogin doesn't need anything, but will reject parameters required by the other providers


dimaryaz:
Actually, disregard - will do this server-side

PR #3327, opened 2023-02-11
Improve XDG Docs

akarve:
# Description

# TODO

- [n/a] Unit tests
- [n/a] Documentation

### DIFF
--- a/docs/api-reference/cli.md
+++ b/docs/api-reference/cli.md
@@ -219,13 +219,16 @@ $ export QUILT_TRANSFER_MAX_CONCURRENCY=20
 ```
 
 ### `XDG_*`
-Quilt uses platformdirs for Python to determine where to write data. You can therefore
-override the following path constants with environment variables using the XDG
-standard (see [platformdirs docs](https://pypi.org/project/platformdirs/)).
+Quilt uses platformdirs for Python to determine where to write data. You can
+override the following environment variables from the XDG
+standard (see [platformdirs docs for more](https://pypi.org/project/platformdirs/)).
 
 For instance, AWS Lambda requires the user to use `/tmp/*` as the scratch
 directory. You can override `quilt3.util.CACHE_PATH`, so that `quilt3 install` will succeed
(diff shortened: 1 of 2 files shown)


docs/api-reference/cli.md
```diff
@@ -219,14 +219,17 @@ $ export QUILT_TRANSFER_MAX_CONCURRENCY=20
 ```
 
 ### `XDG_*`
-Quilt uses platformdirs for Python to determine where to write data. You can therefore
-override the following path constants with environment variables using the XDG
```
robnewman, line 223: It doesn't make sense to say "override the following path constants with...." and then not list the path constants until the very end of the section (i.e. after an un-related example section). Move the "Constants" section immediately below this initial sentence, and then have the lambda example code in an "Example" section beneath the "Constants" section.

gendocs/env_constants.md
```diff
@@ -32,14 +32,17 @@ $ export QUILT_TRANSFER_MAX_CONCURRENCY=20
 ```
 
 ### `XDG_*`
-Quilt uses platformdirs for Python to determine where to write data. You can therefore
-override the following path constants with environment variables using the XDG
-standard (see [platformdirs docs](https://pypi.org/project/platformdirs/)).
+Quilt uses platformdirs for Python to determine where to write data. You can
+override the following environment variables from the XDG
```
robnewman, line 36: Same as the previous comment:

It doesn't make sense to say "override the following path constants with...." and then not list the path constants until the very end of the section (i.e. after an un-related example section). Move the "Constants" section immediately below this initial sentence, and then have the lambda example code in an "Example" section beneath the "Constants" section.

robnewman (changes requested):

docs/api-reference/cli.md
akarve: We never say "path constants" as that was in the old version. One of these .md files is generated so only one reviewer comment is needed per mistake.

"following" is misleading, I'll clean that up.

robnewman (approved):

PR #3286, opened 2023-01-20
Validation of package entries metadata

sir-sigurd:
# Description

# TODO

- [x] Unit tests
- [x] <del>Documentation</del> (added in https://github.com/quiltdata/quilt/pull/3336)
- [x] [Changelog](CHANGELOG.md) entry (skip if change is not significant to end users, e.g. docs only)

PR #3336, opened 2023-02-20
Object-level validation, frontend

fiskus:
* Added one more visible state to files - `"invalid"`
* Moved entries validation to the function used `final-form` API (`<RF.Field validate={...here...} />`)
* Populate `Ajv` error with entry context data, and then check if there are added or existing files with that error

Screenshot from 2023-03-02 14-33-10

- [x] Documentation

---
  - https://app.asana.com/0/0/1201258662284091


docs/advanced-features/workflows.md
```diff
@@ -623,6 +629,31 @@ Requires a README
 }
 ```
 
+##### `s3://bucket/must-have-foo-bar-meta.json`
```
fiskus, line 632: @sir-sigurd please, check if the docs, and example in particular, are correct

sir-sigurd: I guess I need to drop my example in https://github.com/quiltdata/quilt/pull/3286/files, since your doc seems more complete.

docs/advanced-features/workflows.md
```diff
@@ -623,6 +629,31 @@ Requires a README
...
+
+Requires `{ foo: 'bar' }` object as user specified metadata
```
sir-sigurd, line 634: Maybe format this as JSON object?
```suggestion
Requires `{ "foo": "bar" }` object as user specified metadata
```

docs/advanced-features/workflows.md
```diff
@@ -623,6 +629,31 @@ Requires a README
...
+    "properties": {
+      "meta": {
+        "type": "object",
+        "properties": {
+          "foo": {
+            "type": "string",
+            "pattern": "^bar$"
+          }
+        },
+        "required": ["foo"]
```
sir-sigurd, line 650: Just in case you want it to be exactly `{ "foo": "bar" }`, you'd need to set `"additionalProperties": false` or just use `const`.

sir-sigurd (commented):
Doc looks OK but might make sense to use `logical_key` in example to demonstrate how to validate specific object instead of just some object in package.

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -194,7 +195,13 @@ interface DispatchFilesAction {
...
 }
 
-type FilesEntryState = 'deleted' | 'modified' | 'unchanged' | 'hashing' | 'added'
+type FilesEntryState =
+  | 'deleted'
+  | 'modified'
+  | 'unchanged'
+  | 'hashing'
+  | 'added'
+  | 'invalid'
```
nl0, line 204: to me "invalid" property seems orthogonal to the state

nl0: since, for example, both a modified and an added entry can be invalid in the same time

nl0 (approved):
looks sane.
i'm not quite sure if you're still showing the validation errors "not assigned" to any entries (or are they not expected to be present at all times?). e.g. if you have some error related to several fields or smth like that. also, you only get (`.find()`) the first error matching entry path, but can there be more than one per path and they are not shown?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: I agree, and it was my first attempt to implement invalid micro-state for any other state (added/existing), but it was too difficult

Now, I think it's not that hard (probably, because I forgot complexities), and I would like to fix it as a tech/UX dept in a separate task

fiskus:
> i'm not quite sure if you're still showing the validation errors "not assigned" to any entries 

Yes, all entries validation errors are below FilesInput. Unfortunately, they are not very helpful, if you don't know exact JSON Schema. At the same time, I don't show error message in entry. There is a room for improvement here, I address it in the next task.

Probably, it's not so important to invest a lot of time into entries meta validation on frontend

> you only get (.find()) the first error matching entry path, but can there be more than one per path and they are not shown?

With `.find` I only find the fact, the entry is invalid. Actual errors are below the FilesInput block.

fiskus:
This is not very good UX, of course. For example, if Schema requires `README.md` with meta, but user provide two files `foo.txt` and `bar.txt`, then both files are invalid with two errors each (1. this is not README.md, this doesn't have meta). But actually, files are not invalid, user just need one more file.

docs/advanced-features/workflows.md
fiskus: Got it, but my intention was to have any fields AND require `{ "foo": "bar" }`

fiskus:
Updated docs: added `logical_key`

nl0 (approved):

sir-sigurd (approved):

PR #3337, opened 2023-02-20
Object level user meta

fiskus:
* PR mostly consists of type refactoring, specified type for object-level metadata in model/index.ts and re-used everywhere
* Small change in EditFileMeta, where I get only `user_meta`
* Docs change is small, ~and it will be visibly small after https://github.com/quiltdata/quilt/pull/3338 merge~

- [x] Documentation


nl0 (approved):

fiskus:
Reverted docs changes because related PR https://github.com/quiltdata/quilt/pull/3338 diverged significantly. So, I left only necessary changes in docs.

nl0 (approved):

PR #3353, opened 2023-02-28
Add logout button to "Session expired" page

fiskus:
Screenshot from 2023-02-28 16-35-04

### DIFF
--- a/catalog/app/containers/Errors/FinalBoundary.tsx
+++ b/catalog/app/containers/Errors/FinalBoundary.tsx
@@ -4,6 +4,7 @@ import * as M from '@material-ui/core'
 import * as style from 'constants/style'
 import { createBoundary } from 'utils/ErrorBoundary'
 import { CredentialsError } from 'utils/AWS/Credentials'
+import mkStorage from 'utils/storage'
 
 const useFinalBoundaryStyles = M.makeStyles((t) => ({
   root: {
@@ -15,7 +16,11 @@ const useFinalBoundaryStyles = M.makeStyles((t) => ({
     maxHeight: '600px',
   },
   actions: {
-    marginTop: t.spacing(2),
+    color: t.palette.text.primary,
+    marginTop: t.spacing(4),
+  },
+  divider: {
+    margin: t.spacing(0, 2),
   },
   header: {
     color: t.palette.text.primary,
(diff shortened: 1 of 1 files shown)


robnewman:
@fiskus The `get_credentials` endpoint returns a JSON payload with specific errors. We should display those.

nl0 (commented):
if i see "session expired" message, clicking "restart session" seems like a logical step, but in this case it's somewhat misleading.

i think we should change the heading to "Error getting credentials" or smth,
and add a little description that it's probably because of an expired session and that a page reload can possibly fix this -- a logical explanation leading to the "reload" button.
we could also add that if a page reload doesn't help, try restarting your session (logging out and back again) -- leading to the second CTA (consider adjusting the text, making it smth simpler like "sign out"?).

also we should probably show the original error message from the exception, as several people already mentioned, though maybe we want it hidden under some sort of spoiler like "See error details" idk

fiskus:
> also we should probably show the original error message from the exception, as several people already mentioned, though maybe we want it hidden under some sort of spoiler like "See error details" idk

It's impossible without refactoring, and I'm working on it. The original error is in the RegistryCredentials class instance that doesn't exist at the moment FinalBoundary catches it.

fiskus:
I think we can change the title to "Failed to get credentials" and leave button titles "Reload page"/"Restart session".

fiskus:
Additional descriptions can lead to an even more confusion. Because we don't know what's happened for sure and should choose more or less abstract descriptions, that could scare users

nl0:
> > also we should probably show the original error message from the exception, as several people already mentioned, though maybe we want it hidden under some sort of spoiler like "See error details" idk
> 
> It's impossible without refactoring, and I'm working on it. The original error is in the RegistryCredentials class instance that doesn't exist at the moment FinalBoundary catches it.

makes sense. isn't the original error attached as `.originalError` somewhere?

nl0:
> > > also we should probably show the original error message from the exception, as several people already mentioned, though maybe we want it hidden under some sort of spoiler like "See error details" idk
> > 
> > 
> > It's impossible without refactoring, and I'm working on it. The original error is in the RegistryCredentials class instance that doesn't exist at the moment FinalBoundary catches it.
> 
> makes sense. isn't the original error attached as `.originalError` somewhere?

ok i see that it's not

fiskus:
Error lives in RegistryCredentials, but we throw CredentialsError when there is no RegistryCredentials already
https://github.com/quiltdata/quilt/blob/master/catalog/app/utils/AWS/Credentials.js#L82

fiskus:
It's a combination of errors
We catch original error, then use ErrorBoundary that shows `Layout` that depends on S3 credentials to show some elements. And we try to get credentials again, and now show FinalBoundary when there is no original error

nl0:
> It's a combination of errors We catch original error, then use ErrorBoundary that shows `Layout` that depends on S3 credentials to show some elements. And we try to get credentials again, and now show FinalBoundary when there is no original error

so, a proper fix will be smth like not calling `useCredentials` in an element without a proper context, and probably adding an error boundary into an appropriate place in the component tree to catch and display the original credentials errors?

fiskus:
Above steps are not a fix, actually, but providing more debug/error info and fail more gracefully

PR #3365, opened 2023-03-07
WIP

dimaryaz:


api/python/quilt3/packages.py
```diff
@@ -119,6 +119,20 @@ def _write_data_yaml(groups):
...
+    try:
+        existing_lines = path.read_text().splitlines()
+    except FileNotFoundError:
+        existing_lines = []
+
+    existing_set = set(existing_lines)
+
+    for line in lines:
+        if line not in existing_set:
+            existing_lines.append(line)
```
akarve, line 133: Maybe we add a comment `# generated by quilt3`

drernie:
Sigh.  I'm still sad we didn't do this, but we need to move on.

PR #3375, opened 2023-03-20
Quilt3 v5.2.0

sir-sigurd:


robnewman:
Let's wait on merging of the final language changes from Max for the permissive HTML file rendering before we release.

robnewman (approved):

PR #3385, opened 2023-03-24
Markdown html process links images

fiskus:
* Added 'processed' data attribute to links and images, processed with markdown
* Process the rest of the images and links (written using HTML) with the same functions


catalog/app/components/Markdown/Markdown.tsx
```diff
@@ -189,7 +189,7 @@ const linkHandler =
       const rel = nofollow ? ' rel="nofollow"' : ''
       return `<a href="${(Remarkable as unknown as RemarkableWithUtils).utils.escapeHtml(
```
nl0, line 190: let's adjust the formatting here a little bit: move `(Remarkable as unknown as RemarkableWithUtils).utils.escapeHtml(...` to a separate line and assign it to a variable for readability (i know it's kinda out of scope, but anyways, let's "fix" it since we're touching it)

catalog/app/components/Markdown/Markdown.tsx
```diff
@@ -219,6 +246,8 @@ export const getRenderer = memoize(({ images, processImg, processLink }) => {
   md.use(imageHandler({ disable: !images, process: processImg }))
```
nl0, line 246: since we're handling all images and links in the purify's hooks, we can (and should) remove this logic from markdown handlers so we're not doing it twice

nl0 (approved):
looks good, and the solution (using purify's hooks) is very smart, i haven't thought about that. some minor suggestions inline

catalog/app/components/Markdown/Markdown.tsx
fiskus: This makes sense, but I have some resistance, because first we do "right" handling using some AST, and then "hacky" handling using HTML elements mutation

nl0: well, yeah, but
1. we do that "hacky" thing anyways
2. the "right" way doesn't handle html insertions, which are a supported part of the markdown standard, so it's incomplete

also, remarkable's AST thing looks like a joke TBH with all these string manipulations and lists of tokens...

fiskus:
* Replaced processing of the markdown by `dompurify` hook, and made it in one loop cycle
* Dropped most of the escaping, because `setAttribute` is safe enough
* Simplified data structure of the link/image processors: `({ href: string }) => { href: string }` --> `(href: string) => string`
* Removed unused functions and component arguments

catalog/app/components/Markdown/Markdown.tsx
```diff
@@ -113,114 +112,86 @@ const highlight = (str: string, lang: string) => {
...
-      const title = t.title ? ` title="${escape(t.title)}"` : ''
-      return `<img src="${src}" alt="${alt}"${title} />`
-    }
+  const alt = element.getAttribute('alt')
+  if (alt) {
+    element.setAttribute('alt', unescapeMd(alt))
+  }
+  const title = element.getAttribute('title')
+  if (title) {
+    element.setAttribute('title', title)
```
nl0, line 141: what's the purpose of getting the attribute and then setting it to the same value?

catalog/app/components/Markdown/Markdown.tsx
```diff
@@ -113,114 +112,86 @@ const highlight = (str: string, lang: string) => {
...
-      return `<a href="${(Remarkable as unknown as RemarkableWithUtils).utils.escapeHtml(
-        t.href,
-      )}"${rel}${title}>`
-    }
+  return element
+}
+
+function handleLink(process: AttributeProcessor, element: HTMLElement): Element {
+  const attributeValue = element.getAttribute('href')
+  if (!attributeValue) return element
```
nl0, line 149: if we short-circuit here, `setAttribute` calls below won't be executed. is this intentional?

catalog/app/components/Markdown/Markdown.tsx
fiskus: It was leftover, when there was an escaping
I suppose, now, we don't need an escaping because we don't use strings for creating HTML elements

catalog/app/components/Markdown/Markdown.tsx
fiskus: Yes, it was. But I found that we process `href=""` (as relative path to the root of the package), so I changed this condition to `if (typeof attributeValue !== 'string')`

catalog/app/components/Markdown/Markdown.tsx
```diff
@@ -272,21 +243,19 @@ export function Container({ className, children }: ContainerProps) {
...
-  images?: boolean
-  processImg?: () => $TSFixMe
-  processLink?: () => $TSFixMe
+  processImg?: AttributeProcessor
+  processLink?: AttributeProcessor
 }
 
 export default function Markdown({
   data,
-  images = true,
```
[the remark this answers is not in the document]
fiskus: Yes, it's never set to `false`, only `true` or omitted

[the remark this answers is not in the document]
fiskus: It was `false` for
a/catalog/app/containers/Package/Comment.js
b/catalog/app/containers/Package/CommentForm.js

nl0 (approved):

nl0:
it would be nice to add some tests if you have enough capacity and enthusiasm

fiskus:
I'm going to add unit-test for the `getRenderer` in the follow-up PR with slightly changed API for `Markdown#getRenderer`

What is really worth is unit-test for `Markdown` component, but, yeah, this is above my capacity :)

PR #3390, opened 2023-03-31
Refactor GraphQL querying APIs, fix package push failing for unrelated reason

nl0:
- fix package push failing for unrelated reason (missing access counts, for example)
- add `useMutation` wrapper with built-in error handling
  - don't throw on errors when data is present
  - automatically log errors and send them to Sentry
- add the same error handling behavior when calling `fold` for a result of `useQuery`
- add `useQueryS` helper for using queries more conveniently with Suspense
- `.case()` method -> `fold()` / `foldC()` functions

### TODO

- [x] add return types and comments
- [x] changelog

---
  - https://app.asana.com/0/0/1204196412300066


catalog/app/utils/GraphQL/wrappers.ts
```diff
@@ -0,0 +1,145 @@
...
+    const err = result.error || new QueryError(result)
+    if (error) return error(err, result)
+    if (!silent) {
+      log.error('Critical error while executing the query:', err)
+      Sentry.captureException(err, { extra: { result } })
+    }
+    throw error
+  }
+
+export const fold = <Data, OnData, OnFetching, OnError = never>(
```
fiskus, line 68: I think that for core functions like these we can provide a return type as well, even typescript can guess it.
Because it will be some sort of documentation

catalog/app/utils/GraphQL/wrappers.ts
```diff
@@ -0,0 +1,145 @@
...
+  result: urql.UseQueryState<Data, any>,
+  casesAndOpts: FoldOptions<Data, OnData, OnFetching, OnError>,
+) => foldC(casesAndOpts)(result)
+
+export function useQuery<Data, Variables = {}>(
```
fiskus, line 73: In the PR's description:
> add the same error handling behavior for useQuery

But I don't see any error handling here. Maybe, it's a typo: `useQuery` -> `useQueryS`

fiskus (dismissed):
I'm sure code is good. And I see error handling and simplified `useQuery` arguments interface as improvement. But it's difficult to catch everything what is happening under the hood, so, for me, it's worth marking some usage of the API as reference, so I can copy that and use.

---
As usual, I'll grunt about abbreviations, though OK I can get used to them :)
I just want to say that `useQuerySuspended` or `foldConstructor` is much easier to read and spot the difference rather than `useQuery(SAMPLE)/useQueryS(SAMPLE)`

And, I'm not sure `fold` is a good term, if the meaning is something different from `fold/reduce` from functional languages.

catalog/app/utils/GraphQL/wrappers.ts
```diff
@@ -0,0 +1,145 @@
...
+  fetching: (result: urql.UseQueryState<Data, any>) => OnFecthing
+  error?: (
+    error: urql.CombinedError | QueryError<Data>,
+    result: urql.UseQueryState<Data, any>,
+  ) => OnError
+  partial?: boolean
+  silent?: boolean
+}
+
+export const foldC =
```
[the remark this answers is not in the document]
fiskus: I suppose, `C` is for `Constructor`

nl0:
> I'm sure code is good. And I see error handling and simplified `useQuery` arguments interface as improvement. But it's difficult to catch everything what is happening under the hood, so, for me, it's worth marking some usage of the API as reference, so I can copy that and use.

you can refer to the existing calls throughout the application, they have all been "canonized" =)

> As usual, I'll grunt about abbreviations, though OK I can get used to them :) I just want to say that `useQuerySuspended` or `foldConstructor` is much easier to read and spot the difference rather than `useQuery(SAMPLE)/useQueryS(SAMPLE)`

yeah, maybe `Suspended` is more descriptive, but it's just soooo cumbersome =)
i think maybe we can add some dosctrings just to decipher the abbreviations.

> `foldConstructor`

`C` is for "curried" =)
it was a convention in some functional libraries i saw, so it seems quite obvious to me

> And, I'm not sure `fold` is a good term, if the meaning is something different from `fold/reduce` from functional languages.

it's kinda related to this concept, since it comes from the `Foldable` typeclass in io-ts, here's an example what it mean for RemoreData type: https://github.com/devexperts/remote-data-ts#how-to-fold-unwrap-your-data-from-remotedata

catalog/app/utils/GraphQL/wrappers.ts
nl0: i can see the return types by using the "hover" functionality of the language server, it's deduced by TS automatically, and when using a language server or some doc generators, the type should be displayed just fine. tho i agree that it may be a good idea to explicitly add some return type annotations for documentation purposes

nl0:
> Oh, this was unexpected. I worked in DevExperts for one year back in 2009-2010 oO

oh cool)

fiskus (approved):
Awesome, thanks!

nl0:
@fiskus i've merged w/ master and added a changelog entry, pls re-approve

fiskus (approved):

nl0:
@fiskus some last-minute changes

fiskus (approved):

PR #3392, opened 2023-04-03
`Markdown#getRenderer` unit test

fiskus:
* Added `jsdom`
* Lot of new entries in package-lock.json related to `jsdom` and `jest-environment-jsdom`


catalog/setup-jest.ts
```diff
@@ -11,3 +11,5 @@
   mixpanelToken: '',
   passwordAuth: 'ENABLED',
 }
+;(window as any).TextEncoder = require('util').TextEncoder
```
nl0, line 14: why do we need this? pls add a comment

nl0 (approved):

catalog/setup-jest.ts
fiskus: "It's required for jsdom" - is that enough? I don't know real reason why is it required for jsdom, and why it doesn't work out of the box even we set `"testEnvironment": "jsdom"`

nl0 (approved):

PR #3396, opened 2023-04-04
Render status report if no `bucketConfig`

fiskus:
I don't know why, but there is no `bucketConfig` for bucket with status reports

---
  - https://app.asana.com/0/0/1204325167861344


nl0:
> I don't know why, but there is no bucketConfig for bucket with status reports

`bucketConfig` corresponds to the underlying `Bucket` db model, and for utility buckets we don't have those

nl0 (approved):

PR #3397, opened 2023-04-04
Set a cookie for /browse when logging in

dimaryaz:


dimaryaz:
The cookie needs to be set just once for a user session. I didn't want to set it for every iframe since it's unnecessary, so I'm setting it when the user logs in - but not sure if that's the right place.

catalog/app/containers/Auth/saga.js
```diff
@@ -328,6 +345,7 @@ function* handleSignIn(
     const tokensRaw = yield call(signIn, credentials)
     const tokens = adjustTokensForLatency(tokensRaw, latency)
     const user = yield call(fetchUser, tokens)
+    yield call(setBrowseCookie, tokens)
```
nl0, line 348:
```suggestion
    yield fork(setBrowseCookie, tokens)
```

we don't need to block the execution here, `fork` sends it to the background

catalog/app/containers/Auth/saga.js
```diff
@@ -468,6 +486,7 @@ function* handleAuthLost({ forgetTokens, forgetUser, onAuthLost }, { payload: er
 function* handleSignUp({ payload: credentials, meta: { resolve, reject } }) {
   try {
     yield call(signUp, credentials)
+    // TODO: yield call(setBrowseCookie, tokens)
```
nl0, line 489: probably not required here

catalog/app/containers/Auth/saga.js
```diff
@@ -150,6 +150,23 @@ function* signIn(credentials) {
   }
 }
 
+function* setBrowseCookie(tokens) {
```
nl0, line 153: i think maybe we better should put this into `storeTokens()` (and maybe even unset the cookie in `forgetTokens`?) in `Provider.js`

catalog/app/containers/Auth/saga.js
dimaryaz: I'm curious - how does sign up work? (I'm assuming, this is only password signup for public stacks.) Shouldn't it call storeTokens, etc. at some point?

catalog/app/containers/Auth/saga.js
dimaryaz: Hmm. Do you think it's a good idea to make API calls from `forgetTokens`?

As for `forgetTokens`: we could unset it... though as long as the `signOut` call succeeds, the cookie will become invalid anyway.

catalog/app/containers/Auth/saga.js
nl0: afair it creates the user and then shows the message to check your email to complete the registration, and then you can sign-in

catalog/app/containers/Auth/saga.js
nl0: > Hmm. Do you think it's a good idea to make API calls from storeTokens?

maybe it's not perfect, but i think it's the right place, since setting cookies is conceptually similar to writing to the local storage, even though the actual implementation (making an http request) is different

> As for forgetTokens: we could unset it... though as long as the signOut call succeeds, the cookie will become invalid anyway.

well, i guess that's ok

catalog/app/containers/Auth/saga.js
```diff
@@ -150,6 +151,22 @@ function* signIn(credentials) {
...
+  try {
+    yield call(apiRequest, {
+      auth: { tokens, handleInvalidToken: false },
+      url: `${cfg.s3Proxy}/browse/set_browse_cookie`,
+      method: 'POST',
+      credentials: 'include',
+    })
+  } catch (e) {
```
nl0, line 162: i guess we can just `log.warn` and `Sentry.captureException` it, since the error is never awaited anyways

catalog/app/containers/Auth/saga.js
nl0: your approach would work just fine, nevermind

nl0 (dismissed):

catalog/app/containers/Auth/saga.js
dimaryaz: Oh right

catalog/app/containers/Auth/saga.js
```diff
@@ -150,6 +152,19 @@ function* signIn(credentials) {
...
+    log.warn('Unable to set browse cookie', e)
```
nl0, line 164: let's add `Sentry.captureException(e)` here, because this is something we should be aware of

nl0 (approved):

PR #3402, opened 2023-04-07
Multipart checksums / faster hashing (lambdas)

nl0:
## Description

Rewrite the `s3hash` and `pkgpush` lambdas to compute new-style checksums leveraging S3 checksum computation capabilities.

## TODO

- [x] update `quilt_shared` hash after it's merged into master #3849 
- [x] Changelog entry (bootstrap per-lambda changelogs)
- `pkgpush`
  - [x] depend on `quilt_shared`
  - [x] ~consider splitting the code into multiple files
  - [x] unit tests
  - [x] update the description after benchmarking the new code
  - [x] clean up logging
  - [x] move large request handler to shared
  - [x] remove PackageFromFolder code if it's no longer used in the catalog
- `s3hash`
  - [x] get existing checksum when present and compliant
  - [x] use `quilt_shared`
  - [x] fix computing legacy checksums (MULTIPART_CHECKSUMS == False) for files larger than 5 GiB (we currently support files up to 10 GiB)
  - [x] ~ abort upload before lambda times out (or rely on lifecycle policy for cleaning that up)?
  - [x] ~ use ETag to prevent bad hashes on non-versioned buckets
  - [x] get rid of `target` parameter?
  - [x] get rid of `stats` in results? or put boto retries there?
  - [x] don't use per-bucket MPUs
  - [x] wrap timeout error produced by `wait_for()`?
  - [x] use boto retries
  - [x] clean up logging
  - [x] expose perf (timing) stats for benchmarking
  - [x] tune concurrency based on benchmark results
  - [x] unit tests
- Open questions
  - [x] how to reuse code (at least type definitions and constants) between the registry and the lambdas?
    - there's now a `quilt_shared` module (residing in `py-shared`) dir which can be installed from a github url


nl0:
i believe this is ready for an initial review.
i wonder how we can conveniently reuse python code between the lambdas/quilt3/registry codebases

sir-sigurd (commented):
If I'm not missing something s3hash lambda will fail when `MULTIPART_CHECKSUMS` is disabled for files larger than 5 GiB (we currently support files up to 10 GiB).

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -4,219 +4,99 @@
 benchmarks was about 75 MiB/s. To overcome this limitation this function
 concurrently invokes dedicated hash lambda for multiple files.
 """
+# TODO: adjust the decsription above
```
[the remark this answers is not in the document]
nl0: TODO: adjust the TODO above

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -228,138 +108,196 @@ def from_quilt_exception(cls, qe: quilt3.util.QuiltException):
...
     @functools.wraps(f)
-    def wrapper(event, _context):
+    def wrapper(event, context):
+        # XXX: make sure to disable in production to avoid leaking credentials
+        logger.info("event: %s", event)
+        logger.info("context: %s", context)
         try:
-            return {"result": f(event)}
+            result = f(event)
+            logger.info("result: %s", result)
```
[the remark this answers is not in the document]
nl0: yeah probably, if we want to keep that after stabilizing (testing/benchmarking)

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -388,26 +327,68 @@ def get_registry(registry_url):
...
+    copy_data: bool = successor_params.get("copy_data", True)
 
     try:
-        pkg = get_pkg(src_registry, data)
+        pkg = get_pkg(src_registry)
         if copy_data:
             total_size = 0
             total_files = 0
             for lk, e in pkg.walk():
+                assert isinstance(e.size, int)
```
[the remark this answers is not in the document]
nl0: for type-checker mostly

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -475,176 +474,213 @@ def get_pkg(src_registry, data):
...
-                pkg[logical_key]._meta = meta or {}
+                pkg.set(entry.logical_key, str(physical_key))
+                pkg_entry = pkg[entry.logical_key]
+                assert isinstance(pkg_entry, quilt3.packages.PackageEntry)
+                pkg_entry._meta = entry.meta or {}
 
-                size_to_hash += pkg[logical_key].size
+                assert isinstance(pkg_entry.size, int)
+                size_to_hash += pkg_entry.size
                 if size_to_hash > PKG_FROM_FOLDER_MAX_PKG_SIZE:
```
nl0, line 644: i'm not sure why this constant is used here 🤔

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+SCRATCH_KEY_PER_BUCKET = ".quilt/.checksum-upload-tmp"
+
+
+S3: contextvars.ContextVar[S3Client] = contextvars.ContextVar("s3")
+
+
+NonEmptyStr = pydantic.constr(min_length=1, strip_whitespace=True)
+
+
+class AWSCredentials(pydantic.BaseModel):
```
nl0, line 41: XXX: move to shared and reuse between lambdas?

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+    key: NonEmptyStr
+    secret: NonEmptyStr
+    token: NonEmptyStr
+
+
+# Isolated for test-ability.
+def get_user_boto_session(
```
nl0, line 48: XXX: move to shared and reuse?

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+    def __init__(self, name, context=None):
+        super().__init__(name, context)
+        self.name = name
+        self.context = context
+
+    def dict(self):
+        return {"name": self.name, "context": self.context}
+
+
+class ChecksumType(str, enum.Enum):
```
nl0, line 77: XXX: reuse

lambdas/s3hash/index.py
```diff
@@ -1,28 +1,388 @@
...
+# XXX: move decorators to shared?
+# XXX: move reusable types/models to shared?
+@lambda_wrapper
+@pydantic.validate_arguments
+async def lambda_handler(
+    *,
+    credentials: AWSCredentials,
+    location: S3ObjectSource,
+    target: T.Optional[S3ObjectDestination] = None,
+    concurrency: T.Optional[pydantic.PositiveInt] = None,
```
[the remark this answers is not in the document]
nl0: i was planning to use that for benchmarking to find optimal-ish concurrency

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
sir-sigurd: I think at some point its meaning became `MAX_BYTES_TO_HASH`.

sir-sigurd:
>  how to reuse code (at least type definitions and constants) between the registry and the lambdas?

move lambdas closer to registry?

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+                size=actual_part_size,
+            )
+        )
+        offset += actual_part_size
+        part_number += 1
+
+    return parts
+
+
+@tenacity.retry
```
sir-sigurd, line 236: Maybe configure [boto's retries](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html) instead?

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+) -> aiobotocore.session.AioSession:
+    s = aiobotocore.session.get_session()
+    s.set_credentials(credentials.key, credentials.secret, credentials.token)
+    return s
+
+
+@contextlib.asynccontextmanager
+async def aio_context(credentials: AWSCredentials, concurrency: pydantic.PositiveInt):
+    session = get_user_boto_session(credentials)
+    config = aiobotocore.config.AioConfig(max_pool_connections=concurrency)
```
sir-sigurd, line 59: I think you want to limit the number of concurrent requests, but IIUC this just limits number of connections **kept** in the pool i.e. you still can get more than `max_pool_connections` concurrent request, so I think you'd want to use [`semaphore`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore) somewhere.

lambdas/s3hash/index.py
```diff
@@ -1,28 +1,388 @@
...
+    ]
+    if target is not None:
+        dsts.insert(0, target)
+    return dsts
+
+
+MPUDstError = T.Tuple[S3ObjectDestination, botocore.exceptions.ClientError]
+
+
+async def create_mpu(src: S3ObjectSource, target: T.Optional[S3ObjectDestination]) -> MPURef:
```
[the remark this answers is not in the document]
sir-sigurd: Ideally we should also make sure we abort upload before lambda times out.

nl0:
> > how to reuse code (at least type definitions and constants) between the registry and the lambdas?
> 
> move lambdas closer to registry?

not sure how exactly to achieve this. just move to the enterprise repo? this has some downsides as well. maybe just "installing" packages from the quilt repo would be enough

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: well, limiting max_pool_connections fixes "dropouts" (reduces their number to something acceptable at least), so...

> use [semaphore](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore) somewhere

probably this is a good idea

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: yeah, looks promising, i didn't know about it

lambdas/s3hash/index.py
```diff
@@ -1,28 +1,388 @@
...
+
+    def __str__(self):
+        return f"{self.type}:{self.value}"
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self!s})"
+
+
+class S3ObjectSource(pydantic.BaseModel):
+    bucket: str
```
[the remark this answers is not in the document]
nl0: it probably does, yeah

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+    upload_data = await S3.get().create_multipart_upload(
+        **target.boto_args,
+        ChecksumAlgorithm="SHA256",
+    )
+    return MPURef(bucket=target.bucket, key=target.key, id=upload_data["UploadId"])
+
+
+def get_mpu_dsts(src: S3ObjectSource, target: T.Optional[S3ObjectDestination]) -> T.List[S3ObjectDestination]:
+    dsts = [
+        S3ObjectDestination(bucket=src.bucket, key=SCRATCH_KEY_PER_BUCKET),
```
sir-sigurd, line 268: That's a bit risky optimization. It looks like we can't strongly guarantee we call `abort_multipart_upload()` and customers can have buckets without lifecycle policy that aborts uploads automatically (though I think it's advised by AWS), so unfinished uploads will stay forever burning some money. I *think* most customers have their bucket in the same region, it might be wiser to remove this optimization.

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -228,138 +109,196 @@ def from_quilt_exception(cls, qe: quilt3.util.QuiltException):
...
+        max_workers=S3_HASH_LAMBDA_CONCURRENCY
+    ) as pool:
+        credentials = AWSCredentials.from_boto_session(user_boto_session.get())
         fs = [
-            pool.submit(calculate_pkg_entry_hash, get_client_for_bucket, entry)
+            pool.submit(calculate_pkg_entry_hash, entry, credentials)
             for entry in entries
         ]
         for f in concurrent.futures.as_completed(fs):
             f.result()
```
sir-sigurd, line 249: Not quite related to this PR, but I think we can finish faster if we schedule larger tasks first i.e. if we sort `entries` by `physical_key.size`.

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
sir-sigurd: > well, limiting max_pool_connections fixes "dropouts" (reduces their number to something acceptable at least), so...

you're not limiting it but increasing it from default 10
https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html
that can reduce "dropouts", because it needs to open less new connections so there is less overhead

btw what kind of errors did you get?

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: ok, that makes sense

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: > you're not limiting it but increasing it from default 10

i mean i was experimenting with that limit

> btw what kind of errors did you get?

i don't remember rn, but most likely hitting some local connection limits, not AWS rate-limiting or smth

lambdas/s3hash/Dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG FUNCTION_DIR="/function"
+
+FROM python:3.8-slim-bullseye@sha256:5f399249dc208063801e3316bcf98b927be8702ca29023317b792b2bc7c52884 AS base-image
```
sir-sigurd, line 3: I think we prefer AWS based images now, so probably should use https://hub.docker.com/r/amazon/aws-lambda-python/tags

sir-sigurd:
we should add `GetObjectAttributes` and `GetObjectVersionAttributes` [here](https://github.com/quiltdata/quilt/blob/53e4b1784949324eb8f5b152ebd48a07030a43e4/docs/CrossAccount.md?plain=1#L36)

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,416 @@
...
+
+@contextlib.asynccontextmanager
+async def aio_context(credentials: AWSCredentials, concurrency: pydantic.PositiveInt):
+    session = get_user_boto_session(credentials)
+    config = aiobotocore.config.AioConfig(max_pool_connections=concurrency)
+
+    async with session.create_client("s3", config=config) as s3:
+        s3_token = S3.set(s3)
+        yield
+        S3.reset(s3_token)
```
sir-sigurd, line 64: shouldn't we do
```suggestion
        try:
            yield
        finally:
            S3.reset(s3_token)
```
as in the [example](https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager)?

nl0: yeah we probably should

py-shared/quilt_shared/__init__.py
sir-sigurd, line 1: I think I prefer [src-layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout) more.

py-shared/quilt_shared/pkgpush.py
```diff
@@ -0,0 +1,122 @@
...
+
+class TopHash(pydantic.ConstrainedStr):
+    min_length = 64
+    max_length = 64
+    regex = r"^[0-9a-f]+$"
+    strip_whitespace = True
+    to_lower = True
+
+
+class S3ObjectSource(pydantic.BaseModel):
```
sir-sigurd, line 25: Maybe move this and other types to aws?

py-shared/requirements.txt
sir-sigurd, line 1: Do we really need this?

py-shared/quilt_shared/pkgpush.py
nl0: maybe, tho now they only make sense in this context

py-shared/requirements.txt
nl0: for testing purposes mostly

py-shared/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+[project.optional-dependencies]
+pydantic = [
+  "pydantic ~= 1.10",
+  "typing-extensions ~= 4.9",
+]
+boto = [
+  "boto3-stubs[s3] ~= 1.34",
+  "boto3 ~= 1.34",
+  "types-aiobotocore[s3] ~= 2.11",
+]
```
sir-sigurd, line 22: do we need these in production?

nl0: well... no. these extra groups are still open to change, it's just the first approximation

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -4,197 +4,90 @@
...
-    'required': ['logical_key', 'physical_key'],
-}
-
-
-s3 = boto3.client('s3')
-lambda_ = boto3.client('lambda', config=botocore.client.Config(read_timeout=S3_HASH_LAMBDA_READ_TIMEOUT))
+S3_HASH_LAMBDA_CONCURRENCY = int(os.environ["S3_HASH_LAMBDA_CONCURRENCY"])
+S3_HASH_LAMBDA_MAX_FILE_SIZE_BYTES = int(
+    os.environ["S3_HASH_LAMBDA_MAX_FILE_SIZE_BYTES"]
+)
```
sir-sigurd, line 65: how about using 119 line length with black?

nl0: sure, we just need to configure that somewhere

lambdas/pkgpush/setup.py
```diff
@@ -1,8 +1,19 @@
...
     packages=find_packages(where="src"),
     package_dir={"": "src"},
+    install_requires=[
+        "awslambdaric>=2,<3",
+        "boto3~=1.28",
+        "pydantic~=1.10",
+        "quilt3~=5.4",
+        (
+            "quilt_shared[pydantic,boto,quilt] @ git+https://github.com/quiltdata/quilt@"
+            "e95cde71e45458c5eca4aa521678b6ba75b6528f"
```
sir-sigurd, line 15: If we need/want to pin to the specific commit, maybe we should create a separate repo? 🤔

nl0: i don't mind in principle, but seems like unnecessary work. what would we gain?

nl0: like, yeah, i get it, the workflow is somewhat cumbersome. maybe at least make shared vs lambdas changes in different PRs and merge shared first, then lambdas?

sir-sigurd: > what would we gain?

less CI mess

py-shared/quilt_shared/__init__.py
nl0: why?

sir-sigurd: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout

> This layout is very handy when you wish to use automatic discovery, since you don’t have to worry about other Python files or folders in your project root being distributed by mistake. In some circumstances it can be also less error-prone for testing or when using [PEP 420](https://peps.python.org/pep-0420/)-style packages.

also 
https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -0,0 +1,374 @@
...
+
+    async with session.create_client("s3", config=config) as s3:
+        s3_token = S3.set(s3)
+        try:
+            yield
+        finally:
+            S3.reset(s3_token)
+
+
+class S3hashException(Exception):
```
sir-sigurd, line 73: Maybe extract base class for this and PkgpushException to shared? Handling boto errors also makes sense here.

nl0: ok i don't mind

nl0: i'll do it

nl0: done

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -228,138 +92,131 @@ def from_quilt_exception(cls, qe: quilt3.util.QuiltException):
...
+            logger.exception("RequestTooLarge")
+            return {
+                "error": {
+                    "name": "RequestTooLarge",
+                    "context": {
+                        "size": e.size,
+                        "max_size": e.max_size,
+                    },
+                }
+            }
```
sir-sigurd, line 204: Might make sense to move that inside nested try and re-raise as PkgpushException.
Also might make sense to add types for result/error dicts especially if we're going to instantiate error dict explicitly.

lambdas/s3hash/tests/test_compute_checksum.py
```diff
@@ -140,7 +113,148 @@ async def test_legacy(s3_stub: Stubber):
...
+        "TestError",
+        expected_params=EXPECTED_MPU_PARAMS,
+    )
+
+    error = None
+    res = None
+    try:
+        res = await s3hash.compute_checksum(LOC)
+    except s3hash.LambdaError as e:
+        error = e
```
sir-sigurd, line 144: I think you could use [`pytest.raises()`](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions).

sir-sigurd (dismissed):

sir-sigurd (approved):

PR #3403, opened 2023-04-07
Multipart checksums / faster hashing (catalog)

nl0:
Opt-in support of multipart checksums

# TODO

- [x] bump package size limits if multipart checksums are enabled
- [x] treat files from s3 as unchanged if same physical key -- **PUNT**: out of scope
- [x] check upload part size settings and harmonize with boto/quilt3 if they differ
- [x] enable checksums for uploads -- **PUNT**: requires using aws-js-sdk v3
- [x] wait til the backend changes are in


catalog/app/utils/checksums.ts
```diff
@@ -0,0 +1,91 @@
...
+const blobLimit = pLimit(8)
+
+async function hashBlob(blob: Blob) {
+  const buf = await blob.arrayBuffer()
+  return window.crypto.subtle.digest('SHA-256', buf)
+}
+
+const hashBlobLimit = (blob: Blob) => blobLimit(hashBlob, blob)
+
+async function computeFileChecksum(f: File): Promise<Model.PackageEntryHash> {
```
fiskus, line 66: I'd pass `window.crypto.subtle` as an argument to simplify unit-test, because node environment has a slightly different crypto module

fiskus: I mean, `computeFileChecksum(file, window.crypto.subtle)` in browser and `computeFileChecksum(file, crypto.webCrypto.subtle)` in test environment

fiskus (commented):
It looks good, but I want to be confident that *I* understand everything. I'll read one more time tomorrow, to be sure that algorithm is clear for me

catalog/app/utils/checksums.ts
fiskus: But it just me, I don't like to mock `window` or anything else for tests :)

catalog/app/utils/checksums.ts
```diff
@@ -0,0 +1,91 @@
...
+    .map((b) => b.toString(16).padStart(2, '0'))
+    .join('')
+
+const singlepart = (value: ArrayBuffer): Model.PackageEntryHash => ({
+  value: toHex(value),
+  type: Model.HASH_TYPE_SP,
+})
+
+const multipart = (value: ArrayBuffer, partCount: number): Model.PackageEntryHash => ({
+  value: `${B64.fromUint8Array(new Uint8Array(value))}-${partCount}`,
```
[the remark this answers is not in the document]
nl0: that works, tho i had to configure the Buffer polyfill

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -306,7 +304,7 @@ const computeEntries = ({
           // eslint-disable-next-line no-nested-ternary
           state = !a.hash.ready
             ? ('hashing' as const)
-            : a.hash.value === hash
+            : R.equals(a.hash.value, hash)
```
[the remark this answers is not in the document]
nl0: nope, they are `Model.PackageEntryHash` (which is `{type, value}`)

catalog/app/utils/checksums.ts
nl0: i'd just mock that if/when necessary

fiskus (approved):
If backend has unit-tests, it would be good porting them for `app/utils/checksums`. I mean, it would be useful, to have checksums tests with the same values as for backend tests.

fiskus:
Btw, I can write tests later. It will give me confidence, that I'm familiar with this part of code.

nl0:
> If backend has unit-tests

it does not (yet)

> it would be good porting them for app/utils/checksums. I mean, it would be useful, to have checksums tests with the same values as for backend tests.

yep i agree

> Btw, I can write tests later. It will give me confidence, that I'm familiar with this part of code.

cool

fiskus (approved):

PR #3405, opened 2023-04-10
Dropzone only package dialog

fiskus:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation


fiskus:
I'm going to decouple some changes to separate PRs for easier review

fiskus:
`hash` → `hashOrTag`: https://github.com/quiltdata/quilt/pull/3430

docs/Catalog/PackageDialog.md
```diff
@@ -0,0 +1,56 @@
+# Package dialog state
+
+You can use permanent links to the package dialog with pre-defined state.
```
drernie, line 3:
```suggestion
# Opening package dialog via a URL

You can use the 'createPackage' query parameter to create a URL that opens the Quilt Catalog directly to the package dialog with 
certain information already filled in.
```

catalog/app/containers/Bucket/Queries/Athena/CreatePackage.tsx
```diff
@@ -122,26 +122,21 @@ const useStyles = M.makeStyles((t) => ({
...
-    delayHashing: true,
-    disableStateDisplay: true,
-  })
+  const createPackage = useCreatePackage()
   const handleConfirm = React.useCallback(
     (ok: boolean) => {
       if (!ok) return
       addToPackage?.merge(entries.valid)
-      createDialog.open()
+      createPackage?.open()
```
nl0, line 137: this would just silently fail if `createPackage` returns null/undefined. since this can only happen when there's a programmer error (uninitialized context), i think we can add assert/invariant after `const createPackage = useCreatePackage()` to ensure the context is initialized, e.g. `invariant(createPackage, "Package Dialog context wasn't initialized")`

nl0: (this invariant can be added directly to the `use` function so that its return value is not nullable)

docs/Catalog/PackageDialog.md
```diff
@@ -0,0 +1,58 @@
+# Opening package dialog via a URL
+
+You can use the `createPackage` query parameter to create a URL that opens
+the Quilt Catalog directly to the package dialog with certain information
+already filled in.
+
+Available parameters are
+
+* `createPackage=true` to open a package dialog,
```
nl0, line 9: i'd prefer some shorter lowercase params like `edit=` and `dropzone=`, but i guess it would be more ambiguous, so whatever

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
```diff
@@ -724,59 +823,56 @@ interface PackageCreationDialogUIOptions {
 }
 
 interface UsePackageCreationDialogProps {
```
nl0, line 825: this should be renamed to `PackageCreationDialogProps`, since it's no longer related to a hook

nl0 (commented):
looks good, but there's a lot of changes and i can't fully digest them, so i just hope that works =)

catalog/app/containers/Bucket/Dir.tsx
```diff
@@ -429,3 +415,30 @@ export default function Dir({
...
+
+export default function DirWrapper({
+  match: {
+    params: { bucket, path: encodedPath = '' },
+  },
+  location: l,
+}: RRDom.RouteComponentProps<DirRouteParams>) {
+  const path = s3paths.decode(encodedPath)
+  // FIXME: @nl0 is it safe to get only first prefix?
+  const { prefix } = parseSearch(l.search, true)
```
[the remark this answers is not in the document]
fiskus: I think the same. At this moment, we use `parseSearch` without second parameter, so `prefix` can be `string[]`, but it seems like we don't expect it to be an array in the subsequent code

fiskus:
Lost interest in this because of `QuiltSync`

PR #3406, opened 2023-04-10
Extra parameters for quilt3 push

sir-sigurd:
# Description

# TODO

- [ ] it's not clear how these flags should affect `--meta`
- [ ] Unit tests
- [ ] Documentation


api/python/quilt3/main.py
```diff
@@ -202,13 +203,18 @@ def cmd_verify(name, registry, top_hash, dir, extra_files_ok):
         return 1
 
 
-def cmd_push(name, dir, registry, dest, message, meta, workflow, force, dedupe):
+def cmd_push(name, dir_logical_key, dir, registry, dest, message, meta, workflow, force, dedupe, browse):
+    pkg = Package()
+
     try:
-        pkg = Package.browse(name, None)
+        pkg = Package.browse(name, browse)
```
drernie, line 210: I assume you mean "registry=browse"? That would seem to imply the user has to pass the explicit registry to the browse option.
We were imagining something more like:
```suggestion
        browse_registry = (browse == "remote") ? registry : None
        pkg = Package.browse(name, registry=browse_registry)
```

api/python/quilt3/main.py
```diff
@@ -465,6 +471,15 @@ def create_parser():
         action="store_true",
         help="Skip the push if the local package hash matches the remote hash.",
     )
+    optional_args.add_argument(
+        "--browse",
+        help="FIXME", # FIXME
```
drernie, line 476:
```suggestion
        help="By default, `push` first browses the top_hash from the 'local' registry.  Specify 'remote' to explicitly tell push to retrieve the current top_hash from the destination registry, so that it always succeeds.",
        default="local",
        ```

File: api/python/quilt3/main.py [hunk 3]
```diff
@@ -465,6 +471,15 @@ def create_parser():
...
+    )
+    optional_args.add_argument(
+        "--dir-logical-key",
+        help="FIXME", # FIXME
```
drernie, line 480:
```suggestion
        help="By default, the specified `dir` is added to the top level of a package. Use this option to specify a subfolder."
```

drernie (changes requested):

api/python/quilt3/main.py
drernie: Also: could we do `Package.browse(name, registry=browse_registry)` AFTER the push to update the local registry, so it is possible to push twice in a row?

drernie: Also, I am still unclear if (or when) browse() updates the local registry. How can/do we test that?

api/python/quilt3/main.py
drernie: NOTE. Must preserve current behavior of meta:
```
  --meta META          Sets package-level metadata.
```

api/python/quilt3/main.py
drernie: @sir-sigurd You were correct. With these semantics, '--browse' it is identical to '--force' and therefore unnecessary. Please remove.

drernie: Also, Aneesh confirmed that he mean literally "remote" not a registry name.

drernie (approved):

drernie:
Seems obsolete, will close.  Feel free to re-open if you still want to fix.

PR #3408, opened 2023-04-10
Suspense wrapper for ResourceCache

fiskus:
`useVoila` uses `ResourceCache.useData(..., ..., { suspnese: true })`, so it's required to wrap it with `<Suspense />`


nl0 (approved):

catalog/app/components/Preview/loaders/Notebook.tsx
```diff
@@ -53,5 +54,9 @@ export const Loader = function WrappedNotebookLoader({
   handle,
   children,
 }: NotebookLoaderProps) {
-  return <NotebookLoader {...{ handle, children }} />
+  return (
+    <React.Suspense fallback={<M.CircularProgress />}>
```
nl0, line 58: hm, probably we should call `children` with `AsyncResult.Pending()` instead of showing the placeholder directly

nl0 (changes requested):

catalog/app/components/Preview/loaders/Notebook.tsx
fiskus: Done

nl0 (approved):
👍

PR #3422, opened 2023-04-17
Add snapshot test for BucketIcon

fiskus:


catalog/app/components/BucketIcon/__snapshots__/BucketIcon.spec.tsx.snap
```diff
@@ -0,0 +1,41 @@
...
+  className="makeStyles-root-1"
+  src="https://custom-src"
+/>
+`;
+
+exports[`components/BucketIcon should render default when no src 1`] = `
+<img
+  alt="No src"
+  className="makeStyles-root-1"
+  src="IMAGE_MOCK"
```
nl0, line 39: why does it render `IMAGE_MOCK` here? what does that mean?

fiskus: It's a mock for bucket.svg (it's a mock for any media file):
it is set here https://github.com/quiltdata/quilt/blob/master/catalog/jest.config.js#L29
and the content of this mock is https://github.com/quiltdata/quilt/blob/master/catalog/internals/mocks/image.js

fiskus: Probably, we can change to something, that mention exported file name, for example, `bucket.svg-mock`

nl0 (approved):

PR #3437, opened 2023-04-21
Add custom cert docs

robnewman:
# TODO

- [n/a] Unit tests
- [X] Documentation
    - [n/a] Markdown docs for developers


docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+    >>> import certifi
+    >>> certifi.where()
+    '/path/to/site-packages/certifi/cacert.pem'
+    ```
+2. Copy your custom self-signed certificate and paste into the
+`cacert.pem` file.
+
+## Windows
+
+1. Press «Win + R» to open a new command prompt and type `certmgr`.
```
akarve, line 18: Not sure if CP is the right app but the keystroke is an optional detail to actually opening the app.
```suggestion
1. Open Command Prompt («Win + R») to open a new command prompt and type `certmgr`.
```

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+file system (`Path\To\mitm.cer`)
+4. Add the certificate to the list of trusted CA's via the
+`REQUESTS_CA_BUNDLE` environment variable:
+    ```bash
+    > set REQUESTS_CA_BUNDLE=Path\To\mitm.cer
+    ```
+
+## Verification
+Requesting access to a private website should now return a 200 HTTP
+status code:
```
akarve, line 35: It's not about private websites it's about sites that use the cert in question. AWS console is not a private website, the S3 URLs are not even private websites. The issue is they are signed with a private cert.

akarve (changes requested):

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+## Mac and Linux
+
+1. Determine the location of your certificates:
+    ```python
+    >>> import certifi
+    >>> certifi.where()
+    '/path/to/site-packages/certifi/cacert.pem'
+    ```
+2. Copy your custom self-signed certificate and paste into the
+`cacert.pem` file.
```
dimaryaz, line 14: This is actually not a good solution cause that will will be overwritten any time `certifi` gets upgraded. Setting ``REQUESTS_CA_BUNDLE` is better.

On Linux, need set `REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt`.

On OS X, it turns out it's more complicated - need to first create the bundle file. Assuming you're using OS X), can you see if this works? https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/non-standard-certs.html

And then on both Linux and OS X, `export REQUESTS_CA_BUNDLE=...` should be added to `~/.bashrc`.

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Windows
+
+1. Press «Win + R» to open a new command prompt and type `certmgr`.
+This opens the Windows Certificate Manager for the current user.
+2. Search all certificates stores for a certificate issued by `mitm`.
+There should be at least one certificate.
```
dimaryaz, line 21: The user will need to search for whatever certificate that's actually being used - I was just using "mitmproxy" in my test.

You can find the certificate issuer by opening any https website and looking at SSL settings:

VirtualBox_Windows 8_21_04_2023_12_22_00

Or more details here:

VirtualBox_Windows 8_21_04_2023_12_22_34

(And of course this depends on the browser, etc.)

docs/advanced-features/self-signed-certificates.md
dimaryaz: I'd also add, it's whatever status code the website returns - as long as its not an SSL exception.

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+
+   ![MITM certificate](../imgs/certmgr-windows.png)
+
+3. Export the certificate in Base-64 encoded X.509 (.CER) to your
+file system (`Path\To\mitm.cer`)
```
dimaryaz, line 26: Since it's not going to be `mitm`, I'd call it something neutral - `custom.cer` maybe

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+4. Add the certificate to the list of trusted CA's via the
+`REQUESTS_CA_BUNDLE` environment variable:
+    ```bash
+    > set REQUESTS_CA_BUNDLE=Path\To\mitm.cer
+    ```
+
+## Verification
+Requesting access to a private website should now return a 200 HTTP
```
akarve, line 34: We should just say the the test is the requests now succeed. I think when the CA is not trusted we don't even get an HTTP code but rather an exception.

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+
+   ![MITM certificate](../imgs/certmgr-windows.png)
+
+3. Export the certificate in Base-64 encoded X.509 (.CER) to your
+file system (`Path\To\mitm.cer`)
+4. Add the certificate to the list of trusted CA's via the
+`REQUESTS_CA_BUNDLE` environment variable:
+    ```bash
+    > set REQUESTS_CA_BUNDLE=Path\To\mitm.cer
+    ```
```
dimaryaz, line 31: This will make _only_ the custom certificate work. If some customer's VPN only hijacks some websites and not others, then they'll need to trust both the original certifi certificates and their custom one. They'll need to concat the two files, etc.:

```
type ...\custom.cer AppData\Roaming\Python\Python311\site-packages\certifi\cacert.pem > ...\everything.pem
```

where `AppData\Roaming\Python\Python311\site-packages\certifi\cacert.pem` was the output of `certifi.where()`, and then use `REQUESTS_CA_BUNDLE=...\everything.pem`

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
...
+2. Search all certificates stores for a certificate issued by `mitm`.
+There should be at least one certificate.
+
+   ![MITM certificate](../imgs/certmgr-windows.png)
+
+3. Export the certificate in Base-64 encoded X.509 (.CER) to your
+file system (`Path\To\mitm.cer`)
+4. Add the certificate to the list of trusted CA's via the
+`REQUESTS_CA_BUNDLE` environment variable:
+    ```bash
```
akarve, line 29: Maybe just `sh`.

dimaryaz (dismissed):

docs/SUMMARY.md
```diff
@@ -54,6 +54,7 @@
 * [Event-driven packaging](advanced-features/event-driven-packaging.md)
 * [GxP & Quilt](advanced-features/good-practice.md)
 * [Private endpoints](advanced-features/private-endpoint-access.md)
+* [Self-signed certificates](advanced-features/self-signed-certificates.md)
```
akarve, line 57: Probably belongs in the API section instead. Advanced is bloating.

docs/advanced-features/self-signed-certificates.md
```diff
@@ -0,0 +1,42 @@
+<!-- markdownlint-disable-next-line first-line-h1 -->
+To use a self-signed certificate with the `quilt3` API or CLI,
+the certificate must first be added to your Python environment.
```
akarve, line 3: Need to explain that this procedure is useful when using `quilt3` with a client-to-site VPN.

akarve (changes requested):

docs/advanced-features/self-signed-certificates.md
robnewman: Updated as follows to prepend the custom cert:

Mac: `export REQUESTS_CA_BUNDLE=/path/to/converted/certificate.crt:$REQUESTS_CA_BUNDLE`
Linux: `export REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt:$REQUESTS_CA_BUNDLE`
Windows: `set REQUESTS_CA_BUNDLE=%REQUESTS_CA_BUNDLE%;Path\To\mycert.cer`

dimaryaz: This won't actually prepend anything:
- It will only include the previous value of the environment variable, which is almost certainly unset (if the user had already set it to something, he/she probably wouldn't need these instructions)
- The docs don't even mention that the variable allows multiple files, so it probably doesn't. Pretty sure all certs have to be concatenated into a single file.

So on Windows, need to run the `type` command, or on Mac, it should be `cat`.

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,61 @@
...
+access) VPN you may need to use a self-signed certificate to
+securely connect. This requires adding the certificate to your
+Python environment.
+
+## Mac OS X
+
+1. Open Keychain Access.
+2. Select System Keychains > System Roots.
+3. Select the «Certificates» tab.
+4. Select a root certificate (e.g. AAA Certificate Services).
```
dimaryaz, line 13: Should probably make this consistent with the Windows instructions: start with open a browser, go to a website, get the certificate issuer, etc. - and then in this step, look for the certificate from earlier.

docs/advanced-features/self-signed-certificates.md
robnewman: Simplified: removed the whole prepend commands and just keep the `export` command.

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
+<!--pytest-codeblocks:skipfile-->
+<!-- markdownlint-disable-next-line first-line-h1 -->
+When using the `quilt3` API or CLI with a client-to-site (remote
+access) VPN you may need to use a self-signed certificate to
+securely connect. This requires adding the certificate to your
+Python environment.
```
akarve, line 6: Private and self-signed are not the same thing.
```suggestion
When using the `quilt3` API or CLI with a client-to-site
VPN you may need to trust a custom certificate.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Mac OS X
+
+1. Open a new browser window and navigate to any HTTPS website.
```
akarve, line 10:
```suggestion
1. Open a browser tab to any HTTPS website that uses the custom certificate
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+2. Click the «lock icon» in the browser navigation/address bar.
```
akarve, line 11:
```suggestion
2. Click the lock icon in the address bar.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+3. Click «View certificates» and record the certificate name value
```
akarve, line 12:
```suggestion
3. Click View certificates and copy the certificate name to a safe place.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+(e.g. `mycert`).
+4. Open Keychain Access.
+5. Select System Keychains > System Roots.
+6. Select the «Certificates» tab.
```
akarve, line 16:
```suggestion
6. Open the Certificates tab.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+access) VPN you may need to use a self-signed certificate to
+securely connect. This requires adding the certificate to your
+Python environment.
+
+## Mac OS X
+
+1. Open a new browser window and navigate to any HTTPS website.
+2. Click the «lock icon» in the browser navigation/address bar.
+3. Click «View certificates» and record the certificate name value
+(e.g. `mycert`).
```
akarve, line 13:
```suggestion
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+4. Open Keychain Access.
+5. Select System Keychains > System Roots.
+6. Select the «Certificates» tab.
+7. Search all certificates for a certificate name value that equals
+the value recorded (`mycert` in the example).
```
akarve, line 18:
```suggestion
7. Find the certificate that you noted above
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+8. Click «File», select "Export Items...", export the root certificate
```
akarve, line 19:
```suggestion
8. Click File > Export Items... to export the root certificate.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+to your localhost.
```
akarve, line 20:
```suggestion
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+9. In the Terminal, convert the exported certificate with
```
akarve, line 21:
```suggestion
9. Convert the exported certificate in Terminal as follows:
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+[OpenSSL](https://www.openssl.org/):
```
akarve, line 22: If their Mac doesn't have OpenSSL installing it is beyond the scope of this tutorial.
```suggestion
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+    ```sh
+    openssl x509 -inform der -in /path/to/your/certificate.cer -out /path/to/converted/certificate.crt
+    ```
+10. In the Terminal, open your shell profile (e.g. `~/.bashrc` or
+`~/.zshrc`).
+11. Add the certificate to your shell environment:
+    ```ssh
+    export REQUESTS_CA_BUNDLE=/path/to/converted/certificate.crt`
+    ```
```
akarve, line 31:
```suggestion
10. Export the following variable (e.g. in `~/.zshrc`):
    ```sh
    export REQUESTS_CA_BUNDLE=/path/to/converted/certificate.crt`
    ```
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+12. Save the change.
+13. Exit your shell or Terminal app, then reopen.
```
akarve, line 33:
```suggestion
```

akarve: Please update the Linux section for similar concision and style consistency

akarve (changes requested):

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+Python environment.
+
+## Mac OS X
+
+1. Open a new browser window and navigate to any HTTPS website.
+2. Click the «lock icon» in the browser navigation/address bar.
+3. Click «View certificates» and record the certificate name value
+(e.g. `mycert`).
+4. Open Keychain Access.
+5. Select System Keychains > System Roots.
```
akarve, line 15: If you use "1." for every item you don't have to renumber on edit. These bullets can be consolidated in any case.
```suggestion
1. Open Keychain Access and select System Keychains > System Roots
```

docs/api-reference/self-signed-certificates.md
dimaryaz: "Click View certificates" without quotes kind of looks weird?

That said, we don't want to be too specific here since these are instructions for "any browser on OS X" - we're just hoping that any reasonable browser has such a feature.

docs/api-reference/self-signed-certificates.md
robnewman: Fixed the stray `. Leaving bashrc and zshrc in place (but in simplified format per Aneesh)

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,63 @@
...
+in different Linux distributions. For example, in Ubuntu Linux
+distributions the file is located at `/etc/ssh/certs/ca-certificates.crt`.
+
+## Windows
+
+1. Direct your browser to an HTTPS website that uses a custom certificate.
+1. Click the lock icon in the address bar.
+1. Click View certificates and copy the certificate name to a safe place.
+1. Open the Command Prompt («Win + R») and type `certmgr`. This
+opens the Windows Certificate Manager for the current user.
```
akarve, line 42:
```suggestion
1. Open the Command Prompt («Win + R») and type `certmgr` to open
your Windows Certificate Manager.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,63 @@
...
+1. Find the certificate that you noted above.
+
+   ![MITM certificate](../imgs/certmgr-windows.png)
+
+1. Export the certificate in Base-64 encoded X.509 (.CER) to your
+file system (`Path\To\mycert.cer`).
+1. Export the following variable:
+    ```sh
+    > set REQUESTS_CA_BUNDLE=Path\To\mycert.cer
```
akarve, line 51: Confusing to show shell prompt and it makes copy-paste harder.
```suggestion
    set REQUESTS_CA_BUNDLE=Path\To\mycert.cer
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,63 @@
...
+    ```
+> This only affects the current shell session.
```
akarve, line 53: Eliminate or provide a solution as we do with Linux re: startup files.

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Verification
+
+Requesting access to a registry with the `quilt3` API or CLI should
+no longer fail with SSL errors.
```
akarve, line 58:
```suggestion
`quilt3` should no longer fail with SSL errors related to the custom certificate.
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Further information
```
akarve, line 60: Do we have a standard title for this tail section?
```suggestion
## Resources
```

akarve (dismissed):

docs/api-reference/self-signed-certificates.md
robnewman: We've using `References` [elsewhere](https://docs.quiltdata.com/advanced/private-endpoint-access)

docs/api-reference/self-signed-certificates.md
dimaryaz: Here are some instructions:
https://www.computerhope.com/issues/ch000549.htm

Except that we'll want to create a new var instead of editing an existing one.

In any case, I wouldn't go into much more detail than "here's how to open the editor UI".

robnewman: I found out how to permanently define the cert for an miniconda distro of Python and added that.

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,76 @@
...
+    ```
+    - Certificate in `pem` encoding (no conversion necessary):
+    ```sh
+    openssl x509 -in \Path\To/mycert.cer -out \Path\To\Converted/mycert.crt
+    ```
+1. Export the following variable, which affects the current shell session:
+    ```sh
+    set REQUESTS_CA_BUNDLE=Path\To\mycert.cer
+    ```
+1. If using an Anaconda Python distribution, you may wish to permanently define
```
akarve, line 63: I think this is TMI that makes docs harder and more time-consuming to maintain. Not our job to go into shell specifics or persistence.

robnewman: This is providing a permanent solution for Windows as requested here: https://github.com/quiltdata/quilt/pull/3437#discussion_r1179724732

akarve: But it requires Anaconda which is out of scope and not present on all Windows machines. It creates more questions than it answers. Why not just use the exact same language for Windows as we do elsewhere? "Export the following variable. You may wish to do this in a startup file for repeatability."

robnewman: Okay - fixed.

docs/SUMMARY.md
```diff
@@ -22,6 +22,7 @@
 * [quilt3.Bucket](api-reference/Bucket.md)
 * [CLI, environment](api-reference/cli.md)
 * [Known limitations](api-reference/limitations.md)
+* [Self-signed certificates](api-reference/self-signed-certificates.md)
```
akarve, line 25:
```suggestion
* [Custom SSL certs](api-reference/self-signed-certificates.md)
```

docs/api-reference/self-signed-certificates.md
```diff
@@ -0,0 +1,72 @@
+<!--pytest-codeblocks:skipfile-->
```
akarve, line 1: Please update file name to "custom-ssl-certificates.md" or similar.

akarve (changes requested):

docs/SUMMARY.md
robnewman: Updated to:

`* [Custom SSL certificates](api-reference/custom-ssl-certificates.md)`

Better to use the full word instead of an abbreviation for SEO.

akarve (approved):

PR #3439, opened 2023-04-26
Add synonym for quilt3 catalog --detailed_help with sane spelling

sir-sigurd:
IMO we can simply rename it because I think nobody's work depends on the spelling of the flag for showing help message.


drernie (changes requested):
Shouldn't we also update the script in genders to use the new "recommended" flag, so we can deprecate the old one?

drernie (approved):
Sweet, thanks for doing this.

PR #3445, opened 2023-05-09
Documentation: Troubleshooting - Collect logs from browser and ECS cluster

robnewman:
# TODO

- [n/a] Unit tests
- [X] Documentation
    - [n/a] Markdown docs for developers


docs/Troubleshooting.md
```diff
@@ -138,3 +138,99 @@ and is a JSON object generated by the Quilt build system.
...
+        - Google Chrome: **Ctrl + Click** anywhere on the grid of
+        network requests and select **Save all as HAR with content**.
+        - Mozilla Firefox: Select the gear icon () and select **Save
+        All As HAR**.
+        - Apple Safari: Click the **Export** button
+    1. Save the HAR-formatted file to your localhost.
+        ![Save browser Network error logs as HAR content](imgs/troubleshooting-logs-browser.png)
+1. CloudWatch logs for the Quilt cluster Elastic Container Service:
+    1. Install the [AWS CLI](https://aws.amazon.com/cli/) tool.
+    1. Export the AWS profile for your Quilt stack:
```
akarve, line 172: Let's skip both of these steps. It's better not to document that because they might change, the profile is not really for a stack, some users don't have profiles, etc.

docs/Troubleshooting.md
```diff
@@ -138,3 +138,99 @@ and is a JSON object generated by the Quilt build system.
...
+Quilt, please capture the following logs and attach to an email
+to support@quiltdata.io:
+
+1. Browser logs (isolating the most recent request that triggered an error):
+    1. Go to the affected page in your Quilt Catalog.
+    1. Open the browser Developer tools:
+        - Google Chrome: Press **F12**, **Ctrl+Shift+I** or from the
+        Chrome menu select **More tools > Developer tools**.
+        - Mozilla Firefox: Press **F12** or go to **Tools > Browser Tools > Web Developer Tools**.
+        - Apple Safari: Go to **Develop > Show Web Inspector**.
```
akarve, line 153: Thank you for going the extra mile but I'd say docs for multiple browser. It is indeed helpful to say "(e.g. Do X in Chrome)" and use Chrome throughout as an example, but the browser specifics details may change and it's better to let readers figure out what to do on their own. So we mostly need a skeleton with Chrome as the sole example: clear logs, repro issue, save logs. And don't forget that we also want the browser console logs as well, not just browser network logs.

docs/Troubleshooting.md
```diff
@@ -138,3 +138,99 @@ and is a JSON object generated by the Quilt build system.
...
+        - Apple Safari: Click the **Export** button
+    1. Save the HAR-formatted file to your localhost.
+        ![Save browser Network error logs as HAR content](imgs/troubleshooting-logs-browser.png)
+1. CloudWatch logs for the Quilt cluster Elastic Container Service:
+    1. Install the [AWS CLI](https://aws.amazon.com/cli/) tool.
+    1. Export the AWS profile for your Quilt stack:
+        ```
+        export AWS_PROFILE=stage
+        ```
+    1. Execute the `cloudformation` command to retrieve your Quilt stack log group name:
```
akarve, line 176: `cloudformation` isn't the name of the command. Also let's simplify this whole thing and ask the user to first get the name of the affected  Quilt CF stack (`aws cloudformation list-stacks` is a starting point).

And because in the CFT we do `LogGroupName: !Ref 'AWS::StackName'`, for now we can simply assume that the name of the LogGroup IS the stack name and the user can proceed from there.

docs/Troubleshooting.md
```diff
@@ -138,3 +138,99 @@ and is a JSON object generated by the Quilt build system.
...
+                "PhysicalResourceId": "YOUR_QUILT_STACK_LOG_GROUP_NAME",
+                "ResourceType": "AWS::Logs::LogGroup",
+                "LastUpdatedTimestamp": "2022-01-11T04:44:26.778Z",
+                "ResourceStatus": "UPDATE_COMPLETE",
+                "Metadata": "{}",
+                "DriftInformation": {
+                    "StackResourceDriftStatus": "IN_SYNC"
+                }
+            }
+        }
```
akarve, line 195: It's easier and faster to read if we leave the output out of the docs completely.

docs/Troubleshooting.md
```diff
@@ -138,3 +138,99 @@ and is a JSON object generated by the Quilt build system.
...
+        echo "query_id: $query_id"
+ 
+        # Give the query time to run
+        echo "sleep for 10 seconds (let the query run)..."
+        sleep 10
+
+        echo "write results to $file_out"
+        query_results=$(aws logs get-query-results --query-id $query_id)
+        echo "$query_results" | jq -r '.results[] | [.[] | .value] | @tsv' | column -t > $file_out
+        ```
```
akarve, line 236: OK for now, until we have a Python script (which will checked into source control, where we can use Python to get around OS specifics, etc.: Let's do something really simple like this instead, noting the date function may vary by OS (this is for Mac BSD), and the user may need to adjust the time range to capture their event.

```
aws logs filter-log-events \
  --log-group-name $YOUR_LOG_GROUP \
  --start-time "$(( $(date -u -v-30M +%s) * 1000 ))" \
  --end-time "$(( $(date -u +%s) * 1000 ))" > log-events.json
```

docs/Troubleshooting.md
```diff
@@ -138,3 +138,99 @@ and is a JSON object generated by the Quilt build system.
 
     `$ aws cloudformation describe-stack-events --stack-name QUILT_STACK_NAME`
 
+## Collecting logs
```
akarve, line 141:
```suggestion
## Collect logs to report a bug
```

Note that the three steps from the section just above this one are all relevant logs and we should consolidate those steps into this section.

All of these steps are of course laborious for customers but let's get this PR over the line first and then as and when possible move to a single script like `get-quilt-logs --stack-name MY_QUILT_STACK` in GitHub or better.

docs/Troubleshooting.md
akarve: Note the redirect at the end. All commands need to output into a file that's appropriately named.

akarve (changes requested):

docs/Troubleshooting.md
akarve: Oh, lastly, so that we can easily ask for and link to specific logs (not all customers need to send all logs), let's have a clean subsection for Browser Logs, CloudFormation Logs, ECS Logs, etc.

docs/technical-reference.md
```diff
@@ -266,15 +266,15 @@ to Google's OAuth 2.0 server.
 ![](./imgs/google_console.png)
 
 Copy the `Client ID` and `Client secret` to a safe place.
-Add `YOUR_QUILT_CATALOG_URL/oauth-callback` to *authorized redirect URIs*.
+Add `QuiltWebHost/oauth-callback` to *authorized redirect URIs*.
```
akarve, line 269: This changeset is helpful. How can we distinguish literal text from a template? Maybe `$QuiltWebHost` or `< QuiltWebHost>`? Go with the one that you think is most consistent and clear before merging. Pre-emptively approving.

akarve (dismissed):

docs/technical-reference.md
robnewman: This changeset was pushed from the wrong branch. It has been reverted.

docs/Troubleshooting.md
```diff
@@ -138,3 +138,59 @@ and is a JSON object generated by the Quilt build system.
 
     `$ aws cloudformation describe-stack-events --stack-name QUILT_STACK_NAME`
 
+## Collect logs to report a bug
+
+To expedite the resolution of any errors encountered while using
+Quilt, please capture the following logs and attach to an email
+to support@quiltdata.io:
```
akarve, line 145:
```suggestion
Quilt, please capture the following logs and attach share them with Quilt support:
```

docs/Troubleshooting.md
```diff
@@ -138,3 +138,59 @@ and is a JSON object generated by the Quilt build system.
...
+1. Open the browser Developer tools:
+    - Google Chrome: Press **F12**, **Ctrl+Shift+I** or from the
+    Chrome menu select **More tools > Developer tools**.
+1. Select the **Console** tab.
+1. Click the **Clear** button to clear out existing logs.
+1. Perform the action that triggers the error (e.g. clicking the `Download package` button).
+1. Export the logs.
+    - Google Chrome: **Ctrl + Click** anywhere on the grid of
+    network requests and select **Save as...**.
+1. Save the log file to your localhost.
```
akarve, line 177: Please join this with the previous section since there's so much overlap. To make things easier you can remove **Clear** form both sections because more data is usually better.

akarve (commented):
Almost there, left a few small pointers.

akarve (approved):

PR #3446, opened 2023-05-09
Fix indexing bug when dots in file name

akarve:
Infer extensions more correctly when the the filename has more than one dot.

Also update pandas, pyarrow as they were causing tests on an M1 MacBook to fail.

# TODO
- [x] More unit tests


sir-sigurd:
> Also update pandas, pyarrow, and nbconvert as they were causing tests on an M1 MacBook to fail.

I see no changes related to nbconvert.

sir-sigurd (dismissed):
Makes sense.
(I think we should rework extension handling because it seems too scattered and for deep indexing we actually want to infer file type/format, but not extension)

akarve:
> Makes sense.
> (I think we should rework extension handling because it seems too scattered and for deep indexing we actually want to infer file type/format, but not extension)

Yeah good point maybe as part of our search refactor.

lambdas/es/indexer/index.py
```diff
@@ -197,6 +197,9 @@ def infer_extensions(key, ext):
             or ext == ".pq"
     ):
         return ".parquet"
+    # if there are multiple dots take only the last suffix
+    elif ext.count(".") > 1:
+        return pathlib.PurePosixPath(ext).suffix
```
dimaryaz, line 202: I think this is the wrong place to handle it just cause we already have code that is supposed to handle multiple extensions: https://github.com/quiltdata/quilt/pull/3446/files#diff-75e3588577c6f019a82536d2c88e59ca3d77cb035b1d933e8f3d92fda26b8c67R705-R708

I'm not sure what what the best solution is, but we could:
- pass `ext1` and `ext2` instead of `ext`, OR
- move the `.gz` handling up the stack, and guarantee that `ext` is only a single extension

lambdas/es/indexer/requirements.txt
```diff
@@ -17,10 +17,10 @@ jupyter-core==4.11.2
 nbformat==4.4.0
 numpy==1.21.5
 openpyxl==3.0.7
-pandas==1.1.5
+pandas==1.3.5
 pdfminer.six==20201018
 psutil==5.7.0
-pyarrow==2.0.0
+pyarrow==12.0.0
```
dimaryaz, line 23: That's quite a version bump. Should probably check if there are any breaking changes?

akarve: Yeah, I'll check their release notes before merging. The sucky thing is local dev isn't possible without this huge bump. I can put it in a separate PR if you prefer.

akarve:
> I see no changes related to nbconvert.
You're right I fixed that in the description.

akarve:
> I see no changes related to nbconvert.

Fixed. Good eye.

lambdas/es/indexer/index.py
```diff
@@ -187,18 +187,30 @@ def now_like_boto3():
...
     """guess extensions if possible"""
     # Handle special case of hive partitions
     # see https://www.qubole.com/blog/direct-writes-to-increase-spark-performance/
+    long_ext = "".join(exts)
     if (
-            re.fullmatch(r".c\d{3,5}", ext) or re.fullmatch(r".*-c\d{3,5}$", key)
+            re.fullmatch(r".c\d{3,5}", long_ext) or re.fullmatch(r".*-c\d{3,5}$", key)
             or key.endswith("_0")
-            or ext == ".pq"
+            or long_ext == ".pq"
```
akarve, line 198:
```suggestion
            or exts[-1] == ".pq"
```

dimaryaz:
Hmm it looks like it picked up unrelated catalog changes?

akarve:
> Hmm it looks like it picked up unrelated catalog changes?

Yes and lol. Fixed.

lambdas/es/indexer/index.py
```diff
@@ -187,18 +187,42 @@ def now_like_boto3():
...
+
+
+def get_normalized_extensions(key) -> List[str]:
+    """standard function turning keys into a list of (possibly empty) extensions"""
+    path = pathlib.PurePosixPath(key)
+    try:
+        ext_last = path.suffix.lower()
+        ext_next_last = path.with_suffix('').suffix.lower()
+    except ValueError:
+        return ["", ""]
```
dimaryaz, line 202: Don't think this can actually happen?

lambdas/es/indexer/index.py
```diff
@@ -187,18 +187,42 @@ def now_like_boto3():
...
+
+    # return in left-to-right order as they occur in the key
+    return [ext_next_last, ext_last]
```
dimaryaz, line 205: Should really be a tuple, with return type `Tuple[str, str]`

lambdas/es/indexer/index.py
```diff
@@ -187,18 +187,42 @@ def now_like_boto3():
...
     # see https://www.qubole.com/blog/direct-writes-to-increase-spark-performance/
+    long_ext = "".join(exts)
+    # pylint: disable=too-many-boolean-expressions)
     if (
-            re.fullmatch(r".c\d{3,5}", ext) or re.fullmatch(r".*-c\d{3,5}$", key)
+            re.fullmatch(r".c\d{3,5}", long_ext) or re.fullmatch(r".*-c\d{3,5}$", key)
             or key.endswith("_0")
-            or ext == ".pq"
+            or exts[-1] == ".pq"
+            or compression and exts[0] == ".pq"
```
dimaryaz, line 218: Parenthesis around "and"?

lambdas/es/indexer/index.py
```diff
@@ -699,10 +716,9 @@ def handler(event, context):
                     event_["s3"]["object"].get("lastModified") or event_["eventTime"]
                 )
                 # Get two levels of extensions to handle files like .csv.gz
-                path = pathlib.PurePosixPath(key)
-                ext1 = path.suffix
-                ext2 = path.with_suffix('').suffix
-                ext = (ext2 + ext1).lower()
+                ext_next_last, ext_last = get_normalized_extensions(key)
+                compression = get_compression(ext_last)
+                ext = ext_next_last + ext_last
```
dimaryaz, line 721: I can't remember anymore all the details... but if the only reason we support two extensions is to handle `.gz`, then once the compression is factored out, we can just use a single extension, right?

lambdas/es/indexer/index.py
akarve: It happens in only one case that I could find `key="."`, which is evil and why would anyone do that, but S3 will let you have an object named `.` no?

lambdas/es/indexer/index.py
akarve: I didn't want to change any of the downstream dependencies, for instance we send the full extension, both parts, to ElasticSearch documents via `do_index` and I'm leary of changing that behavior (it would change how certain search results are returned).

lambdas/es/indexer/index.py
akarve: Will do. Same effect though because of `and` precedence over `or`.

lambdas/es/indexer/index.py
```diff
@@ -187,18 +187,41 @@ def now_like_boto3():
     return datetime.datetime.now(tz=tzutc())
 
 
-def infer_extensions(key, ext):
+def get_compression(ext: str):
+    """return the compression type or None if not supported"""
+    return "gz" if ext == ".gz" else None
+
+
+def get_normalized_extensions(key) -> Tuple[str]:
```
dimaryaz, line 195: Should be `Tuple[str, str]` - it describes every entry's type

dimaryaz (approved):

dimaryaz (approved):

PR #3448, opened 2023-05-10
Fix search in Python client

sir-sigurd:
# Description

* use registry for search API because it's removed from API gateway at least since 60113c5816b845e8f9005aaaf3b1ca131ad4ca21
* allow using custom ES queries additionally to query string queries 
* align what we return in `quilt3.Bucket.search()` with docstring  and `quilt3.search()`, it's kinda breaking, but it could be considered already broken since return type didn't match docstring
* remove description of result entries shape because it was incorrect and I don't think we should make it part of contract
* I don't like that we return mix of object and package results and don't provide a clear way to check it it's one or another, but I gave up on this, since nobody wants spend time on it  

# TODO

- [x] `limit` doesn't really work
- [x] `api.search()` returns both objects and packages indexes, `Bucket.search()` returns only objects (both APIs returns mix of object and package results)
- [x] check docstrings re what is returned
- [x] Unit tests
- [x] Documentation

### DIFF
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -197,5 +197,5 @@ def search(query, limit=10):
     """
     # force a call to configure_from_default if no config exists
     _config()
-    raw_results = search_api(query, '*', limit)
+    raw_results = search_api(query, '_all', limit)
     return raw_results['hits']['hits']
--- a/api/python/quilt3/api.py
+++ b/api/python/quilt3/api.py
@@ -178,22 +178,7 @@ def search(query, limit=10):
 
 
     Returns:
-        a list of objects with the following structure:
-        ```
-        [{
-            "_id": <document unique id>
-            "_index": <source index>,
-            "_score": <relevance score>
-            "_source":
-                "key": <key of the object>,
-                "size": <size of object in bytes>,
-                "user_meta": <user metadata from meta= via quilt3>,
-                "last_modified": <timestamp from ElasticSearch>,
-                "updated": <object timestamp from S3>,
-                "version_id": <version_id of object version>
-            "_type": <document type>
-        }, ...]
-        ```
+        a list of dicts
     """
     # force a call to configure_from_default if no config exists
     _config()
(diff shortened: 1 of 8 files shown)


api/python/quilt3/api.py
```diff
@@ -164,38 +166,23 @@ def _disable_telemetry():
...
+        query: query string to search if passed as `str`, ES query body if passed as `dict`
+        limit: maximum number of results to return. Defaults to 10
 
     Query Syntax:
-        [simple query string query](
-            https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-simple-query-string-query.html)
-
+        [Query String Query](
+            https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-query-string-query.html)
+        [Query DSL](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl.html)
```
nl0, line 180: is it actually 6.8 or 6.7?

nl0 (dismissed):
i don't know much about this, but the change looks reasonable

drernie:
Ah, so you use the session instead of credentials. Great!

drernie (dismissed):

drernie:
@sir-sigurd I assume you were actually able to authenticate and run a query? What else needs to be done before marking it un-draft?

api/python/quilt3/api.py
sir-sigurd: we use 6.7 but I don't think it's any different to 6.8 here

nl0 (approved):

api/python/quilt3/api.py
```diff
@@ -164,38 +166,23 @@ def _disable_telemetry():
...
 @ApiTelemetry("api.search")
-def search(query, limit=10):
+def search(query: T.Union[str, dict], limit: int = 10) -> T.List[dict]:
     """
     Execute a search against the configured search endpoint.
 
     Args:
-        query (str): query string to search
-        limit (number): maximum number of results to return. Defaults to 10
+        query: query string to search if passed as `str`, ES query body if passed as `dict`
```
akarve, line 174:
```suggestion
        query: query string to query if passed as `str`, DSL query body if passed as `dict`
```

nl0 (approved):

PR #3451, opened 2023-05-11
Fix lambdas' requirements.txt

dimaryaz:
requirements.txt is supposed to pin everything, including sub-dependencies.

Update CI to not install sub-dependencies.


.github/workflows/py-ci.yml
```diff
@@ -187,7 +187,8 @@ jobs:
             python -m pip install -e lambdas/shared
             python -m pip install -e lambdas/${{ matrix.path }}
           fi
-          python -m pip install -r lambdas/${{ matrix.path }}/requirements.txt -r lambdas/${{ matrix.path }}/test-requirements.txt
+          python -m pip install --no-deps -r lambdas/${{ matrix.path }}/requirements.txt
```
akarve, line 190: I think it's important to spell out that without `--no-deps` second order dependencies can float and cause bugs and inconsistencies, not that anyone would read comments in the CI file. Maybe /lambdas needs a README with, for now, only that statement. Up to you.

.github/workflows/py-ci.yml
```diff
@@ -187,7 +187,8 @@ jobs:
...
+          python -m pip install -r lambdas/${{ matrix.path }}/test-requirements.txt
```
akarve, line 191: Should we not do test-requirements.txt first so that requirements.txt has the last word (in case dependencies overlap; but maybe you checked that and again in that case a note in lambdas/README.md would be nice).

lambdas/pkgpush/requirements.txt
```diff
@@ -1,2 +1,25 @@
...
+importlib-metadata==6.6.0
+jmespath==1.0.1
+jsonlines==1.2.0
 jsonschema==3.2.0
-https://github.com/quiltdata/quilt/archive/e96df90afc80ea2e3820a57a2a7ef4477df24204.tar.gz#egg=quilt3&subdirectory=api/python
+platformdirs==3.5.1
+pyrsistent==0.19.3
+python-dateutil==2.8.2
+PyYAML==6.0
+quilt3 @ https://github.com/quiltdata/quilt/archive/e96df90afc80ea2e3820a57a2a7ef4477df24204.tar.gz#subdirectory=api/python
```
akarve, line 16: @sir-sigurd is this still the right pin?

akarve (approved):

.github/workflows/py-ci.yml
dimaryaz: The problem is, that could easily hide a missing second-order dependency from `requirements.txt`. In general, though, `test-requirements` only contain additional requirements, without any overlap - though it's not exactly enforced anywhere.

lambdas/pkgpush/requirements.txt
dimaryaz: This was generated by `pip freeze`, and seems to work the same. But I could revert to the original. Not sure what the "correct" way is.

.github/workflows/py-ci.yml
dimaryaz: Wait, sorry, that doesn't make sense. It'll hide the missing dependency regardless. I guess I could switch them.

lambdas/pkgpush/requirements.txt
akarve: I’m not worried about the syntax just do we really need that weird version or was that a temporary hack and we since updated PyPI?

.github/workflows/py-ci.yml
akarve: Up to you but it does seem more correct to do test-requirements first, assuming pip does things left to right because lambdas never see any of those libs.

akarve (approved):

lambdas/pkgpush/requirements.txt
dimaryaz: Ohh. Probably don't need that version - I guess if we can specify a normal release number, that would be better. Let me see.

sir-sigurd: https://github.com/quiltdata/quilt/pull/3452

sir-sigurd (commented):
I think we should put more efforts here and move direct dependencies to setup.py (or better migrate to pyproject.toml) and generate requirements.txt with [pip-compile](https://pip-tools.readthedocs.io/en/latest/).

akarve (approved):

PR #3456, opened 2023-05-13
Lambdas: more efficient memory use and estimation for parquet

akarve:
* [x] Python unit tests

### DIFF
--- a/lambdas/shared/t4_lambda_shared/preview.py
+++ b/lambdas/shared/t4_lambda_shared/preview.py
@@ -1,17 +1,17 @@
 """
 Shared helper functions for generating previews for the preview lambda and the ES indexer.
 """
-import math
 import re
 import tempfile
 import zlib
 from io import BytesIO
+from typing import Tuple
 
 import fcsparser
 import pandas
 from xlrd.biffh import XLRDError
 
-from .utils import get_available_memory
+from .utils import get_available_memory, get_quilt_logger
 
 AVG_PARQUET_CELL_BYTES = 100  # a heuristic to avoid flooding memory
 # CATALOG_LIMIT_BYTES is bytes scanned, so acts as an upper bound on bytes returned
@@ -26,7 +26,7 @@ READ_CHUNK = 1024
 # common string used to explain truncation to user
 TRUNCATED = (
     'Rows and columns truncated for preview. '
-    'S3 object contains more data than shown.'
+    'S3 object may contain more data than shown.'
 )
 
 
@@ -129,27 +129,32 @@ def extract_fcs(file_, as_html=True):
     return body, info
 
 
-def extract_parquet(file_, as_html=True, skip_rows=False, *, max_bytes: int):
+def extract_parquet(file_, as_html=True, skip_rows: bool = False, *, max_bytes: int) -> Tuple[str, str]:
     """
     parse and extract key metadata from parquet files
 
     Args:
         file_ - file-like object opened in binary mode (+b)
 
+    Observations & assumptions:
+        * ParquetFile and iter_batches sizes in memory are negligible in practice
+        * Deserializing and converting batches to dataframes and strings consumes the most memory
+        * First row group contains more than enough data for search, preview
+        * This function will hit system bytes limits (DOC_LIMIT_BYTES, etc.) long
+            before it runs out of RAM (except maybe for pathologically large rows)
+
     Returns:
-        dict
+        tuple
             body - summary of main contents (if applicable)
-            info - metdata for user consumption
+            info - metadata for user consumption
     """
-    # TODO: generalize to datasets, multipart files
-    # As written, only works for single files, and metadata
-    # is slanted towards the first row_group
-    # local import reduces amortized latency, saves memory
+    logger_ = get_quilt_logger()
     import pyarrow.parquet as pq  # pylint: disable=C0415
 
     pf = pq.ParquetFile(file_)
     meta = pf.metadata
 
+    body = ""
     info = {}
     info['created_by'] = meta.created_by
     info['format_version'] = meta.format_version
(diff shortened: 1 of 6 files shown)


lambdas/shared/t4_lambda_shared/utils.py
```diff
@@ -61,8 +62,18 @@ def get_quilt_logger():
 
 
 def get_available_memory():
-    """how much virtual memory is available to us (bytes)?"""
-    from psutil import virtual_memory
+    """how much virtual memory is available to us (in bytes)"""
+    try:
+        # this import fails in AWS Lambda in spite of requirements.txt
```
dimaryaz, line 67: Huh? Why does it fail? And if that's the case, how does any lambda that calls `get_available_memory` work at all?

Though if it does fail, I'd just drop `psutil` completely in favor of the env var.

lambdas/shared/tests/test_utils.py
```diff
@@ -94,6 +106,7 @@ def test_separated_env_to_iter(self):
         """ensure the function that infers overrides from the env works:
             always returns a valid set(), perhaps empty, lowercases extensions
         """
+        assert not set(), "expected empty sets to be falsy"
```
dimaryaz, line 109: Why would this ever fail?..

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -166,13 +169,14 @@ def extract_parquet(file_, as_html=True, skip_rows=False, *, max_bytes: int):
...
     # the entire parquet file into a BytesIO by the time we get here
     if meta.num_row_groups:
-        # guess because we meta doesn't reveal how many rows in first group
+        # guess because meta doesn't reveal how many rows in first group
         num_rows_guess = math.ceil(meta.num_rows / meta.num_row_groups)
         size_guess = num_rows_guess * meta.num_columns * AVG_PARQUET_CELL_BYTES
         if skip_rows or (size_guess > get_available_memory()):
             # minimal dataframe with all columns and one row
             dataframe = pandas.DataFrame(columns=meta.schema.names)
             info['warnings'] = 'Large file: skipped rows to conserve memory, only showing column names'
```
dimaryaz, line 178: Where does this one end up? The file's metadata?

lambdas/shared/tests/test_utils.py
akarve: Lol. It shouldn't. Will remove.

lambdas/shared/t4_lambda_shared/utils.py
akarve: Yeah this was not the problem. Though I don't see need to revert this new code but I can if you want. The problem is that the first row-group is huge but the new pyarrow has an iterator instead of reading the whole row group, so I'm gonna try that.

lambdas/shared/t4_lambda_shared/preview.py
akarve: It gets returned to the caller.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,22 +129,28 @@ def extract_fcs(file_, as_html=True):
...
     parse and extract key metadata from parquet files
 
     Args:
         file_ - file-like object opened in binary mode (+b)
 
+    Observations & assumptions:
+        * Parquet file in memory in batches is a small constant factor (as observed
+        in practice)
+        * Deserializing and converting batches to dataframes and strings is,
+        asymptotically, where memory is consumed
```
akarve, line 143:
```suggestion
        * Parquet file in memory in batches is a negligible (in practice)
        * Deserializing and converting batches to dataframes and strings consumes the most memory
```

akarve:
Silly bugs, lemme fix and re-request.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,81 +129,80 @@ def extract_fcs(file_, as_html=True):
...
-    if as_html:
-        body = remove_pandas_footer(dataframe._repr_html_())  # pylint: disable=protected-access
+
+    available = get_available_memory()
+    iter_batches = None
+    # 10MB heuristic; should never happen, e.g. with current default of 512MB
+    if (available < 10E6) or skip_rows:
+        logger_.warning("Insufficient memory to index parquet file: %s", info)
+        info['warnings'] = "Skipped rows; insufficient memory"
+    elif meta.num_rows and meta.num_row_groups:
```
dimaryaz, line 175: Is `num_row_groups` still relevant? You don't call `read_row_group` anymore.

I'm trying to understand if batches and row groups are related - but I don't think they are? So you could always use `iter_batches`?

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,81 +129,80 @@ def extract_fcs(file_, as_html=True):
...
+                for column in row.astype(str):
+                    encoded = column.encode()
+                    encoded_size = len(encoded) + 1  # +1 for \t
+                    if (size + encoded_size) < max_bytes:
+                        buffer.append(encoded)
+                        buffer.append(b"\t")
+                        size += encoded_size
+                    else:
+                        done = True
+                        break
```
dimaryaz, line 198: This will break out of the inner for loop, but then still run the middle for loop to the end before it gets to `if done`.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,81 +129,80 @@ def extract_fcs(file_, as_html=True):
...
-                    buffer.append(encoded)
-                    buffer.append(b"\t")
-                    size += encoded_size
-                else:
-                    done = True
-                    break
-            buffer.append(b"\n")
-            size += 1
+        for batch in iter_batches:
+            df = batch.to_pandas()
```
dimaryaz, line 184: If I remember correctly, `to_pandas` is useful for the `_repr_html_()` method - but do we still need it in the non-HTML case? If you can access parquet values directly, that's probably more efficient. But I don't know much about all this.

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,81 +129,80 @@ def extract_fcs(file_, as_html=True):
...
+                    else:
+                        done = True
+                        break
+                buffer.append(b"\n")
+                size += 1
             if done:
                 break
         body = b"".join(buffer).decode()
+    elif as_html:
+        body = remove_pandas_footer(pandas.DataFrame(columns=meta.schema.names)._repr_html_())
```
dimaryaz, line 205: Do we need the 'Large file: skipped rows to conserve memory, only showing column names' here?

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,81 +129,80 @@ def extract_fcs(file_, as_html=True):
...
```
dimaryaz, line 206: No `else`? No preview if there are no row groups and `as_html` is false?

lambdas/shared/tests/test_utils.py
```diff
@@ -13,13 +13,20 @@
...
     separated_env_to_iter,
 )
 
 
+def test_get_available_memory():
+    mem = get_available_memory()
+    assert isinstance(mem, int), "Expected an int"
+    assert mem > 1, "Expected some memory"
+
+
```
dimaryaz, line 29: Doesn't seem that useful - should be safe to assume that `psutil` works as advertised. I'd just remove it.

dimaryaz (changes requested):

lambdas/shared/t4_lambda_shared/preview.py
akarve: It is because it has to be non-zero or this `row_groups` argument will fail: `iter_batches = pf.iter_batches(batch_size=128, row_groups=[0])`.

lambdas/shared/t4_lambda_shared/preview.py
akarve: Good question. Let me think about it. Not sure if the resulting effort will save us memory in that ostensibly arrow has a very efficient way to convert to Pandas. Is there anything I can do to make sure batches get garbage collected better?

lambdas/shared/t4_lambda_shared/preview.py
akarve: In theory num_rows > 0 implies num_row_groups > 0 but I don't see a good reason to take a chance. Of course num_row_groups > 0 does not imply num_rows > 0.

lambdas/shared/t4_lambda_shared/preview.py
akarve: Gah how does Python not have labeled break already...

lambdas/shared/t4_lambda_shared/preview.py
akarve: No. If `as_html` is false we should return `body = ""` which was set at the top of the routine.

lambdas/shared/t4_lambda_shared/preview.py
dimaryaz: Ok. Anyways, let's leave it alone - I was curious, but not worth it trying to rewrite everything at once. Garbage collection, etc. should be fine.

lambdas/shared/t4_lambda_shared/preview.py
dimaryaz: If you want a concise solution that no one will be able to understand, here's one:

```
for batch in iter_batches:
  ...
  for _, row in df.iterrows():
    for column in row.astype(str):
      ...
      if (size + encoded_size) < max_bytes:
        ...
      else:
        break
    else:
      buffer.append(b"\n")
      size += 1
      continue
    break
  else:
    continue
  break
```

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -1,17 +1,17 @@
...
+from typing import Tuple
 
 import fcsparser
 import pandas
 from xlrd.biffh import XLRDError
 
-from .utils import get_available_memory
+from .utils import get_available_memory, get_quilt_logger
 
 AVG_PARQUET_CELL_BYTES = 100  # a heuristic to avoid flooding memory
```
dimaryaz, line 16: Can delete `AVG_PARQUET_CELL_BYTES` and `MAX_PREVIEW_ROWS` now

lambdas/shared/t4_lambda_shared/preview.py
```diff
@@ -129,81 +129,80 @@ def extract_fcs(file_, as_html=True):
...
-                else:
-                    done = True
-                    break
-            buffer.append(b"\n")
-            size += 1
+        for batch in iter_batches:
+            df = batch.to_pandas()
+            if as_html:
+                body = remove_pandas_footer(df._repr_html_())
+                return body, info
```
dimaryaz, line 187: I think this changes the behavior of the preview lambda? It used to be up to `MAX_PREVIEW_ROWS` rows, which is 1000 - but now it's one batch, which is 128. An easy compromise might be to use a batch size of a 1000 rows - it's probably fine as far as memory goes.

dimaryaz (commented):
I think I understand it better now, so your change seems fine. If anything, the existing "size_guess" logic is pretty sketchy.

lambdas/shared/t4_lambda_shared/preview.py
akarve: I have a slightly uglier change that is more readable incoming.

lambdas/shared/tests/test_utils.py
akarve: I see your point but then that function will lose test coverage and it's used in multiple lambda functions so I'd prefer to leave it as is. It's also useful if e.g. we want to switch to getting memory from an AWS env variable.

lambdas/shared/t4_lambda_shared/preview.py
akarve: I don't think so because if this _function_ skips rows we triggered the condition on line 172. So if we reached this far we know there are no batches and that can only be because the file itself had no rows.

lambdas/shared/t4_lambda_shared/preview.py
akarve: Luckily now because the fronted I think uses /preview (new lambda) for everything. In any case the _repr_html() is just 10+1 rows. I'll double check that we note the repr is truncated. 
Screenshot 2023-05-15 at 10 59 50 PM

akarve: Some measurements for this file (sysgetsizeof):
* all batches with batch_size=128: 18,232 bytes
* all batches with batch_size=1000: 2,520 bytes
* Size of dataframe from first batch as string (batch_size=128): 21,204 bytes
* Size of dataframe from first batch as string (batch_size=1000): 164212 bytes

So I can gladly make the batch size larger but I didn't because in some cases have very low size limits (like 10kb for docs to elastic), but I'm fine other way (100, or 1000 batch_size); lmk.

PR #3459, opened 2023-05-16
Fix oauth callback example syntax & use QuiltWebHost variable

robnewman:
# TODO

- [n/a] Unit tests
- [X] Documentation


docs/technical-reference.md
```diff
@@ -321,7 +321,7 @@ for further details.
     Do not allow wildcard `*` in the login URI redirect. This will be something like the following:
 
         ```
-        https://quilt.<MY_COMPANY>.com/
+        https://quilt.<MY_COMPANY>.com/oauth-callback
```
akarve, line 324: Let's get down to one sentence instead of saying the same thing twice. In other words we have `YOUR_QUILT_CATALOG_URL/oauth-callback` right above this, and then again in this section, but with different wording. So let's consolidate into a single, clear statement that is consistent with the rest of the docs. We probably don't need a code block for this.

Another issue is the "Do not allow" statement is confusing because that's an actual checkbox. We should call it out as such or omit that statement altogether.
Screenshot 2023-05-16 at 12 02 53 PM

akarve (changes requested):
See inline comment

docs/technical-reference.md
akarve: Clarified the above comment.

akarve (approved):

PR #3476, opened 2023-06-01
Fix command for getting ECS logs, make it work on Linux

sir-sigurd:


docs/Troubleshooting.md
```diff
@@ -177,11 +177,11 @@ stacks (in your default region, which is listed in your
...
+the last 30 minutes:
     <!--pytest.mark.skip-->
     ```bash
     aws logs filter-log-events \
-        --log-group-name YOUR_QUILT_STACK_NAME
-        --start-time "$(( $(date -u -v-30M +%s) * 1000 ))" \
-        --end-time "$(( $(date -u +%s) * 1000 ))" > log-quilt-ecs-events.json
+        --log-group-name YOUR_QUILT_STACK_NAME \
+        --start-time "$(( ($(date +%s) - 1800) * 1000 ))" \
+        --end-time "$(( $(date +%s) * 1000 ))" > log-quilt-ecs-events.json
```
sir-sigurd, line 186: I removed `-u` because it doesn't seem to affect `%s` format which means `seconds since the Epoch (1970-01-01 00:00 UTC)`.

akarve (approved):

PR #3480, opened 2023-06-07
Filter for Admin/Users and Admin/Buckets tables

fiskus:
Screenshot from 2023-06-07 15-51-25

Screenshot from 2023-06-07 14-34-51


nl0 (approved):
looks good

PR #3488, opened 2023-06-12
Error sand image

fiskus:
I guess sand metaphor means "desert", "nothing". But since it was a low-res stretched image, it was looked broken to me.

I replaced with free-to-use repeatable texture (https://www.deviantart.com/jb1992/art/Desert-sand-texture-323267113)

There are two variants. I prefer the one with white inner shadows, but maybe it has too much visual noise.

Screenshot from 2023-06-12 16-54-05
Screenshot from 2023-06-12 16-54-26


akarve (dismissed):
First presentation looks better to me. Please crush the image; I was able to get sand.jpg down another 6.1% with ImageOptiim.

fiskus:
I already crushed with `imagemin`. Seems, `ImageOptiim` does the better job

But, I just converted .jpg to .webp and got 4 times smaller file with no visual difference

akarve (approved):

PR #3496, opened 2023-06-14
Add links to documentation and re-use code samples

fiskus:
Screenshot from 2023-06-14 14-37-24

---
  - https://app.asana.com/0/0/1202880605767757


docs/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Entries inside each section should be ordered by type:
 * [Fixed] Fix Header's orange flash on load ([#3487](https://github.com/quiltdata/quilt/pull/3487))
 * [Fixed] Fix code sample for package push ([#3499](https://github.com/quiltdata/quilt/pull/3499))
 * [Added] Add filter for users and buckets tables in Admin dashboards ([#3480](https://github.com/quiltdata/quilt/pull/3480))
+* [Added] Add links to documentation and re-use code samples ([#/3496](https://github.com/quiltdata/quilt/pull/3496))
```
nl0, line 28:
```suggestion
* [Added] Add links to documentation and re-use code samples ([#3496](https://github.com/quiltdata/quilt/pull/3496))
```

nl0 (approved):

nl0 (approved):

fiskus:
Sorry, one small fix: removed `--recursive` for `aws cp` file code sample

catalog/app/containers/Bucket/CodeSamples/Dir.tsx
```diff
@@ -0,0 +1,85 @@
...
+      b.ls("${path}")
+    `,
+  CLI_LIST: (bucket: string, path: string) =>
+    dedent`
+      # List files [[https://docs.aws.amazon.com/cli/latest/reference/s3/ls.html]]
+      aws s3 ls "s3://${bucket}/${path}"
+    `,
+  CLI_DOWNLOAD: (bucket: string, path: string, dest: string) => {
+    const cliDest = dest ? `"./${dest}"` : '.'
+    const recursiveFlag = dest ? ` --recursive` : ''
```
nl0, line 37: why is this based on the presence of `dest`? i think it should be based on whether the `path` is a prefix or an object (catalog determines that by the presence of a trailing slash afair)

nl0: tho this is the `Dir.tsx` file, so i guess it's only meant for directories, and therefore it should always be recursive

catalog/app/containers/Bucket/File.js
```diff
@@ -521,7 +498,7 @@ export default function File({
                 {
                   Ok: ({ ui: { blocks } }) => (
                     <>
-                      {blocks.code && <Code>{code}</Code>}
+                      {blocks.code && <DirCodeSamples {...{ bucket, path }} />}
```
nl0, line 501: it's quite confusing tbh. why call the component `DirCodeSamples` and then use it in both file and dir contexts with an additional `isDirectory` flag?

catalog/app/containers/Bucket/CodeSamples/Dir.tsx
nl0: ok, i'm just confused by the naming. i think if you are to keep this function signature, you should at least add a comment clarifying that `dest` is only present for directories, therefore we can infer `recursiveFlag` from that

fiskus: I started with `CodeSamples/Dir.tsx` and `CodeSamples/File.tsx`, but they have too many things in common, and I decided to group them in one component
But, I guess, it doesn't worth it, and it's better to split them again.

nl0: > But, I guess, it doesn't worth it, and it's better to split them again.

well, i think just adjusting the naming a little bit for clarity would do, but it's up to you

fiskus: Now the code samples for directory and for file are all in one template each, not interrupted by conditions. And, components doing simple things are simple themselves :)

nl0 (approved):

PR #3497, opened 2023-06-14
Google Tag Manager tracker

fiskus:
I think, the very simple code makes sense, because we probably don't need anything more sophisticated as done for Intercom and Mixpanel

---
  - https://app.asana.com/0/0/1204730372005692


catalog/app/utils/gtm.tsx
```diff
@@ -0,0 +1,40 @@
...
+  React.useEffect(() => {
+    if (!gtmId) return
+    loadScript(`https://www.googletagmanager.com/gtag/js?id=${gtmId}`)
+    addScript(`
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', '${gtmId}');
+    `)
+  }, [gtmId])
```
nl0, line 38: i'd say we should add a cleanup function removing the script elements added above, but it doesn't seem like a plausible use case for that exists

catalog/app/utils/gtm.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react'
+
+import cfg from 'constants/config'
+
+function loadScript(src: string) {
+  const s = window.document.createElement('script')
+  s.type = 'text/javascript'
+  s.async = true
```
nl0, line 8: i'd merge these two functions into one accepting a dict of attributes, so that it can be called like `addScript({ async: true, src: '...' })` and `addScript({ innerText: '...' })`, but that's up to you, since the code is good enough as it is

catalog/app/utils/gtm.tsx
```diff
@@ -0,0 +1,40 @@
...
+  s.src = src
+
+  const x = window.document.getElementsByTagName('script')[0]
+  x.parentNode?.insertBefore(s, x)
```
nl0, line 12: this would fail silently if `x.parentNode` is null-ish (which shouldn't happen, but...), so i'd rather add `assert x.parentNode` before this call to fail loudly and make TS happy

catalog/app/utils/gtm.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+interface GTMLoaderProps {
+  children: React.ReactNode
+}
+
+export default function GTMLoader({ children }: GTMLoaderProps) {
+  const gtmId = cfg.gtmId
```
nl0, line 28: maybe it would make sense to get this as a prop, and pass it in `app.tsx`, but... not sure what that would buy us)

nl0 (approved):

catalog/app/utils/gtm.tsx
fiskus: It's tricky to support adding **any** script attribute, so I've added only `async`, `innerText`, `src`

catalog/app/utils/gtm.tsx
fiskus: Ok, I added `invariant`

catalog/app/utils/gtm.tsx
fiskus: I don't know how to explain it clearly, but I feel it would be better as well

catalog/app/utils/gtm.tsx
fiskus: All initialized code will be preserved, so it's rather ritual, I guess. Anyway I've added removing scripts

nl0 (approved):

PR #3500, opened 2023-06-16
Fix collaborators background

fiskus:
Found correct Theme constant for that color


akarve (approved):

fiskus:
It uses NavBar theme. So, this color is orange, not blue.
It's always confusing to use navbar theme. Probably, it makes sense to use only two themes website and product, but use correct override styles to M.AppBar in product theme.

PR #3501, opened 2023-06-16
Simplify and correct admin email update instructions

akarve:


docs/Troubleshooting.md
```diff
@@ -84,33 +84,13 @@ after clicking the `RELOAD` button in the Quilt Catalog.
...
-field value will be `Modify` and the "Logical ID" field value will
-be `Migration` for approximately four resources. Click the "Submit"
-button.
-
-After the deployment update is successfully completed, login to the
-Catalog with the new administrator credentials. Create other Admin
-users as needed.
+If you need to change the admin or have accidentally broken your admin user, try
+the following:
+1. Change the value of the `AdminEmail` CloudFormation parameter.
```
dimaryaz, line 89: I think it's still worth pointing out that it's creating an admin only if the email is not taken. I.e., if a normal user has already logged in with that email through SSO or whatever, it will not be made admin.

docs/Troubleshooting.md
```diff
@@ -84,33 +84,13 @@ after clicking the `RELOAD` button in the Quilt Catalog.
...
+    > If you are using SSO, `AdminEmail` must match the admin's SSO email address.
```
akarve, line 90:
```suggestion
    > If you are using SSO, `AdminEmail` must match the admin's SSO email address.
    
    > Note: If and admin user was created via just-in-time provisioning through SSO
    > setting the admin password has no effect whatsoever and _password login will never work_.
```

docs/Troubleshooting.md
```diff
@@ -84,33 +84,17 @@ after clicking the `RELOAD` button in the Quilt Catalog.
...
-field value will be `Modify` and the "Logical ID" field value will
-be `Migration` for approximately four resources. Click the "Submit"
-button.
-
-After the deployment update is successfully completed, login to the
-Catalog with the new administrator credentials. Create other Admin
-users as needed.
+If you need to change the admin or have accidentally broken your admin user, try
+the following:
+1. Change the value of the `AdminEmail` CloudFormation parameter.
```
akarve, line 89:
```suggestion
1. Change the value of the `AdminEmail` CloudFormation parameter.
> Only _net new_ email addresses that have never logged in before will create new admins.

```

docs/Troubleshooting.md
akarve: Done.

dimaryaz (approved):

PR #3506, opened 2023-06-23
Preserve selection for create package from directory

fiskus:
After https://github.com/quiltdata/quilt/pull/2931 and https://github.com/quiltdata/quilt/pull/2990 it turns out we need to keep selection for Create Package dialog

* Moved `selection/setSelection` state to upper level, so it can be kept for both directory listing and package dialog
* Explicitly reset path, prefix, selection on bucket change and navigation, instead of using `useEffect` "listener"


fiskus:
2,3: I think you are confused with future work that is still under shaping. This one PR fixes losing selection when you select something in one directory, then click "Create package from directory": you see the same directory (inside Create Package dialog) but without selection.
[Screencast from 2023-06-26 16-50-09.webm](https://github.com/quiltdata/quilt/assets/533229/d4ae9adf-d8fc-4e0e-9b53-2520ded32fa2)

---

1. I think selection should be "consumed" for the current implementation. But actually, it should be preserved, if I change my mind and want to add more files, or want to create another package with slightly different set of files. But if it preserved, then we need "Clear selection" button. Anyway, I would leave it as is - selection will be reset on any navigation
2. I'm still working on it. I think we need to allow users to edit logical keys (and paths) and do something similar to OSes: downloaded.txt, downloaded (1).txt, downloaded(2).txt
3, 4. Yes, I think so.

nl0:
> 2,3: I think you confused with future work that is still under shaping. This one PR fixes losing selection when you select something in one directory, then click "Create package from directory": you see the same directory (inside Create Package dialog) but without selection.
[Screencast from 2023-06-26 16-50-09.webm](https://github.com/quiltdata/quilt/assets/533229/d4ae9adf-d8fc-4e0e-9b53-2520ded32fa2)

oh wow i didn't realize we have this. makes more sense now.

nl0 (approved):

fiskus:
https://github.com/quiltdata/quilt/pull/3527

PR #3508, opened 2023-06-23
Docs: Example of converting logical ID to resource ID

akarve:


sir-sigurd (commented):
I think you have to add `<!--pytest.mark.skip-->`.
IMO it would be more logical to move that before `Event source mapping` section.

docs/Troubleshooting.md
```diff
@@ -192,6 +192,25 @@ aws s3api list-object-versions --bucket "$BUCKET" --prefix "$PREFIX"
...
+STACK_NAME="YOUR_QUILT_STACK"
+RESOURCE_ID="YOUR_LOGICAL_ID"
+SG_ID=$(
+  aws cloudformation describe-stack-resource \
+    --stack-name "${YOUR_QUILT_STACK}" \
+    --logical-resource-id "${YOUR_LOGICAL_ID}" \
+    --query 'StackResourceDetail.PhysicalResourceId' \
+    --output text
+)
+aws ec2 describe-security-groups --group-ids "${SG_ID}"
```
sir-sigurd, line 211: I'm not sure we should have this line because it's specific to security groups and if you want to have "full" example you should mention that it's example for SG maybe refer to some existing SG.

sir-sigurd (dismissed):
See comment, it's OK to ignore, otherwise LGTM.

docs/Troubleshooting.md
```diff
@@ -192,6 +192,25 @@ aws s3api list-object-versions --bucket "$BUCKET" --prefix "$PREFIX"
 aws s3api get-object-tagging --bucket "$BUCKET" --key "$PREFIX"
 ```
 
+### Specific logical resources
+
+Sometimes you may wish to find an ID or other information from a logical resource
+in a Quilt stack.
```
akarve, line 198:
```suggestion
Sometimes you may wish to find an ID or other information from a logical resource
in a Quilt stack. The following example is for security groups. Modify the commands as needed
for other resource types.
```

sir-sigurd (approved):

PR #3512, opened 2023-06-26
Hide columns without values in files listings

fiskus:
* Hide "modified date" column if empty. We can fix backend model separately ("Size" column is also hidden in directories without files)
* Fixed text aligning for Size column in Packages tab


catalog/app/containers/Bucket/Listing.tsx
```diff
@@ -985,16 +986,9 @@ export function Listing({
...
-      //   field: 'type',
-      //   headerName: 'Type',
-      //   type: 'string',
-      //   hide: true,
-      //   // TODO: custom filter operators
-      //   // filterOperators: GridFilterOperator[]
-      // },
-      {
+    ]
+    if (items.some(({ size }) => !!size)) {
```
nl0, line 990:
```suggestion
    if (items.some(({ size }) => size != null)) {
```
zero size is a valid case i believe

nl0 (approved):

nl0 (approved):

PR #3517, opened 2023-06-29
Add `jsx-key` ESLint rule, and fix missing key

fiskus:


catalog/app/components/JsonEditor/Note.tsx
```diff
@@ -30,9 +30,13 @@ function getExamples(examples: JsonValue[]) {
...
       'Examples:',
-      ...examples.map((example) => <Code>{printObject(example)}</Code>),
+      ...examples.map((example) => {
+        const str = printObject(example)
+        return <Code key={str}>{str}</Code>
+      }),
     ]
-  return ['Example:', <Code>{printObject(examples[0])}</Code>]
+  const singleExample = printObject(examples[0])
+  return ['Example:', <Code key={singleExample}>{singleExample}</Code>]
```
nl0, line 39: seems like the only difference between single and multiple examples is the leading text, so i'd consolidate these codepaths

nl0 (dismissed):

catalog/app/components/JsonEditor/Note.tsx
fiskus: Done

fiskus:
Also, fixed duplicated keys

nl0 (approved):

PR #3525, opened 2023-07-08
Docs: make stack log collection clear, consistent; use AWS CLI

akarve:


docs/Troubleshooting.md
```diff
@@ -111,19 +111,32 @@ reasons. To expedite resolution of stack deployment issues, it's helpful to
...
-![TemplateBuildMetadata Screenshot](imgs/cloudformation-stack-metadata.png)
-
-2. Quilt stack drift:
-
-    `$ aws cloudformation describe-stack-resource-drifts --stack-name QUILT_STACK_NAME`
+1. Quilt stack outputs:
+    <!--pytest.mark.skip-->
+    ```sh
+    STACK_NAME="YOUR_QUILT_STACK"
+    AWS_REGION="YOUR_REGION"
```
sir-sigurd, line 118: 1. I think both stack name and region vars don't have to be quoted, but that's OK
2. I'm not sure we should mention regions here because that seems to be a part of basic stuff like we're assuming they have credentials and we don't mention `--profile` option
3. it looks like `--region` would be unnecessary if you use `AWS_DEFAULT_REGION` instead of `AWS_REGION`

docs/Troubleshooting.md
```diff
@@ -111,19 +111,28 @@ reasons. To expedite resolution of stack deployment issues, it's helpful to
...
+    aws cloudformation describe-stacks \
+        --stack-name "$STACK_NAME" \
+        --query 'Stacks[].Outputs'
+    ```
 
-3. Recent events for the Quilt stack:
+2. Quilt stack drift status:
+    <!--pytest.mark.skip-->
+    ```sh
+    aws cloudformation describe-stack-resource-drifts \
```
sir-sigurd, line 126: I think prior to this strict detection had to be started with `detect-stack-drift` and use should wait till it finished.

sir-sigurd (dismissed):

docs/Troubleshooting.md
akarve: Good point; fixed

sir-sigurd (approved):

PR #3531, opened 2023-07-14
Replace deprecated webpack loaders

fiskus:
`raw-loader`, `file-loader` and `url-loader` are deprecated for webpack@v5: https://webpack.js.org/guides/asset-modules/

The new build is mostly compatible with the old one, but filename hashes for images and fonts are smaller now


nl0 (approved):
nice

PR #3532, opened 2023-07-14
Update some `devDependencies`

fiskus:
Updated dependencies that don't require changing code.

Some of them has major version update, but mostly because of dropping support of old Node or webpack.


nl0 (approved):

fiskus:
Merged updates using renovate bot

PR #3586, opened 2023-07-19
Remove tslib

fiskus:
It's not used. Tests, build and watcher all work without it.
`tslib` is a dependency at least for GQL codegen, but it is mentioned as dependency in package-lock


nl0:
it is used in the built code. check out the docs: https://www.npmjs.com/package/tslib

> This is a runtime library for [TypeScript](https://www.typescriptlang.org/) that contains all of the TypeScript helper functions.
>
> This library is primarily used by the --importHelpers flag in TypeScript. When using --importHelpers, a module that uses helper functions like __extends and __assign in the following emitted file:

fiskus:
Yeah, thanks, I forgot

PR #3596, opened 2023-07-20
Resolve name conflicts in packages

fiskus:
* Added "Rename" button and prompt for a new file name
* Added new properties to files:
  * `conflict` is this file conflicted with some other file
  * `changed` what was changed in this file
* Removed unused FilesSelector

// TODO:
* [x] get rid of @ts-expect-error
* [x] re-use clodeDomFile for adding/editing meta


catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -88,32 +110,103 @@ export type FilesAction = tagged.InstanceOf<typeof FilesAction>
...
+  return fileCopy
+}
+
+function setKeyValue<T>(key: string, value: T, file: AnyFile): AnyFile {
   if (file instanceof window.File) {
-    const fileCopy = new window.File([file as File], (file as File).name, {
-      type: (file as File).type,
+    const fileCopy = cloneDomFile(file, key)
+    Object.defineProperty(fileCopy, key, {
+      value,
```
nl0, line 161: did you mean to return `fileCopy` here?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -312,7 +433,13 @@ const computeEntries = ({
         }
         return { state, type, path, size: a.size, meta }
       }
-      return { state: 'unchanged' as const, type: 'local' as const, path, size, meta }
+      return {
+        state: changed ? 'modified' : ('unchanged' as const),
```
nl0, line 437: `'modified' as const`?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -1671,182 +1812,3 @@ export function FilesInput({
...
-    dirty: boolean
-    error?: string
-    submitError?: string
-    initial: FilesSelectorState
-  }
-  title: React.ReactNode
-  truncated?: boolean
-}
-
-export function FilesSelector({
```
nl0, line 1722: could you please move this change to a separate PR to reduce the size of this one?

catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
```diff
@@ -612,12 +617,21 @@ export function useEntriesValidator(workflow?: workflows.Workflow) {
 
   return React.useCallback(
     async (entries: ValidationEntry[]) => {
+      const conflictsErrors = entries
+        .filter((e) => !!e.conflict)
+        .map((e) => {
+          const error = new Error(
+            `"${e.logical_key}" is conflicting with "${e.conflict}". Please, rename it or delete`,
```
nl0, line 624:
```suggestion
            `"${e.logical_key}" is conflicting with "${e.conflict}". Rename or remove it please.`,
```

nl0 (commented):
I haven't quite digested the changes in this PR.
Let's move out the component deletion into its own PR and converge/merge #3527 first.
Then, let's deploy this to the dev stack and see how it works live.

fiskus:
I deployed it to searchminimal

nl0 (commented):
First, I'd like to point out that this feature is essential and we were missing it for so long. However, I feel like the current implementation is not yet ready for the prime time.

1. Here's some UX issues / suggestions that are (probably) easy to address:

   1. In the file rename dialog enter triggers package push -- probably should trigger only rename instead.

   2. In the file rename dialog "submit" button label seems too generic, "rename" or smth would be more appropriate.

   3. Instead of always succeeding, I think it's better to check if the name is already taken and disallow submitting to prevent automatic rename (which is kinda confusing when it happens).

   4. Instead of treating conflicts as errors, maybe it would make more sense to just show warnings when auto-renaming files, and don't block submit.

   5. When there are conflicts, "files should match schema" error message is shown, which is confusing and doesn't reflect reality.

   6. Auto-renaming is sometimes buggy and leads to smth like `file (conflict 1) (conflict 1).ext` instead of `file (conflict 2).ext`.

   7. (Nice-to-have, tangential) Consider adding a tooltip describing the state of an entry (e.g. added from s3, removed, renamed from $key, etc. showing the physical key may be useful as well).

   8. (Nice-to-have, tangential) Adding an option to toggle deleted files visibility may be useful to reduce visual clutter, tho it feels orthogonal and somewhat out of scope.

2. My main complaint, tho, is I feel like we're stretching the model too far and the abstractions start leaking -- this makes the logic harder to follow and less predictable, ultimately leading to unexpected errors and difficulties maintaining this code and developing it further. So my advice is to adjust the model to accommodate the new requirements and make it more consistent. Some suggestions:

   1. Use composition: contain / reference File objects instead of mutating (extending) them directly.

   2. "renamed" state implementation is inconsistent with the model, since the model only has the notion of existing / added / deleted entries. Instead of tracking this as "changed" prop, maybe on first rename we should (1) "add" an entry with the new logical key and set "renamedFrom" (or smth like that) prop on it pointing to the original logical key; and (2) mark the original entry as deleted -- this feels more in-line with our current state model and UX. The added entry with "renamedFrom" prop may be rendered with the "revert" action instead of "remove".

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -86,32 +109,107 @@ export type FilesAction = tagged.InstanceOf<typeof FilesAction>
...
+  oldPath: string,
+  newPath: string,
+  mainItems: Record<string, AnyFile>,
+  itemsToCheck: Record<string, AnyFile>,
+  reverted = false,
+) {
+  const file = mainItems[oldPath]
+  if (!file) return mainItems
+  const itemsWithOldNameRemoved = R.dissoc(oldPath, mainItems)
+  const changedFile = setKeyValue<ChangedDict | undefined>(
```
nl0, line 207: there's no need to pass the type arguments explicitly, TS can infer those by itself

fiskus:
> Instead of treating conflicts as errors, maybe it would make more sense to just show warnings when auto-renaming files, and don't block submit.
I think it just unacceptable to submit files with auto-renaming files. It's unexpected for user, that he added a file with one name, but it was submitted with another name. Warning is easy to miss.

> Instead of always succeeding, I think it's better to check if the name is already taken and disallow submitting to prevent automatic rename (which is kinda confusing when it happens).
Yes, good idea and probably easy. First, I interpreted it differently, and it could take a week :)

> Auto-renaming is sometimes buggy and leads to smth like file (conflict 1) (conflict 1).ext instead of file (conflict 2).ext.
Ok, but I couldn't reproduce, I'll try to find it in code

> nice to have: tooltip
Good idea, I'll implement it in standalone PR

> nice to have: hide deleted
Hmm, could be useful, but I never felt I need this

fiskus:
On model

Yes, I agree, I tried to fit into model something that it wasn't designed for. I have two excuses and two solutions.

I wanted to do a minimal change, to have better chances to fit into horizon myself and to fit review into horizon. And I wanted to do a feature separately from refactoring. It's a large separate topic why it's a good idea to make it work first, then make it good (but of course, only if we don't forget and invest time in making good).

I suggest merging this (after addressing UX issues and bugs) and work on refactoring in a separate PR. Also, there is a spec on refactoring file types https://github.com/quiltdata/specs/pull/53, it, for example, has the same proposal to contain file in a separate property. I'll split the spec into three specs as noted in description

catalog/app/containers/Bucket/PackageDialog/EditFileMeta.tsx
```diff
@@ -117,25 +105,26 @@ export default function EditFileMeta({
...
-  const color = React.useMemo(
-    () => (state === 'invalid' || R.isEmpty(value) ? 'inherit' : 'primary'),
-    [state, value],
-  )
-
-  if (!onChange) return null
+  const color = React.useMemo(() => {
+    if (!disabled && state !== 'invalid' && !R.isNil(value) && !R.isEmpty(value)) {
+      return 'primary'
+    }
```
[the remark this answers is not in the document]
fiskus: Just saved a couple of symbols, and done this consistently with `EditFileName.tsx`

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: And what type use for argument in function definition?

```ts
function setKeyValue(key: string, value: any, file: AnyFile): AnyFile {
```
or
```ts
function setKeyValue(key: string, value: string|ChangedDict|Model.EntryMeta, file: AnyFile): AnyFile {
```
?

I'm not sure, I have preferences now over these three variants, just asking

nl0: i don't mean changing function signature, i mean don't pass the type args explicitly, bc TS is smart enough to infer them in most cases, i.e. just use `setKeyValue(...)` instead of `setKeyValue<ChangedDict | undefined>(...)`

fiskus:
* fixed form submit on rename file
* "Submit" → "Rename"
* "files should match schema" → "There are files with identical names. Rename or remove them please"

---
* treat conflict errors as warnings -- skipped
* nice-to-haves -- TODO
* model adjustments -- TODO
---
* Check if user renames to existing file -- providing data to the editing component would be too hacky, and moving editing component up looks a good thing (also, for performance), but it will take more time
* `"file (conflict 1) (conflict 1).ext"` -- couldn't find this bug. Though, it can happen when you rename `"file (conflict 2).ext"` to `"file (conflict 1).ext"` (not to `"file.ext"`), manually

catalog/app/components/Dialog/Prompt.tsx
```diff
@@ -82,8 +96,8 @@ export function usePrompt({ initialValue, title, onSubmit, validate }: PromptPro
   }, [])
   const close = React.useCallback(() => setOpened(false), [])
   const handleSubmit = React.useCallback(
-    (value: string) => {
-      onSubmit(value)
+    (...args: [React.FormEvent, string]) => {
```
nl0, line 99: you can make it a little DRYer here:
```suggestion
    (...args: Parameters<typeof onSubmit>) => {
```

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -181,8 +181,8 @@ function addFile<T extends AnyFile, M extends Record<string, AnyFile>>(
     ...mainItems,
     [resolvedName]:
       resolvedName === path
-        ? setKeyValue<undefined>('conflict', undefined, file)
-        : setKeyValue<typeof path>('conflict', path, file),
+        ? setKeyValue('conflict', undefined, file)
+        : setKeyValue('conflict', path, file),
```
nl0, line 185: maybe just `setKeyValue('conflict', resolvedName === path ? undefined : path, file)`?

nl0:
> I wanted to do a minimal change, to have better chances to fit into horizon myself and to fit review into horizon

well, in the end it did not fit, so i don't think we should rush and try to fit it anyways in some half-baked form

> It's a large separate topic why it's a good idea to make it work first, then make it good (but of course, only if we don't forget and invest time in making good).

it's ok (in my book) to make it work and then make it good _before shipping_.

i can't agree that breaking things, shipping them, and then fixing them is a valid approach.

i'm kinda ok with expanding the model and postponing a proper refactoring, but i'm not ok with breaking the model, so i still feel like this requires addressing:

> "renamed" state implementation is inconsistent with the model, since the model only has the notion of existing / added / deleted entries. Instead of tracking this as "changed" prop, maybe on first rename we should (1) "add" an entry with the new logical key and set "renamedFrom" (or smth like that) prop on it pointing to the original logical key; and (2) mark the original entry as deleted -- this feels more in-line with our current state model and UX. The added entry with "renamedFrom" prop may be rendered with the "revert" action instead of "remove".

since the current approach is too brittle and has edge cases that aren't handled

nl0:
> > I wanted to do a minimal change, to have better chances to fit into horizon myself and to fit review into horizon
> 
> well, in the end it did not fit, so i don't think we should rush and try to fit it anyways in some half-baked form
> 

i realize that i am accountable for this as well to some extent (as a reviewer).

@drernie @akarve you can override and approve this PR if you think it's more important to get this out the door sooner and backlog the polish / refactoring, but i'm convinced that we should not compromise on quality. scope hammering is not compromising on quality, it's compromising on, well, scope, and this feature didn't make it.

nl0:
> * fixed form submit on rename file

👍 

>     * Check if user renames to existing file -- providing data to the editing component would be too hacky, and moving editing component up looks a good thing (also, for performance), but it will take more time

just pass a callback to check if the file exists in the manifest?

>     * `"file (conflict 1) (conflict 1).ext"` -- couldn't find this bug. Though, it can happen when you rename `"file (conflict 2).ext"` to `"file (conflict 1).ext"` (not to `"file.ext"`), manually

not sure how i managed that lol



Screenshot 2023-08-04 at 13 57 57

catalog/app/components/Dialog/Prompt.tsx
```diff
@@ -25,10 +29,11 @@ function Dialog({
   const error = React.useMemo(() => validate(value), [validate, value])
   const handleChange = React.useCallback((event) => setValue(event.target.value), [])
   const handleSubmit = React.useCallback(
-    (event) => {
+    (event: React.FormEvent) => {
+      // event.stopPropagation()
```
nl0, line 33: pls uncomment or remove

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -19,9 +21,29 @@ import useMemoEq from 'utils/useMemoEq'
...
 
+function resolveNameConflictRudely(name: string, attempt: number) {
+  const ext = extname(name)
+  const base = name.split(ext)[0]
+  return `${base} (conflict ${attempt})${ext}`
+}
+
+function resolveName(
+  name: string,
+  obj: Record<string, any>,
```
nl0, line 36: this argument could probably have a more descriptive name

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -19,9 +21,29 @@ import useMemoEq from 'utils/useMemoEq'
...
+  attempt: number = 1,
+): string {
+  if (!obj[name]) return name
+
+  const newName = resolveNameConflictRudely(name, attempt)
+  if (!obj[newName]) return newName
+
+  return resolveName(name, obj, attempt + 1)
```
nl0, line 44: i doubt that would be an issue in practice, but in theory this can lead to error by exceeding the recursion depth (since js doesn't have a tail-call optimization, unfortunately), so if we are being pedantic, this function should be rewritten using a loop instead of recursion

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -86,32 +109,108 @@ export type FilesAction = tagged.InstanceOf<typeof FilesAction>
 
 export type LocalFile = FileWithPath & FileWithHash
 
+// TODO: queue/array
```
nl0, line 112: what does this comment mean?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -86,32 +109,108 @@ export type FilesAction = tagged.InstanceOf<typeof FilesAction>
...
+) {
+  const resolvedName = resolveName(
+    path,
+    itemsToCheck ? { ...mainItems, ...itemsToCheck } : mainItems,
+  )
+  return {
+    ...mainItems,
+    [resolvedName]: setKeyValue(
+      'conflict',
+      resolvedName === path ? undefined : path,
```
nl0, line 184: if i read it correctly, this won't unset the prop on a File instance, because undefined is ignored (see line 160). or am i missing smth?

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -86,32 +109,108 @@ export type FilesAction = tagged.InstanceOf<typeof FilesAction>
...
+
+function addFile<T extends AnyFile, M extends Record<string, AnyFile>>(
+  path: string,
+  file: T,
+  mainItems: M,
+  itemsToCheck?: Record<string, AnyFile>,
+) {
+  const resolvedName = resolveName(
+    path,
+    itemsToCheck ? { ...mainItems, ...itemsToCheck } : mainItems,
```
nl0, line 178: `{...undefined}` works just fine, so you can skip the check

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
```diff
@@ -86,32 +109,108 @@ export type FilesAction = tagged.InstanceOf<typeof FilesAction>
...
     return fileCopy
   }
-  return R.assoc('meta', meta, file)
+  return R.assoc(key, value, file)
+}
+
+function addFile<T extends AnyFile, M extends Record<string, AnyFile>>(
+  path: string,
+  file: T,
+  mainItems: M,
```
nl0, line 173: why do you need this type var here? you're not using it. or is it needed for the return type?

catalog/app/containers/Bucket/PackageDialog/PackageCreationForm.tsx
```diff
@@ -625,6 +624,8 @@ function PackageCreationForm({
                       validate={validateFiles as FF.FieldValidator<$TSFixMe>}
                       validateFields={['files']}
                       errors={{
+                        conflict:
+                          'There are files with identical names. Rename or remove them please',
```
nl0, line 628:
```suggestion
                          'There are entries with conflicting logical keys. Rename or remove them please',
```
not sure if this is more user-friendly for less-tech-savvy people, but it seems more correct

catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
```diff
@@ -575,16 +576,26 @@ function useFetchEntriesSchema(workflow?: workflows.Workflow) {
...
 
 export interface ValidationEntry {
+  conflict?: string
   logical_key: string
   size: number
   meta?: JsonRecord
 }
 
+export class EntryNameError extends BaseError {
+  static data: ValidationEntry
```
nl0, line 586: i don't quite get it why this is needed (or at least why it's `static`)

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: It's here for keys only, so it doesn't matter what kind of object is here. Just some object with keys

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: Use a list of objects instead of a single one. So, we can use infinite "revert"

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: It creates a copy of a file, first without 'conflict' (at cloning step), and then 'conflict' is ignoring (at setting step)

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: Do you mean you don't want to use recursion in JS in general, or in this particular case?

catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
fiskus: This is used for showing in which file there was an error. Under the FilesInput
`static` is a mistyping copypasta, thanks

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
fiskus: Yes, I can skip it now. Earlier types were stricter, but I gave up

catalog/app/containers/Bucket/PackageDialog/FilesInput.tsx
nl0: i don't "don't want" to, it's just not safe (may lead to an exception) if the depth is unbounded, and here it well may be, since we don't control what files (and how many) a user adds.
idk, maybe it's kinda ok in practice in this particular case ([source](https://javascript.info/recursion)):

> The maximal recursion depth is limited by JavaScript engine. We can rely on it being 10000, some engines allow more, but 100000 is probably out of limit for the majority of them. There are automatic optimizations that help alleviate this (“tail calls optimizations”), but they are not yet supported everywhere and work only in simple cases.

but in general, in js recursion shouldn't be used for possibly unbounded iterative processes (tho it's ok to use it to traverse recursive data structures, ofc)

fiskus: Ok, but IMO it's an overkill to sacrifice elegance of recursion to cover edge case with 10000 duplicate names. Also, we already have a limit on number of files: 10000

nl0, line 44: i'm not sure what sacrifice do you mean:
```suggestion
function resolveName(name: string, obj: Record<string, any>): string {
  let newName = name, attempt = 1
  while (newName in obj) newName = resolveNameConflictRudely(name, attempt++)
  return newName
```

drernie:
Is this obsoleted by @fiskus' work on Selection's last Horizon?

fiskus:
No, it's still relevant. Alexei couldn't approve this due to his disagreement on tradeoffs I made in design. To reduce tradeoffs and made more suitable design for the tasks, I need to refactor this code first. Also, this unification of types can help me refactor https://github.com/quiltdata/quilt/pull/3681

fiskus:
Buried in a review hell. It would be easier for me to start from scratch

PR #3613, opened 2023-07-21
Unify search, TSify

nl0:
when searching from a bucket page, use global search and pre-select the bucket

## TODO


catalog/app/containers/NavBar/Search.tsx
```diff
@@ -0,0 +1,362 @@
...
+    setHelpOpen(false)
+    onBlur?.()
+  }, [onBlur])
+
+  const handleHelpOpen = React.useCallback(() => setHelpOpen(true), [])
+
+  const handleHelpClose = React.useCallback(() => setHelpOpen(false), [])
+
+  const onQuery = React.useCallback(
+    (strPart: string) => change(`${value} ${strPart}`),
```
fiskus, line 228: Just a micro-optimizaton:
`onQuery` is triggered every `value` change, but Help actually can be independent of `value`
```tsx
const onQuery = useCallback((strPart) => setValue((v) => `$x{v} ${strPart}`), [])
```

catalog/app/containers/NavBar/Search.tsx
```diff
@@ -0,0 +1,362 @@
...
+              </M.Icon>
+            </Lab.ToggleButton>
+          )}
+        </M.InputAdornment>
+      }
+    />
+  )
+}
+
+function useSearchUrlState(bucket: string | undefined) {
```
fiskus, line 159:
```suggestion
function useSearchUrlState(bucket?: string) {
```

fiskus (dismissed):

catalog/app/containers/NavBar/Search.tsx
nl0: that actually means something different, but in this case it would work as well

fiskus (approved):

PR #3617, opened 2023-07-25
hotfix embed

nl0:


fiskus (approved):

fiskus:
I wonder could we catch it if Embed was tsified?

PR #3633, opened 2023-07-26
catalog settings: suppress confusing error message, log to sentry

nl0:


catalog/app/utils/CatalogSettings.tsx
```diff
@@ -39,10 +40,14 @@ async function fetchSettings({ s3 }: { s3: S3 }) {
     const text = res.Body!.toString('utf-8')
     return JSON.parse(text) as CatalogSettings
   } catch (e) {
+    const { code } = e as any
+    // assuming this is caused by missing settings file, which is expected
+    if (code === 'AccessDenied' || code === 'NoSuchKey') return null
```
fiskus, line 45: I don't expect that 'AccessDenied' is expected :)

nl0: it is when a user doesn't have a ListBucket permission, so in production we are only getting these 403 AccessDenied errors. i've added NoSuchKey here just in case, in normal setting (with quilt-managed permissions) it doesn't happen

fiskus: I mean, why we're hiding this error. It looks like legitimate error - user should fix his permission?

nl0: no, if the settings file is missing, s3 returns 403 AccessDenied, not 404 NoSuchKey, because a quilt user normally does not have a ListBucket permission for service bucket. also, a user cannot (and should not) fix this, because these permissions are managed by quilt

fiskus (approved):

PR #3634, opened 2023-07-27
Hide .quilt directory

fiskus:
Added `preferences.blocks.browser.hidden` property to show hidden files. Probably, it is confusing, and is better to use `showHidden`. Or maybe it's ok in context of entire config where `true` shows and `false` hides

Also, memoized Listing in PackageTree


nl0 (changes requested):
i think this should be a catalog-level option, not bucket-level. @akarve wdyt?

fiskus:
We could return to this if anyone express interest in hiding `.quilt` once more.

PR #3655, opened 2023-08-03
Remove @mui/styles @mui/system deps from package.json; they are deps of @mui/core

fiskus:


akarve (approved):
How did this work before? Did dependency paths change?

fiskus:
`styles` are re-exported from `core/styles` and they are included in `core`, so no reason to keep them in package.json as standalone dependency. But this works both ways with `styles` or with `core/styles`

Also, we already use `core/styles` by such imports:
```
...
const useStyles = M.makeStyles()
```

Because, `core` re-exports all `styles/*`. So, this probably could be broken if we use `styles` with some diverged API, but we always use synced versions of `core/styles` and `styles`.

catalog/app/website/components/Illustration.js
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { Box } from '@material-ui/core'
-import { styled } from '@material-ui/styles'
+import { styled } from '@material-ui/core/styles'
```
nl0, line 3: these two are not strictly identical, tho probably it should be okay for our use case

PR #3656, opened 2023-08-03
Index metadata fields

sir-sigurd:
# TODO

- [x] use logger instead of print
- [x] <del>handle null values?</del> (seems unneeded)
- [x] handle properties with dots
- [x] <del>support dates in US format?</del> (punt)
- [x] <del>do we really need type field?</del> (seems so)
- [x] support text arrays?
- [x] <del>ignore fields with very long names (e.g. > 256)? </del> (maybe, but not now)
- [x] tests


lambdas/es/indexer/index.py
```diff
@@ -297,6 +297,63 @@ def do_index(
         logger_.debug("%s indexed as package (%s)", key, event_type)
 
 
+def _try_parse_date(s: str) -> datetime.datetime:
+    # XXX: do we need to support more formats?
+    # XXX: do we need to preserve UTC timezone?
```
nl0, line 302: why not?

nl0 (commented):
looks fine for a first step of PoC =)

lambdas/es/indexer/index.py
sir-sigurd: 1. `fromisoformat()` can't parse ISO format string that ends on `Z`
2. if ES treats timestamps without timezone the same way as with UTC then probably there is no point to complicate things preserving timezone

lambdas/es/indexer/index.py
```diff
@@ -299,6 +304,94 @@ def do_index(
...
+            raise Exception(f"Bad workflow config URL {config_url}")
+
+        config_url_parsed = urllib.parse.urlparse(config_url)
+        query = urllib.parse.parse_qs(config_url_parsed.query)
+        version_id = query.pop('versionId', [None])[0]
+        if query:
+            raise Exception(f"Unexpected S3 query string: {config_url_parsed.query!r}")
+
+        return {
+            "config_version_id": version_id,  # XXX: how to handle None?
```
nl0, line 380: well i guess None is a valid case (non-versioned buckets), so we just use it as is. or is it "null" string? which is also okay, probably, untils it works as expected with s3 apis. anyways, it's not used in the current version of faceted search, so doesn't matter that much

lambdas/es/indexer/index.py
```diff
@@ -299,6 +304,94 @@ def do_index(
         logger_.debug("%s indexed as package (%s)", key, event_type)
 
 
+def _try_parse_date(s: str) -> Optional[datetime.datetime]:
+    # XXX: do we need to support more formats?
```
nl0, line 308: i'd say we _want_ that, but don't _need_ for the time being

lambdas/es/indexer/index.py
```diff
@@ -299,6 +304,94 @@ def do_index(
...
+            else:
+                logger.warning("ignoring value of type %s", type(v))
+                continue
+
+            yield path + (k,), type_, raw_value, v
+
+
+def get_metadata_fields(meta):
+    if not isinstance(meta, dict):
+        # XXX: can we do something better?
```
nl0, line 351: i think it's good enough, at least i don't have better ideas

lambdas/es/indexer/test/test_index.py
```diff
@@ -1829,3 +1838,174 @@ def test_extract_pptx():
...
+                "text": json.dumps(True),
+                "boolean": True,
+            },
+        ),
+    ],
+)
+def test_get_metadata_fields_values(src_value, expected_field):
+    field_name = "a"
+
+    assert index.get_metadata_fields(
```
nl0, line 1924: oh black, what are doing to me

nl0 (approved):
lgtm

PR #3684, opened 2023-08-09
Revert react-redux update

fiskus:
Last react-redux@v8 was a little problematic. All went well, we weren't forced to change anything, and everything worked as before.
Except one error, that appeared on every page load. Something relevant to coupling Redux, S3 Provider and Graphql Provider.
So it makes sense to revert this update, and figure out how to fix it then.


sir-sigurd (approved):
Might make sense to add PR description or commit message to make clear why it's needed.

PR #3685, opened 2023-08-09
Add a region config variable to support GovCloud

dimaryaz:
GovCloud doesn't have us-east-1, so we need to explicitly set a valid region for S3. It's also a potential optimization for any non-us-east-1 stacks (assuming most S3 buckets are in the same region as the stack).

Also use it for Athena: seems less hacky that extracting the region from apiGatewayEndpoint.

### DIFF
--- a/catalog/app/utils/AWS/Athena.tsx
+++ b/catalog/app/utils/AWS/Athena.tsx
@@ -1,5 +1,4 @@
 import Athena from 'aws-sdk/clients/athena'
-import * as R from 'ramda'
 import * as React from 'react'
 
 import cfg from 'constants/config'
@@ -8,13 +7,7 @@ import useMemoEqLazy from 'utils/useMemoEqLazy'
 import * as Config from './Config'
 import * as Credentials from './Credentials'
 
-const getRegion: (input: string) => string = R.pipe(
-  R.match(/\.([a-z]{2}-[a-z]+-\d)\.amazonaws\.com/),
-  R.nth(1),
-  R.defaultTo('us-east-1'),
-)
-
-const region = getRegion(cfg.apiGatewayEndpoint)
+const region = cfg.defaultRegion
 
 const AthenaContext = React.createContext<() => Athena | null>(() => null)
 
--- a/catalog/app/utils/AWS/Athena.tsx
+++ b/catalog/app/utils/AWS/Athena.tsx
@@ -7,7 +7,7 @@ import useMemoEqLazy from 'utils/useMemoEqLazy'
 import * as Config from './Config'
 import * as Credentials from './Credentials'
 
-const region = cfg.defaultRegion
+const region = cfg.region
 
 const AthenaContext = React.createContext<() => Athena | null>(() => null)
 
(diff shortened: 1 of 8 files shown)


catalog/config.js.example
```diff
@@ -1,4 +1,5 @@
 window.QUILT_CATALOG_CONFIG = {
+  "defaultRegion": "us-east-1",
```
sir-sigurd, line 2: Shouldn't it be required property?

sir-sigurd (dismissed):

catalog/config.js.example
dimaryaz: Not sure what you mean?

sir-sigurd: sorry, I meant in JSON schema

dimaryaz: Oh, ok, will fix it

fiskus:
* It's a little confusing for me to name it `defaultRegion`, because I don't know how to override this default. So, maybe, just `"region"`?
* Is this a breaking change for every stack? If so, maybe, we can default this region to 'us-east-1' if it's missing in config?
* Is there corresponding PR to for adjusting config for quilt-t4-staging? Or, it will be done somewhere in AWS?

catalog/app/utils/AWS/Athena.tsx
```diff
@@ -8,13 +7,7 @@ import useMemoEqLazy from 'utils/useMemoEqLazy'
...
 import * as Credentials from './Credentials'
 
-const getRegion: (input: string) => string = R.pipe(
-  R.match(/\.([a-z]{2}-[a-z]+-\d)\.amazonaws\.com/),
-  R.nth(1),
-  R.defaultTo('us-east-1'),
-)
-
-const region = getRegion(cfg.apiGatewayEndpoint)
+const region = cfg.defaultRegion
```
fiskus, line 10: Yes, this is definitely better!

fiskus (approved):

sir-sigurd (dismissed):

dimaryaz:
> It's a little confusing for me to name it defaultRegion, because I don't know how to override this default. So, maybe, just "region"?

The S3 buckets can be in different regions, and the S3 client handles that automatically - but it needs to try one of the regions first. So that's why I didn't want to call it "region". (Though, the S3 client's parameter is called "region"...)

@sir-sigurd - what are your thoughts? I'm fine with either name.

> Is this a breaking change for every stack? If so, maybe, we can default this region to 'us-east-1' if it's missing in config?

The S3 client already defaults to 'us-east-1'. I think we could just make "defaultRegion" not required.

> Is there corresponding PR to for adjusting config for quilt-t4-staging? Or, it will be done somewhere in AWS?

Yes, every stack will have it: https://github.com/quiltdata/deployment/pull/1503/files

akarve:
> * It's a little confusing for me to name it `defaultRegion`, because I don't know how to override this default. So, maybe, just `"region"`?

How about we call this `stackRegion`?

sir-sigurd:
> > * It's a little confusing for me to name it `defaultRegion`, because I don't know how to override this default. So, maybe, just `"region"`?
> 
> How about we call this `stackRegion`?

I actually thought about that, but maybe simple `region` could be good enough.

dimaryaz:
Ok, I'll just make it `region` then

sir-sigurd (approved):

PR #3686, opened 2023-08-10
Use ResizeObserver third-party hook

fiskus:
It fixes some edge cases, when ResizeObserver emits error.

I can consistently reproduce this error: open file with metadata (JsonDisplay should be there), scroll back and forth while page is loading (it is not required to do actual scroll, just scroll events by mousewheel).

We can fix this in our code, or just use that library, that we already use for Status page.

### DIFF
--- a/catalog/app/components/JsonDisplay/JsonDisplay.js
+++ b/catalog/app/components/JsonDisplay/JsonDisplay.js
@@ -1,6 +1,7 @@
 import cx from 'classnames'
 import * as R from 'ramda'
 import * as React from 'react'
+import useResizeObserver from 'use-resize-observer'
 import * as M from '@material-ui/core'
 
 import * as NamedRoutes from 'utils/NamedRoutes'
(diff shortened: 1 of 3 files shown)


akarve (approved):

catalog/app/components/JsonDisplay/JsonDisplay.js
```diff
@@ -336,7 +322,7 @@ export default function JsonDisplay({
 }) {
   const ref = React.useRef(null)
   const classes = useStyles()
-  const currentBPWidth = useElementWidth(ref)
+  const { width: currentBPWidth } = useResizeObserver({ ref })
   const computedKeys = React.useMemo(() => {
     if (showKeysWhenCollapsed === true) return Number.POSITIVE_INFINITY
     if (showKeysWhenCollapsed === false) return Number.POSITIVE_INFINITY
```
akarve, line 328: This looks awkward. We return the same thing not matter the state of `showKeysWhenCollapsed`?

fiskus: Yes, probably, should be NEGATIVE_INFINITY or `0`. The code was never changed since created https://github.com/quiltdata/quilt/pull/1806, but, this condition was not used when it was added, and it is not used today :)
I'll remove it in standalone PR

PR #3689, opened 2023-08-11
Show "invalid username" error

fiskus:
Maybe it's also safe to remove "is not valid" check

I think it broke after https://github.com/quiltdata/enterprise/pull/535

### DIFF
--- a/catalog/app/containers/Admin/Users/Users.js
+++ b/catalog/app/containers/Admin/Users/Users.js
@@ -89,6 +89,11 @@ function Invite({ close, roles, defaultRoleId }) {
             username: 'invalid',
           }
         }
+        if (APIConnector.HTTPError.is(e, 400, /Username is invalid/)) {
+          return {
+            username: 'invalid',
+          }
+        }
         if (APIConnector.HTTPError.is(e, 409, /Username already taken/)) {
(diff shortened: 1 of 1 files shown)


catalog/app/containers/Admin/Users/Users.js
```diff
@@ -89,6 +89,11 @@ function Invite({ close, roles, defaultRoleId }) {
             username: 'invalid',
           }
         }
+        if (APIConnector.HTTPError.is(e, 400, /Username is invalid/)) {
```
nl0, line 92: why not just adjust the regex above to merge these cases? and since we're not going to use it with the older registry, i think it's safe to keep only the current option

nl0:
> I think it broke after https://github.com/quiltdata/enterprise/pull/535

yeah seems so

catalog/app/containers/Admin/Users/Users.js
fiskus: > why not just adjust the regex above to merge these cases?

Because, it's easier to scan two strings instead of one regex

>  i think it's safe to keep only the current option

Ok, I removed old one

nl0 (approved):

PR #3690, opened 2023-08-14
Add filter to Policies and Roles tables, refactor filter

fiskus:
* ~Moved repeatable filter code to `Table.useOrdering`~ Created `Table.useFiltering`
* Converted the most of the `Table` to Typescript (splitted code to typed.tsx and untyped.js)
* Moved `<Filter />` to `<Table.Filter />` and used it as one of toolbar `actions`
* Added filter for list of attached policies, associated roles, buckets. Re-use `Table.useFiltering` for these components

Screenshot from 2023-08-15 16-28-26
Screenshot from 2023-08-15 16-30-20
Screenshot from 2023-08-15 16-30-47
Screenshot from 2023-08-15 16-31-01

### DIFF
--- a/catalog/app/containers/Admin/Buckets/Buckets.tsx
+++ b/catalog/app/containers/Admin/Buckets/Buckets.tsx
@@ -29,7 +29,6 @@ import { useTracker } from 'utils/tracking'
 import * as Types from 'utils/types'
 import * as validators from 'utils/validators'
 
-import Filter from '../Filter'
 import * as Form from '../Form'
 import * as Table from '../Table'
 
@@ -1186,7 +1185,7 @@ function CustomBucketIcon({ src }: CustomBucketIconProps) {
   return <BucketIcon alt="" classes={classes} src={src} title="Default icon" />
 }
 
-const columns = [
+const columns: Table.Column<BucketConfig>[] = [
   {
     id: 'name',
     label: 'Name (relevance)',
@@ -1272,17 +1271,14 @@ interface CRUDProps {
 
 function CRUD({ bucketName }: CRUDProps) {
   const { bucketConfigs: rows } = GQL.useQueryS(BUCKET_CONFIGS_QUERY)
-  const [filter, setFilter] = React.useState('')
-  const filtered = React.useMemo(
-    () =>
-      filter
-        ? rows.filter(({ name, title }) =>
-            (name + title).toLowerCase().includes(filter.toLowerCase()),
-          )
-        : rows,
-    [filter, rows],
-  )
-  const ordering = Table.useOrdering({ rows: filtered, column: columns[0] })
+  const filtering = Table.useFiltering({
+    rows,
+    filterBy: ({ name, title }) => name + title,
+  })
+  const ordering = Table.useOrdering({
+    rows: filtering.filtered,
+    column: columns[0],
+  })
   const pagination = Pagination.use(ordering.ordered, {
     // @ts-expect-error
     getItemId: R.prop('name'),
@@ -1346,8 +1342,7 @@ function CRUD({ bucketName }: CRUDProps) {
       </M.Dialog>
 
       <Table.Toolbar heading="Buckets" actions={toolbarActions}>
-        {/* @ts-expect-error */}
-        <Filter value={filter} onChange={setFilter} />
+        <Table.Filter {...filtering} />
       </Table.Toolbar>
       <Table.Wrapper>
         <M.Table size="small">
@@ -1361,9 +1356,7 @@ function CRUD({ bucketName }: CRUDProps) {
                 style={{ cursor: 'pointer' }}
               >
                 {columns.map((col) => (
-                  // @ts-expect-error
                   <M.TableCell key={col.id} align={col.align} {...col.props}>
-                    {/* @ts-expect-error */}
                     {(col.getDisplay || R.identity)(col.getValue(i), i)}
                   </M.TableCell>
                 ))}
--- a/catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
+++ b/catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
@@ -1,10 +1,13 @@
 import * as React from 'react'
+import * as R from 'ramda'
 import * as RF from 'react-final-form'
 import * as M from '@material-ui/core'
 
 import * as GQL from 'utils/GraphQL'
 import StyledLink from 'utils/StyledLink'
 
+import * as Table from '../Table'
+import Filter from './Filter'
 import { MAX_POLICIES_PER_ROLE } from './shared'
 
 import POLICIES_QUERY from './gql/Policies.generated'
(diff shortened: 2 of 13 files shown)


nl0 (changes requested):
i see a fundamental design issue here -- filtering should not be an "extension" for ordering. instead, it should be a standalone composable unit (a hook in this case), which gets applied before ordering and its application result is be passed down to ordering as an input

catalog/app/containers/Admin/Table/typed.tsx
```diff
@@ -95,73 +163,14 @@ export function Toolbar({
         )}
       </div>
       <div className={classes.spacer} />
-      {children}
+      {ordering?.filter && <Filter {...ordering.filter} />}
```
nl0, line 166: i feel very strongly against such "abstracting" in favor of simple composition. like, if we have a filtering model exposed by the hypothetical `useFiltering` hook, we can just render this Filter component explicitly using that model when we need it, and not overload the generalized Toolbar component with very specific functionality which can be easily "composed" instead

nl0:
>  (splitted code to typed.tsx and untyped.js)

i'd suggest to rename `typed.tsx` into `Table.tsx` following the implicit convention observed throughout the codebase (keeping the untyped stuff in `untyped.js` until it's converted)

catalog/app/containers/Admin/Table/typed.tsx
fiskus: Or, this filter can be in any Table and be part of both Table and Table.useOrdering?

fiskus: Maybe, by "order" you mean "sort" only?

fiskus: @nl0 ping

nl0: the original intent behind that design was to make individual pieces composable and not inter-dependent whenever possible. 

> Maybe, by "order" you mean "sort" only?

well, these terms are almost interchangeable in the current context, and filtering is orthogonal to order/sorting

fiskus: So, can I rename it to `useSorting`? Because in my head `useOrdering` means to keep items in order: sort and filter them.

nl0: well, the words "order" / "ordering" feel natural to me to describe the concept of, well, elements (i.e. of a set) having a particular order, following one another, while "sorting" being a specific operation of sorting a list or smth (maybe because of me thinking in terms of functional languages where "Order" quite often is a literal name of a typeclass providing specific properties related to ordering things). it correlates well with the often used phrase "order by". but in the end ordering feels distinct from filtering. order doesn't imply including / excluding elements, only their sequence. filtering is orthogonal to ordering.

fiskus:
I rewrote the code, moving filter outside `Table.useOrdering` and `Table.Toolbar`

catalog/app/containers/Admin/RolesAndPolicies/AttachedPolicies.tsx
```diff
@@ -38,34 +41,52 @@ function PolicySelectionDialog({
...
-              </M.ListItemText>
-            </M.ListItem>
-          ))
-        ) : (
-          <M.DialogContent dividers>
-            <M.Typography>No more policies to attach</M.Typography>
-          </M.DialogContent>
-        )}
-      </M.List>
+      {policies.length && (
```
nl0, line 56:
```suggestion
      {!!policies.length && (
```

catalog/app/containers/Admin/RolesAndPolicies/BucketsPermissions.tsx
```diff
@@ -39,33 +43,51 @@ function BucketAddDialog({ open, onClose, buckets, addBucket }: BucketAddDialogP
...
-                </M.Box>
-              </M.ListItemText>
-            </M.ListItem>
-          ))}
-        </M.List>
-      ) : (
-        <M.DialogContent>
-          <M.Typography>No more buckets to add</M.Typography>
-        </M.DialogContent>
+      {buckets.length && (
```
nl0, line 58:
```suggestion
      {!!buckets.length && (
```

catalog/app/containers/Admin/Table/Table.tsx
```diff
@@ -68,20 +136,29 @@ const useToolbarStyles = M.makeStyles((t) => ({
...
     flex: '0 0 auto',
   },
 }))
 
+interface ToolbarProps {
+  heading: React.ReactNode
+  selected?: number
+  actions?: Action[]
+  selectedActions?: Action[]
+  filtering?: { value: string; onChange: (f: string) => void }
```
nl0, line 153: this one seems unused

catalog/app/containers/Admin/Table/Table.tsx
```diff
@@ -6,15 +6,58 @@ import { lighten } from '@material-ui/core/styles/colorManipulator'
...
+
+const changeDirection = (d: Direction) => (d === 'asc' ? 'desc' : 'asc')
+
+export type Column<Row> = {
+  align?: string
+  getValue: $TSFixMe
+  hint?: M.TooltipProps['title']
+  id: string
+  label: React.ReactNode
+  // props?: M.TableCellProps
```
nl0, line 39: is this still needed?

catalog/app/containers/Admin/Buckets/Buckets.tsx
```diff
@@ -1266,23 +1265,25 @@ const columns = [
...
-    () =>
-      filter
-        ? rows.filter(({ name, title }) =>
-            (name + title).toLowerCase().includes(filter.toLowerCase()),
-          )
-        : rows,
-    [filter, rows],
-  )
-  const ordering = Table.useOrdering({ rows: filtered, column: columns[0] })
+  const filtering = Table.useFiltering<BucketRow>({
```
nl0, line 1279: afaict you don't have to provide an explicit type arg (`<BucketRow>`) here, bc `rows` var is typed, and so `useFiltering` should infer the type of the argument just fine

catalog/app/containers/Admin/Buckets/Buckets.tsx
```diff
@@ -1293,6 +1294,7 @@ function CRUD({ bucketName }: CRUDProps) {
   const history = RRDom.useHistory()
 
   const toolbarActions = [
+    <Table.Filter {...filtering} key="toolbar-filter" />,
```
nl0, line 1297: to me previous approach (passing extra controls as children) seemed somewhat cleaner, bc actions and arbitrary controls weren't mixed together

catalog/app/containers/Admin/Table/Table.tsx
```diff
@@ -6,15 +6,58 @@ import { lighten } from '@material-ui/core/styles/colorManipulator'
...
+  rows: readonly Row[]
+  filterBy: (r: Row) => string
+}
+
+export function useFiltering<Row>({ rows, filterBy }: UseFilteringProps<Row>) {
+  const [value, onChange] = React.useState('')
+  const filtered = React.useMemo(
+    () =>
+      value
+        ? R.filter<Row>(
```
nl0, line 19: won't using `rows.filter()` be cleaner here?

catalog/app/containers/Admin/Buckets/Buckets.tsx
fiskus: Yes, it works without type args. It works for type checking, but resulting types are barely readable for my taste.

catalog/app/containers/Admin/Table/Table.tsx
```diff
@@ -6,15 +6,56 @@ import { lighten } from '@material-ui/core/styles/colorManipulator'
...
 
-export function useOrdering({ rows, ...opts }) {
+type Direction = 'asc' | 'desc'
+
+const changeDirection = (d: Direction) => (d === 'asc' ? 'desc' : 'asc')
+
+export type Column<Row> = {
+  align?: M.TableCellProps['align']
+  getDisplay: $TSFixMe
+  getValue: $TSFixMe
```
nl0, line 33: these two can be made at least a little bit stricter by typing them as `(r: Row) => $TSFixMe`

catalog/app/containers/Admin/Table/Table.tsx
```diff
@@ -186,36 +208,25 @@ export function InlineActions({ actions = [], children = undefined, ...props })
   )
 }
 
+const useWrapperStyles = M.makeStyles({
+  root: {
+    overflowX: 'auto',
+  },
+})
+
+export function Wrapper({ className = undefined, ...props }) {
```
nl0, line 217: doesn't TS complain about missing type annotations?

nl0 (approved):

catalog/app/containers/Admin/Table/Table.tsx
fiskus: I added these types:
```ts
+  getDisplay?: (v: $TSFixMe, r: Row, opts?: $TSFixMe) => React.ReactNode
+  getValue: (r: Row) => $TSFixMe
```

catalog/app/containers/Admin/Table/Table.tsx
fiskus: No. I think it's because `className` has inferred `undefined` type, and destructured object - `Record<string, any>`

I added this:
```tsx
+export function Wrapper({
+  className = undefined,
+  ...props
+}: React.HTMLAttributes<HTMLDivElement>) {
+  const classes = useWrapperStyles()
+  return <div className={cx(classes.root, className)} {...props} />
+}
```

fiskus:
Resolved CHANGELOG conflict, added types, and squashed commits. I described new commits in comments

catalog/app/containers/Admin/Table/Table.tsx
fiskus: Also, changed `const columns = [...] as Table.Column[]` to `const columns: Table.Column[] = [...]`

nl0 (approved):

PR #3691, opened 2023-08-14
Index S3 tags

sir-sigurd:
# Description
This indexes S3 object tags as text i.e. like this `"key1 value1 key2 value2 ..."` into `s3_tags` field.

# TODO

- [x] Unit tests


lambdas/es/indexer/index.py
```diff
@@ -801,7 +807,13 @@ def handler(event, context):
                     last_modified=last_modified,
                     size=size,
                     text=text,
-                    version_id=version_id
+                    version_id=version_id,
+                    s3_tags=get_object_tagging(
```
akarve, line 811: Very minor and up to you: this is the only inline call. All other calls are made above. This also made me question certain things like "is this the only S3 operation that we don't retry?" (it's not).

akarve (dismissed):

lambdas/es/indexer/index.py
sir-sigurd: I'm not sure what we should retry on and if we should do that at all:
1. boto3 has its own [retry logic](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html)
2. some of retrying logic is for dealing with eventual consistency:
https://github.com/quiltdata/quilt/blob/2e5d4e7e1bd34fd908dd324c0749b25f2ab91bf2/lambdas/es/indexer/index.py#L828-L830
I'm not sure if that logic was really needed in the first place, but obviously it's not needed now.

drernie: @akarve Did you address his above questions in a different comment? Or just want him to use his own judgment?

nl0 (approved):
lgtm, tho indexing tags as `key1 value1 key2 value2 ...` feels like a questionable decision tbh

sir-sigurd:
> lgtm, tho indexing tags as `key1 value1 key2 value2 ...` feels like a questionable decision tbh

yes, though it was promoted by @akarve

nl0:
> > lgtm, tho indexing tags as `key1 value1 key2 value2 ...` feels like a questionable decision tbh
> 
> yes, though it was promoted by @akarve

i don't urge anyone to re-evaluate the solution atm, but we may consider revisiting that some time in the future

PR #3697, opened 2023-08-16
Download and bookmark on listing item mouseover

fiskus:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation


catalog/app/containers/Bucket/ListingActions.tsx
```diff
@@ -0,0 +1,240 @@
...
+function BucketDirectory({ className, location: { bucket, key } }: BucketButtonProps) {
+  return (
+    <FileView.ZipDownloadForm suffix={`dir/${bucket}/${key}`} className={className}>
+      <Button icon="arrow_downward" title="Download" type="submit" />
+    </FileView.ZipDownloadForm>
+  )
+}
+
+function BucketFile({ className, location }: BucketButtonProps) {
+  return AWS.Signer.withDownloadUrl(location, (url: string) => (
```
nl0, line 64: it would be more idiomatic to use the hook (`useDownloadUrl`)

catalog/app/containers/Bucket/ListingActions.tsx
```diff
@@ -0,0 +1,240 @@
...
+    flexDirection: 'row-reverse',
+  },
+  container: {
+    '.MuiDataGrid-row:hover &': {
+      animation: `$show 150ms ease-out`,
+      position: 'absolute',
+    },
+    display: 'flex',
+    background: `linear-gradient(to right, ${fade('#f5f5f5', 0)} 0, #f5f5f5 ${t.spacing(
+      4,
```
nl0, line 116: oh how i hate what prettier does in such cases :rage4:

catalog/app/containers/Bucket/Listing.tsx
```diff
@@ -1152,6 +1154,18 @@ export function Listing({
...
       })
     }
+    columnsWithValues.push({
+      field: 'actions',
+      headerName: '',
+      align: 'right',
+      width: 0,
+      renderCell: (params: DG.GridCellParams) =>
+        params.id === '..' ? (
+          <div />
```
nl0, line 1164: RowActions returns null when there's no actions to show, so i'd suggest returning null here as well

nl0 (commented):
looks ok, some nitpicks inline

catalog/app/containers/Bucket/Listing.tsx
fiskus: `null` conflicts with type annotations, I can return `<></>` in both cases

catalog/app/containers/Bucket/ListingActions.tsx
fiskus: I formatted it to multiline notation

nl0 (approved):

catalog/app/containers/Bucket/Listing.tsx
nl0: `null` feels more appropriate, but whatever

catalog/app/containers/Bucket/ListingActions.tsx
nl0: oh, that actually looks better

fiskus:
I've added Changelog and squashed commits, because otherwise `git log` becomes noisy with "merge queue"

nl0 (approved):

fiskus:
Resolved conflict (in Changelog), and fixed download for Package file: zip lambda doesn't work, and I used `physicalKey` for that (Download button in the top of the page uses `physcalKey` as well)

nl0 (approved):

catalog/app/containers/Bucket/ListingActions.tsx
```diff
@@ -210,7 +211,7 @@ function useMatchedParams(to: string) {
         }
       }
       default:
-        return {}
+        assertNever(match.path as never)
```
nl0, line 214: this statement doesn't do anything bc you virtually just telling it to assert that never explicitly casted by you is never, and this will always succeed no matter what. so you either have to remove the type cast (`as ...`) or replace it with a runtime check (assert / invariant)

nl0 (approved):

PR #3701, opened 2023-08-23
Index empty metadata as null, not as "null" or "{}" strings

sir-sigurd:


nl0 (dismissed):
looks good, some suggestions inline

PR #3702, opened 2023-08-23
Workaround to fix us-east-1 S3 signatures

dimaryaz:
In us-east-1, the S3 client sometimes uses the `s3.amazonaws.com` hostname, sometimes `s3.us-east-1.amazonaws.com`. We need to use exactly the same one in s3 proxy in order for the signature to work. S3 proxy already supports region-less hostnames - just need to pass "-" as the region.


dimaryaz:
> is this superseded by #3703 ?

Yes - though this is a quick fix, while #3703 is a bigger and riskier change. But if it looks fine, then I'll close this one.

PR #3703, opened 2023-08-23
Rewrite hostname and region logic in S3 proxy

dimaryaz:
Instead of passing the region, S3 bucket name, and S3 path just pass the original HTTP hostname and path.

It simplifies the logic quite a bit and lets us not worry about things like:
- Virtual hosts vs path-style (aka bucket names with dots problems)
- `s3.amazonaws.com` vs `s3.us-east-1.amazonaws.com`

### DIFF
--- a/catalog/app/utils/AWS/S3.js
+++ b/catalog/app/utils/AWS/S3.js
@@ -83,26 +83,23 @@ function useSmartS3() {
           )
           req.on('sign', () => {
             if (req.httpRequest[PRESIGN]) return
+
             // Monkey-patch the request object after it has been signed and save the original
             // values in case of retry.
+            const origEndpoint = req.httpRequest.endpoint
+            const origPath = req.httpRequest.path
+
             req.httpRequest[PROXIED] = {
-              endpoint: req.httpRequest.endpoint,
-              path: req.httpRequest.path,
+              endpoint: origEndpoint,
+              path: origPath,
             }
             const basePath = endpoint.path.replace(/\/$/, '')
-            // handle buckets with dots in their names
-            if (
-              req.httpRequest.path.startsWith(`/${b}`) &&
-              !req.httpRequest.endpoint.host.startsWith(`${b}.`)
-            ) {
-              req.httpRequest.path = req.httpRequest.path.replace(`/${b}`, '')
-            }
 
             req.httpRequest.endpoint = endpoint
             req.httpRequest.path =
               type === 'select'
-                ? `${basePath}${req.httpRequest.path}`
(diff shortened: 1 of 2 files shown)


catalog/app/utils/AWS/S3.js
```diff
@@ -83,26 +83,23 @@ function useSmartS3() {
...
-            ) {
-              req.httpRequest.path = req.httpRequest.path.replace(`/${b}`, '')
-            }
 
             req.httpRequest.endpoint = endpoint
             req.httpRequest.path =
               type === 'select'
-                ? `${basePath}${req.httpRequest.path}`
-                : `${basePath}/${req.httpRequest.region}/${b}${req.httpRequest.path}`
+                ? `${basePath}${origPath}`
```
dimaryaz, line 101: This likely breaks S3 select - but I can't remember where we use it, so don't know how to test it.

Also, [this hostname](https://github.com/quiltdata/quilt/blob/master/lambdas/s3select/src/t4_lambda_s3select/__init__.py#L37) can't possibly work with bucket names that contain dots, right?.. We need path-style access here.

akarve: > This likely breaks S3 select - but I can't remember where we use it

From memory we use S3 select in two places: package-select lambda, es/indexer to pull package-level metadata.

Try searching for `select_object_content` (the boto3 call).
I believe that is only dispatched from `query_manifest_content` in utils.py.

dimaryaz: This is a lambda called from JS, though. But anyways, figured it out: it's only used to get object access counts when you're not logged in. So https://open.quiltdata.com/ is pretty much the only place this comes up. Bucket names with dots are broken in that case - but we've never added such buckets to the registry of open data, so we wouldn't know.

So turns out my change is actually ok here. I tested it on the registry of open data, and it still works fine (on buckets without dots, that is). Buckets with dots are broken, but they never worked in the first place: the s3select lambda just returns a 500. I'll see if it's an easy fix - but it shouldn't block this PR.

dimaryaz: Well, s3select was not as trivial of a fix as I thought... But, turns out, we only use s3select for the _analytics bucket_ - so we really don't need to worry about dots there.

nl0 (approved):
this makes things simpler. also, s3 select lambda should be removed in near future -- i have plans to replace that call with a dedicated gql query.

dimaryaz:
> s3 select lambda should be removed in near future -- i have plans to replace that call with a dedicated gql query.

Oh nice! I was trying to fix it - but it is such an obscure use case, and annoying to test.

PR #3704, opened 2023-08-25
Hide navbar controls for guests in PRODUCT

fiskus:
image

image

### DIFF
--- a/catalog/app/containers/NavBar/NavBar.tsx
+++ b/catalog/app/containers/NavBar/NavBar.tsx
@@ -589,7 +589,7 @@ export function NavBar() {
   const classes = useNavBarStyles()
   return (
     <Container>
-      {cfg.disableNavigator || (cfg.alwaysRequiresAuth && isSignIn) ? (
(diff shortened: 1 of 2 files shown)


catalog/app/containers/NavBar/NavBar.tsx
```diff
@@ -589,7 +589,7 @@ export function NavBar() {
   const classes = useNavBarStyles()
   return (
     <Container>
-      {cfg.disableNavigator || (cfg.alwaysRequiresAuth && isSignIn) ? (
+      {cfg.disableNavigator || (cfg.alwaysRequiresAuth && !authenticated) ? (
```
nl0, line 592: what's the actual purpose of this change? i'm not quite sure it's in line with the original intent

fiskus: If authentication is required, but the user is not authenticated, then we show spacer instead of controls.
So, the purpose is the same as written with code

I interpreted the current state as a bug: we show search and bucket selector on the signup page. These controls don't work for unauthenticated users.

nl0:
incorporated into #3982

PR #3707, opened 2023-08-31
Metadata dropzone: disable click, fix JSON drop

fiskus:
Click bug introduced in https://github.com/quiltdata/quilt/pull/3647
And JSON drop bug introduced in https://github.com/quiltdata/quilt/pull/2510


catalog/app/containers/Bucket/PackageDialog/MetaInput.tsx
```diff
@@ -284,7 +284,7 @@ export const MetaInput = React.forwardRef<HTMLDivElement, MetaInputProps>(
               onChange(contents)
             } else {
               try {
-                JSON.parse(contents as string)
+                onChange(JSON.parse(contents as string))
```
fiskus, line 287: To actually fix this, I would rewrite this as two steps: first returns the result consistently, and then call `onChange` once

PR #3712, opened 2023-09-07
Faceted Search

nl0:
Screenshot 2023-10-13 at 18 02 39

# TODO


fiskus (commented):
I'd like to add global (initial) extents to each filters to highlight what values are not available. I think it could help the user to not get lost

catalog/app/containers/Search/model.tsx
```diff
@@ -0,0 +1,342 @@
...
+import NEXT_PAGE_QUERY from './gql/NextPage.generated'
+
+function ifChanged<T>(newValue: T) {
+  return (oldValue: T) => (R.equals(newValue, oldValue) ? oldValue : newValue)
+}
+
+enum ResultType {
+  Objects = 'objects',
+  Packages = 'packages',
+  Any = 'any',
```
fiskus, line 24: I'd use verbose but explicit `ObjectsAndPackages` - it's self-descriptive, and not confusing with typescript `any`

catalog/app/containers/Search/model.tsx
```diff
@@ -0,0 +1,342 @@
...
+}
+
+interface ResultTypeFacetState {
+  selected: ResultType
```
fiskus, line 28: Good old `value` is good enough. Also, it can be confusing to read `selected` not as `boolean`

catalog/app/containers/Search/model.tsx
```diff
@@ -0,0 +1,342 @@
...
+}
+
+interface BucketFacetState {
+  selected: string[]
```
fiskus, line 32: Bucket facet doesn't have extents, because we use bucket list API?

catalog/app/containers/Search/model.tsx
nl0: yeah ok

catalog/app/containers/Search/model.tsx
nl0: well, the main reason for this is we don't want to store enum extents (lists of values) in the url, since they may eat up lots of space (and are unbounded, in general). that state should live somewhere else (it's ok to just request it in facet widget components or smth)

catalog/app/containers/Search/model.tsx
nl0: it's not future proof. i'll just keep the actual options here, and we can assume `null` means don't filter by type

nl0: > it's not future proof

i mean we will probably add other types in the future

nl0:
@fiskus pls do a `git pull --rebase` if conflicts while pushing to avoid introducing unnecessary merge commits

nl0:
@fiskus ima merge this

fiskus (approved):

PR #3715, opened 2023-09-11
Update list of logos, remove unused

fiskus:
Consistent use of logos, final list is here https://github.com/quiltdata/quilt/blob/logos-carousel-update/catalog/app/website/pages/Landing/Logos/list.ts


akarve (approved):
You probably already crushed Decibel but I was able to get another 10% out of it; maybe from PNG-8 which I think you explained your asset crusher doesn't support. Just FYI.

nl0 (approved):
inari logo seems to have a lot of surrounding whitespace that could have been trimmed. is that intentional?

fiskus:
I crushed `logo-decibel.png` more using `pngquant` and quality 50.

---
`logo-inari.svg` is rendered as square with empty spaces on top and bottom, I guess, because it has wide `viewBox`. For example, this logo is 1200x1200 size and has empty spaces as well https://github.com/tylerbuchea/gatsby/blob/master/www/static/monogram.svg

nl0 (approved):

PR #3721, opened 2023-09-18
Add meta noindex tag

fiskus:


fiskus:
We don't need it since we now have robots.txt

PR #3723, opened 2023-09-18
Add a robots.txt to the catalog

dimaryaz:
Block all indexing in private catalog deployments.

CloudFront deployments will exclude /robots.txt.

### DIFF
--- a/catalog/firebase.json
+++ b/catalog/firebase.json
@@ -10,7 +10,7 @@
         "destination": "/index.html"
       }
     ],
-    "ignore": ["**/.*"],
+    "ignore": ["**/.*", "robots.txt"],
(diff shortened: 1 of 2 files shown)


catalog/firebase.json
```diff
@@ -10,7 +10,7 @@
         "destination": "/index.html"
       }
     ],
-    "ignore": ["**/.*"],
+    "ignore": ["**/.*", "robots.txt"],
```
dimaryaz, line 13: I'm guessing that this is the way to do it, but I don't know how to test it without actually deploying it?

akarve: I can easily deploy to stage and test there.

PR #3724, opened 2023-09-18
Add a robots.txt to s3-proxy

dimaryaz:
Block all indexing in the s3 proxy.

### DIFF
--- a/s3-proxy/nginx.conf
+++ b/s3-proxy/nginx.conf
@@ -69,6 +69,11 @@ http {
             add_header 'Content-Type' 'text/plain';
             return 200 'S3 Proxy';
         }
+
+        location = /robots.txt {
+            add_header 'Content-Type' 'text/plain';
+            return 200 'User-agent: *\nDisallow: /\n';
+        }
     }
 
     server {


s3-proxy/nginx.conf
```diff
@@ -69,6 +69,11 @@ http {
             add_header 'Content-Type' 'text/plain';
             return 200 'S3 Proxy';
         }
+
+        location = /robots.txt {
```
akarve, line 73: Does this work for head requests? Some bots HEAD before GET, etc.

akarve (approved):

s3-proxy/nginx.conf
dimaryaz: Yep

s3-proxy/nginx.conf
```diff
@@ -69,6 +69,11 @@ http {
...
+            add_header 'Content-Type' 'text/plain';
+            return 200 'User-agent: *\nDisallow: /\n';
```
akarve, line 75: I was wondering if we need a terminating `\n`. Most examples I saw don't have it but I doubt it matters.

dimaryaz: Pretty sure it doesn't matter - but in general, files should end with a `\n`

PR #3725, opened 2023-09-19
Enable SignUp for `mode="OPEN"` only

fiskus:


nl0 (approved):

akarve:
I think we need something more general here, such as a flag to disable the SignUp route via config.js? @dimaryaz can you comment?

SignUp orthogonal to PasswordAuth, as even SSO stacks can sign up. The condition is more like in "PRODUCT" mode and under certain backend circumstances (sign up disabled, which is the common case) disable the route.

dimaryaz (dismissed):
I think this is correct.

In the past, SSO sign up had a second step, where the user had to enter a username - but that's gone now, so the sign in page is all that's necessary.

akarve (approved):

drernie:
@sir-sigurd I'm assuming this is obsoleted by open.quilt.bio.  If not, and worth prioritizing, please re-open.

sir-sigurd:
> I'm assuming this is obsoleted by open.quilt.bio

I'm out of the context but don't see a reason why

PR #3737, opened 2023-09-28
Rebase s3-proxy Docker image from "debian" to "amazonlinux"

dimaryaz:
This fixes vulnerabilities found by the ECR scanner.

Also upgrade nginx to 1.24.0.

### DIFF
--- a/s3-proxy/nginx.conf
+++ b/s3-proxy/nginx.conf
@@ -31,7 +31,7 @@ http {
             proxy_request_buffering off;
             proxy_ssl_verify on;
             proxy_ssl_verify_depth 5;
-            proxy_ssl_trusted_certificate /etc/ssl/certs/ca-bundle.trust.crt;
+            proxy_ssl_trusted_certificate /etc/nginx/certs.pem;
 
             # Add CORS headers.
             add_header 'Access-Control-Allow-Headers' $http_access_control_request_headers always;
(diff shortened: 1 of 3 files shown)


s3-proxy/nginx.conf
```diff
@@ -31,7 +31,7 @@ http {
             proxy_request_buffering off;
             proxy_ssl_verify on;
             proxy_ssl_verify_depth 5;
-            proxy_ssl_trusted_certificate /etc/nginx/certs.pem;
+            proxy_ssl_trusted_certificate /etc/ssl/certs/ca-bundle.trust.crt;
```
dimaryaz, line 34: This one is a change that I don't fully understand, and not even sure how to test properly.

We used to `cat /etc/ssl/certs/*.pem > /etc/nginx/certs.pem` - but the latest version of `amazonlinux` doesn't even have those files anymore. Instead, it has:

```
/etc/ssl/certs/ca-bundle.crt
/etc/ssl/certs/ca-bundle.trust.crt
```

I don't understand the difference; the only thing I can find is this: https://serverfault.com/questions/620003/difference-between-ca-bundle-crt-and-ca-bundle-trust-crt It implies that we should use `ca-bundle.crt`, unless we really care about security? But maybe we do?

dimaryaz: Oh, forgot that we potentially append custom certs to /etc/nginx/certs.pem. Need to undo this.

dimaryaz: Actually, amazonlinux:2023 only has one bundle, so that makes it easy...

nl0 (commented):
seems fine, i don't see anything suspicious

sir-sigurd (approved):

PR #3792, opened 2023-11-17
"Open in desktop" button, also added button for the file view

fiskus:
https://github.com/quiltdata/quilt/assets/533229/8be2299e-3ebb-4be3-8d60-88f9cec9dcd9


fiskus:
`QuiltSync` is not production ready yet.

PR #3805, opened 2023-11-29
Use non linear range for numbers filter

fiskus:
* Use "logarithmic" scale for numbers (n → n^10) bigger than 100


catalog/app/components/Filters/NumbersRange.tsx
```diff
@@ -1,17 +1,71 @@
...
+const ExpSuffix = ({ n }: { n: number }) => (
+  <>
+    ×10<sup>{n}</sup>
+  </>
+)
+
+const createValueLabelFormat =
+  (scale: d3Scale.ScalePower<number, number>) => (number: number) =>
+    formatQuantity(Math.round(scale(number)), {
+      suffixes: [
```
nl0, line 30: default suffixes seem more suitable here, because at least they fit somewhat better into the "bubble"

nl0 (commented):
i'd say we should keep the linear scale for values less then some threshold (100?), bc the UX feels quite confusing for smaller numbers.

fiskus:
Added linear scale for number < 100. Removed custom suffixes and fixed non-linear scale marks

nl0 (approved):

nl0:
failing ci breaks auto-merge(

PR #3808, opened 2023-11-30
Scroll area for filters and search results

fiskus:
Screenshot from 2023-12-05 12-41-25

### DIFF
--- a/catalog/app/components/Filters/List.tsx
+++ b/catalog/app/components/Filters/List.tsx
@@ -9,6 +9,9 @@ import TinyTextField from './TinyTextField'
 // Number of items, when we show search text field
 const TEXT_FIELD_VISIBLE_THRESHOLD = 8
 
+// Number of items, above which we collapse
+const SHOW_MORE_THRESHOLD = 6
+
 function fuzzySearchExtents(extents: string[], searchStr: string): string[] {
   if (!searchStr) return extents
   const fuse = new Fuse(extents, { includeScore: true })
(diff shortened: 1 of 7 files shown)


nl0:
the ux feels too complicated and clumsy. i believe we can make it simpler by doing away with the custom top/bottom indicators and scrollable containers

nl0:
> Containers for keywords filter? 

nope

> It's no longer scrollable, it has "Show more" button, and show the whole list without scroll

that's good

> Top/bottom indicators are additional buttons, just to show that something is hidden at top or at bottom. You can use scroll instead of buttons. 

i don't think they are necessary and i believe they just clutter the UI.

i meant smth like this, without too much custom logic and controls, just some css positioning stuff:

Faceted search scroll

fiskus:
Rewrote layout using sticky app header

drernie:
After reviewing:

- these [PR comments](https://github.com/quiltdata/quilt/pull/3808)
- the test stack on [SearchMinimal](https://searchminimal.quiltdata.com/search?q=inari)
- [Kevin's comments](https://quiltdata.slack.com/archives/C8CCWARNK/p1702053966283079?thread_ts=1702048194.922309&cid=C8CCWARNK) in Slack
- Our [Horizon deliverables](https://searchminimal.quiltdata.com/b/quilt-leadership/packages/horizon/2024-01-18) and strategic needs

I have reluctantly concluded that @fiskus should **not** spend any more time on this PR, so he can start working on MOAB deliverables with Dima. 

My rationale is that successfully shipping MOAB in January is more important than a highly-polished faceted search this month.  @kevinemoore  has already said that the observed cosmetic issues, though serious, are not blockers. The reality is that most customers will only install a December release in their 'staging' stacks, and may well provide feedback that obsolete portions of this PR.

@nl0  I believe your options are:

1. Ignore this PR and ship the release without it
2. Approve and merge this PR even though it does not address your concerns
3. Find some other way to address them (e.g., fix them yourself)
4. Throw an Exception, if you cannot in good conscience sign off on a release with this feature set

Please let me know at Monday's D&D which way you are leaning. Thanks.

fiskus:
I suppose, we change this layout anyway when we adjust packages search result.

PR #3816, opened 2023-12-11
Specify condition for rendering Quilt manifests, allowing to render other types of files in `.quilt/packages/`

fiskus:


drernie (approved):
Thank you!

PR #3817, opened 2023-12-12
Make sure package has the expected hash when browsing and installing

dimaryaz:


api/python/quilt3/packages.py
```diff
@@ -619,6 +619,10 @@ def download_manifest(dst):
 
             pkg = cls._from_path(local_pkg_manifest)
             pkg._origin = PackageRevInfo(str(registry.base), name, top_hash)
+
+            if pkg.top_hash != top_hash:
```
akarve, line 623: OK so doing this check in `_browse` also gets `install` (in PR title)? And works for subpackage install as well? Since the user can install any subpackage and we have the manifest at that point I think we only need to check the installed object hashes and not the tophash.

sir-sigurd, line 623: that might be OK, but are you sure we should put manifest to the cache if this check fails?

sir-sigurd (approved):
well, I'm not sure we shouldn't trust hash, but it feels safer and probably performance penalty is not big

api/python/quilt3/packages.py
dimaryaz: It's safe to do that right now, since it checks the hash regardless of where the package is being loaded from - local, S3, or cache.

I could change it to only check hash when loading from S3, and before putting it into the cache. Then we'll trust any local package (not sure we should?), and anything in the cache (probably fine, except for packages that were already cached before this PR).

akarve:
@dimaryaz I think you can close this one per discussion to focus on ensuring that client and S3 hash the same bits.

drernie:
Closing old dima PRs.  Will revisit with quilt_rs

PR #3821, opened 2023-12-19
Faceted Search: show helpful message in case of search query syntax errors

nl0:
# Description

Screenshot 2023-12-19 at 14 44 36

# TODO

### DIFF
--- a/catalog/app/containers/Search/Results.tsx
+++ b/catalog/app/containers/Search/Results.tsx
@@ -112,7 +112,7 @@ const useEmptyResultsStyles = M.makeStyles((t) => ({
 interface EmptyResultsProps {
   className?: string
   clearTitle?: string
-  description?: string
+  description?: string | React.ReactNode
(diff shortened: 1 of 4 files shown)


catalog/app/containers/Search/Results.tsx
```diff
@@ -112,7 +112,7 @@ const useEmptyResultsStyles = M.makeStyles((t) => ({
 interface EmptyResultsProps {
   className?: string
   clearTitle?: string
-  description?: string
+  description?: string | React.ReactNode
```
fiskus, line 115: I'd rather use `children` prop for free-form react content. That way you can get rid of React.Fragments

fiskus (approved):

catalog/app/containers/Search/Results.tsx
nl0: yeah true, but i don't want to refactor all the instances of this component

PR #3825, opened 2024-01-09
Backward compatible S3 checksums attempt

fiskus:
when `multipartChecksums: false`:
* [x] compare existing hashes
* [ ] type and graphql works


fiskus:
Works as expected, all these changes are unnesessary or wrong

PR #3844, opened 2024-01-23
Make quilt+s3 URL clickable

fiskus:
Screenshot from 2024-01-23 13-05-20

### DIFF
--- a/catalog/app/containers/Bucket/CodeSamples/Code.tsx
+++ b/catalog/app/containers/Bucket/CodeSamples/Code.tsx
@@ -52,7 +52,13 @@ function LineOfCode({ lang, text, help }: LineOfCodeProps) {
   const classes = useLineOfCodeStyles()
   return (
     <div className={classes.root}>
-      {highlight(text, lang)}
+      {lang === 'uri' ? (
+        <StyledLink href={text} target={text.startsWith('http') ? '_blank' : '_self'}>
+          {text}
+        </StyledLink>
+      ) : (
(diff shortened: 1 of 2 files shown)


catalog/app/containers/Bucket/CodeSamples/Code.tsx
```diff
@@ -52,7 +52,11 @@ function LineOfCode({ lang, text, help }: LineOfCodeProps) {
   const classes = useLineOfCodeStyles()
   return (
     <div className={classes.root}>
-      {highlight(text, lang)}
+      {lang === 'uri' ? (
+        <StyledLink href={text}>{text}</StyledLink>
```
nl0, line 56: doesn't that need a target=blank or smth?

nl0 (approved):

catalog/app/containers/Bucket/CodeSamples/Code.tsx
fiskus: It's a URL that a browser can't open. So, with target="_blank" it will open literally blank tab :)
If opening URL on the same page, browser handles URL, asking on how to open that URL, but the page will be preserved.

fiskus: Though, thinking in general, I can add target="blank" when URL starst from 'http'

fiskus:
Added target="_blank" for cases when someone pass http(s) URLs
Rebased

nl0 (approved):

PR #3849, opened 2024-01-29
Extract reusable python code into `quilt_shared` package

nl0:
## TODO

- [ ] revisit extra dependency groups
- [ ] unit tests?
- [ ] changelog


py-shared/requirements.txt
sir-sigurd, line 1: I still don't understand why we need since we don't have tests

py-shared/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[project.optional-dependencies]
+pydantic = [
+  "pydantic ~= 1.10",
+  "typing-extensions ~= 4.9",
+]
+boto = [
+  "boto3-stubs[s3] ~= 1.34",
+  "boto3 ~= 1.34",
+  "types-aiobotocore[s3] ~= 2.11",
```
sir-sigurd: all of these seem to be needed only for type annotations
maybe move them to `dev`?

py-shared/src/quilt_shared/aws.py
```diff
@@ -0,0 +1,34 @@
...
+        return dict(
+            aws_access_key_id=self.key,
+            aws_secret_access_key=self.secret,
+            aws_session_token=self.token,
+        )
+
+    @classmethod
+    def from_boto_session(cls, session: boto3.Session):
+        credentials = session.get_credentials()
+        assert credentials
```
sir-sigurd, line 30: maybe raise `ValueError`? it's really can be `None`

py-shared/requirements.txt
nl0: 1. to examine transitive dependencies
2. to fiddle with the package in venv

tho i can remove it, no problem, bc i see its presence may be confusing

py-shared/pyproject.toml
nl0: well, yes, but that may complicate consumption of this package, since we'll need to list it in both the main section of the dependencies and then in some optional section with the `dev` extra. not sure how to make it convenient enough

py-shared/src/quilt_shared/aws.py
nl0: ok

sir-sigurd (approved):

PR #3852, opened 2024-01-30
Cut long bucket names in bucket grid

fiskus:
Before:
image

After:
Screenshot from 2024-01-30 19-41-44


nl0 (approved):
looks better, tho i would make them strictly one-line (without wrapping on dashes)

fiskus:
I think wrapping is fine, that way some long lines fit into view. And we also wrap bucket names, so it looks consistently

PR #3857, opened 2024-02-02
Deploy lambdas from CI

dimaryaz:
Build and deploy lambda zips and Docker images from CI


lambdas/upload_ecr.sh
```diff
@@ -0,0 +1,33 @@
...
+image_name=$2
+
+regions=$(aws ec2 describe-regions --query "Regions[].{Name:RegionName}" --output text)
+
+for region in $regions
+do
+    if [[ $region == "eu-south-2" ]]
+    then
+        continue
+    fi
```
dimaryaz, line 22: I'll remove this as soon as we fix the repos.

nl0:
> I think we still need the scripts in the deployment repo for manual testing (since CI will only run for the master branch), but I'll try to modify them to call these new scripts.

can't we just call the scripts manually from this repo?
i think it makes sense to keep this build/deploy toolchain here to colocate it with the actual thing (lambdas) being built/deployed

dimaryaz:
> can't we just call the scripts manually from this repo?

I guess we could. But still, what would be the right place for those scripts?

nl0:
> > can't we just call the scripts manually from this repo?
> 
> I guess we could. But still, what would be the right place for those scripts?

idk, it doesn't seem too important, but i'd prefer some subdir either under lambdas or under the repo root

dimaryaz:
Ok, moved helper scripts to `lambdas/scripts/`.

But, we'll still want user-friendly wrappers around those. E.g., the old ones in deployment do a few more things:
- Get the git hash (won't work in CI; it gives us a var instead)
- Check for uncommitted changes (unnecessary in CI)
- Set the region based on account ID (redundant cause CI already has all the info)

If we put them in this repo, then I think we need two directories, one for low-level helper scripts, and one for high-level user-friendly scripts.

lambdas/upload_ecr.sh
sir-sigurd: > as soon as we fix the repos

when we're planning to do that?

.github/workflows/deploy-lambdas.yml
sir-sigurd, line 1: let's use .yaml extension as it's recommended

.github/workflows/deploy-lambdas.yml
```diff
@@ -0,0 +1,99 @@
...
+      - '.github/workflows/deploy-lambdas.yml'
+      - 'lambdas/**'
+
+jobs:
+  deploy-lambda-s3:
+    strategy:
+      matrix:
+        path:
+          - access_counts
+          - es/indexer
```
sir-sigurd, line 17:
```suggestion
          - indexer
```

lambdas/scripts/build_zip.sh
```diff
@@ -0,0 +1,54 @@
...
+
+mkdir out
+cd out
+
+pip3 install -U pip setuptools
+
+# remove pre-installed lambda packages from requirements.txt
+preinstalled=$(pip freeze | cut -d '=' -f 1)
+preinstalled_regex=${preinstalled//$'\n'/|}
+grep -v -E -i "^(${preinstalled_regex})==" /lambda/function/requirements.txt > filtered_requirements.txt || true
```
sir-sigurd, line 23: looks outdated

sir-sigurd:
> Ok, moved helper scripts to `lambdas/scripts/`.

seems fine

> If we put them in this repo, then I think we need two directories, one for low-level helper scripts, and one for high-level user-friendly scripts.

I don't quite understand, why do we need low-level scripts? can't both CI and users use the same scripts?

.github/workflows/deploy-lambdas.yml
```diff
@@ -0,0 +1,99 @@
...
+    runs-on: ubuntu-latest
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build zip
+        run: |
+          BUILDER_IMAGE=quiltdata/lambda:build-3.8
```
sir-sigurd, line 36: would be nice if python version become variable so we could use different versions if we need

.github/workflows/deploy-lambdas.yml
```diff
@@ -0,0 +1,99 @@
...
+
+          docker pull "$BUILDER_IMAGE"
```
sir-sigurd, line 38: well, you don't have to call pull here

lambdas/scripts/upload_ecr.sh
```diff
@@ -0,0 +1,33 @@
...
+}
+
+[[ $# == 2 ]] || error "Usage: $0 account_id image_name"
+
+account_id=$1
+image_name=$2
+
+regions=$(aws ec2 describe-regions --query "Regions[].{Name:RegionName}" --output text)
+
+for region in $regions
```
sir-sigurd, line 17: maybe we should use pre-defined list of regions, so our scripts won't fail when new regions are added?

.github/workflows/deploy-lambdas.yml
dimaryaz: Ok. Does it mean I should also remove `--pull` from the `docker buildx build` command?

lambdas/scripts/build_zip.sh
dimaryaz: Why? This was intentional: the lambda runtime already contains e.g. boto3, so we don't want to try to install it ourselves.

.github/workflows/deploy-lambdas.yml
dimaryaz: TBH, I don't even know where this image comes from. And we probably should use different images, not different tags of the same image, for different Python versions.

dimaryaz:
> I don't quite understand, why do we need low-level scripts? can't both CI and users use the same scripts?

I guess it would be possible, just inefficient: the user scripts build everything and deploy to whatever account the user has selected (from `AWS_PROFILE`, etc.). But in CI, we should build it once, and deploy it to Prod and GovCloud. Running the user script twice would mean building everything twice. So at the very least, we should split the scripts into "build" and "deploy".

(The user scripts also do other things like check for uncommitted changes, but I guess that's harmless in CI.)

lambdas/scripts/upload_ecr.sh
dimaryaz: I don't actually know what's worse. If new regions are added, we probably _should_ support them immediately? CI failing is not great, but CI succeeding and then customer deployments failing is even worse - so maybe it's good to fail here so it's obvious.

lambdas/scripts/build_zip.sh
sir-sigurd: https://github.com/quiltdata/deployment/pull/1692

.github/workflows/deploy-lambdas.yml
sir-sigurd: > And we probably should use different images, not different tags of the same image, for different Python versions.

why? e.g. python images uses tags i.e. python:3.8, python:3.9 etc

but anyway nothing change here, I'll take care about that myself

sir-sigurd: > TBH, I don't even know where this image comes from

https://github.com/quiltdata/lambda

.github/workflows/deploy-lambdas.yaml
```diff
@@ -78,20 +78,18 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build Docker image
         working-directory: ./lambdas/${{ matrix.path }}
-        run: |
-          image_name=quiltdata/lambdas/${{ matrix.path }}:${{ github.sha }}
-          docker buildx build --platform=linux/amd64 -t "$image_name" -f Dockerfile ..
+        run: docker build -t "quiltdata/lambdas/${{ matrix.path }}:${{ github.sha }}" -f Dockerfile ..
```
sir-sigurd, line 81: are you sure `docker build` and `docker buildx` do exactly the same thing?

sir-sigurd: I mean I thought `buildx` is super-duper and works faster.

lambdas/scripts/upload_ecr.sh
sir-sigurd: > we probably should support them immediately?

well, at the moment "support" means skipping a region in a script 😁

sir-sigurd: failing job is OK but we need either
1. provide clear steps how to add a new region (preferred option)
2. deployments to old regions should still work

sir-sigurd:
> > I don't quite understand, why do we need low-level scripts? can't both CI and users use the same scripts?
> 
> I guess it would be possible, just inefficient: the user scripts build everything and deploy to whatever account the user has selected (from `AWS_PROFILE`, etc.). But in CI, we should build it once, and deploy it to Prod and GovCloud. Running the user script twice would mean building everything twice. So at the very least, we should split the scripts into "build" and "deploy".
> 
> (The user scripts also do other things like check for uncommitted changes, but I guess that's harmless in CI.)

OK, makes sense, I just want to avoid code duplicatation

.github/workflows/deploy-lambdas.yaml
dimaryaz: `buildx` just has extra features like cross-compiling for a different OS/arch (so useful for building Linux images on a Mac, or maybe specifically Arm64 Macs?) But looks like it also does parallel builds, so I guess that's a reason to use it here.

dimaryaz:
> OK, makes sense, I just want to avoid code duplicatation

I was planning to re-use the low-level scripts and call them from the existing user-friendly scripts, so there wouldn't be duplication - but they're currently in the deployment repo, so I can't do it in this PR. (Should probably move them to this repo, too - but I don't want to do everything all at once.)

lambdas/upload_ecr.sh
dimaryaz: I've manually added the repos and removed the hack

lambdas/scripts/build_zip.sh
dimaryaz: ah, ok, updated it

lambdas/scripts/upload_ecr.sh
dimaryaz: 2 sounds almost impossible: we either a) ignore a failed upload, let the job succeed, and hide the problem (bad), or b) not return an error immediately, allow subsequent CI steps to run (upload to GovCloud, etc.), but somehow return an error at the very end. But even if we do b), what guarantees do we have? Maybe there was an unrelated network failure in addition to the new region, but it goes unnoticed.

So anyways, I'd rather do 1, but add the instructions once we actually have a good way to configure new regions (TF, CloudFormation, whatever).

sir-sigurd: > So anyways, I'd rather do 1, but add the instructions once we actually have a good way to configure new regions (TF, CloudFormation, whatever).

🆗

sir-sigurd:
> If we put them in this repo, then I think we need two directories, one for low-level helper scripts, and one for high-level user-friendly scripts.

I think it makes sense to have high-level scripts in `lambdas/scripts/` and low-level scripts in `lambdas/scripts/helpers/`

> I was planning to re-use the low-level scripts and call them from the existing user-friendly scripts, so there wouldn't be duplication

well, for example build zip step looks like it will be duplicated

sir-sigurd (approved):

dimaryaz:
> I think it makes sense to have high-level scripts in lambdas/scripts/ and low-level scripts in lambdas/scripts/helpers/

Ok. Let me do that in a follow-up PR, though.

> well, for example build zip step looks like it will be duplicated

It's just mounted inside the docker container, but doesn't matter where it lives. In any case, I'll make it work when I move the high-level scripts.

> I don't see a reason this won't work in CI

You're right; for some reason, I thought CI didn't do a full git checkout.

PR #3860, opened 2024-02-05
JsonEditor: fix editing collections keys with `additionalProperties`

fiskus:
image


nl0 (approved):

fiskus:
Also, fixed minor issues with removing and editing:
* added remove button for items with `.additionalProperties` and `.items`
* fixed editing keys in arrays (they should be disabled, probably, but this would be more involved)

nl0 (approved):

PR #3863, opened 2024-02-06
Add `&referrer=` to the quilt+s3 URI

fiskus:
* Added `&referrer=` to the `quilt+s3://` URI as a string modification, not making at a part of URI struct
* Simplified code by
  * replacing weak assumptions (`if lang == 'uri'`) with types: used `PackageURI` type instead
  * using two components for Uri and code lines, moving parsing lines to that component
* Visually nothing changed and "Copy" button copies URI without `&referrer=`


fiskus:
Decided to store Stack in QuiltSync and not pollute Quilt URI

PR #3865, opened 2024-02-06
Merge pylint/pycodestyle and isort

dimaryaz:
No reason why pylint/pycodestyle should be one job, but isort should be separate.


sir-sigurd (commented):
well, why not split them instead?

nl0:
> well, why not split them instead?

yeah, at least in that case we'll have all the errors, not just the the first ones

drernie:
@sir-sigurd Is this still worth doing? If not, can you close?

drernie:
Not needed

PR #3866, opened 2024-02-06
Split the workflows into catalog.yml, client.yml, and lambdas.yml

dimaryaz:
It makes more sense than JS vs Python, and lets us filter by paths and skip unnecessary testing.

### DIFF
--- a/.github/workflows/catalog.yml
+++ b/.github/workflows/catalog.yml
@@ -1,4 +1,9 @@
-on: [push]
+on:
+  push:
+    paths:
+      - '.github/workflows/catalog.yml'
+      - 'catalog/**'
+      - 'shared/**'
 jobs:
   test-catalog:
     runs-on: ubuntu-latest
--- a/.github/workflows/lambdas.yml
+++ b/.github/workflows/lambdas.yml
@@ -1,0 +1,90 @@
+on:
+    push:
+      paths:
+        - '.github/workflows/lambdas.yml'
+        - 'lambdas/**'
+jobs:
+  linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.7'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip setuptools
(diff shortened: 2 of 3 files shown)


.github/workflows/catalog.yml
```diff
@@ -1,4 +1,9 @@
-on: [push]
+on:
+  push:
+    paths:
+      - '.github/workflows/catalog.yml'
+      - 'catalog/**'
+      - 'shared/**'
```
dimaryaz, line 6: @nl0: does anything besides the catalog depend on this?

.github/workflows/lambdas.yml
```diff
@@ -0,0 +1,90 @@
+on:
+    push:
+      paths:
+        - '.github/workflows/lambdas.yml'
+        - 'lambdas/**'
+jobs:
+  linter:
```
sir-sigurd, line 7: Instead of copy-paste you might want this: https://docs.github.com/en/actions/using-workflows/reusing-workflows

.github/workflows/catalog.yml
sir-sigurd, line 1: we might want to use `yaml` extension as it's recommended in https://yaml.org/faq.html

sir-sigurd, line 1: `lint-docs` job is not really related to catalog

.github/workflows/catalog.yml
nl0: no, i don't think so

sir-sigurd, line 6: I think that doesn't work well with required jobs (you can check them (you can see them in repo settings -> branches -> master)
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
> If a workflow is skipped due to branch filtering, [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

you can check registry workflow in enterprise for idea on how to implement filtering

also I suggest testing this CI stuff in a fork

drernie:
Closing old Dima PRs.

PR #3872, opened 2024-02-09
Deploy catalog from CI

dimaryaz:


.github/workflows/deploy-catalog.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - 'shared/**'
+
+jobs:
+  deploy-catalog-ecr:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: catalog
+    env:
+      NODE_OPTIONS: --max-old-space-size=4096
```
dimaryaz, line 19: No idea if this is useful - I just copied it from js-ci.yml

dimaryaz:
I'm intentionally pushing only to ECR, not S3. S3 is only used for RODA - and updating RODA is such a manual process that I don't think there's any benefit in slowing down CI for that.

.github/workflows/deploy-catalog.yaml
sir-sigurd: I think it's not

sir-sigurd (approved):

sir-sigurd:
> I'm intentionally pushing only to ECR, not S3. S3 is only used for RODA - and updating RODA is such a manual process that I don't think there's any benefit in slowing down CI for that.

well, yes
though I think we want to make it automated one day, so I'm not sure we should make it more manual
but actually do we really know that using CloudFront for it is really something useful for us?

nl0:
> actually do we really know that using CloudFront for it is really something useful for us?

@sir-sigurd i'd vote for getting rid of this discrepancy and keeping a single deployment / serving approach for the sake of simplicity. i doubt it very much serving w/ nginx would be a perf bottleneck. maybe we would save a couple $ tho, idk. @akarve wdyt?

nl0 (commented):
shouldn't it depend on the tests passing?

.github/workflows/deploy-catalog.yaml
```diff
@@ -0,0 +1,56 @@
...
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '16.11'
+          cache: 'npm'
+          cache-dependency-path: 'catalog/package-lock.json'
```
nl0, line 30: i'm not sure this is useful. afaict it's supposed to save the npm cache, which holds the packages fetched from the registry, not `node_modules`, and that means we shouldn't invalidate it on lockfile changes, bc even if several out of thousands of packages have changed, we would benefit hugely from still getting the majority of the packages from cache. it may get somewhat bloated over time, but i guess we can reset it somehow from time to time, or just ignore that, since i believe the overhead would be negligible.
(tho i'm not quite sure it would work properly without this option specified)

dimaryaz:
> shouldn't it depend on the tests passing?

Well, that's why I wanted to split everything up in #3866 - but that looks more complicated than I expected.

But in any case, if we already require tests to pass before merging into `master`, then I don't think there's much point in running tests again in `master`.

.github/workflows/deploy-catalog.yaml
dimaryaz: I don't think I actually understand this. Without this option, it won't cache anything at all, right?

Also, do we want to keep this in `js-ci.yml`? What's the difference?

nl0: > I don't think I actually understand this. Without this option, it won't cache anything at all, right?

i'm saying we should keep `cache: 'npm'`, but drop `cache-dependency-path`

.github/workflows/deploy-catalog.yaml
```diff
@@ -0,0 +1,52 @@
...
+        working-directory: catalog
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '16.11'
```
nl0, line 26:
```suggestion
          node-version: '16.11'
          cache: 'npm'
```

nl0 (approved):

.github/workflows/deploy-catalog.yaml
sir-sigurd: > even if several out of thousands of packages have changed, we would benefit hugely from still getting the majority of the packages from cache. it may get somewhat bloated over time, but i guess we can reset it somehow from time to time, or just ignore that, since i believe the overhead would be negligible.

I think we should use what's recommended unless we understand all consequences e.g. I'm not sure if we pay for the cache, if there is a way to reset cache etc.

https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data

> The action defaults to search for the dependency file (package-lock.json, npm-shrinkwrap.json or yarn.lock) in the repository root, and uses its hash as a part of the cache key. Use cache-dependency-path for cases when multiple dependency files are used, or they are located in different subdirectories.

nl0: yeah it doesn't work w/o that key, so im adding it back, sorry: #3880

.github/workflows/deploy-catalog.yaml
```diff
@@ -0,0 +1,53 @@
...
+          cache: 'npm'
```
dimaryaz, line 27: But does this do anything by itself, without the `cache-dependency-path`? From the docs, it sounds like it will only look for `package-lock.json` in the root directory, and not in `catalog/package-lock.json`.

nl0: > But does this do anything by itself, without the cache-dependency-path? 

i assumed that it would just never invalidate the cache, but that turned out wrong. i fixed it by bringing back `catalog/package-lock.json`

akarve:
> > actually do we really know that using CloudFront for it is really something useful for us?
> 
> @sir-sigurd i'd vote for getting rid of this discrepancy and keeping a single deployment / serving approach for the sake of simplicity. i doubt it very much serving w/ nginx would be a perf bottleneck. maybe we would save a couple $ tho, idk. @akarve wdyt?

I see this is merged but I support at least trying to remove the CloudFront special casing for open b/c I agree it's complex. The CDN probably does help SEO scores (page load times) but how much I don't know. Worth measuring the perf difference, as I don't think we ever did. And even if nginx ends up slower maybe we can just scale up those container services.

PR #3877, opened 2024-02-12
quilt3: simplify write_yaml() a bit

sir-sigurd:


dimaryaz (commented):
An even better solution would be to write to a temporary file, then move it to the real config. Handling errors and hoping to undo the config change will never be as reliable.

(For the temporary file, I'd use something in the same dir, not the normal temp dir, so it's guaranteed to be on the same filesystem - something like [this](https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/packages.py#L607-L611))

sir-sigurd:
> An even better solution would be to write to a temporary file, then move it to the real config. Handling errors and hoping to undo the config change will never be as reliable.
> 
> (For the temporary file, I'd use something in the same dir, not the normal temp dir, so it's guaranteed to be on the same filesystem - something like [this](https://github.com/quiltdata/quilt/blob/master/api/python/quilt3/packages.py#L607-L611))

yes, I'm planning to use temporary file but in the following PR

nl0 (approved):

PR #3883, opened 2024-02-14
Add .python-version files for lambdas

sir-sigurd:
This is mostly to allow lambdas use different python versions.


.github/workflows/py-ci.yml
```diff
@@ -166,11 +166,8 @@ jobs:
           - shared
           - status_reports
           - tabular_preview
+          - thumbnail
```
dimaryaz, line 169: Huh... we just forgot about it?

dimaryaz (approved):

.github/workflows/py-ci.yml
sir-sigurd: no, it's moved from lines below

nl0 (approved):

PR #3884, opened 2024-02-15
WIP: speedup promote in pkgpush

sir-sigurd:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation


lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -147,6 +151,66 @@ def calculate_pkg_hashes(pkg: quilt3.Package):
             f.result()
 
 
+# TODO: implement
```
nl0, line 154: isn't it implemented? 🤔

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -20,13 +20,9 @@
...
-from quilt_shared.pkgpush import (
-    ChecksumResult,
-    ChecksumType,
-    MPURef,
-    S3ObjectDestination,
-    S3ObjectSource,
-)
+from quilt_shared.pkgpush import ChecksumResult, ChecksumType, CopyResult
+from quilt_shared.pkgpush import MPURef as MPURefBase
+from quilt_shared.pkgpush import S3ObjectDestination, S3ObjectSource
```
nl0, line 25: isort? 🤔

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -222,47 +218,110 @@ async def upload_part(mpu: MPURef, src: S3ObjectSource, etag: str, part: PartDef
...
+
+class MPURef(MPURefBase):
+    _completed: bool = pydantic.PrivateAttr(default=False)
+
+    @property
+    def completed(self):
+        return self._completed
+
+    async def complete(self, parts: T.Sequence[PartUploadResult]):
+        if self.completed:
```
nl0, line 285: seems like this should be a promise-like thing, or at least a three-state thing (true/false/pending), otherwise we risk running into race conditions

py-shared/src/quilt_shared/pkgpush.py
```diff
@@ -46,6 +46,13 @@ class S3ObjectDestination(pydantic.BaseModel):
     bucket: str
     key: str
 
+    @classmethod
+    def from_pk(cls, pk: PhysicalKey):
+        # XXX: do we need this?
```
nl0, line 51: 🤷 we can just ignore it

nl0 (commented):
looks reasonable

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: or maybe even a four-state thing, i.e. init / pending / completed / aborted 🤔

sir-sigurd: > seems like this should be a promise-like thing, or at least a three-state thing (true/false/pending), otherwise we risk running into race conditions

can that really happen with asyncio?

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
sir-sigurd: isort

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: well, seems like in your design there's no places where a race can occur

py-shared/src/quilt_shared/pkgpush.py
nl0: well, technically yes, so this check is quite reasonable, tho maybe not very practical. i don't have a strong preference for keeping vs removing this, so use your judgement

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -148,6 +162,57 @@ def calculate_pkg_hashes(pkg: quilt3.Package):
...
+    version_id = invoke_copy_lambda(credentials, src, dst)
+    return idx, PhysicalKey(bucket=dst.bucket, path=dst.path, version_id=version_id)
+
+
+def copy_file_list(
+    file_list: T.List[T.Tuple[PhysicalKey, PhysicalKey, int]],
+    message=None,
+    callback=None,
+) -> T.List[PhysicalKey]:
+    # TODO: Copy single part files directly, because using lambda for that just adds overhead,
```
sir-sigurd, line 193: @nl0 
I don't want to use `XXX` because I find it quite unclear. By [wikipedia](https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags) its means
> warn other programmers of problematic or misguiding code

I agree that `TODO` is not quite clear in terms whether it should be done in scope of this PR or in future. (or it is if compare with `FIXME`)?

So maybe I should start using `FUTURE` for improvements to be done in bright future?

nl0 (approved):

nl0 (approved):

PR #3887, opened 2024-02-21
Catalog: adjust checksums-related constants and logic to adhere to the spec

nl0:
## Description

- Consistently refer to the new checksums as "modern"
- Adhere to the spec (use `sha2-256-chunked` type/name and always double-hash)

## TODO

- [x] Unit tests
- [x] ?? Documentation

### DIFF
--- a/catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
+++ b/catalog/app/containers/Bucket/PackageDialog/PackageDialog.tsx
@@ -34,8 +34,8 @@ import PACKAGE_EXISTS_QUERY from './gql/PackageExists.generated'
 export const MAX_UPLOAD_SIZE = 20 * 1000 * 1000 * 1000 // 20GB
 // XXX: keep in sync w/ the backend
 // NOTE: these limits are lower than the actual "hard" limits on the backend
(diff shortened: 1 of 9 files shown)


sir-sigurd:
scope is unclear

dimaryaz:
I don't know if using the word "modern" is a good idea, especially in user-visible places like docs. What happens if we change the format again - will we call it "post-modern"?

nl0:
> I don't know if using the word "modern" is a good idea, especially in user-visible places like docs.

well, yeah, maybe. any ideas? "multipart" seems somewhat imprecise, since it only refers to one particular case

> will we call it "post-modern"?

i like that 😄

nl0:
smth like "adaptive" maybe? or just "multi" referring to multiformats

fiskus:
Maybe, if it's default format, and legacy has its `legacy` prefix, then no additional name is good? Just checksums

dimaryaz (dismissed):
Looks good - just remember to undo the CI changes

sir-sigurd (dismissed):
I didn't dig into tests, otherwise LGTM.

sir-sigurd (approved):

PR #3888, opened 2024-02-21
shared: Tweak checksum types

nl0:
## Description

Use the recently-agreed-upon `sha2-256-chunked` checksum type

## TODO

- [x] Changelog

### DIFF
--- a/py-shared/src/quilt_shared/pkgpush.py
+++ b/py-shared/src/quilt_shared/pkgpush.py
@@ -72,8 +72,8 @@ class S3CopyLambdaParams(pydantic.BaseModel):
 
 
 class ChecksumType(str, enum.Enum):
-    MODERN = "sha2-256-chunked"
-    LEGACY = "SHA256"
+    SHA256 = "SHA256"  # legacy
+    SHA256_CHUNKED = "sha2-256-chunked"
 
 
 class Checksum(pydantic.BaseModel):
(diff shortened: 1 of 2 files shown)


py-shared/src/quilt_shared/pkgpush.py
```diff
@@ -72,8 +72,8 @@ class S3CopyLambdaParams(pydantic.BaseModel):
 
 
 class ChecksumType(str, enum.Enum):
-    MP = "QuiltMultipartSHA256"
-    SP = "SHA256"
+    MODERN = "sha2-256-chunked"
+    LEGACY = "SHA256"
```
sir-sigurd, line 76: I'd say we should use `SHA256` and `SHA256_CHUNKED`

nl0: i'm not opposed to that in principle, but you probably should have said that earlier, since we've already painted all our bike sheds this color

sir-sigurd: well, I've said that I don't like "modern" at the very beginning and we agreed that chunked is good especially if we'll decide to always double hash

nl0: do you propose to use these ids on the python level? e.g.
```suggestion
    SHA256_CHUNKED = "sha2-256-chunked"
    SHA256 = "SHA256"
```

nl0: or for both enum variants and the actual strings?

sir-sigurd: > do you propose to use these ids on the python level? e.g.

yes

sir-sigurd (approved):

PR #3889, opened 2024-02-21
Lambdas: chunked checksums, adhere to the spec

nl0:
# Description

Use new checksum algo and type

# TODO

- [x] Unit tests

### DIFF
--- a/lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
+++ b/lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
@@ -138,12 +138,13 @@ def get_compliant_checksum(attrs: GetObjectAttributesOutputTypeDef) -> T.Optiona
             checksum_value = object_parts["Parts"][0]["ChecksumSHA256"]
 
         checksum_bytes = base64.b64decode(checksum_value)
-        if not MODERN_CHECKSUMS:
-            return Checksum.legacy(checksum_bytes)
 
-        # double-hash
-        checksum_bytes = hashlib.sha256(checksum_bytes).digest()
-        return Checksum.modern(checksum_bytes)
+        return (
+            # double-hash
+            Checksum.modern(hashlib.sha256(checksum_bytes).digest())
+            if MODERN_CHECKSUMS
+            else Checksum.legacy(checksum_bytes)
+        )
 
     if object_parts is None:
         return None
--- a/lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
+++ b/lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
@@ -128,9 +128,6 @@ async def get_obj_attributes(location: S3ObjectSource) -> T.Optional[GetObjectAt
 
 
 def get_compliant_checksum(attrs: GetObjectAttributesOutputTypeDef) -> T.Optional[Checksum]:
-    if attrs["ObjectSize"] == 0:
-        return Checksum.empty()
-
     checksum_value = attrs.get("Checksum", {}).get("ChecksumSHA256")
     if checksum_value is None:
         return None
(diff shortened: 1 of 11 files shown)


sir-sigurd:
haven't we agree on using "chunked" vs "modern"?

nl0:
> haven't we agree on using "chunked" vs "modern"?

well, kinda sorta. we use the `chunked` in the hash type string

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -306,12 +311,15 @@ async def compute_checksum_legacy(location: S3ObjectSource) -> Checksum:
...
-    return Checksum.singlepart(hashobj.digest())
+    return Checksum.legacy(hashobj.digest())
 
 
 async def compute_checksum(location: S3ObjectSource) -> ChecksumResult:
     obj_attrs = await get_obj_attributes(location)
     if obj_attrs:
+        if obj_attrs["ObjectSize"] == 0:
+            return ChecksumResult(checksum=Checksum.empty())
+
```
sir-sigurd, line 322: I think you don't need this
you could just unindent the check below

nl0: i don't quite get what you mean

sir-sigurd: you could check object size with a single condition regardless whether you have it from get_object_attributes() or head_object()

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -142,7 +143,14 @@ def get_compliant_checksum(attrs: GetObjectAttributesOutputTypeDef) -> T.Optiona
             assert "ChecksumSHA256" in object_parts["Parts"][0]
             checksum_value = object_parts["Parts"][0]["ChecksumSHA256"]
 
-        return Checksum.singlepart(base64.b64decode(checksum_value))
+        checksum_bytes = base64.b64decode(checksum_value)
+
+        return (
+            # double-hash
+            Checksum.sha256_chunked(hashlib.sha256(checksum_bytes).digest())
```
sir-sigurd, line 146: maybe?
```suggestion
            Checksum.for_parts([checksum_bytes])
```

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
nl0: sure, but i need to do that before calling `get_compliant_checksum`, bc that may return a wrong checksum (double-hashed)

lambdas/s3hash/tests/test_compute_checksum.py
```diff
@@ -76,19 +76,55 @@ async def _get_s3():
...
+        },
+        EXPECTED_GETATTR_PARAMS,
+    )
+
+    res = await s3hash.compute_checksum(LOC)
+
+    assert res == s3hash.ChecksumResult(checksum=s3hash.Checksum.empty())
+
+
+async def test_empty_no_access(s3_stub: Stubber):
```
sir-sigurd, line 112: IMO we should test that both with enabled/disabled chunked checksums and not rely on `Checksum.empty()` for expected value

lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
sir-sigurd: well, then it might make sense to have that check in `get_compliant_checksum()` or add assert 
if you go we assert you can do here something like

1. object_size, etag = ...
2. if object_size == 0: ...
3. if obj_attrs is not None: ...

sir-sigurd (approved):

PR #3895, opened 2024-02-22
Add copy_file_list_fn parameter for Package._build()

sir-sigurd:
Needed for https://github.com/quiltdata/quilt/pull/3884

### DIFF
--- a/api/python/quilt3/packages.py
+++ b/api/python/quilt3/packages.py
@@ -12,6 +12,7 @@ import sys
 import tempfile
 import textwrap
 import time
+import typing as T
 import uuid
 import warnings
 from collections import deque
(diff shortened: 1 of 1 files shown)


api/python/quilt3/packages.py
```diff
@@ -70,6 +71,16 @@
 )
 
 
+class CopyFileListFn(T.Protocol):
```
nl0, line 74: @sir-sigurd what's the point of using this instead of `Callable[args, return]`?

nl0 (approved):

api/python/quilt3/packages.py
sir-sigurd: AFAICT it doesn't allow to specify parameters with defaults.

PR #3903, opened 2024-02-28
s3hash: always stream bytes in legacy mode

nl0:
to prevent cross-region vpc endpoint issue

### DIFF
--- a/lambdas/s3hash/CHANGELOG.md
+++ b/lambdas/s3hash/CHANGELOG.md
@@ -16,6 +16,7 @@ where verb is one of
 
 ## Changes
 
+- [Changed] Always stream bytes in legacy mode ([#3903](https://github.com/quiltdata/quilt/pull/3903))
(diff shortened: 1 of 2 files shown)


drernie:
Approved. However, to validate this I would ask for:

- two test stacks (one with and without checksums)
- SSO login to both (so we can all verify)
- the s3://quilt-bake bucket to be added to both

drernie:
Can @dimaryaz verify if Sergey is past EOD?

nl0:
>     * two test stacks (one with and without checksums)
>
>     * SSO login for both (so we can all verify)
> 
>     * the s3://quilt-bake bucket (or equivalent) to be added to both (so we can check fallback behavior)

i think that's too much for a trivial change like that.
i'd just deploy it to the current legacy checksums testing stack and confirm it works there

dimaryaz (approved):

dimaryaz (approved):

PR #3905, opened 2024-03-02
catalog push limits (to review)

drernie:
Are the limits  on `number of files` still accurate?

```markdown
    Maximum number of files per push (soft limit)              | 1,000
    Maximum number of files per push (hard limit)              | 5,000
```

Do NOT spend more than 15 minutes on this, but if you have a quick answer please let us know.

This matters [because](https://quiltdata.slack.com/archives/C03RRFEAVLY/p1709319449267729?thread_ts=1709319337.460759&cid=C03RRFEAVLY):

> Not only the total file size but the number of files (catalog hard limit of 5K) is a usage-stopping issue for Customers and causes them to do weird things like zipping up files before syncing to S3 so they are within Quilt file size limits. That's an example where Quilt is hurting data liquidity.


nl0:
> Are the limits on number of files still accurate?

yes they are

PR #3924, opened 2024-04-03
Omics fixes

drernie:
1. Use 'list_objects' if cannot call 'list_object_versions'
2. Do NOT fail if the user cannot read the workflow config folder (causes many errors)
3. Add integration test to check Omics access (will need to be removed, as it requires sales credentials)


drernie:
1. Can we add some sort of standard option to NOT always have to read the workflow config? This drove me crazy at the NextFlow summit.
2. The integration test is horrible, in that it requires the sales profile to test against Omics.  Should I just delete it, or is there a saner way to test the "versions not allowed" use case?

api/python/quilt3/data_transfer.py
```diff
@@ -741,15 +741,25 @@ def list_object_versions(bucket, prefix, recursive=True):
     s3_client = S3ClientProvider().find_correct_client(S3Api.LIST_OBJECT_VERSIONS, bucket, list_obj_params)
     paginator = s3_client.get_paginator('list_object_versions')
 
-    for response in paginator.paginate(**list_obj_params):
-        versions += response.get('Versions', [])
-        delete_markers += response.get('DeleteMarkers', [])
-        prefixes += response.get('CommonPrefixes', [])
+    try:
```
sir-sigurd, line 744: I'm not sure that's the right place to have this logic, might be a better idea to have it `Package.set_dir()`

api/python/tests/integration/test_access.py
sir-sigurd, line 1: I see no reason to not having unit tests instead

api/python/quilt3/workflows/__init__.py
```diff
@@ -294,11 +294,17 @@ def validate(*, registry: PackageRegistry, workflow, name, pkg, message):
...
-    workflow_config = registry.get_workflow_config()
-    if workflow_config is None:
-        if workflow is ...:
+    try:
+        workflow_config = registry.get_workflow_config()
+        if workflow_config is None:
+            if workflow is ...:
+                return
+            raise util.QuiltException(f'{workflow!r} workflow is specified, but no workflows config exist.')
+    except Exception as e:
```
sir-sigurd, line 303: the exception is too broad and logic is not clear
`registry.get_workflow_config()` already returns `None` if workflow config doesn't exist
IMO we should fail in case (and other cases?) default workflow is to be used but workflow config can't be read because of lacking permissions or network error

sir-sigurd (commented):
this PR needs to be split at least into two parts: list objects logic change and workflow logic changes

sir-sigurd:
@drernie 
why don't you use master branch as base of your PR?

api/python/quilt3/workflows/__init__.py
drernie: Reverting, not needed.

api/python/tests/integration/test_access.py
drernie: I have no idea how to test permission errors.

api/python/quilt3/data_transfer.py
```diff
@@ -741,15 +741,25 @@ def list_object_versions(bucket, prefix, recursive=True):
...
+        for response in paginator.paginate(**list_obj_params):
+            versions += response.get("Versions", [])
+            delete_markers += response.get("DeleteMarkers", [])
+            prefixes += response.get("CommonPrefixes", [])
 
+        if recursive:
+            return versions, delete_markers
+        else:
+            return prefixes, versions, delete_markers
+    except Exception as e:
```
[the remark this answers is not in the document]
drernie: Using ClientError instead.

api/python/tests/integration/test_access.py
sir-sigurd: https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html#botocore.stub.Stubber.add_client_error

you could find some example of its usage in the existing tests

api/python/quilt3/data_transfer.py
drernie: Fair point. Moving.

api/python/quilt3/packages.py
```diff
@@ -910,11 +911,19 @@ def set_dir(self, lkey, path=None, meta=None, update_policy="incoming"):
             src_path = src.path
             if src.basename() != '':
                 src_path += '/'
-            objects, _ = list_object_versions(src.bucket, src_path)
+            try:
+                objects, _ = list_object_versions(src.bucket, src_path)
+            except botocore.exceptions.ClientError as e:
```
sir-sigurd, line 916: it's still too broad
```suggestion
            except botocore.exceptions.ClientError as e:
                if e.response["Error"]["Code"] != "AccessDenied":
                    raise
```

drernie: Fixed in new PR

drernie:
Let me just start over, rather than risk messing up git with my lack of skill:
https://github.com/quiltdata/quilt/pull/3927

PR #3927, opened 2024-04-03
Fix #3931: Add `unversioned` flag to `Package.set()` and `Package.set_dir()`

### LINKED ISSUE #3931: Support omics / unversionable buckets
1. The default (current) behavior is that `set_dir` fails if called on a bucket that lacks versioning permissions
2. Proposed change: If passed the flag, 'allow_unversioned_bucket', it should instead call `list_objects`

drernie:
Rewrite of #3924


drernie:
First stab at trying to mock S3 calls.  Any advice appreciated...

sir-sigurd:
@drernie 
WDYT about having explicit flag to use non-versioned physical keys instead of fallback?

I think it would useful in these cases:
1. `ListBucketVersions` is not granted *by accident*
2. you have `ListBucketVersions` permission, but don't have `GetObjectVersion` permission, in that case physical keys with version IDs are useless for you

api/python/tests/integration/test_packages.py
```diff
@@ -631,6 +642,23 @@ def test_set_dir_wrong_update_policy(self):
...
+        """Verify that set_dir uses list_objects if list_object_versions fails."""
+        # uses setup_s3_stubber_cannot_list_object_versions
+        registry = 's3://test-bucket'
+        pkg_registry = self.S3PackageRegistryDefault(PhysicalKey.from_url(registry))
+        pkg_name = 'Quilt/test'
+        pkg = Package()
+
+        with patch('quilt3.packages.list_object_versions', side_effect=Exception):
+            pointers = {'latest': 'latest'}
+            self.setup_s3_stubber_list_pkg_pointers(pkg_registry, pkg_name, pointers=pointers)
```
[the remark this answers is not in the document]
drernie: Rewrote

api/python/tests/integration/test_packages.py
```diff
@@ -631,6 +642,23 @@ def test_set_dir_wrong_update_policy(self):
...
 
+    def test_set_dir_cannot_list_object_versions(self):
+        """Verify that set_dir uses list_objects if list_object_versions fails."""
+        # uses setup_s3_stubber_cannot_list_object_versions
+        registry = 's3://test-bucket'
+        pkg_registry = self.S3PackageRegistryDefault(PhysicalKey.from_url(registry))
+        pkg_name = 'Quilt/test'
+        pkg = Package()
+
+        with patch('quilt3.packages.list_object_versions', side_effect=Exception):
```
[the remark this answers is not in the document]
drernie: Yes, thanks.

drernie:
Okay, this seems a minimal PR that addresses the most urgent use case for Omics.
Does this work for you @sir-sigurd ?

Also: is there any way we could put out a 5.4.1 pre-release next Monday, so we can announce "Omics support" at my Tuesday session (right before the AWS Omics talk)?

sir-sigurd:
> Okay, this seems a minimal PR that addresses the most urgent use case for Omics. Does this work for you @sir-sigurd ?

my idea was that we do **only** `list_objects` or **only** `list_object_versions` depending on the value of the new flag
with current code if you have `ListBucketVersions` you get versioned physical keys which are useless if you don't have `GetObjectVersion`

IMO we should also fix `Package.set()` in this PR as well because it's quite related (it should fix `Package.set()` with Omics)

> Also: is there any way we could put out a 5.4.1 pre-release next Monday, so we can announce "Omics support" at my Tuesday session (right before the AWS Omics talk)?

I think yes, though it should be 5.5.0 because it's a new feature

drernie:
Ah.  So should the flag actually be "unversioned=True", so that it explicitly forces the code to ignore all versioning?

api/python/quilt3/packages.py
```diff
@@ -910,11 +915,20 @@ def set_dir(self, lkey, path=None, meta=None, update_policy="incoming"):
...
+                if not allow_unversioned or e.response["Error"]["Code"] != "AccessDenied":
+                    raise e
+                print(f"Warning[s3://{src.bucket}/{src_path}]: {e}.")
+                objects = list_objects(src.bucket, src_path, recursive=True)
+                for obj in objects:
+                    obj["IsLatest"] = True
+
             for obj in objects:
                 if not obj['IsLatest']:
                     continue
```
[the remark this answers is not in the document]
drernie: I don't understand.   I am not filtering on the objects WITH VersionId.  I am *adding* a pseudo-version to the objects *without* VersionId, so that I don't have to change the downstream code.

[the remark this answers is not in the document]
drernie: Ah.  So you are changing the downstream code.  VS Code mangled the suggestion, but I'll try to implement that.

drernie:
@sir-sigurd  Are we good? Should I bump the version to 5.5.0a1, or will you do that when you push the pre-release?

api/python/quilt3/packages.py
```diff
@@ -1162,9 +1175,18 @@ def set(self, logical_key, entry=None, meta=None, serialization_location=None, s
...
 
-    def _set(self, logical_key, entry=None, meta=None, serialization_location=None, serialization_format_opts=None):
+    def _set(
+        self,
+        logical_key,
+        entry=None,
+        meta=None,
+        serialization_location=None,
+        serialization_format_opts=None,
+        unversioned: bool = False
```
sir-sigurd, line 1188:
```suggestion
        unversioned: bool = False,
```

api/python/quilt3/packages.py
```diff
@@ -910,11 +912,13 @@ def set_dir(self, lkey, path=None, meta=None, update_policy="incoming"):
...
+            if not unversioned:
+                objects, _ = list_object_versions(src.bucket, src_path)
+                objects = filter(lambda obj: obj["IsLatest"], objects)
+            else:
+                objects = list_objects(src.bucket, src_path, recursive=True)
+
             for obj in objects:
-                if not obj['IsLatest']:
-                    continue
-                # Skip S3 pseduo directory files and Keys that end in /
```
sir-sigurd, line 917: not sure why you removed this comment

docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Entries inside each section should be ordered by type:
 
 # unreleased - YYYY-MM-DD
 ## Python API
+* [Added] New 'unversioned' parameter to `set_dir` for use with HealthOmics and other S3 URIs that lack permissions for `list_object_versions` ([#3927])
```
sir-sigurd, line 20:
```suggestion
* [Added] New 'unversioned' parameter to `Package.set_dir()` and `Package.set()` for use with HealthOmics and other S3 URIs that lack permissions for `ListBucketVersions` and/or `GetObjectVersion` ([#3927](https://github.com/quiltdata/quilt/pull/3927))
```
I'm not sure that `S3 URIs that lack permissions` is semantically correct

sir-sigurd:
> @sir-sigurd Are we good? Should I bump the version to 5.5.0a1, or will you do that when you push the pre-release?

we need a branch without breaking changes to cut 5.x release

drernie:
Recommended: 
- Review with Aneesh
- Release at 6.0.0b1
-

docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Entries inside each section should be ordered by type:
 
 # unreleased - YYYY-MM-DD
 ## Python API
+* [Added] New 'unversioned' parameter to `Package.set_dir()` and `Package.set()` for use with HealthOmics and other S3 URIs that lack permissions for `ListBucketVersions` and/or `GetObjectVersion` ([#3927](https://github.com/quiltdata/quilt/pull/3927))
```
sir-sigurd, line 20: I'm not sure that S3 URIs that lack permissions is semantically correct

drernie: Any suggestions? The term "buckets that lack sufficient permissions" feels even less accurate, since Omics has S3 URIs but no actual "buckets" :-(

akarve, line 20:
```suggestion
* [Added] New 'unversioned' parameter to `Package.set_dir()` and `Package.set()` for use with S3 URIs, such as HealthOmics, that do not support `ListBucketVersions` and/or `GetObjectVersion` ([#3927](https://github.com/quiltdata/quilt/pull/3927))
```

akarve:
> And should the default be None, so unversioned=False throws an error for unversioned buckets? Is there even an easy way to test for that?

Should the default for _what_ be `None`?

There is a [GetBucketVersioning ](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketVersioning.html) call possible but not everyone will have permissions to it. There are in fact three possible states for the _bucket_ versioning: Suspended, Enabled, <no status>. The latter means bucket versioning has never been on.

As for Sergey's comment from memory (but I didn't try this recently) I believe that if you turn versioning on, or on then off, objects present before or created after versioning was off indeed have a version id of `null`. I have very bad Internet where I am. Let me read the PR a second time and see if I can be more helpful.

akarve:
Although we _should_ in a separate PR distinguish between versioned and unversioned buckets in various ways throughout the product, we currently do not and have not, therefore this changeset should not have any conditionals to warn about versioned or unversioned state for reasons of consistency and backwards compatibility.

drernie:
> this changeset should not have any conditionals to warn about versioned or unversioned state

Okay, so it sounds like you are okay with this PR as it is, right? Can @sir-sigurd move forward with merging this and cutting a pre-release?

akarve:
@sir-sigurd this is ready for pre-release

drernie:
> Although we _should_ in a separate PR distinguish between versioned and unversioned

Created a new issue to track that. Please add additional comments there.

docs/CHANGELOG.md
sir-sigurd: I mean that it's user lacks permissions to "S3 URIs"

sir-sigurd (approved):

PR #3950, opened 2024-04-17
Move Listing's pagination to the bottom

fiskus:
* Moved pagination to the bottom
* Added message how many rows are rendered
* Slightly increased pagination controls size
* Added rows fillers on the last page, so layout doesn't jump

[simplescreenrecorder-2024-04-17_18.18.01.webm](https://github.com/quiltdata/quilt/assets/533229/9b33058c-a7b2-4ee6-b453-8f62c02e790e)

### DIFF
--- a/catalog/app/containers/Bucket/Listing.tsx
+++ b/catalog/app/containers/Bucket/Listing.tsx
@@ -284,14 +284,11 @@ const usePaginationStyles = M.makeStyles((t) => ({
   select: {
     alignItems: 'center',
     display: 'flex',
-    height: 24,
-    paddingBottom: 0,
     paddingLeft: t.spacing(0.5),
-    paddingTop: 0,
   },
   input: {
     fontSize: 'inherit',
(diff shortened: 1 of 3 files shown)


catalog/app/containers/Bucket/Listing.tsx
```diff
@@ -572,7 +580,9 @@ function Toolbar({
           )}
         </div>
       )}
-      <Pagination truncated={truncated} loadMore={loadMore} />
+      <div className={classes.summary}>
+        You see {from}—{to} out of {items.length}
```
nl0, line 584:
```suggestion
        Showing {from}—{to} out of {items.length}
```

nl0 (approved):
i'm not quite sure about this ux tbh, but whatever

fiskus:
Workaround for layout jump doesn't always work, so I removed it

nl0 (approved):

PR #3961, opened 2024-04-22
Make containers readonlyrootFileSystem=true compatible

sir-sigurd:
## Description
- s3-proxy: adjust some paths to be aligned with other nginx containers
- write runtime generated configs to /tmp


catalog/Dockerfile
```diff
@@ -16,18 +16,23 @@ RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
...
 # Copy pre-built catalog assets to nginx
-RUN rm -rf /usr/share/nginx/html
-COPY build /usr/share/nginx/html
+RUN rm -rf $NGINX_STATIC_DIR
+COPY build $NGINX_STATIC_DIR
 
 # Copy config file
 COPY config.json.tmpl config.json.tmpl
 
+# XXX: use subdir in /tmp?
```
nl0, line 27: what for? do you foresee any problems here?

nl0 (approved):
lgtm

catalog/Dockerfile
sir-sigurd: it kinda feels safer, but I don't think it's really needed

sir-sigurd, line 27:
```suggestion
```

nl0 (approved):

PR #3962, opened 2024-04-23
Robust handling of PFS cookies

nl0:
## Description

Lazily (re)set PFS cookie when required

## TODO


catalog/app/utils/PFSCookieManager.tsx
```diff
@@ -0,0 +1,114 @@
...
+  if (state.expires.getTime() + params.histeresis < params.expires.getTime()) return true
+  return false
+}
+
+export function PFSCookieManager({ children }: React.PropsWithChildren<{}>) {
+  const stateRef = React.useRef<State>()
+
+  const req = useApi()
+
+  const setPFSCookie = React.useCallback(
```
fiskus, line 45: Maybe, `requestPFSCookie`, so not to confuse with `React.setState` and highlight that this function is not cheap

catalog/app/utils/PFSCookieManager.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const selectToken = createSelector(
+  tokensSelector,
+  (tokens) => tokens?.token as string | undefined,
+)
+
+interface State {
+  token: string | undefined
+  expires: Date
+  promise: Promise<void>
```
fiskus, line 23: Would be nice to have some clue what is it for. For example,

```ts
State {
 ...
 settingCookie: Promise<void>
}
```

fiskus (commented):
For some reason, it still doesn't work for me
Screenshot from 2024-04-25 15-53-52

nl0:
> For some reason, it still doesn't work for me

1. you need an updated registry for that to work (PR in review)
2. it won't work on localhost bc the cookie has strict policy -- it only works for the same (sub?)domain

fiskus (dismissed):

catalog/app/utils/PFSCookieManager.tsx
fiskus: I think, you can make this callback an ordinary function (more likely two functions), but probably it's not worth it, and not much time left

fiskus (approved):

catalog/app/utils/PFSCookieManager.tsx
nl0: if i split them, nothing except the actual request call will remain in the "ordinary" function

fiskus (approved):

PR #3978, opened 2024-05-16
Double the expiration time of URL signing

fiskus:
5 minutes is too small amount of time for downloading some IGV tracks assets.

### DIFF
--- a/catalog/app/utils/AWS/Signer.js
+++ b/catalog/app/utils/AWS/Signer.js
@@ -10,7 +10,7 @@ import { handleToHttpsUri } from 'utils/s3paths'
 import * as Credentials from './Credentials'
 import * as S3 from './S3'
 
-const DEFAULT_URL_EXPIRATION = 5 * 60 // in seconds
+const DEFAULT_URL_EXPIRATION = 10 * 60 // in seconds
 const POLL_INTERVAL = 10 // in seconds
 const LAG = POLL_INTERVAL * 3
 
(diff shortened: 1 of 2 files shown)


catalog/app/components/Preview/loaders/useSignObjectUrls.ts
```diff
@@ -65,13 +65,15 @@ const createObjectUrlsSigner =
     return traverseUrls((idx: number): string => results[idx], jsonWithPlaceholders)
   }
 
+// Increased time to download linked data assets
+const URL_EXPIRATION = 10 * 60
```
nl0, line 69: is this needed tho? you've already bumped the default value

fiskus: Sorry, meant to bump only this value. Forgot to remove the default or messed up when rebasing

fiskus: Yes, forgot to commit.

Though, singing only used in renderers. So, maybe, it's better to increase default value to 10 minutes, but use the old default for thumbnails and logo? What do you think?

nl0: i think it's ok to just use the new default for everything

fiskus: Yes, it's probably the most reasonable choice. Thanks

fiskus:
I think, we no longer need this after https://github.com/quiltdata/quilt/pull/3979

PR #3979, opened 2024-05-17
Use promises for URLs in IGV to have fresh signing each time they used

fiskus:

### DIFF
--- a/catalog/app/components/Preview/loaders/Igv.ts
+++ b/catalog/app/components/Preview/loaders/Igv.ts
@@ -73,6 +73,13 @@ export const Loader = function IgvLoader({ gated, handle, children }: IgvLoaderP
       const tail = data.tail.join('\n')
       try {
         const options = JSON.parse([head, tail].join('\n'))
+        if (!options.reference?.indexFile && options.reference?.fastaURL) {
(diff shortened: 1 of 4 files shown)


nl0 (dismissed):
i don't quite get it -- is IGV able to accept async thunks instead of strings as URLs?

fiskus:
> is IGV able to accept async thunks instead of strings as URLs?

Yes, it is. Sorry, I didn't add this to the PR description.
They use their own helper module for every URL (though in documentation they said only about `tracks` URLs), and it nailed down to that function where it is a check if URL is a string or a function, and those functions or strings could be `async`:
https://github.com/igvteam/igv-utils/blob/master/src/igvxhr.js#L91

fiskus:
I can pass thunks but if `indexURL` or `indexFile` is omitted, they try to guess it, and it produces invalid URL like `function(){}.fai`. I filed a bug and fix https://github.com/igvteam/igv.js/pull/1821
Hope they accept it, so I can remove our hacky fix.

nl0:
> Hope they accept it

looks like they did. not sure when they're going to cut their next release tho

nl0 (approved):

PR #3983, opened 2024-05-23
TSify `utils/Dialogs` and some related code

nl0:
Extracted from #3982


catalog/app/utils/Dialogs.tsx
```diff
@@ -0,0 +1,67 @@
...
+import * as M from '@material-ui/core'
+
+import defer from 'utils/defer'
+import type { Resolver } from 'utils/defer'
+
+type DialogState = 'open' | 'closing' | 'closed'
+
+type ExtraDialogProps = Omit<M.DialogProps, 'open' | 'onClose' | 'onExited'>
+
+export type Close<R> = [R] extends [never] ? () => void : Resolver<R>['resolve']
```
fiskus, line 11: I don't understand this syntax :(

catalog/app/utils/Dialogs.tsx
```diff
@@ -0,0 +1,67 @@
...
+      const { resolver, promise } = defer<R>()
+      setDialog({ render, props, resolver })
+      setState('open')
+      return promise
+    },
+    [setDialog, setState],
+  )
+
+  const close = React.useCallback(
+    (reason: any) => {
```
fiskus, line 36: Aren't there just three "reasons" 'cancel' | 'empty' and 'readme'?

fiskus: So, maybe, you can use `M.DialogReason` (I don't remember type exactly, but there is something like this), and then assert for "reasons" we support.

fiskus (approved):

catalog/app/utils/Dialogs.tsx
nl0: this might be helpful: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types

basically i wanted to allow calling the `close()` function without any arguments by default, so i used a conditional type here, but it wasn't working out of the box for some cases until i used the workaround described in the docs

catalog/app/utils/Dialogs.tsx
nl0: the idea is to make the return value configurable

> 'cancel' | 'empty' and 'readme'?

yeah, exactly, these are the "reasons" for this exact dialog -- they may be different for the other cases (in most cases we don't need that at all -- defaulting to "never")

catalog/app/utils/Dialogs.tsx
fiskus: Thanks!

PR #3988, opened 2024-05-27
Don't fold GQL `Me` data, split code to `MobileSignIn`/`DesktopSignIn`

fiskus:
* Don't fold data early, so we can show error
* Move GQL `Me` call down into where it is used (in further refactoring, on mobile we can call it only when menu is opened)
* To achieve both these, I split code more explicitly into `MobileSignIn` and `DesktopSignIn`


nl0:
@fiskus is this kinda ready?

fiskus:
Thank you, yes. I waited for CI checks to complete, and then forgot to request review

nl0 (approved):

PR #3990, opened 2024-06-04
GraphQL-based quilt3.admin API

sir-sigurd:
# Description

# TODO

- [x] better exceptions
- [x] fix `ImportError: cannot import name 'Annotated' from 'typing'` (drop Python 3.8?)
- [x] wrap dates?
- [x] docstrings
- [x] type docstrings
- [x] adminql -> admin
- [x] add requirements.txt
- [x] CI
- [x] Unit tests
- [x] Documentation


drernie:
This looks great!  I assume in the final version it will be something like "quilt3.admin"; 'adminql' is too much information

My main concern is that I find the name and type of "role" confusing; when do I use a role_id vs a role_name?

Is it possible to:
 - standardize on only one
 - change the argument names to be clearer which is which
 - make the type explicit enough to distinguish (rather than just "str")

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+from datetime import datetime
+from typing import Any, Annotated, Literal, Union, Optional, List
+
+from pydantic import Field, TypeAdapter
+
+from ._graphql_client import *
+from ._graphql_client.base_model import UNSET, UnsetType
+
+
+class ManagedRole(BaseModel):
```
sir-sigurd, line 14: I added these types to wrap what generated code returns, because otherwise we end up with 2 types for user, 3 types for managed roles and 3 types for unmanaged roles in signatures (and docs).

@nl0 
any better idea?

docs/api-reference/Admin.md
```diff
@@ -1,29 +1,102 @@
...
+
+`self` is explicitly positional-only to allow `self` as a field name.
+
+## User(\*, name: str, email: str, date\_joined: datetime.datetime, last\_login: datetime.datetime, is\_active: bool, is\_admin: bool, is\_sso\_only: bool, is\_service: bool, role: Optional[Annotated[Union[quilt3.admin.ManagedRole, quilt3.admin.UnmanagedRole], FieldInfo(annotation=NoneType, required=True, discriminator='typename\_\_')]], extra\_roles: List[Annotated[Union[quilt3.admin.ManagedRole, quilt3.admin.UnmanagedRole], FieldInfo(annotation=NoneType, required=True, discriminator='typename\_\_')]]) -> None  {#User}
+Create a new model by parsing and validating input data from keyword arguments.
+
+Raises [`ValidationError`][pydantic_core.ValidationError] if the input data cannot be
+validated to form a valid model.
+
+`self` is explicitly positional-only to allow `self` as a field name.
```
sir-sigurd, line 29: That's not great but might be OK for now.
I don't think we should try to do something with https://github.com/quiltdata/pydoc-markdown/ and should switch to something something commonly used e.g. sphinx which has extension for pydantic: https://sphinx-pydantic.readthedocs.io/en/latest/

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+
+class UserNotFoundError(Quilt3AdminError):
+    def __init__(self):
+        super().__init__(None)
+
+
+def _handle_errors(result: BaseModel) -> Any:
+    if isinstance(result, (InvalidInputSelection, OperationErrorSelection)):
+        raise Quilt3AdminError(result)
+    return result
```
sir-sigurd, line 59: is this sane?

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+
+def get_user(name: str) -> User:
+    """
+    Get a specific user from the registry.
+
+    Args:
+        name: Username of user to get.
+    """
+    result = _get_client().get_user(name=name)
+    # XXX: should we really throw an exception here?
```
nl0, line 73: no. i think returning None is adequate

api/python/quilt3/admin/_graphql_client/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+from .get_user import GetUser, GetUserAdmin, GetUserAdminUser, GetUserAdminUserGet
+from .get_users import GetUsers, GetUsersAdmin, GetUsersAdminUser, GetUsersAdminUserList
+from .input_types import UserInput
+from .remove_roles import (
+    RemoveRoles,
+    RemoveRolesAdmin,
+    RemoveRolesAdminUser,
+    RemoveRolesAdminUserMutate,
+    RemoveRolesAdminUserMutateRemoveRolesInvalidInput,
+    RemoveRolesAdminUserMutateRemoveRolesOperationError,
```
nl0, line 54: maybe we can tell the codegen to separate parts with an underscore to make it a little bit more readable? e.g. `RemoveRoles_Admin_User_Mutate_RemoveRoles_OperationError`

docs/api-reference/Admin.md
nl0: > switch to something something commonly used e.g. sphinx which has extension for pydantic

yeah that would be nice

nl0 (commented):
looks good. some suggestions inline

api/python/quilt3-admin/exceptions.py
```diff
@@ -0,0 +1,84 @@
+# pylint: disable=super-init-not-called
+from typing import Any, Dict, List, Optional, Union
+
+import requests
+
+
+class GraphQLClientError(Exception):
```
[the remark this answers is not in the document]
sir-sigurd: yes, I've just added comment about that

api/python/quilt3/admin/_graphql_client/__init__.py
sir-sigurd: nope, we can't AFAICT

api/python/quilt3-admin/queries.graphql
```diff
@@ -0,0 +1,133 @@
...
+          ...OperationErrorSelection
+        }
+      }
+    }
+  }
+}
+
+query getRoles {
+  roles {
+      ...RoleSelection
```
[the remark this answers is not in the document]
sir-sigurd: fixed

.github/workflows/test-quilt3-admin-codegen.yaml
```diff
@@ -0,0 +1,34 @@
+name: Test quilt3.admin code generation
+
+on:
+  push:
+    paths:
+      - '.github/workflows/test-quilt3-admin-codegen.yaml'
+      - 'api/python/quilt3-admin/**'
+      - 'api/python/quilt3/admin/_graphql_client/**'
```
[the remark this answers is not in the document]
sir-sigurd: yes, good catch, fixed

api/python/quilt3-admin/queries.graphql
```diff
@@ -0,0 +1,133 @@
...
+  isSsoOnly
+  isService
+  role {
+    ...RoleSelection
+  }
+  extraRoles {
+    ...RoleSelection
+  }
+}
+fragment InvalidInputSelection on InvalidInput {
```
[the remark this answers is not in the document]
sir-sigurd: no, that doesn't help

api/python/quilt3-admin/base_client.py
```diff
@@ -0,0 +1,211 @@
...
+from .exceptions import (
+    GraphQLClientGraphQLMultiError,
+    GraphQLClientHttpError,
+    GraphQLClientInvalidResponseError,
+)
+
+Self = TypeVar("Self", bound="BaseClient")
+
+
+class BaseClient:
```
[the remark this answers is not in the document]
sir-sigurd: > can we import and subclass this instead of rewriting? maybe that would lessen maintenance burden idk

to import it we need to depend on ariadne-codegen and I don't think we want it

> or at least mark the modified parts

I don't think someone will want to modify it soon and just doing diff against upstream version makes more sense IMO

api/python/quilt3/admin/__init__.py
sir-sigurd: > if nothing else, i guess you could just import the fragment types and use them directly -- they should be equivalent to the types you construct here (tho maybe names would be more noisy, but i guess you can just subclass them to fix this)

well, if I do something like
```python
class User(UserSelection):
    pass
```

1. it would still refer to `UserSelectionRoleUnmanagedRole`, `UserSelectionRoleManagedRole`, `UserSelectionExtraRolesUnmanagedRole` and `UserSelectionExtraRolesManagedRole`
2. field names end up spelled in camel case instead of snake case in type constructors and I think we should provide API with snake case

see https://github.com/quiltdata/quilt/blob/134c93747ed96098bd471262c4c550fc50bff9fd/api/python/quilt3/admin/_graphql_client/fragments.py#L41-L64

docs/api-reference/Admin.md
```diff
@@ -1,29 +1,102 @@
 
 # quilt3.admin
+
 APIs for Quilt administrators. 'Registry' refers to Quilt stack backend services, including identity management.
 
-## create\_user(\*, username: str, email: str)  {#create\_user}
+
+## ManagedRole(\*, \_\_typename: Literal['ManagedRole'], id: str, name: str, arn: str) -> None  {#ManagedRole}
+Create a new model by parsing and validating input data from keyword arguments.
```
akarve, line 8: We need to introduce the model concept first if we're going to talk about it.

api/python/quilt3/admin/__init__.py
nl0: another idea: since you're using custom classes anyways, just use plain dataclasses -- they will show up better (simpler?) in the docs. since the data is expected to be valid at that point, it's ok to trust it and not validate it again

sir-sigurd: > since the data is expected to be valid at that point, it's ok to trust it and not validate it again

but I guess I would have to wrap nested non-scalars (e.g. `role` and `extra_roles`) manually?

sir-sigurd: > but I guess I would have to wrap nested non-scalars (e.g. role and extra_roles) manually?

well, I think I don't

sir-sigurd: ok, I switched to dataclasses, now doc looks a bit better

sir-sigurd:
@nl0 
do you think anything needs to be done other than tests?

nl0:
> @nl0 do you think anything needs to be done other than tests?

idk, seems good enough. maybe add some more management commands? e.g. make and admin, edit an email, etc?

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def get_users() -> List[User]:
+    """
+    Get a list of all users in the registry.
+    """
+    return [User(**u.model_dump()) for u in _get_client().get_users()]
+
+
+def create_user(name: str, email: str, role: str, extra_roles: Optional[List[str]] = None) -> None:
```
nl0, line 90: shouldn't we return the created user data here?

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    return [role_adapter.validate_python(r.model_dump()) for r in _get_client().get_roles()]
+
+
+def set_role(
+    name: str,
+    role: str,
+    extra_roles: Optional[List[str]] = None,
+    *,
+    append: bool = False,
+) -> None:
```
nl0, line 132: return the updated user data?

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    Set the active and extra roles for a user.
+
+    Args:
+        name: Username of user to update.
+        role: Role to be set as the active role.
+        extra_roles: Additional roles to assign to the user.
+        append: If True, append the extra roles to the existing roles. If False, replace the existing roles.
```
nl0, line 140: if you add the comments directly to the arguments, will the doc gen use them?

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    Get a specific user from the registry. Return `None` if the user does not exist.
+
+    Args:
+        name: Username of user to get.
+    """
+    result = _get_client().get_user(name=name)
+    if result is None:
+        return None
+    return User(**result.model_dump())
```
nl0, line 80: i'd prefer making this a classmethod on User, tho not sure it's worth it

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        extra_roles: Additional roles to assign to the user.
+        append: If True, append the extra roles to the existing roles. If False, replace the existing roles.
+    """
+    result = _get_client().set_role(name=name, role=role, extra_roles=extra_roles, append=append)
+    if result is None:
+        raise UserNotFoundError
+    _handle_errors(result.set_role)
+
+
+def add_roles(name: str, roles: List[str]) -> None:
```
nl0, line 148: same here: probably worth it to return the updated user data

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        name: Username of user to update.
+        roles: Roles to add to the user.
+    """
+    result = _get_client().add_roles(name=name, roles=roles)
+    if result is None:
+        raise UserNotFoundError
+    _handle_errors(result.add_roles)
+
+
+def remove_roles(
```
nl0, line 162: same here: probably worth it to return the updated user data

api/python/quilt3/admin/__init__.py
sir-sigurd: I'm not sure that's really useful

api/python/quilt3/admin/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    Args:
+        name: Username of user to delete.
+    """
+    result = _get_client().delete_user(name=name)
+    if result is None:
+        raise UserNotFoundError
+    _handle_errors(result.delete)
+
+
+def get_roles() -> List[Union[_graphql_client.GetRolesRolesUnmanagedRole, _graphql_client.GetRolesRolesManagedRole]]:
```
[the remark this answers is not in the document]
sir-sigurd: I guess I'll have to handle `__typename` manually in that case or do you have a better idea?

api/python/quilt3/admin/__init__.py
sir-sigurd: there is no way to put docstrings directly to parameters
that way of documenting parameters is just one of the conventions on top of function docstrings (specifically from [Google one](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings))

api/python/quilt3/admin/__init__.py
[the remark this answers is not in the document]
nl0: oh actually i see that you _are_ wrapping the result, but the type hint doesn't quite match that -- it should be `List[AnnotatedRole]` (or maybe `List[Role]`)

api/python/quilt3/admin/__init__.py
nl0: there is data not directly inferrable from the input params (generated/inferred by the backend), like isAdmin flag and such -- tho i'm not sure it's that useful either

api/python/quilt3/admin/__init__.py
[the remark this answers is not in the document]
sir-sigurd: ok, fixed typing, not sure why type checker haven't complain

api/python/quilt3/admin/__init__.py
sir-sigurd: yes, and `User.set_admin()` and `User.set_email()` and so on

nl0: i'm not against that in principle, but it would require thorough design considerations.
this kind of "resource"-oriented api feels kinda more complex, bc you have to distinguish between entity identity vs attributes/props.
e.g. do you have to fetch the user before mutating it, like `User.get(name).set_role(role)`?
or can you just construct a "resource object" by providing an identifier, e.g. `User(name).set_role(role)`?

sir-sigurd: > e.g. do you have to fetch the user before mutating it, like User.get(name).set_role(role)?

this

sir-sigurd: > i'm not against that in principle, but it would require thorough design considerations.
this kind of "resource"-oriented api feels kinda more complex, bc you have to distinguish between entity identity vs attributes/props.

yes, so I'd probably leave things as is and we could rework it in future

sir-sigurd:
CI currently fail because of https://github.com/jd/tenacity/issues/471

sir-sigurd:
@nl0 
does it make sense to rename some functions
set_role -> set_user_role
add_roles -> add_user_roles
remove_roles -> remove_user_roles
?

or maybe use prefixes for consistency? i.e.

get_user -> user_get
get_users -> user_list
set_role -> user_set_role
?

nl0:
> does it make sense to rename some functions?

yeah i think so

> or maybe use prefixes for consistency? i.e.

that feels like a good intent, but the resulting names don't look too nice, so... your call.
imo it would look nice if we had some sort of a namespace (e.g. `admin.users.list` or smth, idk)

sir-sigurd:
> imo it would look nice if we had some sort of a namespace (e.g. `admin.users.list` or smth, idk)

I made this change

nl0 (approved):
looks good

nl0 (approved):

PR #3994, opened 2024-06-05
Use `meta.validating` instead of manually control state

fiskus:
This fixes `Warning: Cannot update a component ('PackageCreationForm') while rendering a different component ('ForwardRef(Field)'). To locate the bad setState() call inside 'ForwardRef(Field)', follow the stack trace as described in https://reactjs.org/link/setstate-in-render`


nl0 (approved):
nice

PR #4008, opened 2024-06-18
Fix thumbnail lambda build error caused by numpy@2 release

nl0:
Thumbnail lambda has a transitive dependency (via `aicsimageio`) on `tifffile @ 0.15.1`,
which states `numpy >= 1.x` as a build dependency.
`pip` resolves this to a recently released `numpy @ 2` (even tho the top-level requirements file has `numpy < 2`),
and that breaks the build (which is only compatible with v1).
Workaround is quite simple -- install `numpy @ 1` before trying to install / build `tifffile`,
satisfying the build requirement and avoiding installation of `numpy @ 2`.

### DIFF
--- a/.github/workflows/py-ci.yml
+++ b/.github/workflows/py-ci.yml
@@ -181,22 +181,21 @@ jobs:
           python-version-file: lambdas/${{ matrix.path }}/.python-version
       - name: Install dependencies
         run: |
-          # Due to behavior change in pip>=23.1 installing tifffile==0.15.1
-          # from thumbnail lambda fails whithout installed wheel.
-          # See https://github.com/pypa/pip/issues/8559.
-          python -m pip install wheel
-
-          if [ ${{ matrix.path }} == "shared" ]
-            python -m pip install -e lambdas/shared[tests]
-          then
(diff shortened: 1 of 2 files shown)


.github/workflows/py-ci.yml
```diff
@@ -185,6 +185,8 @@ jobs:
           # from thumbnail lambda fails whithout installed wheel.
           # See https://github.com/pypa/pip/issues/8559.
           python -m pip install wheel
+          # HACK: Pre-install numpy v1 as a build dependency for tifffile to prevent it from using v2 and failing to build
+          python -m pip install 'numpy<2'
```
sir-sigurd, line 189: I'd put this under `if` similar to the one below.

sir-sigurd (approved):

.github/workflows/py-ci.yml
nl0: ok. btw i'm not quite sure why that `if [ ... ] ... then ... else ... fi` statement has something between if and then? is it a bug?

fiskus (approved):

PR #4024, opened 2024-06-21
Fix getting `writeFile` config property

fiskus:


drernie (approved):

drernie:
@akarve FYI, this is a showstopper bug that must be fixed and validated before we can ship.

nl0 (approved):

nl0:
🤦 how i missed this

fiskus:
Python was patch-updated in `py-ci` job, and it causes the `test_brackets` test fail. I don't know how to properly fix it.
And I can't merge when test fails.

drernie:
Weird. And test_brackets only seem to fail on Windows 12.  Can we revert the Python version?

```
>       assert pkg['asdf']['qwer'].get() == LOCAL_MANIFEST.as_uri()
E       AssertionError: assert 'file://///D:...anifest.jsonl' == 'file:///D:/a...anifest.jsonl'
E         - file:///D:/a/quilt/quilt/api/python/tests/integration/data/local_manifest.jsonl
E         + file://///D:/a/quilt/quilt/api/python/tests/integration/data/local_manifest.jsonl
E         ?      ++
```

drernie:
It looks like things went wrong with Log file maintenance c07d062784fc0f3ab6d64811cf96c941d66c10e2

First, it failed, then it succeeded, and then everything went wrong...

Screenshot 2024-06-22 at 13 54 23

drernie:
Ouch.  This Amazon Linux commit  273189e that previously completed successfully now fails on Windows 12.

https://github.com/quiltdata/quilt/actions/runs/9611572379

drernie:
Trying the prior AmazonLinux commit 14b2f04...
That also fails.  So, it sure seems like something OTHER than the code is causing the failure...

PR #4025, opened 2024-06-22
Hotfix: `test_brackets` test

fiskus:
Python got a patch update. And, now `test_brackets` is failing. I guess, it is related to https://github.com/python/cpython/issues/67693


fiskus:
Fixed here https://github.com/quiltdata/quilt/pull/4027

PR #4027, opened 2024-06-23
Fix PhysicalKey to URI conversion in Python 3.12 on Windows

dimaryaz:
On Windows, `pathname2url` returns three leading slashes for some reason. `urlunparse` has been ignoring them - until Python 3.12, where they result in a URL like `file://///C:/...`.

We can instead use `pathlib.PurePath.as_uri` for this. It is certainly cleaner than using `pathname2url` (whose documentation basically says "don't use it") - however, we are still stuck with using the `url2pathname` for the opposite conversion, since there's no `pathlib` solution for that.

### DIFF
--- a/api/python/quilt3/util.py
+++ b/api/python/quilt3/util.py
@@ -13,7 +13,7 @@ from urllib.parse import (
     urlparse,
     urlunparse,
 )
-from urllib.request import pathname2url, url2pathname
+from urllib.request import url2pathname
 
 import requests
 # Third-Party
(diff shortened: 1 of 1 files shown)


nl0 (approved):
oh wow thanks for fixing this. i was just re-running the test until it succeeded (and it was succeeding _sometimes_, not sure why lol)

dimaryaz:
> it was succeeding sometimes

Huh, that seems impossible... Unless CI uses some machines that still use an older release of Python?

nl0:
> > it was succeeding sometimes
> 
> Huh, that seems impossible... Unless CI uses some machines that still use an older release of Python?

yeah i guess it's related to some kind of caching idk

drernie (approved):

PR #4030, opened 2024-06-24
Bump pandocfilters to fix build

sir-sigurd:
Currently it fails with:
KeyError: "linkname 'pandocfilters-1.4.2/README' not found"

See https://github.com/quiltdata/quilt/actions/runs/9642202413/job/26589474567#step:3:210


sir-sigurd:
it might be a good idea to remove this line

https://github.com/quiltdata/quilt/blob/1594d8412c7cfdf461a05fe1e20e14964c3f524c/lambdas/scripts/build_zip.sh#L18

or maybe upgrade pip in tests job? that way we'd have the same pip version in tests and build jobs.

nl0 (approved):

PR #4038, opened 2024-06-29
Remove polyfill.io

nl0:


fiskus:
Polyfill was added when we replaced Babel with Typescript, as I understand https://github.com/quiltdata/quilt/pull/2043

fiskus (approved):

fiskus:
I checked that `ResizeObserver` and `RelativeTimeFormat`, which explicitly polyfiled, work fine in Firefox without polyfill.io

nl0:
> I checked that `ResizeObserver` and `RelativeTimeFormat`, which explicitly polyfiled, work fine in Firefox without polyfill.io

yeah, they are widely supported nowadays according to caniuse

PR #4062, opened 2024-07-31
s3hash lambda: Fix invalid checksum for some non-canonical objects with existing checksum

sir-sigurd:
# TODO

- [x] Unit tests

### DIFF
--- a/lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
+++ b/lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
@@ -176,7 +176,9 @@ def get_compliant_checksum(attrs: GetObjectAttributesOutputTypeDef) -> T.Optiona
     assert "Parts" in object_parts
     # Make sure we have _all_ parts.
     assert len(object_parts["Parts"]) == num_parts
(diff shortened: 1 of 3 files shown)


lambdas/s3hash/src/t4_lambda_s3hash/__init__.py
```diff
@@ -176,7 +176,9 @@ def get_compliant_checksum(attrs: GetObjectAttributesOutputTypeDef) -> T.Optiona
     assert "Parts" in object_parts
     # Make sure we have _all_ parts.
     assert len(object_parts["Parts"]) == num_parts
-    if all(part.get("Size") == part_size for part in object_parts["Parts"][:-1]):
+    expected_num_parts, remainder = divmod(attrs["ObjectSize"], part_size)
+    expected_part_sizes = [part_size] * expected_num_parts + ([remainder] if remainder else [])
+    if [part.get("Size") for part in object_parts["Parts"]] == expected_part_sizes:
```
sir-sigurd, line 181: I think it would be sufficient to just add check either for last part or number of parts, but it feels safer to check both.

nl0 (approved):
lgtm

PR #4071, opened 2024-08-06
Drop parsed.netloc check for files

akarve:
Network drives have a netloc of the drive IP. There's no reason to fail from_url just because of that.


drernie:
Closing old @akarve PRs

PR #4096, opened 2024-08-14
Add docs for SSO permissions mapping

sir-sigurd:


docs/advanced-features/sso-permissions.md
```diff
@@ -0,0 +1,53 @@
...
+
+> This feature requires Quilt stack version 1.54.0 or higher
+
+## Overview
+
+This feature allows Quilt admin to configure what roles and admin flag value are
+assigned to the user who logs in via SSO given on the user's ID token claims.
+
+The configuration can be initially set with admin `quilt3.admin.sso_config.set()`
+and subsequently can be update with admin UI as well.
```
sir-sigurd, line 11: we don't display "SSO config" button in UI when there is no SSO config set
@fiskus says that might be a feature not a bug because only the most advanced users can access it

fiskus: There is a PR for enabling it https://github.com/quiltdata/quilt/pull/4097
But there is no validation, so, users can put some nonsense content into config, and, potentially break the stack

drernie:
Ah, I only thought we were supporting API upload configuration, in order to perform validation and erroring.

We should not provide an alternative way to update the configuration without validation.

sir-sigurd:
> Ah, I only thought we were supporting API upload configuration, in order to perform validation and erroring.
> 
> We should not provide an alternative way to update the configuration without validation.

all validation is done on backend so I don't understand the problem

fiskus:
Ernie, there is no frontend validation (except handling errors from backend), I apologize for confusion. And users could make the same mistakes in Web editor as in `quilt3.admin.sso_config.set()`.
But, since, the config is something sensitive, probably, it makes sense to not allow creating it from web, because it's too easy.

docs/advanced-features/sso-permissions.md
```diff
@@ -0,0 +1,63 @@
...
+
+![admin UI for setting SSO permissions mapping](../imgs/admin-sso-config.png)
+
+> Note: Roles used by configuration can't be removed or renamed.
+
+> Note: User who sets configuration last can't be revoked admin flag.
```
sir-sigurd, line 17: but currently this user still can be disabled, that doesn't look right, does it?

docs/advanced-features/sso-permissions.md
```diff
@@ -0,0 +1,63 @@
...
+        - groups
+    roles:
+      - ReadWriteQuiltBucket
+```
+
+1. user with email `admin@example.com` will have `ReadWriteQuiltBucket` role and
+admin flag set to true
+1. user with group `rw` will have `ReadWriteQuiltBucket` role and admin flag set
+to false (except the user with `admin@example.com` email)
+1. all other users will have `ReadQuiltBucket` role
```
sir-sigurd, line 63: currently the users who aren't matched by any mapping have admin flag untouched, shouldn't we set it to false?

docs/advanced-features/sso-permissions.md
drernie: I don't think it matters a huge amount either way, but I am fine with us requiring first-use to happen via the API.

docs/advanced-features/sso-permissions.md
drernie: That is less than ideal, but not a showstopper.  Worth documenting, though.

docs/advanced-features/sso-permissions.md
drernie: I wouldn't think so.  If they went through the effort to create a manual user account and give it Admin privileges, they may well have a good reason.

drernie (dismissed):

drernie:
> But, since, the config is something sensitive, probably, it makes sense to not allow creating it from web, because it's too easy.

Actually, now that I am writing the release notes, it seems odd not to allow Admins to upload a configuration file from the GUI for the first time but allow it later.  The main reason I wanted API-only uploads was because I didn't think we had time for a GUI.  Since both are equally validated, I don't see any particular risk.

This may be too late too change, but I wanted to weigh in.

fiskus:
PR for enabling _adding_ SSO config from Admin UI is here https://github.com/quiltdata/quilt/pull/4097

drernie (approved):
Please review these grammar changes

docs/advanced-features/sso-permissions.md
```diff
@@ -0,0 +1,65 @@
...
+This feature allows Quilt admin to configure what roles and admin flag value are
+assigned to the user who logs in via SSO based on the user's ID token claims.
+
+The configuration can be set with `quilt3.admin.sso_config.set()` or with admin UI.
+
+![admin UI for setting SSO permissions mapping](../imgs/admin-sso-config.png)
+
+> Note: Roles used by configuration can't be removed or renamed.
+
+> Note: User who sets configuration last can't be revoked admin flag.
```
drernie, line 16:
```suggestion
> Note: The user who sets the configuration will never have their admin flag revoked.
```

drernie (approved):
Two minor changes.

drernie (approved):

docs/advanced-features/sso-permissions.md
akarve: Ever or just in the edit AFTER they set the configuration. Agree this needs to be documented.

docs/advanced-features/sso-permissions.md
akarve: I think we should set it to `False`. It is far to dangerous to make someone admin by default. Admins should be explicit.

docs/advanced-features/sso-permissions.md
akarve: I read the new string and I think this is permanent. It seems like a really odd side-effect. So even if someone sets the config after they do you can't revoke the old admin? That doesn't sound right indeed.

docs/advanced-features/sso-permissions.md
drernie: Nobody is set to admin by default, except the user who uploaded the config file (with admin permissions).

This simply does NOT remove Admin permissions that were manually added to users that don't march the SSO mappings.

docs/advanced-features/sso-permissions.md
drernie: @sir-sigurd Can a new admin create a mapping that disables the Admin Flag for the previous Admin?

sir-sigurd: > So even if someone sets the config after they do you can't revoke the old admin? 

admin flag can't be revoked only from use who set config ***last***

my concern was that config uploader still can be disabled, so AFAICT you still end up without a single active admin in the stack under some circumstances

PR #4113, opened 2024-08-20
Wider markdownlint settings

fiskus:
Instead of linting only `./docs`, lint everything and use `.markdownignore` to ignore the rest.

Anyway, even `./docs` didn't work well, glob should have "quotes"


docs/advanced-features/sso-permissions.md
sir-sigurd, line 1: I think we should disabled `MD028` because currently it's rendered as intended as separate blockquotes
image
vs single blockquote after change
image

drernie: Concur

drernie (dismissed):

docs/advanced-features/sso-permissions.md
sir-sigurd: @drernie 
why did you mark this as resolved?

drernie: Sorry, should have let you.  But I agree with you.

fiskus: As I see, blockquotes are used because of how they look. It's not a blockquote by meaning.

Can we use the same "Note" hints? https://docs.gitbook.com/content-editor/blocks/hint

sir-sigurd: > Can we use these "Note" hints?

I don't think we should do that because I think we should migrate from GitBook 🤷‍♂️

drernie: Please check with @QuiltSimon , as I think he was planning to double-down on GitBook.  Is there a better alternative?

fiskus: I disabled MD028 for this particular file

sir-sigurd (approved):

.markdownlintignore
```diff
@@ -0,0 +1,12 @@
+api
+catalog/node_modules
```
sir-sigurd, line 2: might make sense together to put it in the group with `.git` and `venv`

docs/CHUNKED_CHECKSUMS.md
```diff
@@ -34,7 +34,9 @@ as of at least [boto3 1.34.44](https://pypi.org/project/boto3/1.34.44/)
 * Prefix: 0xb510
 * Status: draft
 * Type: multihash
-* Description: Hash of concatenated SHA2-256 digests of 8*2^n MiB source chunks;
-  n = ceil(log2(source_size/(10^4 * 8MiB)))
-* Registrar: [multiformats/multicodec](https://github.com/multiformats/multicodec) 
-* Registration Date: [2024-02-23](https://github.com/multiformats/multicodec/pull/343)
\ No newline at end of file
+* Description: Hash of concatenated SHA2-256 digests of `8*2^n MiB` source chunks;
+`n = ceil(log2(source_size/(10^4 * 8MiB)))`
```
sir-sigurd, line 38: I'm not sure using backticks here is justfied

fiskus: I don't understand why this whole section is here. But if it is there, you may be right, that it should be identical to https://github.com/multiformats/multicodec/pull/343/files

However, ` * ` is treated like the start of *emphasizing*, and, also, it is a formula, so, it can be marked as code.

fiskus: I'm disabling the failing lint rule here

.markdownlintignore
fiskus: Agree

sir-sigurd (approved):

PR #4116, opened 2024-08-21
Mention removed top_hash parameter in `push()` docs

sir-sigurd:


api/python/quilt3/packages.py
```diff
@@ -1386,6 +1386,7 @@ def push(
             name: name for package in registry
             dest: where to copy the objects in the package. Must be either an S3 URI prefix (e.g., s3://$bucket/$key)
                 in the registry bucket, or a callable that takes logical_key and package_entry, and returns an S3 URI.
+                (Changed in 6.0.0a1) previously top_hash was passed to the callable dest as a third argument.
```
sir-sigurd, line 1389: can't use colon here because docgen start to think it's another function parameter

docs/api-reference/Package.md
```diff
@@ -303,6 +303,7 @@ __Arguments__
 * __name__:  name for package in registry
 * __dest__:  where to copy the objects in the package. Must be either an S3 URI prefix (e.g., s3://$bucket/$key)
     in the registry bucket, or a callable that takes logical_key and package_entry, and returns an S3 URI.
+    (Changed in 6.0.0a1) previously top_hash was passed to the callable dest as a third argument.
```
sir-sigurd, line 306: I wanted to format it like `Changed in 6.0.0a1: ...` but docgen start to think it's another function parameter because of colon

drernie (approved):

PR #4117, opened 2024-08-22
Service role for stack installation and deletion

akarve:
Tested for stack creation and stack deletion with "all features on" using TF on a pure CF deployment

### DIFF
--- a/docs/cfn-service-role.yaml
+++ b/docs/cfn-service-role.yaml
@@ -69,6 +69,7 @@ Resources:
                   - ec2:CreateVpc
                   - ec2:DeleteInternetGateway
                   - ec2:DeleteLaunchTemplate
+                  - ec2:DeleteNetworkInterface
                   - ec2:DeleteRoute
                   - ec2:DeleteRouteTable
                   - ec2:DeleteSecurityGroup
@@ -90,6 +91,7 @@ Resources:
                   - ec2:DescribeVpcAttribute
                   - ec2:DescribeVpcs
                   - ec2:DetachInternetGateway
+                  - ec2:DetachNetworkInterface
                   - ec2:DisassociateRouteTable
                   - ec2:DisassociateVerifiedAccessInstanceWebAcl
                   - ec2:GetVerifiedAccessInstanceWebAcl
--- a/docs/cfn-service-role.yaml
+++ b/docs/cfn-service-role.yaml
@@ -228,7 +228,7 @@ Resources:
                   - rds:ListTagsForResource
                   - rds:ModifyDBInstance
                   - route53:CreateHostedZone
-                  - s3:* # could be improved but canary buckets failing on access denied even with GetObject
+                  - s3:*  # Stack has this anyway; Canary buckets get Access Denied even with GetObject
                   - secretsmanager:CreateSecret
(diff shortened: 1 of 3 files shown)


docs/cfn-service-role.yaml
```diff
@@ -11,31 +11,51 @@ Resources:
...
-                  - apigateway:GetResources
                   - apigateway:PATCH
                   - apigateway:POST
                   - apigateway:PUT
+                  - apigateway:SetWebACL
                   - apigateway:UpdateRestApiPolicy
+                  - appsync:SetWebACL
+                  - apprunner:AssociateWebAcl
+                  - apprunner:DescribeWebAclForService
+                  - apprunner:DisassociateWebAcl
```
sir-sigurd, line 30: where does this come from? 🤔

sir-sigurd (approved):
I still think it might make sense to just have `<service>:*` actions at this point

docs/cfn-service-role.yml
```diff
@@ -91,19 +117,30 @@ Resources:
...
                   - es:DeleteElasticsearchDomain
                   - es:DescribeElasticsearchDomain
+                  - es:UpdateElasticsearchDomainConfig
                   - events:DeleteRule
                   - events:DescribeRule
+                  - events:ListTargetsByRule
                   - events:PutRule
                   - events:PutTargets
                   - events:RemoveTargets
+                  - firehose:CreateDeliveryStream
```
QuiltSimon, line 145: What are we using Kinesis Firehose for?

docs/cfn-service-role.yml
```diff
@@ -135,42 +185,96 @@ Resources:
...
+                  - s3:GetBucketTagging
+                  - s3:GetBucketVersioning
+                  - s3:GetBucketWebsite
+                  - s3:GetEncryptionConfiguration
+                  - s3:GetIntelligentTieringConfiguration
+                  - s3:GetInventoryConfiguration
+                  - s3:GetLifecycleConfiguration
+                  - s3:GetMetricsConfiguration
                   - s3:GetObject
+                  - s3:GetReplicationConfiguration
```
QuiltSimon: Is this a currently used IAM policy? Or potentially future state?

QuiltSimon (approved):
ok

docs/cfn-service-role.yaml
akarve: That is apparently how WAF operations are dispatched. Specifically it was reported by IAM > Generate Policy.

docs/cfn-service-role.yml
akarve: Canaries events AFAICT.

docs/cfn-service-role.yml
akarve: We take * now (as the stack has). Believe it or not this is likely a current operation as reported by CloudTrail and Generate Policy.

docs/cfn-service-role.yaml
sir-sigurd: won't it work without this?

docs/cfn-service-role.yml
sir-sigurd: audit trail

PR #4122, opened 2024-08-27
Move bucket settings to a separate page

fiskus:
* Added two routes `/admin/buckets/edit/:bucketName` and `/admin/buckets/add`
* Old route `buckets?bucketName` redirected to new one
* Dialog components removed
* ~Form components are mostly intact, but hierarchy of component slightly changed so the diff is bigger than actual change~

"Add bucket" page. No preview cards, single form.
Screenshot from 2024-09-03 14-47-01

---

Bucket page. Bucket has avatar
Screenshot from 2024-09-03 14-48-20

---

Bucket page. No avatar, many tags.
Screenshot from 2024-09-03 15-11-19

---

Edit metadata on bucket page. All other cards are in preview mode.
Screenshot from 2024-09-03 15-11-29

---

Showing error
Screenshot from 2024-08-29 18-46-40


fiskus:
Further refactoring of the page into separate cards is in https://github.com/quiltdata/quilt/pull/4126

catalog/app/constants/routes.ts
```diff
@@ -262,11 +262,23 @@ export const admin: Route = {
...
+export const adminBuckets: Route<AdminBucketLegacyArgs> = {
   path: '/admin/buckets',
-  url: (bucket) => `/admin/buckets${mkSearch({ bucket })}`,
+  url: (legacyBucketParam) => `/admin/buckets${mkSearch({ bucket: legacyBucketParam })}`,
+}
+
+export type AdminBucketEditArgs = [bucket: string]
+
+export const adminBucketEdit: Route<AdminBucketEditArgs> = {
+  path: '/admin/buckets/edit/:bucketName',
```
nl0, line 275: this url suggests that there should be something under `/admin/buckets/edit/` route, but there is nothing,
which is confusing

nl0 (dismissed):
i think this is mostly ok

catalog/app/constants/routes.ts
fiskus: Do you have another idea for a route?

Also, just out of my head several examples of such routes:

* https://github.com/quiltdata/quilt/pull/
* https://nightly.quilttest.com/b/allencell/packages/aics/
* https://nightly.quilttest.com/b/

I mean, I don't think this is ideal, but also I don't think it is expected that every "slash" has something

fiskus: > i was thinking about `/buckets/$name` vs `/buckets?add`

Ok, let's do this

nl0 (approved):

nl0 (approved):

PR #4127, opened 2024-09-03
Ignore line-length lint rule for CHANGELOG

fiskus:
Changelog is not a regular prose, and I think it is more readable and easier to add new entries without line-length restriction.

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable line-length -->
 # Changelog
 
 Changes are listed in reverse chronological order (newer entries at the top).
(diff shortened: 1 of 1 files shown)


nl0 (commented):
why? it seems ok with wrapping lines

fiskus:
> Changelog is not a regular prose

So, I don't read it word after word, but scan it line after line. And regular rule for readability (80 letters per line) don't work here for me.

> I think it is more readable

As I said, it's easier to read. I clearly see when every item is ended. When they wrapped, it's a wall of text, that is more difficult to scan.

> easier to add new entries without line-length restriction

Less formatting, less mistakes

nl0:
ok makes sense

nl0 (approved):

PR #4135, opened 2024-09-09
Tabulator, Admin: UI for modifying tabulator configs

fiskus:
* Fix `<TextEditor />` hardcoded height. Make props and look of an error similar to `<M.TextField />`
* `OnDirty` module based on `ReactProvider` and `RF.FormSpy` to observe dirty states outside `RF.Form`
* Use JSON Schema for config validation
* Page structure:
  * Re-usable form fields are in `<*Form />` components: `<PrimaryForm />, `<MetadatForm />` etc.
  * `Add` page has a root `<RF.Form>` and lists all form inside. There is a simple route guard on leaving the page
  * `Edit` page lists all these forms wrapped with `<*Card />` component each: `<PrimaryCard />`, `<MetadataCard />` etc.
  * Card re-uses and shows Form wrapped with `<RF.Form>` each, also adding form buttons via `StickyActions/CardActions`

#### Edit bucket
Edit bucket

#### Add bucket
Add Bucket

#### List tables
Tabulator list

#### Edit/Add tables
Tabulator edit


nl0 (commented):
### Some general nitpicks re: new layout:

1. I'd prefer something like this for the header and first section:

Screenshot 2024-09-24 at 15 30 25

(arrow on the left, then the name of the bucket, since it's not editable, and then a fixed title to avoid excessive jumpiness -- it would translate well to the "add" screen as well)

2. in general, i feel like the layout is too jumpy -- the contents just change significantly without any transition when activating/deactivating a card. this is especially noticeable on the indexing section.

3. indexing section: "Default ..." wording feels confusing. i think we should rather say "File extensions to deep index: ... (default)" or smth like that

4. permissive html rendering: i don't think it makes sense to have buttons and minimized state for this section altogether: just make it an always visible toggle switch.

5. spacing and rounded corners are somewhat off here:

Screenshot 2024-09-24 at 15 39 00

6. maybe it would look better without the margins between sections in collapsed state? like the stock behavior (or just always use some fixed margin idk)

### Tabulator configs

1. i think we should call the section "Tabulation (Longitudinal Querying)" @drernie wdyt?

2. when no tables configured, show "No tables configured" + an "add table" button

3. layout / spacing here seems a little off
Screenshot 2024-09-24 at 15 43 01

also, the resize handles are not working -- we should either hide or fix them

4. we should allow renaming the tables

5. "add config" -> "add table", "config name" -> "table name"

6. i feel like table name should be on top of the config, not the other way around

7. "delete" button looks like it's disabled
Screenshot 2024-09-24 at 15 46 18

8. considering that the configs are edited/saved individually, maybe it's worth it to make the edit mode togglable per table?

fiskus:
I think, I can ditch the whole idea of previewing settings, and just leave the forms. I believe, simplifying by removing unnecessary UI could address major UX concerns

fiskus:
Addressed almost all feedback, thank you. Mostly by removing UI that has UX problems.
Except allowing editing table name. API doesn't support that, though I can delete and create a new table in the background. But I'd rather implement this on the backend.

fiskus:
I think, it's optional. I need @sir-sigurd for the backend work. If we used IDs instead of names and keep the same API for changing name as for changing config, frontend change should be trivial.

nl0 (commented):
looks good.

1. let's not capitalize the bucket name (it was capitalized in my sketch accidentally)
2. maybe make section headers just a little bit larger?
3. i think "reindex and repair" belongs into the indexing section
4. imo "enable permissive html rendering" looks better inside a section as well
5. "no tabulator tables configured" -> "no tables configured", "create table" -> "add table". also maybe it would look better if not centered?
6. not sure it makes much sense to show the longitudinal querying placeholder/hint on the bucket addition screen
7. let's maybe add the table renaming UI so that it's ready by the time it's implemented on the backend?
8. i feel like it might be nicer if table configs were somehow collapsed by default, since they take quite a lot of space expanded

fiskus:
> 1. let's not capitalize the bucket name (it was capitalized in my sketch accidentally)

Digits don't look nice, actually, so I thought, you intentionally capitalized. I'll uncapitalize.

> 2. maybe make section headers just a little bit larger?

I think, they have the right size finally :) But I'm changing them to `h6` for consistency

> 3. i think "reindex and repair" belongs into the indexing section

I agree it should be under the "Indexing and notifications" title. But it shouldn't belong to this form. Placing it inside the form with "Cancel"/"Submit" buttons is a source of frustration, because this action is neither cancellable nor requires submit.
I'm going to play with titles, or place this button on the left side of "Cancel"/"Submit" section.

> 4. imo "enable permissive html rendering" looks better inside a section as well

I'll try to play with the layout, but in general, I don't think it is right to use the same layout for the form that should be submitted and the action that takes place immediately.

> 5. let's maybe add the table renaming UI so that it's ready by the time it's implemented on the backend?

I was expecting API with IDs, so I can just enable table name input. But if it is a standalone request, I keep the UI the same, anyway. I'll call rename method under-the-hood, if table name changed.

> 6. i feel like it might be nicer if table configs were somehow collapsed by default, since they take quite a lot of space expanded

They occupy space, but this space is at the bottom, and nothing else under. So, it's not a big deal. But text editor is heavy, it would be nice to lazy load it. I'll see what can I do without adding much complexity.

nl0:
> > 2. maybe make section headers just a little bit larger?
> 
> I think, they have the right size finally :) But I'm changing them to `h6` for consistency

i just noticed that the headers are smaller than the field labels
 
> I agree it should be under the "Indexing and notifications" title. But it shouldn't belong to this form.

agree

> place this button on the left side of "Cancel"/"Submit" section.

this sounds about right, let's see how it looks/feels in practice

> > 6. i feel like it might be nicer if table configs were somehow collapsed by default, since they take quite a lot of space expanded
> 
> They occupy space, but this space is at the bottom, and nothing else under. So, it's not a big deal. But text editor is heavy, it would be nice to lazy load it. I'll see what can I do without adding much complexity.

cool, thanks

fiskus:
> > > 3\. maybe make section headers just a little bit larger?
> > I think, they have the right size finally :) But I'm changing them to `h6` for consistency
> i just noticed that the headers are smaller than the field labels

Yes. Something like this https://v4.mui.com/components/lists/#nested-list, for example.

fiskus:
I have an idea of a different UI. UI is almost ready, but finishing state and requests work will take time. I'm not sure I can finish it today

Screenshot from 2024-09-30 15-23-04

fiskus:
New tabulator layout:
Screenshot from 2024-10-02 19-35-09
Screenshot from 2024-10-02 19-35-38

shared/graphql/schema.graphql
```diff
@@ -141,6 +146,7 @@ type BucketConfig {
   associatedPolicies: [PolicyBucketPermission!]! @admin
   associatedRoles: [RoleBucketPermission!]! @admin
   collaborators: [CollaboratorBucketConnection!]!
+  tabulatorTables: [TabulatorTable!]! @admin
```
nl0, line 149:
```suggestion
  tabulatorTables: [TabulatorTable!]!
```

this has been changed in the registry

shared/schemas/tabulatorTable.yml.json
```diff
@@ -0,0 +1,94 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://quiltdata.com/federator/config/1",
```
nl0, line 3:
```suggestion
  "$id": "https://quiltdata.com/tabulator/config/1",
```

shared/schemas/tabulatorTable.yml.json
```diff
@@ -0,0 +1,94 @@
...
+            "Timestamp(Millisecond)",
+            "Timestamp(Microsecond)",
+            "Timestamp(Nanosecond)"
+          ]
+        }
+      },
+      "required": ["name", "type"]
+    }
+  },
+  "properties": {
```
nl0, line 73: shouldn't all the top-level props be required?

nl0 (commented):
look good in general, some comments inline

nl0:
this feedback doesn't seem very helpful, tho i'm not sure what we can do about it (use smth like `@effect/schema`?):
Screenshot 2024-10-03 at 14 43 47

catalog/app/containers/Admin/Buckets/Buckets.tsx
```diff
@@ -1493,7 +1420,7 @@ function Reindex({ bucket, open, close }: ReindexProps) {
     try {
       // TODO: use graphql mutation
       await req({
-        endpoint: `/admin/reindex/${bucket}`,
+        endpoint: `/ admin / reindex / ${bucket}`,
```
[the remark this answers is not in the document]
fiskus: Interesting. Probably, I did automatic formatting when syntax was invalid

fiskus:
Errors: 
Screenshot from 2024-10-03 17-30-44

nl0 (approved):
i think this is good to go now. don't forget to revert the workflow change (and wait til the registry pr is in)

fiskus:
Thanks!

shared/schemas/tabulatorTable.yml.json
sir-sigurd, line 1: @nl0 
shouldn't we generate this with pydantic?

nl0:
@fiskus how hard it would be to populate the config editor with some basic boilerplate when adding a new config? smth like:
```yaml
schema:
  - name: column1 # specify the schema
    type: Utf8
source:
  type: quilt-packages
  package_name: "" # specify a RegEx for matching packages
  logical_key: ".*\\.csv$" # specify a RegEx for matching logical keys
parser:
  format: csv # or parquet

fiskus:
> populate the config editor with some basic boilerplate 

Easy, good idea

---

Also, I moved `NameForm` for renaming table to the popup::

* it kinda re-uses UX pattern from renaming user
* but I can't re-use the UI entirely, because list item is clickable

shared/schemas/tabulatorTable.yml.json
fiskus: If that what I could do, I don't know how

nl0 (approved):

PR #4138, opened 2024-09-13
Add note on changelog tags

fiskus:


nl0 (commented):
i don't think it's a rule to be documented, just a common-sense approach to make the message less verbose and more descriptive in some cases

fiskus:
Ok, let it be undocumented rule :)

PR #4146, opened 2024-09-17
Fix `ui.package_description` example

fiskus:
- [x] Documentation
- [x] ~[Changelog](../tree/master/docs/CHANGELOG.md) entry (skip if change is not significant to end users, e.g. docs only)~

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Fixed] Docs: Fix `ui.package_description` example ([#4146](https://github.com/quiltdata/quilt/pull/4146))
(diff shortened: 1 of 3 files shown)


catalog/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Fixed] Docs: Fix `ui.package_description` example ([#4146](https://github.com/quiltdata/quilt/pull/4146))
```
nl0, line 20: i don't think we need a catalog changelog entry for this

fiskus: Do we need a changelog entry somewhere else?

fiskus: Ideally, I would like to have this logged somewhere, because in my opinion, the change is significant.

drernie (approved):

nl0 (approved):

PR #4149, opened 2024-09-19
rewrite SUMMARY (doc reorg)

drernie:
# Description

Reorganize the SUMMARY document to:

1. Refactor content by use case
2. Prioritize Catalog user documentation over developer APIs
3. Match our new product terminology

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -1,72 +1,72 @@
 # Quilt Documentation
 
 * [About Quilt](README.md)
-* [Architecture](Architecture.md)  
+* [Architecture](Architecture.md)
 * [Mental Model](MentalModel.md)
-* [Metadata Management](Catalog/Metadata.md)  
+* [Metadata Management](Catalog/Metadata.md)
 * [Metadata Workflows](advanced-features/workflows.md)
 
 ## Quilt Platform Usage (Catalog UI)
 
-* [About the Catalog](walkthrough/working-with-the-catalog.md)  
-* [Bucket Browsing](Catalog/FileBrowser.md)  
-* [Document Previews](Catalog/Preview.md)  
-* [Embeddable iFrames](Catalog/Embed.md)  
-* [Search & Query](Catalog/SearchQuery.md)  
+* [About the Catalog](walkthrough/working-with-the-catalog.md)
+* [Bucket Browsing](Catalog/FileBrowser.md)
+* [Document Previews](Catalog/Preview.md)
+* [Embeddable iFrames](Catalog/Embed.md)
+* [Search & Query](Catalog/SearchQuery.md)
 * [Visualization & Dashboards](Catalog/VisualizationDashboards.md)
 * **Advanced**
   * [Athena](advanced-features/athena.md)
-  * [Elasticsearch](walkthrough/working-with-elasticsearch.md)  
+  * [Elasticsearch](walkthrough/working-with-elasticsearch.md)
 
 ## Quilt Platform Administration
 
-* [Admin Settings UI](Catalog/Admin.md)  
+* [Admin Settings UI](Catalog/Admin.md)
 * [Catalog Configuration](Catalog/Preferences.md)
-* [Cross-Account Access](CrossAccount.md)  
+* [Cross-Account Access](CrossAccount.md)
 * [Enterprise Installs](technical-reference.md)
 * [quilt3.admin Python API](api-reference/Admin.md)
-* **Advanced**  
-  * [Package Events](advanced-features/package-events.md)  
-  * [Private Endpoints](advanced-features/private-endpoint-access.md)  
-  * [Restrict Access by Bucket Prefix](advanced-features/s3-prefix-permissions.md)  
-  * [S3 Events via EventBridge](EventBridge.md)  
+* **Advanced**
+  * [Package Events](advanced-features/package-events.md)
+  * [Private Endpoints](advanced-features/private-endpoint-access.md)
+  * [Restrict Access by Bucket Prefix](advanced-features/s3-prefix-permissions.md)
+  * [S3 Events via EventBridge](EventBridge.md)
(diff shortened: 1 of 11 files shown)


drernie:
The goal of this PR is simply to fix the SUMMARY without changing the content.
Cleaning up the actual documents to align with this new structure is a follow-on task (suggestions welcome).

QuiltSimon (approved):
Looks like the recommendation we put together. Looks good to me.

docs/SUMMARY.md
```diff
@@ -1,66 +1,72 @@
...
-# Summary
+# Quilt Documentation
 
-* [Introduction](README.md)
-* [Installation](Installation.md)
-* [Quick start](Quickstart.md)
-* [Mental model](MentalModel.md)
-* [Architecture](Architecture.md)
+* [About Quilt](README.md)
+* [Architecture](Architecture.md)
```
akarve, line 4: You've got a few trailing spaces in here and below that are not needed.

docs/SUMMARY.md
```diff
@@ -1,66 +1,72 @@
...
 
-### Walkthrough
-* [Editing a Package](walkthrough/editing-a-package.md)
-* [Uploading a Package](walkthrough/uploading-a-package.md)
-* [Installing a Package](walkthrough/installing-a-package.md)
-* [Getting Data from a Package](walkthrough/getting-data-from-a-package.md)
-* [Working with the Catalog](walkthrough/working-with-the-catalog.md)
-* [Working with a Bucket](walkthrough/working-with-a-bucket.md)
-* [Working with Elasticsearch](walkthrough/working-with-elasticsearch.md)
+## Quilt Platform Usage (Catalog UI)
```
akarve, line 9: Doesn't "Catalog" and not "(Catalog UI)" deserve to be a top-level section under platform? Where do we explain the components of the platform. How do visitors know the API is free forever and the catalog is open source but offers a paid product?

akarve (commented):
This is plausible just need to cleanup spaces, make sure no section links are dead (I spot tested at least 10) and we aren't changing page names since existing names already have SEO. Some inline comments on section names and potentially missing info.

docs/SUMMARY.md
drernie: So, I could use help here.  Marketing-wise, we talk about the 'Quilt Platform'.  But for a user, all they see (and call it) is the "Quilt Catalog."  Which technically is open source, but for 99% of our users is actually the thing they think they are paying for.

My concern is that if we just call it "Platform" users won't know where to look, which is why I added the parenthetical.

Do we want to refactor our documentation so there is a hard split between the open source Catalog, and those features that are only part of the paid product? 

Or should we focus on the "Platform" vs "SDK" split, and let the README explain the nuances?

docs/SUMMARY.md
drernie: Doh! I didn't realize markdown lint didn't warn about double-spaces at the end. Thanks.

docs/SUMMARY.md
drernie: Perhaps call this section: "Quilt Platform User Catalog"?

docs/SUMMARY.md
drernie: Also fixed the broken links, thanks.  Somehow, the file extensions were lost in translation.

drernie: FYI, I tested all the links from the branch, and only found three that were dead.
https://github.com/quiltdata/quilt/blob/doc-reorg/docs/SUMMARY.md

Did I miss something?

docs/SUMMARY.md
drernie: @QuiltSimon What do you think of renaming each section by the people who should use it:

## Quilt Platform Catalog User
## Quilt Platform Administrator
## Quilt Python SDK Developers

A little wordy, but it seems like it would make it easier for people to scan for what they need.
Though that does make this one a bit awkward:
## Quilt Ecosystem Integrations

QuiltSimon: I'm trying to understand the nuances of this discussion. I'm going to bring us back to first principles to get us back on track.

I think of Quilt as being our entire product, Consisting of a Platform (the catalog) an SDK that interacts with the platform (python and CLI) and then ecosystem products that extend Quilt. Id like to see the breakdown as follows

Intro
Quilt Platform
- Users
- Admin
Quilt SDK
- Python
- CLI
Platform Ecosystem
- QuiltSync
- Benchling Packager
- Nextflow Integration (nf-quilt)
- Technology Previews

I believe we all can get aligned on the fact that 

Platform == Catalog
SDK == Python and or CLI

I can be swayed on replacing the word "Platform" from "Platform Ecosystem" to make this more logically consistent if others can suggest a better replacement. Perhaps "Quilt Ecosystem".

Does this help put this issue on firmer ground? I'd be happy to pick up this PR and rearrange the docs to fit the schema I suggest here.

drernie: I agree with the 'Concept' that Platform = Catalog.  The challenges I am concerned about are:

1.  Users think they are using the Catalog. When they come here, how do they know they should be looking under Platform?
2. Are you using "Platform" as synonymous with the paid product?  We (for better and worse) have an open source Catalog.  We don't need to talk about it much, but we do need to be able to refer to it without causing extra confusion.

I'm fine with you revising (or forking) this PR to help us see what you mean.

QuiltSimon:
@drernie I'm suggesting a new readme style here, which better defines each of the sections and the terminology. Take a look and let me know what you think.

docs/SUMMARY.md
drernie: Rewrite looks good.

docs/SUMMARY.md
drernie: Triple-checked links, look good.

drernie:
@QuiltSimon Looks good to me; I just made some formatting cleanup.  Approve when you are ready, and it will auto-merge.

QuiltSimon (approved):
lgtm

PR #4153, opened 2024-09-24
Preview markdown files

fiskus:
[markdown-preview.webm](https://github.com/user-attachments/assets/d9382d36-5ce9-4c32-ad41-02a1fd9df09c)

# TODO

- [ ] Unit tests
- [ ] Documentation

### DIFF
--- a/catalog/app/components/FileEditor/Controls.tsx
+++ b/catalog/app/components/FileEditor/Controls.tsx
@@ -16,6 +16,51 @@ export function AddFileButton({ onClick }: AddFileButtonProps) {
   )
 }
 
+interface PreviewButtonProps extends EditorState {
+  className?: string
+}
+
+export function PreviewButton({ className, preview, onPreview }: PreviewButtonProps) {
+  const handleClick = React.useCallback(
+    (event) => {
+      if (!onPreview) {
+        // eslint-disable-next-line no-console
+        console.error('Unsupported preview')
+        return
+      }
+      event.stopPropagation()
+      onPreview(!preview)
+    },
+    [onPreview, preview],
+  )
+  if (!onPreview) return null
+  if (preview) {
+    return (
+      <M.Button
+        className={className}
+        color="primary"
+        size="small"
+        variant="contained"
+        onClick={handleClick}
+      >
+        Continue writing
+      </M.Button>
+    )
+  }
+  return (
+    <M.Button
+      className={className}
+      color="default"
+      size="small"
+      startIcon={<M.Icon fontSize="inherit">check_box_outline_blank</M.Icon>}
+      variant="outlined"
+      onClick={handleClick}
+    >
+      Preview
+    </M.Button>
+  )
+}
+
 interface ControlsProps extends EditorState {
   className?: string
 }
--- a/catalog/app/components/FileEditor/FileEditor.tsx
+++ b/catalog/app/components/FileEditor/FileEditor.tsx
@@ -1,5 +1,8 @@
+import cx from 'classnames'
 import * as React from 'react'
+import * as M from '@material-ui/core'
 
+import Markdown from 'components/Markdown'
 import * as PreviewUtils from 'components/Preview/loaders/utils'
 import PreviewDisplay from 'components/Preview/Display'
 import type * as Model from 'model'
@@ -82,10 +85,26 @@ function EditorSuspended({
   })
 }
 
+const useStyles = M.makeStyles({
+  tab: {
+    display: 'none',
+    width: '100%',
+  },
+  active: {
+    display: 'block',
+  },
+})
+
 export function Editor(props: EditorProps) {
+  const classes = useStyles()
   return (
     <React.Suspense fallback={<Skeleton />}>
-      <EditorSuspended {...props} />
+      <div className={cx(classes.tab, { [classes.active]: !props.preview })}>
+        <EditorSuspended {...props} />
+      </div>
+      <div className={cx(classes.tab, { [classes.active]: props.preview })}>
+        <Markdown data={props.value} />
+      </div>
     </React.Suspense>
   )
 }
--- a/catalog/app/components/FileEditor/State.tsx
+++ b/catalog/app/components/FileEditor/State.tsx
@@ -32,7 +32,9 @@ export interface EditorState {
   onCancel: () => void
   onChange: (value: string) => void
   onEdit: (type: EditorInputType | null) => void
+  onPreview: ((p: boolean) => void) | null
   onSave: () => Promise<Model.S3File | void>
+  preview: boolean
   saving: boolean
   types: EditorInputType[]
   value?: string
@@ -48,6 +50,7 @@ export function useState(handle: Model.S3.S3ObjectLocation): EditorState {
   const [editing, setEditing] = React.useState<EditorInputType | null>(
     edit ? types[0] : null,
   )
+  const [preview, setPreview] = React.useState<boolean>(false)
   const [saving, setSaving] = React.useState<boolean>(false)
   const writeFile = useWriteData(handle)
   const redirect = useRedirect()
@@ -80,11 +83,13 @@ export function useState(handle: Model.S3.S3ObjectLocation): EditorState {
       onCancel,
       onChange: setValue,
       onEdit: setEditing,
+      onPreview: editing?.brace === 'markdown' ? setPreview : null,
       onSave,
+      preview,
       saving,
       types,
       value,
     }),
-    [editing, error, onCancel, onSave, saving, types, value],
+    [editing, error, onCancel, onSave, preview, saving, types, value],
   )
 }
--- a/catalog/app/containers/Bucket/File.js
+++ b/catalog/app/containers/Bucket/File.js
@@ -310,6 +310,15 @@ const useStyles = M.makeStyles((t) => ({
     marginBottom: t.spacing(2),
     flexWrap: 'wrap',
   },
+  editTitle: {
+    display: 'flex',
+    alignItems: 'center',
+    flexGrow: 1,
+  },
+  editButton: {
+    marginLeft: 'auto',
+    marginRight: '8px',
+  },
   preview: {
     width: '100%',
   },
@@ -531,7 +540,21 @@ export default function File() {
                 prefs,
               )}
               {editorState.editing ? (
-                <Section icon="text_fields" heading="Edit content" defaultExpanded>
+                <Section
+                  icon="text_fields"
+                  heading={({ expanded }) => (
+                    <div className={classes.editTitle}>
+                      Edit content
+                      {expanded && editorState.onPreview && (
+                        <FileEditor.PreviewButton
+                          className={classes.editButton}
+                          {...editorState}
+                        />
+                      )}
+                    </div>
+                  )}
+                  defaultExpanded
+                >
                   <FileEditor.Editor {...editorState} handle={handle} />
                 </Section>
               ) : (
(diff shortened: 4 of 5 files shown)


fiskus:
Worth mention https://github.com/quiltdata/quilt/pull/3310

catalog/app/components/Preview/quick/Markdown/Render.tsx
```diff
@@ -0,0 +1,28 @@
...
+  value: string
+  handle: Model.S3.S3ObjectLocation
+}
+
+export default function Render({ value, handle }: RenderProps) {
+  const result = useMarkdownRenderer(AsyncResult.Ok(value), handle)
+  return AsyncResult.case({
+    _: () => null,
+    Err: (error: Error) => <Lab.Alert severity="error">{error.message}</Lab.Alert>,
+    Pending: () => <Skeleton />,
```
nl0, line 21: it's probably better to just show skeleton on `_` (combining Pending and Init)

catalog/app/components/Preview/quick/Markdown/Skeleton.tsx
```diff
@@ -0,0 +1,23 @@
...
+const useSkeletonStyles = M.makeStyles((t) => ({
+  line: {
+    height: t.spacing(3),
+    marginBottom: t.spacing(1),
+  },
+}))
+
+export default function Skeleton() {
+  const classes = useSkeletonStyles()
+  const lines = [80, 50, 100, 60, 30, 80, 50, 100, 60, 30, 20, 70]
```
nl0, line 15: you can move this out of component to avoid allocating/computing this on every render

catalog/app/components/Preview/quick/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+type HandledType = HandledNone | HandledMarkdown // Json | Yaml
+
+function convertToTypeUnion(type: FileEditor.EditorInputType | null): HandledType {
+  if (!type) return { tag: 'none' }
+  switch (type.brace) {
+    case 'markdown':
+      return { tag: 'markdown' }
+    default:
+      return { tag: 'none' }
+  }
```
nl0, line 35:
```suggestion
  switch (type?.brace) {
    case 'markdown':
      return { tag: 'markdown' }
    default:
      return { tag: 'none' }
  }
```

nl0 (commented):
some nitpicks from me and a potential bug from copilot

catalog/app/components/FileEditor/Controls.tsx
```diff
@@ -16,6 +16,32 @@ export function AddFileButton({ onClick }: AddFileButtonProps) {
...
+
+export function PreviewButton({ className, preview, onPreview }: PreviewButtonProps) {
+  const handleClick = React.useCallback(() => onPreview(!preview), [onPreview, preview])
+  return (
+    <M.FormControlLabel
+      onClick={(event) => event.stopPropagation()}
+      className={className}
+      control={
+        <M.Switch
+          onClick={stop}
```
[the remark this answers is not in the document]
nl0: i wonder why it's not caught by the type checker

nl0:
also, maybe 'back to editing' or 'continue editing' instead of 'continue writing'?

catalog/app/components/Preview/quick/index.tsx
fiskus: Technically correct, but I think this variant is less clear

catalog/app/components/Preview/quick/Markdown/Skeleton.tsx
fiskus: Ok, but probably it makes sense to compute it on render only (only when the quick preview is toggled), otherwise it is computed on every Bucket page even without Markdown.

fiskus:
> also, maybe 'back to editing' or 'continue editing' instead of 'continue writing'?

The most recent UI is on screenshot: switch with "Preview" label. So, the preview is whether enabled or not. Instead of buttons with action verbs

catalog/app/components/Preview/loaders/Markdown.js
```diff
@@ -66,21 +65,29 @@ function useLinkProcessor(handle) {
...
-    [processImg, processLink],
-  )
+  const contents = AsyncResult.mapCase({
+    Ok: (r) => r.Body.toString('utf-8'),
+  })(data.result)
+  const markdowned = useMarkdownRenderer(contents, handle)
+  const processed = AsyncResult.mapCase({
+    Ok: (rendered) =>
+      PreviewData.Markdown({ rendered, modes: [FileType.Markdown, FileType.Text] }),
+  })(markdowned)
```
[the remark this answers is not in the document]
fiskus: Thanks

catalog/app/components/Preview/quick/Markdown/Render.tsx
fiskus: Actually, `Pending` is redundant, because we always pass `AsyncResult.Ok`

But if there was a real request, it seems like a shortcut to combine `Init` and `Pending`.
Practically, `Init` is usually followed by `Pending` in requests. But what if someone put `noAutoFetch` somewhere in requests, and request will stack in `Init`? Probably, it would not be the desired UI state: showing skeleton when request is not pending.

catalog/app/components/FileEditor/Controls.tsx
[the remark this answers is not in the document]
nl0: are sure we need this?

catalog/app/components/Preview/quick/Markdown/Render.tsx
nl0: > it seems like a shortcut to combine Init and Pending.

well, exactly

> But what if someone put noAutoFetch somewhere in requests, and request will stack in Init? Probably, it would not be the desired UI state: showing skeleton when request is not pending.

well, that depends on the context. if you want to handle "init" state in a special way, e.g. show a distinct message or a special control -- that's ok. but if you know that there's no practical gain in handling those individually, then it's ok to combine them, and it's definitely better than just showing nothing (returning null).

ofc it's also ok to return null for any of those if you're absolutely sure they are not used, tho i'd prefer to use smth like `absurd` / `assertNever` in such cases to explicitly highlight the expected invariant

catalog/app/components/Preview/quick/Markdown/Skeleton.tsx
nl0: if you move it out of the component, it will be allocated/computed once during the first time module is imported. that's far better that doing it on every render (tho in this particular case the perf impact is negligible ofc)

catalog/app/components/Preview/quick/index.tsx
nl0: 🤷

catalog/app/components/FileEditor/Controls.tsx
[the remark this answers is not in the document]
fiskus: I deleted this line, and added the rule for eslint to catch this case next time

https://github.com/quiltdata/quilt/pull/4153/commits/1e6baeb63a0e7d5f9fc76989342110b1cd85a9ed

nl0:
> Upd: screencast updated

ok i see the new ui

tho i saw "continue writing" string in the code -- is it used for smth else?

catalog/app/components/FileEditor/Controls.tsx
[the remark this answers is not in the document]
nl0: yeah just noticed that

nl0:
@fiskus should i review/approve this now or are you going to work on it further?

fiskus:
> tho i saw "continue writing" string in the code -- is it used for smth else?

I replaced "continue writing" button with Switch on 15th of November https://github.com/quiltdata/quilt/pull/4153/commits/ff5c00d67b68064330b622f40119739c3305a433, and I can't find it in the code now

catalog/app/components/Preview/quick/Markdown/Skeleton.tsx
fiskus: I'm ok with both variants. But I don't think one way is far better than the other, because on the one hand we have array on every Bucket page, and on another only for those who edit markdown files and preview them with "Preview" switch. And skeleton renders for a short amount of time when the js module is loaded.

nl0:
> > tho i saw "continue writing" string in the code -- is it used for smth else?
> 
> I replaced "continue writing" button with Switch on 15th of November [ff5c00d](https://github.com/quiltdata/quilt/commit/ff5c00d67b68064330b622f40119739c3305a433), and I can't find it in the code now

ok, tho i was sure i saw it today while expanding the diff 🤷

nl0 (dismissed):

catalog/app/components/Preview/quick/Markdown/Skeleton.tsx
fiskus: Yes, on every bucket page, loaded once. It doesn't recompute on navigation inside SPA.
So, if we have 1000 customers that open bucket page once, we have 1000 invocations.
How much of these 1000 customers will ever use "Preview", I don't know.

If that wasn't a tiny array, but some heavy computation, costing 0.5 sec to process, would it be better to load every time Bucket page is loaded, or delay Preview function once for 0.5 sec?

nl0: i think you're missing the point, it's not about delaying the Preview or Bucket components, it's about unnecessarily executing the same computation over and over.
especially if it was a heavy computation, you should've put that outside of the render function (or inside a useMemo at least), bc it runs on every render, and you don't control how many times react calls that render.
so, if we really cared about perf here, we should've made that initialization **lazy** and **memoized**, so that it's computed once and only on-demand

fiskus: There are unnecessary executions in both variants. In the first variant, execution happens for everyone opening the page, the second is somewhat heavier but happens only in rare cases and interim. Why is first variant is superior?

```
// contributes to everyone

const once = [1, 2, 3]

function RareCase1() {
  return once
}

---

// contributes more heavily but only in rare cases

function RareCase2() {
  const everytime = [1, 2, 3]
  return everytime
}
```

nl0 (approved):

nl0:
@fiskus are we going to release this today?

fiskus:
I would wait and release next time. I'm not comfortable to release something in the last minute. Especially, if I made some changes in the PR.
So, I'm going to merge it after the release.

catalog/app/components/Preview/quick/Markdown/Skeleton.tsx
nl0: you allocate a constant once (deterministically) and then you use it further in the code (or you don't, but that doesn't matter that much, bc the perf impact of doing that during module loading/init is negligible in the grand scheme of things, given there are other things, e.g. functions in that module, that are being initialized as well).

in the second approach, you allocate that array again and again on every render.
i repeat: it doesn't matter much in this particular case, but it would be very undesirable for heavier computations

fiskus: You’re right that React can re-render a component any number of times, especially if it has state or props. In this case, the component doesn't have props and includes state via the Material-UI `useStyles` hook. While these hooks are usually memoized, re-renders are still technically possible.

That said, my point was based on practical observations. In this specific usage, the component renders only once. I didn’t mean it will always behave this way, just that it does in this context.

With that in mind, and for clarification, when you say “agree on all the other points,” does that include the point that putting the array outside the component is not inherently 100% better than keeping it inside? Specifically, considering this component’s current usage and characteristics, both approaches have trade-offs, and one is not definitively superior.

nl0 (dismissed):

nl0 (approved):

PR #4155, opened 2024-09-26
Config json key

drernie:
# Description

Updated CONFIG_TEMPLATE to expose more configuration keys inside QuiltConfig

# TODO

- [X] Unit tests

### DIFF
--- a/api/python/quilt3/util.py
+++ b/api/python/quilt3/util.py
@@ -81,9 +81,6 @@ binaryApiGatewayEndpoint:
 
 default_registry_version: 1
 
-""".format(BASE_PATH.as_uri() + '/packages')
-
-CONFIG_EXTRA = """
 # AWS Region
 region:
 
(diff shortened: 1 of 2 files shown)


drernie:
@sir-sigurd  Is there a good reason (e.g., security) to not expose all these keys?
How much detail should I put in the CHANGELOG?

sir-sigurd:
> @sir-sigurd Is there a good reason (e.g., security) to not expose all these keys? 

I don't think it affects security because these values already can be accessed directly
though I don't think we should add anything beside region because other values would be unused and a lot of them don't make sense in context of quilt3

I think we should add public API to make boto3 session i.e.

```python
def get_boto3_session():
    return boto3.Session(botocore_session=quilt3.session.create_botocore_session(), region_name=quilt3.config()["region"])
```

drernie:
@sir-sigurd Is this better?

drernie:
@sir-sigurd Can you review and/or fix, so we can merge in time for this release?

drernie:
This is obsolete, right?

PR #4164, opened 2024-10-09
Update tabulator config schema

sir-sigurd:
# Description

This updates tabulator config schema with one generated by pydantic.

# TODO

### DIFF
--- a/shared/schemas/tabulatorTable.yml.json
+++ b/shared/schemas/tabulatorTable.yml.json
@@ -1,85 +1,27 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://quiltdata.com/tabulator/config/1",
+  "title": "Config",
   "type": "object",
-  "definitions": {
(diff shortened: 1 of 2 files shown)


shared/schemas/tabulatorTable.yml.json
```diff
@@ -1,94 +1,155 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://quiltdata.com/tabulator/config/1",
```
sir-sigurd, line 3: @fiskus 
do we need these to have it working?

nl0: let's add that back just in case

nl0 (approved):

shared/schemas/tabulatorTable.yml.json
fiskus: I think, it's not required, but I'm not sure

nl0 (approved):

PR #4168, opened 2024-10-10
Cleanup CHANGELOG.md

sir-sigurd:
This is mostly to remove empty section from the previous release, but also decided to cleanup it with markdownlint


fiskus (approved):
Maybe, makes sense to ignore default `no-duplicate-heading` in Changelog only

sir-sigurd:
> > Maybe, makes sense to ignore default `no-duplicate-heading` in Changelog only
> 
> anchors seem to be generated correctly both in GitBook and GitHub, so I don't see a reason for this also I haven't found how to set that parameter in the file

OK, I found `markdownlint-configure-file` directive, but still IMO it should configured globally.

nl0 (approved):

PR #4169, opened 2024-10-10
add quilt3.get_boto3_session() function

sir-sigurd:
# Description

This provides public API for creation of boto3 session with stack credentials/region.
This also should reduce number of S3 requests (1 per non us-east-1 bucket (assuming buckets are in the stack region)) 🚀.

# TODO

- [x] Unit tests
- [x] Documentation


api/python/quilt3/session.py
```diff
@@ -297,3 +298,20 @@ def create_botocore_session():
...
+
+def get_boto3_session() -> boto3.Session:
+    """
+    Return a Boto3 session with Quilt credentials.
+    If no Quilt credentials are found, return a session that uses normal credentials provider chain.
+
+    > Note: you need to call `quilt3.config("https://your-catalog-homepage/")` to have region set,
+    if you previously called it in quilt3 < 6.1.0.
+    """
+    # XXX: maybe it makes more sense to return None if
```
nl0, line 311: yeah i'd do this (or maybe add an optional argument to fallback to the "system" session?)

nl0 (approved):

api/python/quilt3/data_transfer.py
```diff
@@ -151,10 +151,8 @@ def find_correct_client(self, api_type, bucket, param_dict):
                 else:
                     raise S3NoValidClientError(f"S3 AccessDenied for {api_type} on bucket: {bucket}")
 
-    def get_boto_session(self):
-        botocore_session = create_botocore_session()
-        boto_session = boto3.Session(botocore_session=botocore_session)
-        return boto_session
+    def get_boto_session(self) -> boto3.Session:
+        return get_boto3_session() or boto3.Session()
```
sir-sigurd, line 155: @nl0 
it feels like it's not right to have fallback logic here
I don't quite like adding `fallback` parameter to `get_boto3_session()` because it would have `Optional[]` as return type and you have to check `assert session is not None` even if `fallback=True` (though maybe some type annotations magic exists to handle that)
maybe it's not useful to have `get_boto3_session()` returning `None` after all? 🤔

api/python/quilt3/session.py
```diff
@@ -288,14 +290,34 @@ def load(self):
...
+    if you previously called it in quilt3 < 6.1.0.
+    """
+    if not (credentials := _load_credentials()):
+        if fallback:
+            return boto3.Session()
+        raise QuiltException("No Quilt credentials found.")
+    return boto3.Session(
+        botocore_session=create_botocore_session(credentials=credentials),
+        region_name=get_from_config("region"),
+    )
```
nl0, line 323: i'd swap the branches to 1) avoid negation in the condition, 2) make things less nested

```suggestion
    if credentials := _load_credentials():
        session = create_botocore_session(credentials=credentials)
        return boto3.Session(botocore_session=session, region_name=get_from_config("region"))

    if fallback:
        return boto3.Session()

    raise QuiltException("No Quilt credentials found.")
```

nl0 (approved):
lgtm

docs/api-reference/api.md
```diff
@@ -43,6 +43,16 @@ __Arguments__
 * __top_hash (str)__:  Optional. A package hash to delete, instead of the whole package.
 
 
+## get\_boto3\_session(\*, fallback: bool = True) -> boto3.session.Session  {#get\_boto3\_session}
+
+Return a Boto3 session with Quilt stack credentials and AWS region.
+In case of no Quilt credentials found, return a "normal" Boto3 session if `fallback` is `True`,
```
nl0, line 49:
```suggestion
In case if no Quilt credentials found, return a "normal" Boto3 session if `fallback` is `True`,
```

api/python/quilt3/session.py
sir-sigurd: I like to handle "special" cases first

docs/api-reference/api.md
sir-sigurd: AFAICT "In case of" is correct English (actually it's written by copilot)
default value is specified in signature

PR #4173, opened 2024-10-10
Select files and directories in a package to download

fiskus:
Backend: https://github.com/quiltdata/enterprise/pull/779

Screenshot from 2024-10-11 17-22-23
Screenshot from 2024-10-11 17-22-51

Also, some refactoring and fixing minor bugs:

* `SelectionItem { logicalKey: string }` type instead of a `string`
* Better aligning (fixing MUI bug)
* Re-use components


catalog/app/containers/Bucket/Meta.tsx
```diff
@@ -81,6 +81,7 @@ export function PackageMetaSection({ meta, preferences }: PackageMetaSectionProp
   const classes = usePackageMetaStyles()
   if (!meta || R.isEmpty(meta)) return null
   const { message, user_meta: userMeta, workflow } = meta
+  if (!message && !userMeta && !workflow) return null
```
nl0, line 84: i think you should remove the `R.isEmpty(meta)` check above if you're adding this

catalog/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Added] Select files and directories in packages to download ([#4173](https://github.com/quiltdata/quilt/pull/4173))
```
nl0, line 20:
```suggestion
- [Added] Selective package downloading ([#4173](https://github.com/quiltdata/quilt/pull/4173))
```
maybe sth like this?

nl0 (approved):
looks reasonable

catalog/CHANGELOG.md
fiskus: I started with this, but tried to find more specific and simpler phrasing. However, using verbs in my variant is hanging in the air.

catalog/app/containers/Bucket/Meta.tsx
fiskus: Oh, thanks

fiskus:
Used `React.Provider` for passing selection data instead of `props`. It was required to use `utils/Dialog`, because it is initialized in callback and doesn't react to `props` changes.

Also, fixed route guarding: confirming navigation from package page when something is selected. The condition turned out to be quite sprawling: https://github.com/quiltdata/quilt/pull/4173/files#diff-95c902495a06ff31c5e3a2bede51af727048f9151bede5b5195a4f2cb8fc5090R106

nl0 (approved):

PR #4191, opened 2024-10-15
Test quilt3 on Python 3.13

sir-sigurd:


sir-sigurd:
currently fails because there are no Python 3.13 wheels for pyarrow

fiskus (approved):

PR #4194, opened 2024-10-17
Athena: Add messages to skeletons

fiskus:
[simplescreenrecorder-2024-10-23_12.13.21.webm](https://github.com/user-attachments/assets/8e365d5e-f9b7-495c-b471-047bafe56b81)

### DIFF
--- a/catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
+++ b/catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
@@ -19,6 +19,63 @@ import Results from './Results'
 import History from './History'
 import Workgroups from './Workgroups'
 
+const useRelieveMessageStyles = M.makeStyles((t) => ({
+  root: {
+    padding: t.spacing(2),
+  },
+  text: {
+    animation: '$show 0.3s ease-out',
+  },
+  '@keyframes show': {
+    from: {
+      opacity: 0,
+    },
+    to: {
+      opacity: 1,
+    },
+  },
+}))
+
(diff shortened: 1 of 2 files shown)


catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
```diff
@@ -19,6 +19,63 @@ import Results from './Results'
...
+}
+
+function RelieveMessage({ className, messages }: RelieveMessageProps) {
+  const classes = useRelieveMessageStyles()
+  const [relieve, setRelieve] = React.useState('')
+  const timersData = React.useMemo(
+    () =>
+      messages.map((message, index) => {
+        const increasedTimeout = RELIEVE_INITIAL_TIMEOUT * (index + 1) ** 2
+        const randomShift = (Math.random() * increasedTimeout) / 2
```
nl0, line 53: if i read this correctly, there's no guarantee that the series of timeouts will be ascending (due to randomness).
it doesn't seem like big deal, but i think it would be somewhat more "correct" to just offset every new timeout by the length of the previous one

fiskus: I think timeouts are always ascending. Every next timeout is bigger than a random shift, that added to the timeout.
The random shift is needed only because there are two popup messages, and it looks weird when they are changing synchronously.

But in a new PR I replaced skeleton with faded input element, and I left only one "relieving" message, so random shift is no longer needed.

fiskus:
Superseeded by #4208

PR #4202, opened 2024-10-24
access counts gql

nl0:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation

### DIFF
--- a/shared/graphql/schema.graphql
+++ b/shared/graphql/schema.graphql
@@ -215,11 +215,22 @@ type User {
 type AccessCountForDate {
   date: Datetime!
   value: Int!
+  sum: Int!  # running sum
 }
 
 type AccessCounts {
   total: Int!
   counts: [AccessCountForDate!]!
+}
+
+type AccessCountsGroup {
+  ext: String!
+  counts: AccessCounts!
+}
+
+type BucketAccessCounts {
+  byExt(groups: Int): AccessCountsGroup[]
+  combined: AccessCounts
 }
 
 type PackageDir {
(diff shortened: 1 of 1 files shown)


shared/graphql/schema.graphql
```diff
@@ -215,13 +215,24 @@ type User {
 type AccessCountForDate {
   date: Datetime!
   value: Int!
+  sum: Int!  # running sum
```
nl0, line 218: not sure this is required, since it can be easily calculated on the consuming side, but it feels quite convenient to have this

shared/graphql/schema.graphql
```diff
@@ -215,13 +215,24 @@ type User {
...
   counts: [AccessCountForDate!]!
 }
 
+type AccessCountsGroup {
+  ext: String!
+  counts: AccessCounts!
+}
+
+type BucketAccessCounts {
+  byExt(groups: Int): AccessCountsGroup[]
```
nl0, line 232: don't merge / collapse the groups by default
(maybe we don't have to do that at all, just keep that logic on the front-end, not sure)

shared/graphql/schema.graphql
```diff
@@ -556,6 +567,9 @@ type Query {
   searchMorePackages(after: String!, size: Int = 30): PackagesSearchMoreResult!
   subscription: SubscriptionState!
 
+  bucketAccessCounts(bucket: String!, window: Int!): BucketAccessCounts
+  objectAccessCounts(bucket: String!, key: String!, window: Int!): AccessCounts
```
nl0, line 571: we don't have a proper "bucket" entry point currently, and i don't feel comfortable designing / adding that now, so let's just add separate queries.
tho feel free to add a simple container type to group this if you like

shared/graphql/schema.graphql
sir-sigurd: I'm not quite sure what you mean

nl0: [catalog code](https://github.com/quiltdata/quilt/blob/master/catalog/app/containers/Bucket/requests/requestsUntyped.js#L76) "collapses" access counts into a finite number of "bands", so if there's too many extensions, less frequent ones are merged into a single "other" band.
this proposed api design assumes there's no such collapsing performed by default (when `groups` are null or zero)

sir-sigurd:
I guess this can be closed now

PR #4207, opened 2024-11-05
Fix #4206: allow root folders named '/'

### LINKED ISSUE #4206: Allow root folders named '/'
Not recommended, but totally possible to create them.

% aws s3 cp README.md s3://example-bucket//       
upload: ./README.md to s3://example-bucket//README.md   
ernest@c-208-82-100-121 nf-quilt % python
Python 3.12.4 (main, Jun 25 2024, 14:04:15) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import quilt3 as q3
>>> p = q3.Package()
>>> p.set_dir(".","s3://example-bucket//")
  File "<stdin>", line 1, in <module>
  File "/Users/ernest/GitHub/quilt/api/python/quilt3/packages.py", line 885, in set_dir
    src = PhysicalKey.from_url(fix_url(path))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ernest/GitHub/quilt/api/python/quilt3/util.py", line 165, in from_url
    return cls(bucket, path, version_id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ernest/GitHub/quilt/api/python/quilt3/util.py", line 143, in __init__
    assert not path.startswith('/'), "S3 paths must not start with '/'"
           ^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: S3 paths must not start with '/'

drernie:
# Description

set_dir: Allow '/' folder in root.


drernie:
We have a customer with a lot of data in a path like "s3://bucket//raw".  I presume the check was in place to prevent accidentally entering an absolute path. Should we drop it entirely, or provide a warning?

api/python/quilt3/util.py
```diff
@@ -140,7 +144,9 @@ def __init__(self, bucket, path, version_id):
             if os.name == 'nt':
                 assert '\\' not in path, "Paths must use / as a separator"
         else:
-            assert not path.startswith('/'), "S3 paths must not start with '/'"
+            if path.startswith('/'):
+                # S3 paths are not absolute, but there may be a "/" folder in root.
+                logger.warning("S3 paths should not start with '/': %r", path)
```
sir-sigurd, line 149: do we really need this warning given it's seems to be a valid case?
I think we had that assert to make sure we parse S3 URIs in `from_url()` correctly, but now that seems too defensive

api/python/quilt3/util.py
```diff
@@ -140,7 +144,9 @@ def __init__(self, bucket, path, version_id):
             if os.name == 'nt':
                 assert '\\' not in path, "Paths must use / as a separator"
         else:
-            assert not path.startswith('/'), "S3 paths must not start with '/'"
+            if path.startswith('/'):
+                # S3 paths are not absolute, but there may be a "/" folder in root.
```
sir-sigurd, line 148: this comment makes my brain hurt a lot because of leaky abstractions (S3 don't really have paths, it has object keys)

sir-sigurd (commented):
generally seems right

needs tests

nl0 (commented):
looks reasonable in general. i defer to @sir-sigurd on this one

api/python/quilt3/util.py
```diff
@@ -511,7 +509,7 @@ def validate_key(key):
         )
 
     for part in key.split('/'):
-        if part in ('', '.', '..'):
+        if part in ('.', '..'):
```
sir-sigurd, line 512: why did you do that?
that seems wrong

api/python/tests/integration/test_packages.py
```diff
@@ -623,6 +623,8 @@ def test_s3_set_dir(self):
 
             list_object_versions_mock.assert_called_with('bucket', 'foo/')
 
+            pkg.set_dir('bar', 's3://bucket//foo')  # top-level '/' folder
```
sir-sigurd, line 626: you should add a separate test case for that

api/python/quilt3/util.py
drernie: Because the (now valid) path '//foo' splits to ['', 'foo'], and causes an error.

To be fair, that was only an error for one of the two occurences. I will revert the other.

api/python/tests/integration/test_packages.py
drernie: test_set_dir_root_folder_named_slash

api/python/quilt3/util.py
drernie: My bad. Reverted.

api/python/tests/integration/test_packages.py
```diff
@@ -623,6 +623,15 @@ def test_s3_set_dir(self):
...
             list_object_versions_mock.assert_called_with('bucket', 'foo/')
 
+    def test_set_dir_root_folder_named_slash(self):
+        with patch('quilt3.packages.list_object_versions') as list_object_versions_mock:
+            list_object_versions_mock.return_value = ([
+                dict(Key='/foo/a.txt', VersionId='xyz', IsLatest=True, Size=10),
+            ], [])
+            pkg = Package()
+            pkg.set_dir('bar', 's3://bucket//foo')  # top-level '/' folder
+            assert True, "No exception raised for top-level folder named '/'"
```
sir-sigurd, line 633: that's not enough
at least you need to check physical key, because you could for example just strip leading slash and it won't raise exception but physical key would be wrong
I suggest check the whole entry like [here](https://github.com/quiltdata/quilt/blob/b0c1e92fb4426c4eab37e3f624dccd211a9cca13/api/python/tests/integration/test_packages.py#L620-L622)

api/python/quilt3/data_transfer.py
```diff
@@ -876,7 +876,7 @@ def copy_file(src: PhysicalKey, dest: PhysicalKey, size=None, message=None, call
     """
     def sanity_check(rel_path):
         for part in rel_path.split('/'):
-            if part in ('', '.', '..'):
+            if part in ('.', '..'):
```
sir-sigurd, line 879: is this really needed to have `set_dir()` working?
I don't think so, as I understand it's only needed for file uploading/copying
I think that change makes some sense for src (maybe we even shouldn't check src at all), but I'm not sure it's a good idea to allow empty path segments in dst
if we go with this change it needs to be tested and to be reflected in changelog and PR title
might make sense to move it to a separate PR

api/python/tests/integration/test_packages.py
sir-sigurd: also you should call `list_object_versions_mock.assert_called_once_with()`

api/python/quilt3/data_transfer.py
drernie: Huh, I thought it was. Maybe that was a side effect of a malformed test. My bad.

api/python/tests/integration/test_packages.py
drernie: Ouch, good catch. Looks like `urlunparse` automatically strips the 'invalid' leading '/'.
This is the simplest way I found to detect and revert it. Any better suggestions?
```
            result = urlunparse(('s3', self.bucket, quote(self.path), None, urlencode(params), None))
            missing_slash = f"s3://{self.bucket}{quote(self.path)}"
            if self.path and result.startswith(missing_slash):
                result = result.replace(missing_slash, f"s3://{self.bucket}/{quote(self.path)}")
            return result
```

sir-sigurd: > Looks like urlunparse automatically strips the 'invalid' leading '/'.

it doesn't, it appears to add missing slash instead and we pass path without leading slash because we remove it earlier here
https://github.com/quiltdata/quilt/blob/b6c708faf6a3bc09c886bcddccf78198b3c85f45/api/python/quilt3/util.py#L156


so it looks like you should simply always add leading slash

docs/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ Entries inside each section should be ordered by type:
 ## CLI
 !-->
 
+# unreleased - YYYY-MM-DD
+
+## Python API
+
+* [Fixed] Allow S3 paths starting with `/` in `Package().set_dir()` ([#4207](https://github.com/quiltdata/quilt/pull/4207))
```
sir-sigurd, line 20: I think this fixes `Package.set()` as well
if so might make sense to add test for it as well

drernie: You're right!

api/python/tests/integration/test_packages.py
drernie: I _think_ line 156 is because "urlparse always returns absolute paths" , as otherwise `self.path` wouldn't have a leading slash in my case.

At any rate, if you can think of a better way to fix the error, feel free to improve the code and merge it.

sir-sigurd: > I think line 156 is because "urlparse always returns absolute paths"

and we remove single leading slash so it become S3 key

> as otherwise self.path wouldn't have a leading slash in my case.

in your case there are *two* leading slashes in path, so after removing one you still have another

> if you can think of a better way to fix the error

I suggested one above

> feel free to improve the code and merge it.

okay

nl0 (approved):
lgtm

PR #4208, opened 2024-11-07
Athena: refactoring and minor UI/UX improvements (alphabetize lists, better loading/error visuals)

fiskus:
The main ideas for refactoring are

1. replacing
```tsx
const data = useData()
return <>{data.case{Ok: () => <Element />, Err: () => <Alert /> }}</>
```
with
```tsx
const data = useData()
if (isError(data)) return <Alert />
if (isOk(data)) return <Element />
```

2. use state `Provider` instead of passing data as props

---

Besides refactoring, there are numerous changes and improvements:
* add message from #4194 
* fade input elements during query execution instead of replacing them with skeletons
* fix minor bugs: exception when parsing invalid data, and several minor inconsistencies for error and loading states
* save catalog name and database in local storage and restore it if nothing selected
* alphabetize workgroups, catalog names and databases

---

I suggest to start reading the code from Athena.tsx:
```tsx
  <Model.Provider preferences={ui.athena}>
    <AthenaContainer />
  </Model.Provider>
```
and then into `Model.Provider` → `app/containers/Buckets/Athena/model/state.tsx`

---

image

- [x] Unit tests


nl0 (commented):
a huge undertaking!

some bugs / suggestions:

1. blocking pop-ups on errors encountered while trying to run a query (try "list tables" or some other invalid sql) feel quite odd. i'd much prefer a persistent error message that doesn't block the ui and doesn't disappear

2. getting an error (`can't access property 0, rows[0] is undefined`) while viewing empty results (try "show tables" while a tabulator db is selected for a quick repro). after that the ui is locked in "loading" state (can be worked around by navigating away, e.g. to query executions list)

3. something's wrong with determining whether the result set is complete or truncated. e.g. when i run "show tables" on some quilt-maintained db (e.g. `userathenadatabase-*`), i see all the bucket names in the result list, but the ui displays the truncation indicator ("only showing XX results") and the "load more" button, which, when clicked, loads the same data again and appends it to the table (this can be done multiple (infinite?) amount of times)

4. workgroup selection control looks a little bit off, with some elements sticking out. maybe just use a stock control for that, same as for db / catalog?
Screenshot 2024-11-11 at 11 19 02
Screenshot 2024-11-11 at 11 19 14

5. "query executions" heading seems too small (especially considering that it gets larger when navigated to)
Screenshot 2024-11-11 at 11 21 21

6. it feels somewhat confusing that in query executions table only the completion date is clickable (quite easy to miss imo)

7. query execution ui jumps when expanded / collapsed
<video alt="" src="https://github.com/user-attachments/assets/906cb093-4faa-4fa3-a2d3-0c19bc02ec6c"></video>

8. ux suggestion: a button to insert the query into the editor, not just copy it

9. layout jumps when switching to / from individual executions
<video src="https://github.com/user-attachments/assets/153e104f-c047-4691-a8fe-b0e9246b386f"></video>

i haven't tested saved queries and package creation

fiskus:
> 1.blocking pop-ups on errors encountered while trying to run a query (try "list tables" or some other invalid sql) feel quite odd. i'd much prefer a persistent error message that doesn't block the ui and doesn't disappear

I know how to fix this. And it, probably, even makes the state/data more consistent. Added TODO notes, and going to implement this, if I have time.

> 2. getting an error (can't access property 0, rows[0] is undefined)

Thanks! I thought, I fixed that. Fixed now

> 3. "load more" button...loads the same data again

Fixed

> 4. workgroup looks a off ... use a stock control

Done

> 5. "query executions" heading seems too small (especially considering that it gets larger when navigated to)

That's intentional, but probably overthought. Index page has the main element "Query body" and this header is the biggest, other headers are equal sized, including "Query executions". The results page has two main elements, "Query body" and results with breadcrumbs (including "Query executions"). I think, I'm going to do 2 things:
1. make all headers equal
2. make "headers" for selects as part of Select, as for Catalog/Database, as you proposed

Done

> 6. it feels somewhat confusing that in query executions table only the completion date is clickable (quite easy to miss imo)

Noted and understood. But, I think, it is more or less familiar pattern from social media. I'll make the whole row clickable, if I have time.

> 7. query execution ui jumps when expanded / collapsed

Fixed

> 8. ux suggestion: a button to insert the query into the editor, not just copy it

I'm not sure about this. The copy and paste is, probably, the main use case, so it can be replaced by insertion. But I don't know what icon I can use ("edit" icon, I guess). And it is somewhat off, that I click here, but UI is changed in a different place. Also, user can go to execution query (click on "Date completed" link), and edit it there. I need to experiment with this.

> 9. layout jumps when switching to / from individual executions

Fixed

nl0:
> Also, user can go to execution query (click on "Date completed" link), and edit it there. I need to experiment with this.

probably that's good enough (i only noticed that "feature" after posting the review)

fiskus:
Persistent query error - done.

The whole execution row clickable - done.

fiskus:
Re-tested manually, and wrote some more tests.

nl0 (approved):
i haven't done the testing this time, but the code changes look good

fiskus:
Fixed two bugs:
1. when there are no results (but there are no errors either), I showed only one row with the current execution. Now I show the message saying there are no results.
2. the execution contains the lowercased catalog and database name, so I'm comparing lowercased values now

Also, added a couple of tests more

catalog/app/containers/Bucket/Queries/Athena/Athena.tsx
```diff
@@ -237,23 +237,17 @@ function ResultsContainer({ className }: ResultsContainerProps) {
           <SeeDocsForCreatingPackage />
         )}
       </ResultsBreadcrumbs>
-      {/* eslint-disable-next-line no-nested-ternary */}
-      {results.data.rows.length ? (
+      {!results.data.rows.length && execution?.error ? (
+        <Alert
+          error={execution?.error || new Error("Couldn't fetch query results")}
```
nl0, line 242: `execution?.error` here can't be nullish because of the condition above, so you don't need the fallback.
i can't quite wrap my head around what this condition is supposed to mean -- only show the error if there's no data?

fiskus: Thank you, I examined the code deeper, and found out this condition will never happen. `execution.error` could be only inside executions list

fiskus: Fixed

nl0 (approved):

PR #4212, opened 2024-11-08
Stop using S3 Select in indexer

sir-sigurd:
# Description

# TODO

- [x] revert CI


nl0 (dismissed):
i don't see anything suspicious

nl0 (approved):

PR #4213, opened 2024-11-12
Add 'catalog' to Quilt+ URIs

drernie:
# Description

This is my formal request to allow specifying the Catalog as part of the Quilt+ URI. This will allow applications that take a Quilt+ URI (nf-quilt, QuiltSync) to generate and share human-readable URLs.

Adding it as a fragment appears backwards-compatible; adding it as a query string, on the other hand, breaks the resolver.


nl0 (commented):
i feel somewhat ambivalent about that proposal, but i think i can live with this

catalog/app/utils/PackageUri.ts
```diff
@@ -108,5 +108,6 @@ export function stringify({ bucket, name, hash, tag, path }: PackageUri) {
     pkgSpec += `:${tag}`
   }
   const pathPart = path ? `&path=${encodeURIComponent(path)}` : ''
-  return `quilt+s3://${bucket}#package=${pkgSpec}${pathPart}`
+  const catalogPart = window.location.hostname
```
nl0, line 111: even tho it may seem like a nice shortcut, i deeply disagree with this approach.
i believe the `catalog` part should be added to the model and handled explicitly

docs/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ Entries inside each section should be ordered by type:
 ## CLI
 !-->
 
+# unreleased - YYYY-MM-DD
+
+## Python API
+
+* [Changed] Add `catalog` fragment to Quilt+ URIs (and to documentation). ([#4213](https://github.com/quiltdata/quilt/pull/4213))
```
nl0, line 20: how is this related to the python api?

nl0 (changes requested):

docs/CHANGELOG.md
drernie: Doh! Dumb copy-and-paste error.

catalog/app/utils/PackageUri.ts
drernie: Okay, I've interpreted `model` as "stringify takes it as an argument", which is passed from:
- Dashboard
- PackageTreeWrapper (on through to PackageCodeSample)

Is that what you meant? If not, any chance you could fix it for me so it makes this release?

drernie:
@fiskus Can you help me out with the "informational" test coverage failure? I don't see any good examples of how to test React.

fiskus:
We tend to ignore these failures.

I can write the test for `app/containers/Bucket/CodeSamples/Package.tsx`, but I'm afraid that writing tests for `app/containers/Bucket/PackageTree/PackageTree.tsx` would be too difficult.

catalog/app/utils/PackageUri.ts
nl0: > Is that what you meant?

yep

> If not, any chance you could fix it for me so it makes this release?

ok ill try

docs/CHANGELOG.md
nl0: we have a separate changelog for the catalog

nl0 (approved):

PR #4224, opened 2024-11-18
Rewrite File/Analytics.tsx with no Effect

fiskus:
So, something seemed off for me in using Effect here, but I wasn't sure what exactly. And started to dig, because the code is trivial, and it's easy to experiment with it.

To justify using some library, it's useful to implement the same with and without it. I tried to remove Effect to see a difference.

With Effect there are `None` and `Some` that doesn't explain intention. And I don't know if an error and lack of results are treated the same way is not a mistake, because it **looks** like a mistake. (Though, returning `Eff.Option.none()` explicitly can help)
When I read Some/None statements, I need to jump every time to the location where they are defined.
I can't combine multiple statements into one "if errors or no results" to highlight that this is my intention. (Actually, I can, like you did with `defaultExpanded`)
I need to decipher, that `defaultExpanded` is true only when there is data. No simple visible statement "when we show data, we show it in an expanded section".

Three major downsides that handled with Effect:
* I need to create an additional component where data is defined, because typescript doesn't recognize that object has data when I handled all other states
* I need to duplicate `Section` wrapper
* I have to use `React.useEffect` to log error

---

In the conclusion, I see that my main pain point is weak None/Some abstraction. I want to work with states that have descriptive names: https://github.com/quiltdata/quilt/pull/4225
Also, I noted above that using an additional component is a disadvantage, but it's not https://github.com/quiltdata/quilt/pull/4226


fiskus:
Was created for demonstration only

PR #4225, opened 2024-11-18
Rewrite File/Analytics.tsx with enum

fiskus:
So, I didn't like to work with Some/None, and rewrote the code with enum, enumerating states explicitly. It turns out I don't need Effect for this. 

Cons:
* using self-invoking function in component
* tuples types are somewhat bulky

Pros:
* states are named
* clearly see that we handle error the same way as no results, and it's not a mistake
* clear statement that `defaultExpanded` is true when there is a data

Using Typescript tuples makes sure data exists when `State.Data`.

Also, if we use an additional component, we can get rid of self-invoking function, but duplicate `<Section />`


nl0:
this kind of thing is more commonly done with disjoint union rather than with plan TS enums, e.g.
```ts
interface Loading {
  _tag: 'Loading'
}
interface NoAnalytics {
  _tag: 'NoAnalytics'
}
interface Data {
  _tag: 'Data'
  data: ...
}
type State = Loading | NoAnalytics | Data
```
that way you keep the tag and the data in one place.
there are helpers for this in effect:
```ts
import { Data } from "effect"

// Define a union type using TaggedEnum
type State = Data.TaggedEnum<{
  Loading: {}
  NoData: {}
  Data: { readonly data: ... }
}>
```

also, for such a general case, i'd use an abstraction like RemoteData or smth.

oh, and having different states for data / no data seems excessive -- `Data: Option<...>` expresses this notion just fine imo.

fiskus:
> having different states for data / no data

It's not about if the request has data or not. It's about whether to show data or not. That's type for UI, View-data.
That particular component shows "No analytics" when the request is failed. That looks like a mistake. But if this is intentional, we can have comment like "it's legit to have error, that means we don't have analytics" or some named variable.

fiskus:
Was created for demonstration only

PR #4226, opened 2024-11-18
Keep `Eff` in `File/Analytics.tsx`, but move request and state to container

fiskus:
In two examples (without Effect and additional code https://github.com/quiltdata/quilt/pull/4224/ and with state enum https://github.com/quiltdata/quilt/pull/4225) I have to split UI component from data component.
That was necessary without Effect and wasn't with Effect. But it seems that it gives the most relief.

With two components I have one component with UI only, and where I don't think about handling other states and the second component that deals with higher-level state and requests.
I can write separate tests for these components:
* make snapshots for UI component
* mock `useQuery` and UI component and write tests for handling states


nl0:
> With two components I have one component with UI only, and where I don't think about handling other states and the second component that deals with higher-level state and requests.

that sounds reasonable in general, but the benefit of doing this in such a tiny component seems negligible

fiskus:
Was created for demonstrations purpose only

PR #4234, opened 2024-11-21
Qurator: use Claude 3.5 Sonnet v2

nl0:
# TODO


sir-sigurd (commented):
lgtm, might make more sense to pass it via config though

nl0:
> lgtm, might make more sense to pass it via config though

that would be a next step, yeah. via config or a registry endpoint / gql query

PR #4236, opened 2024-11-21
Fix embed listing

fiskus:
Add selection provider for embed listing and make selection optional

# TODO

- [x] Make `Selection.Provider` optional. Not sure, it's worth it. It is in https://github.com/quiltdata/quilt/pull/4236/commits/24d8c20b52416d4a041898f1d930a8fc5b056d6e, so it's easily revertible 


nl0:
(tested)

PR #4239, opened 2024-11-22
Athena: check access for catalog

fiskus:
* fetch each data catalog to be sure user has access to it
* fixed catalog list load, there was missing workgroup argument

- [x] Unit tests


nl0 (dismissed):

nl0:
@fiskus what about the changelog entry?

nl0 (approved):

PR #4241, opened 2024-11-27
update lambdas to Python 3.11

sir-sigurd:
# Description

changes:

* add explicit deps for t4_lambda_shared
* add explicit dep on t4_lambda_shared in lambdas
* update all lambdas to Python 3.11 (except thumbnail)
* regenerate/update all requirements.txt with pip-compile
* add test-requirements.in and regenerate test-requirements.txt for every lambda
* clean up deps installation in test CI job
* stop passing shared to builder context
* add missing changelogs

notes
* dep on t4_lambda_shared is specified using hash from this PR, it doesn't looks like t4_lambda_shared changes can be made as separate PR, so hash needs to be updated after PR
* <del>molecule lambda is not really used, so I haven't tested it manually, maybe makes sense to remove it?</del> removed

# TODO


lambdas/preview/test/test_index.py
```diff
@@ -68,7 +68,12 @@ def test_fcs(self):
         extended = False
         if (
                 set(os.path.split(f)[1] for f in fcs_files)
-                != set(['accuri-ao1.fcs', 'bad.fcs', '3215apc 100004.fcs'])
+                != {
```
sir-sigurd, line 71: that test failed because one of the files is now parsed correctly by newer fcsparser
probably test should be reworked but I decided to add just another file to not deal with it now

nl0 (dismissed):
this is so much cleaner

nl0:
> molecule lambda is not really used, so I haven't tested it manually, maybe makes sense to remove it?

yeah probably. maybe in a clean PR?

.github/workflows/deploy-lambdas.yaml
```diff
@@ -2,8 +2,8 @@ name: Deploy lambdas to S3 and ECR
 
 on:
   push:
-    branches:
-      - master
+    # branches:
+    #   - master
```
nl0, line 6: @sir-sigurd still testing smth?

nl0 (approved):
lgtm

.github/workflows/deploy-lambdas.yaml
sir-sigurd, line 6:
```suggestion
    branches:
      - master
```

sir-sigurd: forgot about that

nl0 (approved):

PR #4252, opened 2024-12-03
Enable "html" in `quilt_summarize.json`

fiskus:
This is a quick fix. I've started to work on a more general solution, but this small piece, I think, self-sustainable.

More scalable solutions require more involvement, but I have several ideas already:
* enable everything we have Previews for
* make "name" optional, and treat it like fallback type
* improve `JsonEditor`, so users can create with config with some helper and make "name" a plain string but with select from supported types
* make custom editor for writing `quilt_summarize.json`
* or, change (make v2) Schema (but I'd go with custom editor first, so customers wouldn't care about Schema)

- [x] Documentation

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Added] Added "html" type for `quilt_summarize.json` ([#4252](https://github.com/quiltdata/quilt/pull/4252))
 - [Fixed] Resolve caching issues where changes in `.quilt/{workflows,catalog}` were not applied ([#4245](https://github.com/quiltdata/quilt/pull/4245))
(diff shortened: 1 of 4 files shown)


catalog/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Added] Added "html" type for `quilt_summarize.json` ([#4252](https://github.com/quiltdata/quilt/pull/4252))
```
nl0, line 20:
```suggestion
- [Added] Support "html" type in `quilt_summarize.json` ([#4252](https://github.com/quiltdata/quilt/pull/4252))
```

nl0 (dismissed):

nl0 (approved):

PR #4254, opened 2024-12-05
`quilt_summarize.json` editor

fiskus:
Screenshot from 2024-12-09 15-21-12

Added `__quiltSummarize` type for `QuiltConfigEditor`. But the plan is to replace the universal `QuiltConfigEditor` component with three custom forms.

`QuiltSummarize` is the first one.

- [x] Unit tests


catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
```diff
@@ -0,0 +1,782 @@
...
+    ) =>
+      onChangeValue('type', {
+        ...((file.type || {}) as Summarize.TypeExtended),
+        [key]: value,
+      }),
+    [onChangeValue, file.type],
+  )
+
+  const onRemove = React.useCallback(
+    () => onChange(removeColumn(row.id, column.id)),
```
nl0, line 327: the main reason i prefer namespace imports is it provides context and makes it easier to distinguish local symbols from imported ones. here, for example, it would make it easier to guess that `removeColumn` is a state modification function if it was prefixed with `State.` or smth

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
```diff
@@ -0,0 +1,782 @@
...
+                    label="Width"
+                    name="width"
+                    onChange={(event) =>
+                      onChangeValue('width', event.currentTarget.value)
+                    }
+                    value={file.width || ''}
+                    fullWidth
+                    className={classes.field}
+                    size="small"
+                    helperText="Width in pixels or percents"
```
nl0, line 436:
```suggestion
                    helperText="Width in pixels or percent"
```

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
```diff
@@ -0,0 +1,782 @@
...
+                    name="height"
+                    onChange={(event) =>
+                      onChangeType('style', { height: event.currentTarget.value })
+                    }
+                    value={file.type.style?.height || ''}
+                    fullWidth
+                    className={classes.field}
+                    size="small"
+                    placeholder="Ex., 1000px"
+                    helperText="Height as an absolute value (in `px`, `vh`, `em` etc.)"
```
nl0, line 473: are `vh` and `em` absolute?

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
```diff
@@ -0,0 +1,782 @@
...
+                    label="Perspective config"
+                    name="config"
+                    onChange={(event) => {
+                      try {
+                        onChangeType('config', JSON.parse(event.currentTarget.value))
+                      } catch (error) {
+                        Log.error(error)
+                      }
+                    }}
+                    helperText="Restores renderer to a state previously returned by saving config. Should be valid JSON object"
```
nl0, line 489: maybe smth like this?
```suggestion
                    helperText="Restores renderer state using a previously saved configuration. Configuration must be a valid JSON object."
```

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
```diff
@@ -0,0 +1,782 @@
...
+                {file.type?.name === 'perspective' && (
+                  <M.TextField
+                    disabled={disabled}
+                    label="Perspective config"
+                    name="config"
+                    onChange={(event) => {
+                      try {
+                        onChangeType('config', JSON.parse(event.currentTarget.value))
+                      } catch (error) {
+                        Log.error(error)
```
nl0, line 486: i'm not sure logging errors on every input change is a good idea. maybe display the error to the user instead?

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/State.tsx
```diff
@@ -0,0 +1,196 @@
...
+  id: crypto.randomUUID(),
+  columns: [createColumn(file)],
+})
+
+export const init = (payload?: Layout) => (): Layout =>
+  payload || {
+    rows: [createRow(emptyFile)],
+  }
+
+function insert<T>(array: T[], index: number, item: T): T[] {
```
nl0, line 48: https://effect-ts.github.io/effect/effect/Array.ts.html#insertat

nl0 (dismissed):
looks good, some nitpicks inline

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
```diff
@@ -0,0 +1,782 @@
...
+  row: Row
+  last: boolean
+}
+
+function AddColumn({ className, column, disabled, last, onChange, row }: AddColumnProps) {
+  const { bucket, initialPath } = useParams()
+
+  const classes = useAddColumnStyles()
+  const { file } = column
+  // TODO: [simple, setSimple] mode instead of advanced
```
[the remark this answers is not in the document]
fiskus: When the user hides the rest of the fields in the form except `path`, I'm planning to clear them (asking confirmation first if they are not empty). Because, if users don't see input fields, they shouldn't expect to submit those values.
But if the user clicks to show these fields again, I'm going to fill these fields with previously entered fields.

So, instead of simple flag "show"/"hide", I wanted to use "show"/"hide, but there were these values here".

However, I don't know if I implement this anytime in the near future, so I'll remove this TODO note.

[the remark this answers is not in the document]
nl0: i tend to use TODOs for short-term action items, which are to be resolved in the scope of the current PR, and XXX for longer term stuff or things requiring more consideration

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
fiskus: They are not in CSS context, but I don't know how to better describe that one can enter number as ratio value in `flexGrow` or some concrete value with units.
Probably, "concrete"?

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/State.tsx
fiskus: And Ramda has `insert` too https://ramdajs.com/docs/#insert

nl0: the plan was to gradually remove ramda

fiskus: Ok, I didn't have such plans, however I'm happy to not use Ramda, since native JavaScript has almost everything I need for arrays manipulation.

catalog/app/components/FileEditor/QuiltConfigEditor/QuiltSummarize/QuiltSummarize.tsx
fiskus: I'll prefix them, however I treat them as essential functions for this component, undisguisable from local functions.

nl0 (approved):

PR #4255, opened 2024-12-06
Catalog: Admin: Tabulator Settings (open query) + docs

nl0:
# Description

admin-tabulator-settings

# TODO

- [x] Documentation
    - [x] run `optipng` on any new PNGs
    - [x] JavaScript: basic explanation and screenshot of new features
    - [x] Markdown somewhere in docs/**/*.md that explains the feature to end users (said .md files should be linked from SUMMARY.md so they appear on https://docs.quiltdata.com)


nl0:
@drernie i'd appreciate some help with the documentation

fiskus (approved):
I have to admit, seeing the `*` in the CSS scared me a little at first. But looking closer, it’s actually clever and even performant!

On the other hand, I’m still struggling to see the benefits of using Effect here. It seems to make the code more verbose and less idiomatic compared to simpler approaches like comparing to `null`. I feel I’ve expressed my concerns about `Effect` clearly at this point, so I won’t bring it up again.

That said, the implementation is solid, and I appreciate the attention to detail.

nl0:
> I have to admit, seeing the `*` in the CSS scared me a little at first. But looking closer, it’s actually clever and even performant!

i was sure it's an established pattern =)

> On the other hand, I’m still struggling to see the benefits of using Effect here. It seems to make the code more verbose and less idiomatic compared to simpler approaches like comparing to `null`. I feel I’ve expressed my concerns about `Effect` clearly at this point, so I won’t bring it up again.

fair enough

> That said, the implementation is solid, and I appreciate the attention to detail.

thanks

nl0:
@sir-sigurd @fiskus pls see if the documentation changes make sense

docs/advanced-features/tabulator.md
```diff
@@ -106,11 +108,27 @@ order to query those tables.  This can be done by users via the per-bucket
 5. **Athena VPC**: If you are using a VPC endpoint for Athena, you must ensure
    it is accessible from the Quilt stack and Tabulator lambda.
 
+### Unrestricted Access
+
+> Available since Quilt Platform version 1.57
```
fiskus, line 113: Good idea!

fiskus (dismissed):

docs/advanced-features/tabulator.md
```diff
@@ -106,11 +108,27 @@ order to query those tables.  This can be done by users via the per-bucket
...
+
+By default, Tabulator is only accessible via a session provided by the Quilt Catalog,
```
sir-sigurd, line 115: you also can use quilt3 for this
or Quilt Catalog == registry here?

docs/advanced-features/tabulator.md
```diff
@@ -106,11 +108,27 @@ order to query those tables.  This can be done by users via the per-bucket
...
+and the access is scoped to the permissions of the Catalog user associated with
+that session. However, an admin can enable **unrestricted access** to Tabulator,
+deferring all access control to AWS. The underlying data in S3 is accessed using
```
sir-sigurd, line 118: maybe we should add some docs on how to grant access to tabulator in this case

docs/advanced-features/tabulator.md
nl0: well, basically, yes

docs/advanced-features/tabulator.md
nl0: yeah i guess that would be helpful. is it just the tabulator data catalog and everything inside it?

sir-sigurd: I think it's lambda and data catalog

docs/advanced-features/tabulator.md
```diff
@@ -188,3 +194,128 @@ if state == 'SUCCEEDED':
...
+   - Read access to the Tabulator bucket for spill files
+   - Read/write access to the Athena results bucket
+
+Here is an example CloudFormation template that creates the necessary resources:
+
+```yaml
+AWSTemplateFormatVersion: 2010-09-09
+Description: "Resources for accessing Tabulator in unrestricted mode"
+
+Parameters:
```
drernie, line 234: I apologize if this is a stupid question, but it seems like we already know all of this.  Could we create a Policy that encapsulates this, so they just have to add that to their Role?

drernie: In particular, if they already have a Role with 'unrestricted' abilty to call Athena and Lambdas, then won't that automatically be able to call Tabulator? If not, then what exactly were we protecting against...

drernie: Another thing that confuses me.  It looks like we now *always* use a split folder, and never return any results inline:
```
    fn should_spill(&self, _df: &DataFrame) -> bool {
        true
    }
```
If so, can't we just use that as the way to gate access, so we don't need restricted mode at all? Or am I just seriously confused?

nl0: > Could we create a Policy that encapsulates this, so they just have to add that to their Role?

yeah we can. i was considering this, but for some reason decided to not follow thru. i think i'm reconsidering this -- i'll add both the policy and the workgroup to the template and export their ids/arns for easier consumption


> In particular, if they already have a Role with 'unrestricted' abilty to call Athena and Lambdas, then won't that automatically be able to call Tabulator? If not, then what exactly were we protecting against...

call tabulator -- sure, but the tabulator itself uses two roles for functioning:
1. its "own" role with access to cache and stuff
2. the role it gets from the registry and assumes to actually access the data (in unrestricted mode it's the special "unrestricted" role, otherwise it's a catalog user's role)

so even if someone is calling tabulator without the proper setup it will just crash -- return an error, not the data

> Another thing that confuses me. It looks like we now always use a split folder, and never return any results inline:

correct

> If so, can't we just use that as the way to gate access, so we don't need restricted mode at all? Or am I just seriously confused?

maybe we could, but that removes an extra layer of security, so i don't think we should do that

nl0:
i think this is now ready

fiskus (approved):

sir-sigurd (approved):

PR #4257, opened 2024-12-11
Entries schema docs

drernie:
Documentation updates, to increase clarity.

### DIFF
--- a/docs/advanced-features/workflows.md
+++ b/docs/advanced-features/workflows.md
@@ -78,9 +78,61 @@ schemas:
     url: s3://quilt-dev-metadata/schemas/validate-secrets.schema.json
 ```
 
-Once you add a configuration that defines workflows, you must specify a workflow
-before you can push (unless you explicitly set `is_workflow_required: False` in
-your `config.yml`).
+### Setting a default workflow
+
+Once a configuration file is present, you usually **must** specify a workflow
+when pushing a package from `quilt3` to that bucket. You can however specify a
+`default_workflow` value at the top-level in your `config.yml` file:
+
+```yaml
+version:
+  base: "1"
+  catalog: "1"
+default_workflow: "experiment"
+is_workflow_required: False
+workflows:
+  experiment:
+    name: Experiment
+    metadata_schema: experiment-universal
+schemas:
+  experiment-universal:
+    url: s3://quilt-dev-metadata/.quilt/workflows/schemas/experiment-universal.json
+```
+
+This specifies which workflow will be used (`experiment`) if a
+`workflow` parameter in the `Package.push()` API call or CLI is not provided.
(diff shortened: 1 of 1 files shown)


drernie:
@fiskus Would you want to add your documentation about "auto-config" to this PR?

fiskus:
Thank you for reminding. Let's add "auto-config" docs in a separate PR: https://github.com/quiltdata/quilt/pull/4258

drernie:
@QuiltSimon @fiskus Any other comments? Okay to merge?

docs/advanced-features/workflows.md
```diff
@@ -187,65 +249,261 @@ quilt3.Package().set_meta({'answer': 42}).push(
...
+}
+```
+
+### Cross-bucket package push (Quilt catalog)
+
+In Quilt, S3 buckets are like git branches but for data. With `quilt3` you can
+`browse` any package and then `push` it to any bucket that you choose.
+
+As a rule, cross-bucket pushes or "merges" reflect change in a package's
+lifecycle. For example, you might push a package from  *my-staging-bucket*
```
fiskus, line 467:
```suggestion
lifecycle. For example, you might push a package from *my-staging-bucket*
```

docs/advanced-features/workflows.md
```diff
@@ -461,243 +725,22 @@ such as `object`, `array`, and compound `enum` types.
 This allows for flexible and extensible schema definition, and hence
 validation, of complex metadata schemas to any depth.
 
-> Quilt currently uses the Draft 4 Json Schema where tuples are
```
fiskus, line 464: This block is not outdated. `prefixItems` keyword still is not supported. Draft 7 was released in 2018

I think, there was a typo: "Draft 4" → "Draft 7"

The reason, we mention what we don't support, is because some customer used `prefixItems`

fiskus: Oh, I found it

fiskus (approved):
Changes look good! However, I think the underlying issue remains: the page is still too large and mixes tutorial content with documentation, as well as common use cases with rare cases. Perhaps we could consider splitting these into separate sections/pages in the future?

docs/advanced-features/workflows.md
```diff
@@ -85,7 +145,7 @@ quilt3.Package().push('test/package', registry='s3://quilt-dev-metadata')
 # QuiltException: Workflow required, but none specified.
 ```
 
-Let's try with the `workflow=` parameter:
+Let's try again, adding the `workflow=` parameter:
```
drernie, line 148:
```suggestion
Let's retry, adding the `workflow=` parameter:
```

fiskus (approved):

PR #4258, opened 2024-12-11
Add documentation about auto-adding source bucket

fiskus:
- [x] Documentation


drernie:
Will this work even if the user is not an Administrator, as long as they can write to the bucket?

docs/Catalog/Preferences.md
```diff
@@ -99,8 +99,17 @@ ui:
   defaultSourceBucket: s3://bucket-b
 ```
 
+Note that the `ui.sourceBuckets` is empty by default.
+So, when the users create a new package they can't add files from any bucket
+including the current one.
+
+However, we have a shortcut for adding the current bucket
+to the list of `ui.sourceBuckets` from the UI.
```
drernie, line 107:
```suggestion
As a result, when users create or revise a package in a new bucket they can't add files from any bucket, including that one.  Instead, you need to manually add a configuration file, or click "auto-add current bucket" (which will create or edit the configuration file to add the current bucket to the list of `ui.sourceBuckets`)
```

drernie (approved):

fiskus:
Yes, I've just created a new non-admin user with `ReadWriteQUuiltBucket` role and the feature works for this user.

PR #4259, opened 2024-12-17
Tabulator open query docs

drernie:
Rewriting to use the term 'open query' instead of 'unrestricted access'.
Also rewrote the docs to see if my understanding is correct.

In particular, my naive understanding was that the reason we originally locked down Tabulator is that any user with full Athena + Lambda permissions would otherwise be able to perform any Tabulator query (I.e., they don't need explicit access to the TabulatorDataCatalog and TabulatorBucket).  Is that not true?


docs/advanced-features/tabulator.md
```diff
@@ -195,41 +197,51 @@ else:
...
+- read and write to that workgroup's output location, e.g.
+  `s3://${UserAthenaResultsBucket}`
+- invoke the Tabulator Lambda function
+- read from the relevant Tabulator bucket (in order to access spill files, if
+  any)
 
-2. Have the following permissions:
+In addition, Tabulator itself must be able to write to the workgroup's output
+location. The easiest way to do this is by adding that bucket to the Quilt
+stack.
```
nl0, line 232: this is not quite correct. adding a bucket to the stack won't solve anything. tabulator's "open access role" has hardcoded permissions for writing results to `${UserAthenaResultsBucket}/athena-results/*/*`, and therefore the provided workgroup **must** have compatible results location

nl0:
> In particular, my naive understanding was that the reason we originally locked down Tabulator is that any user with full Athena + Lambda permissions would otherwise be able to perform any Tabulator query (I.e., they don't need explicit access to the TabulatorDataCatalog and TabulatorBucket). Is that not true?

no, it's not

anyone with athena and/or lambda permissions can call tabulator -- sure, but the tabulator itself uses two roles for functioning:

1. its "own" (lambda) execution role with access to cache and stuff (with no access to the actual data)

2. the role it gets from the registry and assumes to actually access the data (in unrestricted mode it's the special "unrestricted" role, otherwise it's a catalog user's role)

so even if someone is calling tabulator without the proper setup it will just crash -- return an error, not the data

drernie:
So... stupid question: why do we even need "restricted" mode?

nl0:
> So... stupid question: why do we even need "restricted" mode?

to enforce this:

> if someone is calling tabulator without the proper setup it will just crash -- return an error, not the data

in unrestricted mode anyone with access to tabulator data catalog, lambda, a compatible workgroup and special s3 buckets can access data from any stack bucket (with configured tables)

nl0:
partially merged into #4255

PR #4264, opened 2024-12-24
Auto-format `SetTabulatorOpenQuery.graphql`

fiskus:
`prettier`'d ".graphql" file, so `npm run lint` is passed.


sir-sigurd (approved):
probably makes sense to check it in CI

sir-sigurd:
> I agree, I don't know why it (both lint and prettier) is not checked during CI test

well, someone has to do it 🙃

Issue #4266, opened 2024-12-24
Dependabot PR


drernie:
Can we target (non-security) Dependabot updates onto a separate branch, where (if CI succeeds) they are auto-merged into an open PR? Then we periodically just need to merge that one PR.

sir-sigurd:
is this still relevant?

PR #4267, opened 2024-12-24
feat(4266): universal dependabot pr

drernie:
#4266 Auto-merge succesful Dependabot PRs into an integration branch


sir-sigurd (commented):
I'm not sure what problem this solves 
I merge updates quite promptly if they pass tests and currently we have a single non-security dependabot PR

drernie:
Not needed.  Now that we are down to a small number, we can just clean up during Broad Work.

PR #4279, opened 2025-01-07
Split Search and Query

drernie:
Cleaner and easier to add new information

### DIFF
--- a/docs/Catalog/Query.md
+++ b/docs/Catalog/Query.md
@@ -1,0 +1,55 @@
+<!-- markdownlint-disable-next-line first-line-h1 -->
+[Amazon Athena](https://aws.amazon.com/athena/) is an interactive query service
+that makes it easy to analyze data in Amazon S3 using standard SQL. Athena is
+serverless, so there is no infrastructure to manage, and you pay only for the
+queries that you run.
+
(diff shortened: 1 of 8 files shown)


drernie:
@QuiltSimon We had discussed this a long time ago. Does it make sense to you?

docs/Catalog/Query.md
QuiltSimon, line 1: I recommend we mention on the Query page that if they're looking for Elasticsearch, that page is available, and viceversa on the search page mentioning querying. A helpful tidbit would be how teams would choose to use one tool vs. the other.

docs/Catalog/Search.md
QuiltSimon, line 1: Same as the comment on query but adding for completeness:

I recommend we mention on the Query page that if they're looking for Elasticsearch, that page is available, and viceversa on the search page mentioning querying. A helpful tidbit would be how teams would choose to use one tool vs. the other.

QuiltSimon (changes requested):
This looks good, but want to ensure a mention of each capability is added on the other's page. Other than that LGTM

QuiltSimon (approved):
Good enough for now

PR #3765, opened 2023-10-16
Min.io Support - User-Provided S3 Clients

kevinemoore:
# Description
Allow users to pass their own S3 clients for accessing specific buckets to support Min.io and passing in S3 credentials outside of environment variables and config files.

TODO: how will this support bucket-to-bucket copy across different clients?

# TODO

- [ ] Unit tests
- [ ] Automated tests (e.g. Preflight)
- [ ] Confirm that this change meets security best practices and does not violate the security model
- [ ] Documentation
    - [ ] [Python: Run `build.py`](../gendocs/build.py) for new docstrings
    - [ ] JavaScript: basic explanation and screenshot of new features
    - [ ] Markdown somewhere in docs/**/*.md that explains the feature to end users (said .md files should be linked from SUMMARY.md so they appear on https://docs.quiltdata.com)
    - [ ] Markdown docs for developers
- [ ] [Changelog](CHANGELOG.md) entry (skip if change is not significant to end users, e.g. docs only)

PR #4281, opened 2025-01-09
Register event callback

drernie:
# Description

Strawman proposal for how to use [boto3 events](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html#boto3-specific-events) to solve the problem of "include appropriate parameters when writing S3 objects"

# TODO

- [X] Unit tests
- [ X] Documentation
    - [X] Markdown docs for developers
- [X] [Changelog](../tree/master/docs/CHANGELOG.md) entry (skip if change is not significant to end users, e.g. docs only)


drernie:
@sir-sigurd @nl0 The basic idea is:

1. Maintain a list of events and callbacks
2. Add those when creating a new client
3. Unset the clients when updating the list (by passing an event-name/callback pair)
4. Add a convenience method to update all (only) the write_options

api/python/quilt3/data_transfer.py
```diff
@@ -171,6 +178,61 @@ def _build_unsigned_client(self):
...
+        """Register a callback for S3 client events.
+
+        Args:
+            event_name: The name of the event to register for (e.g. 'creating-client-class')
+            callback: The callback function to be called when the event occurs, or None to remove the callback
+        """
+        if callback is None:
+            del self._event_callbacks[event_name]
+        else:
+            self._event_callbacks[event_name] = callback
```
sir-sigurd, line 203: might make sense to not fail when handler being removed doesn't exist (or fail when you add duplicate event handler)

BTW boto3 supports multiple handlers for the same event
we *might want* to support that as well though we don't have to

api/python/quilt3/data_transfer.py
```diff
@@ -171,6 +178,61 @@ def _build_unsigned_client(self):
...
+
+        # Force clients to re-register the callback on the new clients
+        self._standard_client = None
+        self._unsigned_client = None
```
sir-sigurd, line 207: seems unneeded

api/python/quilt3/data_transfer.py
```diff
@@ -171,6 +178,61 @@ def _build_unsigned_client(self):
...
+
+    def register_write_options(self, **kwargs):
+        """Register write options for S3 clients.
```
sir-sigurd, line 210: I'd make users responsible for this

api/python/quilt3/data_transfer.py
```diff
@@ -171,6 +178,61 @@ def _build_unsigned_client(self):
...
+        def create_dynamic_callback(options):
+            def dynamic_callback(params, **kwargs):
+                # Update params with the provided write options
+                # TBD: sanitize first
+                self.add_options_safely(params, options)
+
+            return dynamic_callback
+
+        # Register each event with a dynamically created callback
+        callback = create_dynamic_callback(kwargs)
```
sir-sigurd, line 232: doesn't this work?

```suggestion
        options = kwargs
        def callback(params, **kwargs):
            self.add_options_safely(params, options)
```

sir-sigurd (commented):
it needs some user facing API and I guess event handlers should be stored globally instead of storing them on S3ClientProvider instance which we create almost on every call of `data_transfer` functions

I don't have strong opinion on that but it might make sense to allow users to provide some hook for client creation instead
it looks like we had some users modifying client creation (see https://github.com/quiltdata/quilt/issues/1941#issuecomment-1298711883) and we had attempt to allow user-provided clients (see https://github.com/quiltdata/quilt/pull/3765)

api/python/quilt3/data_transfer.py
drernie: So, I'm nervous about that.  It feels like we'd have to document all the internal APIs we call, or they'd have to use trial and error.  Or is that already true in general, and we should just include this as an example?

api/python/quilt3/data_transfer.py
```diff
@@ -90,6 +90,7 @@ def __init__(self):
         self._use_unsigned_client = {}  # f'{action}/{bucket}' -> use_unsigned_client_bool
         self._standard_client = None
         self._unsigned_client = None
+        self._event_callbacks = {}
```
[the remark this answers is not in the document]
drernie: @greptileai suggest commit

api/python/quilt3/data_transfer.py
drernie: My thought was that the easiest way to remove a callback is to set it to None.  But I guess it is effectively reset after every call, so it doesn't matter.

drernie:
> it needs some user facing API and I guess event handlers should be stored globally instead of storing them on S3ClientProvider instance which we create almost on every call of `data_transfer` functions

By globally do you mean as a class property? I forget the proper idiom for Python.

api/python/quilt3/data_transfer.py
drernie: Added as a doc example

api/python/quilt3/data_transfer.py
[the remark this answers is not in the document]
drernie: made a class variable

drernie:
@sir-sigurd @nl0 Okay, I *think* this basically works.  

1. I added a manual integration test to convince myself this actually works; it is skipped for now. Should I just delete it?
2. I've added API documentation to `quilt3.data_transfer.S3ClientProvider`, but I'm not sure the right way to expose that

api/python/quilt3/data_transfer.py
sir-sigurd: > So, I'm nervous about that. It feels like we'd have to document all the internal APIs we call, or they'd have to use trial and error. Or is that already true in general, and we should just include this as an example?

I just don't think we can make it work in general case e.g. because `SSECustomerKey` likely needs to be set for `complete_multipart_upload()` as well but that API doesn't support `ServerSideEncryption` and users may need to set different set of options for different buckets/keys
also it's quite hard to formulate why "write options" are set for `PutObject`, `CopyObject`, `CreateMultipartUpload` but not for `UploadPart` and `CompleteMultipartUpload`
in case they need to comply SCP I guess they need to figure out operations/options from the policy itself

api/python/quilt3/data_transfer.py
```diff
@@ -85,6 +85,72 @@ class S3ClientProvider:
 
     We assume that public buckets are read-only: write operations should always use S3ClientProvider.standard_client
     """
+    _event_callbacks = {}
```
sir-sigurd, line 88: in [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html) they are called "event handlers" I guess it worth to follow this terminology to avoid confusion

api/python/quilt3/data_transfer.py
```diff
@@ -85,6 +85,72 @@ class S3ClientProvider:
...
+    def reset_event_callbacks(cls):
+        cls._event_callbacks = {}
+
+    @classmethod
+    def register_event_options(cls, event_name: str, **kwargs: dict) -> None:
+        """Register options for S3 client events. A convenience method over register_event_callback
+        for trivial callbacks that always and only add options.
+
+        Args:
+            event_name: The name of the event to register for (e.g. 'creating-client-class')
```
sir-sigurd, line 111: AFAICT `creating-client-class` doesn't make sense because it needs to be registered on session, see https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html#creating-client-class
if we'd allow to set this handler on session I guess it could be used for client modification I mentioned including registration of `provide-client-params` event handlers

sir-sigurd:
> I added a manual integration test to convince myself this actually works; it is skipped for now. Should I just delete it?

I think you could patch something to make it create [stubbed](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html) client (maybe using `creating-client-class` event on session?) to check that calls of some functions in data_transfer results in certain parameters in s3 client calls

> I've added API documentation to quilt3.data_transfer.S3ClientProvider, but I'm not sure the right way to expose that

it doesn't look like a best way to expose it, maybe have that registration function separate from S3ClientProvider, maybe even add a separate submodule for that? 🤔

api/python/quilt3/data_transfer.py
```diff
@@ -85,6 +85,72 @@ class S3ClientProvider:
...
+            cls.add_options_safely(params, kwargs)
+
+        cls.register_event_callback(event_name, callback)
+
+    @classmethod
+    def register_event_callback(cls, event_name: str, callback: Callable) -> None:
+        """Register a callback for S3 client events.
+
+        Args:
+            event_name: The name of the event to register for (e.g. 'creating-client-class')
```
sir-sigurd, line 139: probably we should say that we just wrap boto3 events

api/python/quilt3/data_transfer.py
drernie: changed

drernie:
@sir-sigurd I think I've done all I can (can you do the gendocs?).  Can you talk with @nl0 and finalize whether this is the right direction, and what else needs to be done / tested?

drernie:
Obsoleted by #4289

PR #4282, opened 2025-01-10
Make docs clearer that package handle is a RegExp only

fiskus:
- [x] Documentation

### DIFF
--- a/docs/Catalog/Preferences.md
+++ b/docs/Catalog/Preferences.md
@@ -122,7 +122,7 @@ ui:
       user_meta:
         - $.labels
     # for any package with a handle prefix of foo
-    ^foo/*:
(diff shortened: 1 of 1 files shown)


docs/Catalog/Preferences.md
```diff
@@ -122,7 +122,7 @@ ui:
       user_meta:
         - $.labels
     # for any package with a handle prefix of foo
-    foo/*:
+    ^foo/*:
```
nl0, line 125:
```suggestion
    ^foo/.*:
```

drernie (dismissed):

nl0 (changes requested):
thanks for fixing this

nl0 (approved):

PR #4289, opened 2025-01-16
Build client hooks

sir-sigurd:
# Description

# TODO

- [x] Unit tests
- [x] Documentation

### DIFF
--- a/api/python/quilt3/hooks.py
+++ b/api/python/quilt3/hooks.py
@@ -1,0 +1,25 @@
+import typing as T
+
+import boto3
+
+
+class BuildClientBase(T.Protocol):
+    def __call__(self, session: boto3.Session, client_kwargs: dict): ...
+
+
+class BuildClientHook(T.Protocol):
+    def __call__(self, build_client_base: BuildClientBase, session: boto3.Session, client_kwargs: dict): ...
+
+
+_build_client_hook = None
+
+
+def set_build_client_hook(hook: T.Optional[BuildClientHook]) -> T.Optional[BuildClientHook]:
+    global _build_client_hook
+    old_hook = _build_client_hook
+    _build_client_hook = hook
+    return old_hook
+
+
+def get_build_client_hook() -> T.Optional[BuildClientHook]:
+    return _build_client_hook
--- a/api/python/quilt3/hooks.py
+++ b/api/python/quilt3/hooks.py
@@ -14,12 +14,52 @@ class BuildClientHook(T.Protocol):
 _build_client_hook = None
 
 
-def set_build_client_hook(hook: T.Optional[BuildClientHook]) -> T.Optional[BuildClientHook]:
+def get_build_s3_client_hook() -> T.Optional[BuildClientHook]:
+    """
+    Return build S3 client hook.
+    """
+
+    return _build_client_hook
+
+
+def set_build_s3_client_hook(hook: T.Optional[BuildClientHook]) -> T.Optional[BuildClientHook]:
+    """
+    Set build S3 client hook.
+
+    Example for overriding `ServerSideEncryption` parameter for certain S3 operations:
+
+    ```python
+    def event_handler(params, **kwargs):
+        # Be mindful with parameters you set here.
+        # Specifically we it's not recommended to override/delete already set parameters
+        # because that can break quilt3 logic.
+        params.setdefault("ServerSideEncryption", "AES256")
+
+    def hook(build_client_base, session, client_kwargs):
+        client = build_client_base(session, client_kwargs)
+        for op in (
+            "CreateMultipartUpload",
+            "CopyObject",
+            "PutObject",
+        ):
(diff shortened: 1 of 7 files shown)


nl0 (approved):
looks good

api/python/quilt3/hooks.py
```diff
@@ -0,0 +1,58 @@
...
+
+    ```python
+    def event_handler(params, **kwargs):
+        # Be mindful with parameters you set here.
+        # Specifically we it's not recommended to override/delete already set parameters
+        # because that can break quilt3 logic.
+        params.setdefault("ServerSideEncryption", "AES256")
+
+    def hook(build_client_base, session, client_kwargs):
+        client = build_client_base(session, client_kwargs)
```
sir-sigurd, line 39: maybe we should ask to do this to ensure forward compatibility if we decide to pass additional information (e.g. bucket?)?

```suggestion
    def hook(build_client_base, session, client_kwargs, **kwargs):
        client = build_client_base(session, client_kwargs, **kwargs)
```

docs/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ Entries inside each section should be ordered by type:
 ## CLI
 !-->
 
+# unreleased - YYYY-MM-DD
+
+## Python API
+
+* [Added] `quilt3.hooks`: `get_build_s3_client_hook(()` and `set_build_s3_client_hook()` functions for customization of S3 client building ([#4289](https://github.com/quiltdata/quilt/pull/4289))
```
sir-sigurd, line 20:
```suggestion
* [Added] `quilt3.hooks`: `get_build_s3_client_hook()` and `set_build_s3_client_hook()` functions for customization of S3 client building ([#4289](https://github.com/quiltdata/quilt/pull/4289))
```

docs/api-reference/Hooks.md
```diff
@@ -0,0 +1,38 @@
...
+# set\_build\_s3\_client\_hook(hook: Optional[quilt3.hooks.BuildClientHook]) -> Optional[quilt3.hooks.BuildClientHook]  {#set\_build\_s3\_client\_hook}
+
+Set build S3 client hook.
+
+Example for overriding `ServerSideEncryption` parameter for certain S3 operations:
+
+```python
+def event_handler(params, **kwargs):
+    # Be mindful with parameters you set here.
+    # Specifically we it's not recommended to override/delete already set parameters
```
sir-sigurd, line 16:
```suggestion
    # Specifically it's not recommended to override/delete already set parameters
```

nl0 (approved):
lgtm

nl0 (dismissed):

api/python/quilt3/hooks.py
```diff
@@ -0,0 +1,60 @@
...
+    Example for overriding `ServerSideEncryption` parameter for certain S3 operations:
+
+    ```python
+    def event_handler(params, **kwargs):
+        # Be mindful with parameters you set here.
+        # Specifically it's not recommended to override/delete already set parameters
+        # because that can break quilt3 logic.
+        params.setdefault("ServerSideEncryption", "AES256")
+
+    def hook(build_client_base, session, client_kwargs, **kwargs):
```
drernie, line 40: Sorry if this is a stupid question, but why do we always want the user to manually build the client?  
Do they even have enough information to usefully override that?  Or is it safer to (at least to start with) give them a (dumber) convenience method that updates an existing client?

def hook(client)

api/python/quilt3/hooks.py
```diff
@@ -0,0 +1,60 @@
...
+        client = build_client_base(session, client_kwargs, **kwargs)
+        for op in (
+            "CreateMultipartUpload",
+            "CopyObject",
+            "PutObject",
+        ):
+            client.meta.events.register(f"before-parameter-build.s3.{op}", event_handler)
```
drernie, line 47: I assume `before-parameter-build.s3.*` are the standard boto3 events?   
Is that obvious to everyone, or would it help to mention the documentation?
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html

drernie (approved):

api/python/quilt3/hooks.py
nl0: > Sorry if this is a stupid question, but why do we always want the user to manually build the client?

so that they could build their own client if required, without wasting resources on instantiating the built-in one (it could also throw in some circumstances and there will be no clean way around that, e.g. if a user decides to use a custom client with a different set of parameters)

> Do they even have enough information to usefully override that?

they do: there's the session and all the args used for the client creation, so the instantiation can be as flexible as required.

> Or is it safer to (at least to start with) give them a (dumber) convenience method that updates an existing client?

imo there's no point in complicating things (adding more methods). as long as they follow a very straightforward instruction to call the "super" method, everything's fine.

api/python/quilt3/hooks.py
nl0: adding the link to the boto docs sounds like a good idea

PR #4318, opened 2025-02-14
xlsx-0.20.3

drernie:
XLSX is no longer maintained
Bump to new version from SheetJS to avoid security warning

### DIFF
--- a/catalog/package-lock.json
+++ b/catalog/package-lock.json
@@ -104,7 +104,7 @@
         "vega-lite": "^5.14.1",
         "warning": "^4.0.3",
         "wonka": "^4.0.15",
-        "xlsx": "^0.18.5",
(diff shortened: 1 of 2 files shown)


fiskus:
There are some red flags regarding this library:
1. SheetJS/xlsx is a source of vulnerabilities, so they didn't write the best code in the first place.
2. They silently end hosting their library on npm, and didn't notify about that unpaid open-source users.
3. They host their library on their own server, where they sell the paid version.

Considering all this, I think, we should invest time in moving to other library, for example, https://www.npmjs.com/package/exceljs.
We use it only for reading simple tabular data, and we use it through facade which have tests.
So, it shouldn't be difficult.

drernie:
Switch to. [exceljs](https://www.npmjs.com/package/exceljs).

PR #4319, opened 2025-02-14
Catalog: Search UX adjustments

nl0:
# Description

- Search Results: Don't show package hash in the header
- Default package search to latest revisions only

Screenshot 2025-02-17 at 12 59 30

# TODO

- [x] Unit tests

### DIFF
--- a/catalog/app/containers/Search/Route.ts
+++ b/catalog/app/containers/Search/Route.ts
@@ -38,13 +38,13 @@ const PackageSearchParamsFromSearchParams = S.transform(
       resultType: SearchModel.ResultType.QuiltPackage as const,
       filter: S.decodeSync(Filter.PackageFilter.fromSearchParams)(params),
       userMetaFilters: S.decodeSync(UserMetaFilters.fromSearchParams)(params),
-      latestOnly: params.rev?.[0] === 'latest',
+      latestOnly: params.rev?.[0] !== 'all',
     }),
     // json api to qs
     encode: (toI) => ({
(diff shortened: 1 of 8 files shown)


catalog/app/containers/Search/Route.ts
```diff
@@ -38,13 +38,13 @@ const PackageSearchParamsFromSearchParams = S.transform(
       resultType: SearchModel.ResultType.QuiltPackage as const,
       filter: S.decodeSync(Filter.PackageFilter.fromSearchParams)(params),
       userMetaFilters: S.decodeSync(UserMetaFilters.fromSearchParams)(params),
-      latestOnly: params.rev?.[0] === 'latest',
+      latestOnly: params.rev?.[0] !== 'all',
```
fiskus, line 41: "all" and "latest" are used as just strings literals everywhere, right? It makes sense, to use constants or enum to show all available options and prevent mistyping.

fiskus (dismissed):

catalog/app/containers/Search/Route.ts
nl0: there are no other available options, it's just a boolean flag which i've decided to encode as `rev=all` in querystring for future proofing (and for general aesthetics)

nl0:
@drernie @QuiltSimon pls approve the copy (or suggest edits)

drernie:
> title="Enable to show matches from prior versions, versus only in the latest revision"

That's my text, so it works for me. :-) @QuiltSimon ?

PR #4349, opened 2025-03-07
Quilt3 v6.3.1

sir-sigurd:


sir-sigurd:
@drernie 
I want to confirm we want to release this change as a patch release

https://github.com/quiltdata/quilt/blob/7488a87aa9cdb3ff895b73005048482b413cf997/docs/CHANGELOG.md?plain=1#L20

nl0 (commented):
lgtm

drernie:
Sure. Do we need a new major version then?

sir-sigurd:
> Sorry, didn't read the question correctly. If this is a breaking change, shouldn't it be a major version?

on the one hand yes, it's breaking change, but it wasn't intended to work like this and bugfixes in general kinda break behavior
also these manifests break catalog so I tend to think it should be patch release so it's delivered ASAP

drernie:
Okay, I am okay with making it a patch, as long as we frame it along the lines of "There was a bug that in certain cases enabled creation of invalid JSONL documents, which has now been fixed."

nl0 (approved):

PR #4353, opened 2025-03-10
"Open in desktop"/"Download" buttons with popup

fiskus:
* There is also unrelated customization of the `StyledTooltip` width in PR, because I used it but then opted to `M.Paper` with `position: absolute`
* Removed containers/OpenInDesktop and re-purposed containers/Bucket/Download for download button and download/QuiltSync/code samplse popup
* Simplified code samples, removed `push` section

[1 get-package-edited.webm](https://github.com/user-attachments/assets/4580df5a-90d8-448f-bf1e-c4149d3f52f6)

2 get sub package
3 get package selected files
4 get package file

[5 get directory (1).webm](https://github.com/user-attachments/assets/86de6a6d-1e31-49ff-bb39-1be422ca17cd)

6 get file


drernie:
Looks great. Ship it! :-)

catalog/app/components/Buttons/Iconized.tsx
```diff
@@ -16,6 +16,7 @@ interface ButtonIconizedProps extends M.IconButtonProps {
   label: string
   rotate?: boolean
   variant?: 'text' | 'outlined' | 'contained'
+  endIcon?: React.ReactNode
 }
```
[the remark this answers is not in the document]
fiskus: It is used in `...props` spread

catalog/app/containers/Bucket/Download/BucketCodeSamples.tsx
```diff
@@ -0,0 +1,116 @@
...
+interface FileCodeSamplesProps {
+  className?: string
+  bucket: string
+  path: string
+}
+
+export function FileCodeSamples({ className, bucket, path }: FileCodeSamplesProps) {
+  const classes = useStyles()
+  const apiDest = basename(path)
+  const cliDest = ''
```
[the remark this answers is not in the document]
fiskus: `./` is appended to the empty string later inside components

catalog/app/containers/Bucket/Download/PackageOptions.tsx
```diff
@@ -0,0 +1,184 @@
...
+  hashOrTag,
+  hideCode,
+  selection,
+  uri,
+}: OptionsProps) {
+  if (hideCode)
+    return <DownloadPanel fileHandle={fileHandle} selection={selection} uri={uri} />
+
+  return (
+    <Tabs labels={['Download', 'Code']}>
```
nl0, line 174: maybe just make the Tabs component accept `download` and `code`?
that would probably make the code cleaner, bc this `activeTab === 0` thing looks quite confusing.
also, you don't have to make them into functions w/ that approach, just use plain elements.
it would look like this:

```ts
<Tabs
  download={<DownloadPanel fileHandle={fileHandle} selection={selection} uri={uri} />}
  code={<CodePanel hashOrTag={hashOrTag} uri={uri} />}
/>
```

catalog/app/containers/Bucket/File/File.js
```diff
@@ -467,9 +470,21 @@ export default function File() {
...
+              {
+                Ok: ({ ui: { blocks } }) => (
+                  <Download.Button className={classes.button} label="Get file">
+                    <Download.BucketOptions handle={handle} hideCode={!blocks.code} />
+                  </Download.Button>
+                ),
+                Pending: () => (
+                  <Buttons.Skeleton className={classes.button} size="small" />
+                ),
+                Inint: () => null,
```
nl0, line 484:
```suggestion
                Init: () => null,
```

catalog/app/containers/Bucket/PackageTree/PackageTree.tsx
```diff
@@ -435,18 +415,23 @@ function DirDisplay({
...
-                          <Download.DownloadButton
+                          <Download.Button
                             className={classes.button}
-                            label={downloadLabel}
-                            onClick={openInDesktopState.confirm}
-                            path={downloadPath}
-                            selection={slt.selection}
-                          />
+                            label={
+                              !packageUri.path && slt.isEmpty ? 'Get package' : undefined
```
nl0, line 421: does passing `undefined` here mean that there will be no label?
afair argument default won't be used if you are passing `undefined` explicitly

catalog/app/embed/Dir.js
```diff
@@ -124,14 +124,12 @@ export default function Dir() {
         </div>
         <M.Box flexGrow={1} />
         {!cfg.noDownload && (
-          <FileView.ZipDownloadForm suffix={`dir/${bucket}/${path}`} newTab>
-            <Buttons.Iconized label="Download directory" icon="archive" type="submit" />
-          </FileView.ZipDownloadForm>
+          <Download.Button>
```
nl0, line 127: have you actually tested the embed thing?

catalog/app/utils/Config.ts
```diff
@@ -90,6 +91,7 @@ const transformConfig = (cfg: ConfigJson) => ({
   apiGatewayEndpoint: startWithOrigin(cfg.apiGatewayEndpoint),
   noDownload: !!cfg.noDownload,
   noOverviewImages: !!cfg.noOverviewImages,
+  /** @deprecated */
   desktop: !!cfg.desktop,
```
nl0, line 95: you can remove it from here if it's not used in the code

nl0 (commented):
awesome ui/ux feel, some code-related suggestions inline

catalog/app/containers/Bucket/PackageTree/PackageTree.tsx
fiskus: Then intention here was to:
* use default label provided `Download.Button` in every case, except...
* if the path is "" (root of the package) then use "Get package"
  + but if the selection is not empty then still use the default label

I can write

```tsx
  <Download.Button
    {...(!packageUri.path && slt.isEmpty
      ? { label: 'Get package' }
      : {})}
  >
```
if it's less confusing.

---
  
When the property is not set or set to `undefined`, it will use default value during destructuring. Maybe, you confused with `null`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring#default_value

catalog/app/embed/Dir.js
fiskus: Yes, though, it was tricky because embed authentication doesn't work well. It fails in most of the cases for me, because it seems there is duplication of the auth calls, and the second call is invalid, because auth code is already used in the first one.

catalog/app/utils/Config.ts
fiskus: It is used. I didn't want to clutter that PR with removing the use of `cfg.desktop` in components unrelated to the PR.

catalog/app/embed/Dir.js
nl0: hm, that's strange. it was working fine last time i tried

catalog/app/containers/Bucket/Download/Code.tsx
```diff
@@ -0,0 +1,103 @@
...
+      </M.Typography>
+      <div className={classes.container}>
+        <div className={classes.copy}>
+          <M.IconButton onClick={handleCopy} title="Copy to clipboard" size="small">
+            <M.Icon fontSize="inherit">file_copy</M.Icon>
+          </M.IconButton>
+        </div>
+        {lines.map((line, index) => (
+          <p key={`${line}_${index}`} className={classes.line}>
+            {highlight(line, hl)}
```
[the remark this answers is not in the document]
fiskus: I moved it to component where I can memoize with `useMemo`

catalog/app/containers/Bucket/Download/PackageOptions.tsx
fiskus: Oh, I thought, in `code={<CodePanel hashOrTag={hashOrTag} uri={uri} />}` `CodePanel` will be executed even when not rendered, but it is executed only when rendered

fiskus: I moved labels inside, used `'download' | 'code'` for tabs instead of numbers.

catalog/app/embed/Dir.js
fiskus: I wrote about that in Slack

catalog/app/containers/Bucket/Download/PackageOptions.tsx
nl0: > Oh, I thought, in code={<CodePanel hashOrTag={hashOrTag} uri={uri} />} CodePanel will be executed even when not rendered, but it is executed only when rendered

yeah, that's an important thing to know about react:
its model is actually reminiscing monads (🤫) -- you don't execute computations directly, you "describe" them
(you don't call the `CodePanel` function, you just pass it along with the args to the createElement function or smth like that),
and the runtime later executes them in a way it deems necessary.

catalog/app/containers/Bucket/Download/OptionsTabs.tsx
```diff
@@ -0,0 +1,117 @@
...
+  code: 'Code',
+}
+
+type TabsProps = Record<TabType, React.ReactNode>
+
+export function Tabs({ download, code }: TabsProps) {
+  const classes = useStyles()
+  const [activeTab, setActiveTab] = React.useState<TabType>('download')
+
+  const renderTab = React.useCallback(
```
nl0, line 96: you don't need memoization (useCallback) here, bc you don't pass that function anywhere, you just execute it

nl0 (approved):
looks good. one tiny nitpick inline

nl0 (approved):

PR #4362, opened 2025-03-25
Streamline search results

nl0:
# Streamline search results display

- More compact and lightweight cards
- Outlined paper surfaces instead of elevated ones
- More lightweight JSON display

Screenshot 2025-03-31 at 15 53 24
Screenshot 2025-03-31 at 15 55 09

## Checklist


drernie (approved):

catalog/app/components/JsonDisplay/JsonDisplay.tsx
```diff
@@ -66,14 +66,15 @@ const useStyles = M.makeStyles((t) => ({
   iconBlank: {
     paddingRight: t.spacing(2.5),
   },
-  key: {
-    fontWeight: t.typography.fontWeightBold,
+  key: {},
+  value: {
+    opacity: 0.7,
```
fiskus, line 71:
```suggestion
    color: t.palette.text.secondary
```

fiskus: Visually similar

fiskus: `.separator` and `.hint` opacity rules can also be replaced with `t.palette.text.hint`. Or, was your intention to make component independent of color scheme?

catalog/app/containers/Search/Hit.tsx
```diff
@@ -0,0 +1,535 @@
...
+          <Divider />
+          {hit.deleted ? 'Deleted' : 'Updated'}{' '}
+          <M.Tooltip title={hit.modified.toLocaleString()}>
+            <span style={{ position: 'relative' }}>
+              <Format.Relative value={hit.modified} />
+            </span>
+          </M.Tooltip>
+          <Divider />
+          <M.Tooltip title={`VersionID: ${hit.version}`}>
+            <span style={{ position: 'relative' }}>v.{hit.version.slice(0, 4)}</span>
```
fiskus, line 242:
```suggestion
            <span style={{ position: 'relative' }}>VersionId: {trimCenter(hit.version, 12)}</span>
```

fiskus: I think it should be either completely removed, or added some more info: like "VersionId" instead of "v.", and three dots. You can use `text-overflow: ellipsis` for three dots, or `trimCenter` function (though, it needs to be customized with custom divider without spaces).

fiskus: `@ver...sion` works for me too, but I get used to see a version in quilt+s3 URIs

fiskus (approved):
image

No previews look like something broken, but I don't have suggestion on how to improve this.

catalog/app/components/JsonDisplay/JsonDisplay.tsx
nl0: > Or, was your intention to make component independent of color scheme?

yep. since we don't have hardcoded colors there, i've decided to keep it that way and just use opacity

nl0:
> No previews look like something broken, but I don't have suggestion on how to improve this.

i'll see maybe i can just hide previews for empty files. thanks for spotting this

catalog/app/containers/Search/Hit.tsx
nl0: well, it doesn't seem terribly useful, but still, so i've decided to keep while de-emphasizing as much as possible.
hint has the whole version string displayed, in case anyone really needs that

PR #4366, opened 2025-03-31
Allow forms and popups in iframes when Permissive HTML Rendering enabled

nl0:
## Checklist


nl0:
@fiskus u ok with this?

fiskus (approved):

PR #4371, opened 2025-04-04
Search: Only show package comment and hash when including historical versions

nl0:
## hide

Screenshot 2025-04-08 at 16 23 30

## show

Screenshot 2025-04-08 at 16 24 18

## Checklist

### DIFF
--- a/.github/workflows/deploy-catalog.yaml
+++ b/.github/workflows/deploy-catalog.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - hide-revision-when-latest
     paths:
       - '.github/workflows/deploy-catalog.yaml'
       - 'catalog/**'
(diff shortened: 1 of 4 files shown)


fiskus:
I thought, the idea was to disable Metadata as well, because it looks scary to nontechnical users.

fiskus (approved):
I'd double-check the goal. Because, message looks reasonably helpful, but metadata is what can be hidden because it's too technical.

drernie:
@fiskus That is a fair concern.  My thinking is:

1. Users are still filtering on metadata facets, so they need to see the values of unflitertered facets to know what they are filtering against (and it feels weird, to me, not being able to confirm the packages actually contain the metadata I asked for).

2. If we are only showing one revision, the commit message isn't an important distinguished.

The counter argument, I suppose, is that users could be searching on full text that matches the commit message.  I played with the idea of making the commit message a tooltip, but that might not be worth it.

drernie:
Also, if I recall correctly, prior search results had a "static string" of metadata, so it was scary without being useful.  Now that we have interactivity, it is actually quite useful, so I don't think we want to lose it.  Making it less scary is a great goal for a future horizon...

drernie (approved):

nl0:
someone re-approve pls. i've updated the screenshots and added a changelog entry

PR #4377, opened 2025-04-09
docs: permissive rendering enables forms and popups

drernie:
# Description

Update documentation

### DIFF
--- a/docs/Catalog/Preview.md
+++ b/docs/Catalog/Preview.md
@@ -61,7 +61,7 @@ access private files in S3.
 If you wish for your HTML to access data within the enclosing package or bucket
 (at the viewer's level of permissions) and/or use origin-aware Web APIs such as
 data storage/cookies, you must opt in to `Enable permissive HTML rendering` in
-[Bucket settings](Admin.md#buckets). This explicitly allow cross-origin resource
+[Bucket settings](Admin.md#buckets). This explicitly allows cross-origin resource
 sharing (CORS).
 
 > You should _only enable this feature for buckets where you implicitly
@@ -89,8 +89,7 @@ view), the iframe gets the following origin:
 
 > New in Quilt Platform version 1.59.0 or higher
 
-Permissive HTML now includes the ability to allow forms and popups to work from
-iframes. This feature is currently only supported in Chrome and Firefox.
+Enabling Permissive HTML now allows forms and popups to work from iframes.
 
 ### Package view example with permissive rendering enabled
 
--- a/docs/Catalog/Preview.md
+++ b/docs/Catalog/Preview.md
@@ -84,6 +84,14 @@ the iframe gets the following origin:
 > An important implication of same-origin policy is that the scripts
 > executed under the same origin share LocalStorage data and cookies.
 
+### Cross-Origin Resource Sharing (CORS)
+
+> New in Quilt Platform version 1.59.0 or higher
+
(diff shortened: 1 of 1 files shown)


drernie:
@nl0 Did I phrase that correctly?

I remember it did not work with Safari, but I'm not sure if that is only for this issue.

 Please fix and/or approve.

docs/Catalog/Preview.md
```diff
@@ -84,6 +84,14 @@ the iframe gets the following origin:
 > An important implication of same-origin policy is that the scripts
 > executed under the same origin share LocalStorage data and cookies.
 
+### Cross-Origin Resource Sharing (CORS)
+
+> New in Quilt Platform version 1.59.0 or higher
+
+Enabling permissive rendering also disables CORS protections, so that forms and
```
nl0, line 91: it does not disable protections, it allows cross-origin resource sharing (meaning the scripts in html files can use external CORS-aware APIs). also, this was not changed in this release -- this was the case from the initial PFS implementation.
allowing forms and popups is new in this release, but it's kinda orthogonal to CORS.
also, i don't see any difference in how safari handles this.
i think you should just drop this section, it's misleading.

nl0 (changes requested):

docs/Catalog/Preview.md
drernie: I do think it worth documentating "forms and popups" only work on newer releases.

I will send you my credentials so you can test on Safari.

docs/Catalog/Preview.md
```diff
@@ -54,36 +54,44 @@ currently supported.
...
+default, preview iframes do not have IAM permissions and are therefore unable to
+access private files in S3.
 
 If you wish for your HTML to access data within the enclosing package or bucket
-(at the viewer's level of permissions) and/or use origin-aware Web APIs
-such as data storage/cookies, you must opt in to
-`Enable permissive HTML rendering` in [Bucket settings](Admin.md#buckets).
+(at the viewer's level of permissions) and/or use origin-aware Web APIs such as
+data storage/cookies, you must opt in to `Enable permissive HTML rendering` in
+[Bucket settings](Admin.md#buckets). This explicitly allow cross-origin resource
```
nl0, line 64:
```suggestion
[Bucket settings](Admin.md#buckets). This explicitly allows cross-origin resource
```

docs/Catalog/Preview.md
```diff
@@ -54,36 +54,44 @@ currently supported.
...
 
 > An important implication of same-origin policy is that the scripts
 > executed under the same origin share LocalStorage data and cookies.
 
+### Allowing Forms and Popups
+
+> New in Quilt Platform version 1.59.0 or higher
+
+Permissive HTML now includes the ability to allow forms and popups to work from
+iframes. This feature is currently only supported in Chrome and Firefox.
```
nl0, line 93:
```suggestion
Enabling Permissive HTML now allows forms and popups to work from iframes.
```

nl0:
the limitation in safari i mentioned elsewhere:
if my understanding is correct, it doesn't send cookies from inside iframes unless some specific conditions are met.
not sure if we can work around that in a sensible way, but [here](https://blog.certa.dev/third-party-cookie-restrictions-for-iframes-in-safari) is some more context

nl0 (approved):

PR #4379, opened 2025-04-11
Custom policy warning

drernie:
# Description

Warn users when adding a custom policy.
Point to (new) docs about ManagedUserRoleExtraPolicies

# TODO

- [x] Documentation


docs/technical-reference.md
```diff
@@ -550,6 +550,19 @@ Note the comma after the object. Your trust relationship should now look somethi
...
+The `ManagedUserRoleExtraPolicies` parameter allows you to add additional IAM policies to the managed user role. This is useful for granting additional permissions to users in your Quilt instance, which otherwise would be blocked by the permission boundary.
+
+Go to CloudFormation > Your Quilt Stack -> Update -> Parameters 
+and add the ARN of that IAM policy to  `ManagedUserRoleExtraPolicies` 
+at the bottom of the page:
+
+![](imgs/ManagedUserRoleExtraPolicies.png)
+
+If other policies are already in that field, 
+you will need to add a comma before appending the ARN.
```
sir-sigurd, line 564: that wording made sense in the previous place 
here it might make more sense to just say that it's a comma-separated list

docs/technical-reference.md
```diff
@@ -550,6 +550,19 @@ Note the comma after the object. Your trust relationship should now look somethi
 
 You can now configure a Quilt Role with this role (using the Catalog's admin panel, or `quilt3.admin.create_role`).
 
+### ManagedUserRoleExtraPolicies
+
+The `ManagedUserRoleExtraPolicies` parameter allows you to add additional IAM policies to the managed user role. This is useful for granting additional permissions to users in your Quilt instance, which otherwise would be blocked by the permission boundary.
```
sir-sigurd, line 555: IMO it's not quite clear that you need both to modify `ManagedUserRoleExtraPolicies` and add policy in Quilt Admin

sir-sigurd (commented):
mostly LGTM
needs approval from frontend guys

docs/technical-reference.md
drernie: This applies to more than just User Policies (see KMS below).  I will add clarifying text, though.

drernie:
@nl0 @fiskus Can one of you review and approve?

nl0 (dismissed):

docs/technical-reference.md
sir-sigurd: I believe that currently `ManagedUserRoleExtraPolicies` affects only "managed" user role, in KMS section permissions are granted to other roles with resource policy

drernie: Sorry, I'm not understanding your comment.  I think my PR is extending the current documentation in a correct way.  Am I mis-stating something, or is there a flaw in the original documentation we also need to fix?

sir-sigurd: see my top comment

> This applies to more than just User Policies (see KMS below)

that seems both wrong and irrelevant
KMS permissions are granted to "managed" user roles in the steps 1 and 3, and to other roles in the step 2, these "other" roles don't need step 1
so if you want to extract info about granting permissions to "managed" user roles, you need to extract both steps 1 and 3

drernie: @sir-sigurd Does the latest change address your concern?

docs/technical-reference.md
```diff
@@ -208,7 +208,30 @@ Note the comma after the object. Your trust relationship should now look somethi
...
+The `ManagedUserRoleExtraPolicies` parameter allows you to add additional IAM
+policies to the managed user role. This is useful for granting additional
+permissions to users in your Quilt instance, which otherwise would be blocked by
+the permission boundary. 
+
+This parameter works in conjunction with Quilt role configuration in the Admin panel.
+You need to:
+1. Create the appropriate IAM policy
+2. Add it to `ManagedUserRoleExtraPolicies` (this step)
+3. Create a Quilt role in the Admin panel that uses the managed user role (Source=Quilt)
```
sir-sigurd: I think it should be something like this:

3. create Quilt policy with "Manually set ARN instead of configuring per-bucket permissions" flag and enter policy ARN
4. attach that Quilt policy to managed Quilt roles

drernie:
@sir-sigurd Thanks, you're right. I've updated the `ManagedUserRoleExtraPolicies` steps in `docs/technical-reference.md` (16003c3ec) to reflect how this actually works in the Admin panel:

1. Create the IAM policy
2. Add its ARN to `ManagedUserRoleExtraPolicies`
3. In the Admin panel, create a Quilt policy with "Manually set ARN instead of configuring per-bucket permissions" and enter the same policy ARN
4. Attach that Quilt policy to the managed (Source=Quilt) roles

This replaces the earlier (inaccurate) "create a Quilt role that uses the managed user role" wording.

sir-sigurd (commented):
The revised `ManagedUserRoleExtraPolicies` steps match how managed roles actually work — I traced it through the CloudFormation template and the registry. Managed (Source=Quilt) roles share a single IAM role, and the registry assumes it with a session policy built from the Quilt policies attached to each role; a session policy can only narrow what the underlying role grants, so the extra permission genuinely has to be added in both places — on the role (the parameter) and as an attached Quilt policy. The 4-step list and the custom-role note are accurate, and this resolves my earlier comments on the section.

Two fixes and a description note below; this is likely my last pass on the docs.

**P2 — the in-app link won't land on the section.**
The checkbox warning links to `…/technical-reference#ManagedUserRoleExtraPolicies`, but the docs site lowercases heading anchors and matches them case-sensitively, so the generated id is `manageduserroleextrapolicies`. I checked on the live site: a mixed-case fragment stays at the top of the page instead of scrolling to the heading. Since the link exists to take the admin straight to that section, lowercase the fragment:

```
href="https://docs.quilt.bio/quilt-platform-administrator/technical-reference#manageduserroleextrapolicies"
```

**P3 — "permission boundary" wording in the note.**
There's no IAM permissions boundary on the managed user role; the actual constraint is the session policy the registry applies when assuming the role. Suggest rewording the note (the `**Note:**` paragraph at the end of the section):

> **Note:** This parameter is needed only for managed (Source=Quilt) roles. Quilt scopes each managed role's session to the policies attached to it, so an added permission must be granted in *both* places — on the role (this parameter) and as an attached Quilt policy (steps 3–4). For custom (Source=Custom) roles, attach policies directly in IAM without this parameter.

**PR description.**
No changelog entry needed here — but tick that box in the description's checklist (it qualifies under "skip if not significant to end users") so the TODO is squared away before merge.

Reviewed against 16003c3e

drernie:
Addressed in 006ca5c0. I lowercased the docs fragment in the Admin policy warning link so it lands on the generated `#manageduserroleextrapolicies` anchor, replaced the permission-boundary wording with the session-policy explanation for managed roles, and checked the changelog item in the PR description as docs-only/no changelog needed.

PR #4384, opened 2025-04-17
Catalog: Respect `cfg.packageRoot`: prefix physical keys when creating, and set `destPrefix` when promoting a package

fiskus:

### DIFF
--- a/catalog/config-schema.json
+++ b/catalog/config-schema.json
@@ -100,6 +100,10 @@
     "stackVersion": {
       "type": "string",
       "description": "Stack release version"
+    },
+    "packageRoot": {
+      "type": "string",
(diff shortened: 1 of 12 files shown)


catalog/config-schema.json
```diff
@@ -100,6 +100,10 @@
     "stackVersion": {
       "type": "string",
       "description": "Stack release version"
+    },
+    "packageRoot": {
+      "type": "string",
+      "description": "Package prefix where package files will be uploaded"
```
nl0, line 106:
```suggestion
      "description": "Prefix where package files will be uploaded"
```

"Package prefix" doesn't make sense to me

nl0 (approved):
lgtm sans config field description (not sure anyone will ever read that though)

PR #4398, opened 2025-05-05
Redirect from no-slash directory URL to trailing slash URL

fiskus:
If there is no trailing slash, it may be a directory

For bucket:
* make a `head` object request, to detect if the file exists
* if the file doesn't exist, make a listing request (directories must some content, because they are virtual)

For package:
* request for `dir` field in GraphQL query. If the `dir` property is present, it is a directory

- [x] Unit tests


drernie:
Thanks for turning this around so fast.  Is there any way to get this into a stack for testing?

catalog/app/containers/Bucket/HandleNoSlashDir.tsx
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      requestIsDirectory()
+        .then((isDir) => setPageType(isDir ? 'dir' : 'file'))
+        .catch((e) => setPageType(e instanceof Error ? e : new Error(`${e}`)))
+    }
+    resolvePageType()
+  }, [handle, isObject, requestIsDirectory])
+
+  return pageType
```
nl0, line 81: using state here is error-prone, because you have to catch state transitions and update computed state.
maybe try something like this?

```suggestion
  const isObject = useIsObject(handle)
  const isDirectory = useIsDirectory(handle)

  if (isObject === null) return 'loading'
  if (isObject) return 'file'
  if (isDirectory === null) return 'loading'
  return isDirectory ? 'dir' : 'file'
  // TODO: error handling
```

unlike the original version, it will always make the second (bucketListing) call, but we won't be waiting for it, so performance-wise it should be the same.
you'd have to modify `useIsDirectory` to encapsulate request state similar to `useIsObject`.
to avoid the second request, you can add a parameter (say, `pause` or smth) to `useIsDirectory` to make the request conditional.

catalog/app/containers/Bucket/HandleNoSlashDir.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { displayError } from './errors'
+
+// If object exists, then this is 100% an object page
+function useIsObject(handle: Model.S3.S3ObjectLocation) {
+  const [exists, setExists] = React.useState(null)
+
+  const s3 = AWS.S3.use()
+  React.useEffect(() => {
+    const { bucket, key, version } = handle
+    function resolveObjectExistence() {
```
nl0, line 20: why use a nested function here?

catalog/app/containers/Bucket/HandleNoSlashDir.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return pageType
+}
+
+interface HandleNoSlashDirProps {
+  children: React.ReactNode
+  handle: Model.S3.S3ObjectLocation
+}
+
+export default function HandleNoSlashDir({ children, handle }: HandleNoSlashDirProps) {
```
nl0, line 89: the name feels somewhat awkward. maybe `FallbackToDir` or smth? 🤔

nl0 (commented):
general direction seems fine, some nitpicks inline

catalog/app/containers/Bucket/HandleNoSlashDir.tsx
fiskus: Renamed

catalog/app/containers/Bucket/FallbackToDir.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+// If prefix contains at least something, then it is a directory.
+// S3 can not have empty directories, because directories are virtual,
+//    they are based on paths for existing keys (file)
+function useIsDirectory(handle: Model.S3.S3ObjectLocation, pause: boolean) {
+  const bucketListing = requests.useBucketListing()
+  const [isDir, setIsDir] = React.useState<typeof Loading | boolean | Error>(Loading)
+
+  React.useEffect(() => {
+    if (pause) return
```
nl0, line 54: it's probably worth it to reset state here as well (before `if (pause) ...`)

nl0 (commented):
looks good, tho i'm still somewhat anxious about this manual request handling with promises, useEffect and ad-hoc types.
when a promise resolves, you don't check if the component has been unmounted or if a new request was made. it seems highly unlikely to bump into this issue in practice, but it doesn't feel right anyways

catalog/app/containers/Bucket/FallbackToDir.tsx
```diff
@@ -0,0 +1,100 @@
...
+      .then(({ dirs, files }) => setIsDir(!!dirs.length || !!files.length))
+      .catch((e) => setIsDir(e instanceof Error ? e : new Error(`${e}`)))
+  }, [bucketListing, handle, pause])
+
+  return isDir
+}
+
+function useFallbackToDir(handle: Model.S3.S3ObjectLocation) {
+  const isObject = useIsObject(handle)
+  const isDirectory = useIsDirectory(handle, isObject === Loading)
```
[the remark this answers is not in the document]
fiskus: You're right, I want to pause the request when the handle is not an object. But I want to pause it when it is still `Loading` too.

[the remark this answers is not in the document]
fiskus: Well, rather than `pause`, I can use `proceed`. Then I can "proceed" to the request when it is not object.

catalog/app/containers/Bucket/FallbackToDir.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const File = Symbol('file')
+
+// If object exists, then this is 100% an object page
+function useIsObject(handle: Model.S3.S3ObjectLocation) {
+  const s3 = AWS.S3.use()
+  const [exists, setExists] = React.useState<typeof Loading | boolean | Error>(Loading)
+
+  React.useEffect(() => {
+    const { bucket, key, version } = handle
```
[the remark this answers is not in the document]
fiskus: Added

catalog/app/containers/Bucket/FallbackToDir.tsx
fiskus: Added Loading state

nl0:
> In case you meant using `useData` and `AsyncResult` I have this version #4399 But I don't know how to make it more readable and get rid of spaghetti-style code, and it is not secured with types.

i meant something like [this](https://github.com/quiltdata/quilt/pull/4400) if you want to keep more or less the same style

nl0:
> I don't know how to make it more readable and get rid of spaghetti-style code, and it is not secured with types.

yeah, it lacks ergonomic composition functionality

nl0:
and here's an [attempt](https://github.com/quiltdata/quilt/pull/4401) to make it more ergonomic with useData and AsyncResults

fiskus:
Thanks! I used typed version.
In both versions, tests were severely broken, I couldn't fix them

nl0:
> In both versions, tests were severely broken, I couldn't fix them

have you tested manually?

fiskus:
Yes

nl0 (approved):

PR #4421, opened 2025-06-25
Sync GQL stuff

nl0:
@fiskus i've implemented match locations and adjusted some other GQL stuff


nl0:
oh, i've also added lexicographical (aka alphabetic) sort orders

PR #4429, opened 2025-07-04
Fix image display in GitBook in EventBridge.md

sir-sigurd:


fiskus (approved):
Good. There is also inline image in 10'th list item after `under "Indexing and notifications"`

fiskus (approved):

PR #4431, opened 2025-07-09
Fix indexing large FCS files

sir-sigurd:
# Description

# TODO

- [ ] extract shared changes to its own PR
- [ ] bump hashes
- [ ] Unit tests


QuiltSimon:
Closing on path rot rather than on age, as part of the PR reconciliation announced in #engineering (2026-08-29).

Six of the eight files this patch touches no longer exist on master: `lambdas/indexer/index.py` is now `lambdas/indexer/src/t4_lambda_es_indexer/__init__.py`, both `requirements.txt` and both `setup.py` files were replaced by `pyproject.toml`, and `lambdas/shared/t4_lambda_shared/preview.py` moved to `lambdas/shared/src/t4_lambda_shared/preview.py` (re-verified 2026-08-31). The lambdas packaging layout migrated to uv with a `src/` layout underneath this branch, so there is nothing left to rebase onto — a revival would mean rewriting the fix from scratch against a different tree, at which point the 37 lines here are a reference at best.

Worth separating the patch from the question: if large FCS files still OOM the indexer now that `decompress_stream` landed separately, that is a live bug and deserves a fresh issue with current line numbers. Closing the diff, not the problem — reopen or ping if you disagree, @sir-sigurd.

PR #4442, opened 2025-07-16
Handle OOM in indexer lambda

sir-sigurd:
# Description

# TODO

- [ ] Unit tests
- [ ] Security: Confirm that this change meets security best practices and does not violate the security model
- [ ] Documentation

### DIFF
--- a/.github/workflows/deploy-lambdas.yaml
+++ b/.github/workflows/deploy-lambdas.yaml
@@ -2,76 +2,76 @@ name: Deploy lambdas to S3 and ECR
 
 on:
   push:
-    branches:
-      - master
+    # branches:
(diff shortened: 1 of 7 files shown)


QuiltSimon:
Closing: this patches code that no longer exists. It targets indexer/index.py and setup.py, but master's indexer is now structured as src/ with pyproject.toml and uv.lock. Also destructive as written — it comments out the entire deploy-lambdas.yaml push trigger, which would silently disable lambda deploys if merged. Handling OOM in the indexer lambda is a legitimate concern and may well still be unaddressed; it needs a fresh implementation against the current src/ layout, without the workflow edit. Branch left in place.

PR #4454, opened 2025-07-24
Update search doc

nl0:


docs/Catalog/Search.md
```diff
@@ -33,30 +33,62 @@ The search page in the catalog, accessible from the search button in the top men
...
-syntax](https://www.elastic.co/guide/en/elasticsearch/reference/6.7/query-dsl-query-string-query.html#query-string-syntax).
+NOTE: Quilt uses Elasticsearch 6.8 [query string
+syntax](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-query-string-query.html#query-string-syntax).
 
 The following are all valid search parameters:
 
-#### Fields
+#### Object-specific Fields
 
 | Syntax | Description | Example |
```
sir-sigurd: Maybe it should be `field name` instead of syntax? 🤔

docs/Catalog/Search.md
```diff
@@ -33,30 +33,62 @@ The search page in the catalog, accessible from the search button in the top men
...
+* Manifests (`mnfst`) contain package metadata, including the hash, comment, and
+  workflow information.
+* Pointers (`ptr`) represent named package revisions, and associate Manifests
+  with names and tags.
+* Entries (`entry`) represent individual objects in the package with their metadata.
+
+Top-level hits displayed by the search page are always Pointers, merged with the
+corresponding Manifest and matching Entries.
+A Pointer is considered a hit if it matches the search query, or if any of its
+associated documents, i.e. Manifest or Entries, match the search query.
```
sir-sigurd: maybe we should add a note that entry also matches when its object content matches

docs/Catalog/Search.md
```diff
@@ -33,30 +33,62 @@ The search page in the catalog, accessible from the search button in the top men
...
+
+| Syntax | Description | Example |
```
sir-sigurd: does field type affect allowed queries? if so it might make sense to document field types as well

nl0:
> do we promote these fields as something stable? I hope we don't and if we don't might make sense to make it clear

no, i think we don't

docs/Catalog/Search.md
nl0: maybe 🤔

kevinemoore (dismissed):
These docs changes are very helpful. We should also link to the schema in the docs for the Python search (https://docs.quilt.bio/quilt-python-sdk/api-reference/api#search).

drernie (dismissed):

drernie (approved):

sir-sigurd (approved):

PR #4463, opened 2025-07-31
Search: UI improvements

fiskus:
# Description
* "Create package" button on List view too when no packages
* Show `<NoValue>` when commit message is "None"
* Handle JSON parsing exceptions with `react-error-boundaries`

# TODO

- [ ] Unit tests
- [ ] Security: Confirm that this change meets security best practices and does not violate the security model
- [ ] Documentation


fiskus:
Closing in favour of three focused PRs. This branch had gone stale — its merge base was from October 2025, and `Search/model.ts` alone gained ~450 lines on master since, so the work was re-applied rather than rebased.

| This PR's item | Where it went |
|---|---|
| `<NoValue>` for a `"None"` commit message | #5232 — merged |
| `react-error-boundary` around the search model | #5233 |
| Friendlier message when a filter value won't parse | #5233 |
| Splitting `NoResults.Error`'s `kind` prop into three components | #5238 |
| "Create package" button on List view | already on master — #4539 |
| `react-error-boundary` 3.1.3 → 3.1.4, devDep → dep | obsolete — master is on `^5.0.0` as a production dependency |

Two things changed in the re-application, both deliberate:

**The error boundary does more than it did here.** This branch's fallbacks replaced the page. #5233 keeps the page chrome, and reports which filter could not be read — a malformed filter param currently escapes to `Errors/FinalBoundary`, which renders fixed text and never `error.message`, so the user gets a blank failure page with the real cause only in the console.

**The `Table/index.tsx` restructure is not carried over.** It flattened the error switch but dropped `assertNever(err)` and added a bare `@ts-expect-error` over `error.path`. #5238 keeps the switch and its exhaustiveness check, and moves only the component selection. If the flatter shape is still wanted it belongs in its own change, with the types resolved rather than suppressed.

Also left behind: the copy in this branch predates master's rewording of all three error bodies, and its `variant="h4"` heading predates the `h5` that DESIGN.md's No-Display-Font Rule now calls for. #5238 keeps master's wording and typography, so it is behaviour-preserving — the two specs that mock `NoResults` needed only their mock updated, with no assertion changed.

PR #4464, opened 2025-08-01
Change default selector_fn

kevinemoore:
# Description

Changed the default selector_fn for quilt3 push. Now, when pushing to S3, push will not copy files in the destination registry bucket.

# TODO
- [x] Unit tests
- [x] Documentation
    - [x] Markdown somewhere in docs/**/*.md that explains the feature to end users (said .md files should be linked from SUMMARY.md so they appear on https://docs.quilt.bio)


kevinemoore:
> looks okay, tho it seems like a breaking change. mind adding a changelog entry?

Thanks @nl0. I added a changelog entry. I also updated the docstring in Package.push to describe the new behavior.

nl0 (approved):

api/python/quilt3/packages.py
```diff
@@ -488,6 +488,23 @@ def _create_str(results_dict, level=0, parent=True):
...
+
+    # Copy local files only
+    @staticmethod
+    def selector_fn_copy_local(logical_key, entry):
+        return entry.physical_key.is_local()
+
+    # copy_none is intentionally not implemented to help users avoid
+    # pushing local physical keys to S3
+    # def selector_fn_copy_none(logical_key, entry):
+    #    return False
```
sir-sigurd, line 506: I'd remove it
```suggestion
```

api/python/quilt3/packages.py
```diff
@@ -1360,26 +1377,59 @@ def push(
...
-        physical keys that point to the new copies.
+        Creates a new package, or a new revision of an existing package in a
+        package registry in Amazon S3.
 
-        Note that push is careful to not push data unnecessarily. To illustrate, imagine you have
-        a PackageEntry: `pkg["entry_1"].physical_key = "/tmp/package_entry_1.json"`
+        By default, local files are copied to the destination S3 bucket and path
+        according to logical key structure. After any local objects are copied,
+        a new package manifest is created that points to the objects in their
+        new locations.
```
sir-sigurd: this doesn't describe what happens with S3 objects (but I saw the example below)

api/python/quilt3/packages.py
```diff
@@ -1360,26 +1377,59 @@ def push(
...
-        `selector_fn('entry_1', pkg["entry_1"])` returns.
+        The optional parameter `selector_fn` allows callers to choose which
+        files are copied to the destination bucket, and which retain their
+        existing physical key. When using selector functions, it is important to
+        always copy local files to S3, otherwise the resulting package will be
+        inaccessible to users accessing it from Amazon S3.
+
+        The Package class includes two additional built-in selector functions:
+
+        * `Package.selector_fn_copy_all` copies all files to the destination path.
```
sir-sigurd: > destination path

destination S3 bucket?

also might make sense to say "copies all files regardless of their current location" to make things even more clear

api/python/quilt3/packages.py
```diff
@@ -1360,26 +1377,59 @@ def push(
...
+        * `Package.selector_fn_copy_local` copies only local files to the
+          destination path. Any PackageEntry's with physical keys pointing to
+          objects in other buckets will retain their existing physical keys in
+          the resulting package.
+
+        If we have a package with entries:
+
+        * `pkg["entry_1"].physical_key = s3://bucket1/user/pkg_name/entry_1`
```
sir-sigurd: might make sense to use different key to avoid giving impression we preserve key here

api/python/quilt3/packages.py
```diff
@@ -1448,6 +1494,15 @@ def selector_fn(*args):
...
                 )
 
+        if selector_fn is None:
+            # When pushing to S3, do not copy files if they are in the same
+            # bucket as the destination registry.
+            def selector_fn(logical_key, entry):
+                if not registry_parsed.is_local():
+                    if entry.physical_key.bucket == registry_parsed.bucket:
+                        return False
+                return True
```
sir-sigurd: AFAICT we only allow push to S3 bucket so it might make sense to do this

```suggestion
        assert not registry_parsed.is_local()

        if selector_fn is None:
            # Do not copy files if they are in the same bucket as the destination registry.
            def selector_fn(logical_key, entry):
                return entry.physical_key.bucket != registry_parsed.bucket
```

sir-sigurd (commented):
generally LGTM but there are some nitpicks

PR #4467, opened 2025-08-08
Qurator devtools: model override, session recording

nl0:
This PR adds enhanced developer tools to the Qurator AI assistant interface, focusing on debugging and configuration capabilities. The changes introduce model ID override functionality and recording/logging capabilities for debugging conversations.

- Adds model ID override functionality, allowing developers to test different Bedrock models
- Implements session recording with download capabilities for debugging
- Updates the UI to surface these new developer tools in the assistant interface

Screenshot 2025-08-22 at 15 05 51

## Checklist

### DIFF
--- a/catalog/app/components/Assistant/UI/Chat/DevTools.tsx
+++ b/catalog/app/components/Assistant/UI/Chat/DevTools.tsx
@@ -1,7 +1,11 @@
 import * as Eff from 'effect'
 import * as React from 'react'
 import * as M from '@material-ui/core'
-import { Clear as ClearIcon, Delete as DeleteIcon, GetApp as GetAppIcon } from '@material-ui/icons'
+import {
+  Clear as ClearIcon,
+  Delete as DeleteIcon,
+  GetApp as GetAppIcon,
+} from '@material-ui/icons'
 
 import JsonDisplay from 'components/JsonDisplay'
 


drernie (approved):

nl0:
@fiskus ping

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
```diff
@@ -1,11 +1,144 @@
 import * as Eff from 'effect'
 import * as React from 'react'
 import * as M from '@material-ui/core'
+import ClearIcon from '@material-ui/icons/Clear'
+import DeleteIcon from '@material-ui/icons/Delete'
+import GetAppIcon from '@material-ui/icons/GetApp'
```
fiskus: I'd prefer this variant for consistency

```ts
import { Clear as IconClear, Delete as IconDelete, GetApp as IconGetApp } from '@material-ui/icons'
```

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
```diff
@@ -1,11 +1,144 @@
...
+        variant="contained"
+        color="primary"
+        size="small"
+      >
+        {enabled ? 'Stop' : 'Start'} Recording
+      </M.Button>
+
+      {(enabled || log.length > 0) && (
+        <div className={classes.label}>
+          {log.length > 0 ? `${log.length} item(s) recorded` : 'Recording...'}
```
fiskus, line 116:
```suggestion
          {log.length > 0 ? `${log.length} item(s) recorded` : 'Recording…'}
```

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
nl0:
```suggestion
import { Clear as ClearIcon, Delete as DeleteIcon, GetApp as GetAppIcon } from '@material-ui/icons'
```

nl0: tbh i'd prefer `import * as Icons` and let the bundler do its job

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
nl0: im still on the fence re: using "exotic" symbols directly in code

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
fiskus: wildcard-import makes it more difficult to control which icons we use

nl0: how and why do you want to control that?

fiskus: I want an easy way to know which icons we use and where. So, choosing an icon could be consistent and predictable.
If we decide to migrate to other icons set, it can be done programmatically

nl0: i don't see why it can't be done with wildcard imports. a little bit more involved, yes, but still pretty much doable

fiskus: I'm ok with `import * as Icons`, though. If Icons is a component in our codebase.
Like this https://github.com/quiltdata/quilt/blob/4d5d6cb9b86beb8da447eaa1191cb4c5bccaf1e3/catalog/app/components/Icon.tsx

fiskus (approved):

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
```diff
@@ -1,11 +1,146 @@
 import * as Eff from 'effect'
 import * as React from 'react'
 import * as M from '@material-ui/core'
+import {
+  Clear as ClearIcon,
```
fiskus, line 5: ClearIcon → IconClear: from less specific to more specific.

fiskus: At least, because there are`*Outlined` icons, also

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
nl0: > I'm ok with import * as Icons, though. If Icons is a component in our codebase.

that would require maintaining our subset of icons and remembering to review and prune it continuously, which seems like a maintenance burden with no clear benefit

catalog/app/components/Assistant/UI/Chat/DevTools.tsx
nl0: that makes sense, but seems quite inconsequential in this case

PR #4469, opened 2025-08-11
Qurator list models

drernie:
# Description

Dynamically generate list of available models to select from in Qurator Dev Tools

## Core Infrastructure Changes:

- Added AWS Bedrock SDK dependencies (@aws-sdk/client-bedrock and @aws-sdk/client-bedrock-runtime) to support dynamic model querying
- Created a new ListModels.ts utility that discovers available Bedrock models by region, filtering for TEXT input/output capable models and testing access permissions

## UI/UX Enhancements:

- Expanded the DevTools component with a model selection dialog that allows developers to browse and select from discovered models


drernie:
Would you be open to using a popup of available models, rather than requiring users to enter them manually?

nl0:
> Would you be open to using a popup of available models, rather than requiring users to enter them manually?

i'm not convinced this adds enough value.
this was designed as a rudimentary dev tool to test other models, and i don't feel like we should spend more effort evolving it, adding all this unnecessary complexity (new dependencies, calls, filtering, ui state).
also, the code in the PR does not even list the inference profile we use by default, since it only list the models, not the inference profiles.
is it that bad of a DX to just copypaste a model id from the docs?

QuiltSimon:
Closing: the catalog owner declined the premise on the record. nl0 (2026-08-12): 'i'm not convinced this adds enough value. this was designed as a rudimentary dev tool to test other models, and i don't feel like we should spend more effort evolving it' — followed by four conditions that would need meeting first. That is a product decision, not a staleness call, so closing rather than leaving it to rot. Same reasoning applies to its sibling #4562, closed alongside. Branch left in place in case those four conditions are ever met.

PR #4470, opened 2025-08-12
Catalog: Correctly sign and proxy S3 requests and URIs in OPEN mode

nl0:
## Checklist

- [x] Open and Embed: Confirm that this change doesn't break Open variant and Embed widget


sir-sigurd:
> @sir-sigurd i don't remember all the details around the original decision to not sign/proxy on OPEN, so i'm just defaulting to always sign if the user is authenticated and the bucket belongs to the stack.

I had *impression* that there is an easy way to allow anonymous access but (unintentionally?) not allow authenticated principals and that we have such bucket from third party on open
but now I can't find any doc references to support this
maybe I had this impression because of this logic in quilt3 which I think I never understood

https://github.com/quiltdata/quilt/blob/620302ba52e64f155db2979b15f2f0970ab1bea4/api/python/quilt3/data_transfer.py#L78-L79

> it _should_ always succeed because each stack user has specific permissions to access the allowed buckets, right?

I think yes, unless bucket has **weird** bucket policy that allows anonymous but denies some authenticated principals


so your change does make sense but I think we should test it with *all* buckets on open

catalog/app/utils/AWS/S3.js
```diff
@@ -25,41 +22,19 @@ const FORCE_PROXY = Symbol('forceProxy')
...
+  const fnRef = React.useRef()
+  fnRef.current = fn
+  return (...args) => fnRef.current(...args)
 }
 
 function useSmartS3() {
-  const isAuthenticated = useTracking(redux.useSelector(authSelectors.authenticated))
-  const isInStack = useTrackingFn(BucketConfig.useIsInStack())
-  const statusReportsBucket = useStatusReportsBucket()
+  const shouldSign = usePassThruFn(useShouldSign())
```
fiskus, line 41: It's not obvious to me why this is not `const shouldSign = useShouldSign()`, so some comment would help

fiskus (dismissed):

catalog/app/utils/AWS/S3.js
nl0: because we return a constant S3 object from this function, and that object has to be able to access a variable from the component's scope, a variable that will be changing, so we need to wrap it with an immutable reference (ref), and then access the contents of that reference dynamically

fiskus: Could you add a comment in code?

nl0: added

fiskus: Thanks

nl0:
> I think yes, unless bucket has weird bucket policy that allows anonymous but denies some authenticated principals

afair we used to hit an access denied when we tried to access a public bucket out of stack, and our iam policy did not allow that, so we had to use an unsigned request

sir-sigurd:
> afair we used to hit an access denied when we tried to access a public bucket out of stack, and our iam policy did not allow that, so we had to use an unsigned request

that looks like explanation

nl0:
> so your change does make sense but I think we should test it with all buckets on open

i did that and haven't noticed any problems (except for sometimes failing search, but that's unrelated)

PR #4477, opened 2025-08-14
Catalog: Handle search timeout

nl0:
Screenshot 2025-08-14 at 16 42 47

- [x] Open and Embed: Confirm that this change doesn't break Open variant and Embed widget

---
  - https://app.asana.com/0/0/1211092918202662


shared/graphql/schema.graphql
```diff
@@ -507,7 +507,9 @@ type SearchHitPackage {
 
   # specific props
   comment: String
-  meta: String # user_meta
+  # `meta` is the full metadata dict for entry that includes optional `user_meta` property,
+  # see PackageEntry._meta vs PackageEntry.meta.
+  meta: String
```
fiskus: Are you sure this comment is relevant to this property, and not to `SearchHitPackageMatchingEntry[meta]`?

nl0: not quite sure, but the comment was carried over from the registry

nl0: fixed, thanks

PR #4527, opened 2025-09-03
Provide `name` for `Me` query, since we use it as a cache key

fiskus:


nl0 (approved):
oh wow great find. does it fix the auth state display issue?

PR #4528, opened 2025-09-03
Fix Athena query body flickering during execution loading

fiskus:
Preserve query body during query execution loading: 

```ts
useQuery() {
   …
   if (Model.hasData(v) && queries.list.includes(v)) {
      return v
   }
   …
}
```

* Added comments
* Added spec
* Added tests

- [x] Unit tests
- [x] Documentation

---
  - https://app.asana.com/0/0/1209121754668134

### DIFF
--- a/catalog/app/containers/Bucket/Queries/Athena/spec.md
+++ b/catalog/app/containers/Bucket/Queries/Athena/spec.md
@@ -29,7 +29,7 @@ interface ValueController<T> {
 
 ### Story 1: View Execution Results
 
-**URL:** `/bucket/{bucket}/athena/workgroup/{workgroupId}/execution/{executionId}`
+**URL:** `/b/{bucket}/queries/athena/{workgroupId}/{executionId}`
 
 ```xml
 <Athena :bucket :workgroupId :executionId>
@@ -57,7 +57,7 @@ interface ValueController<T> {
 
 ### Story 2: Main Query Interface
 
-**URL:** `/bucket/{bucket}/athena/workgroup/{workgroupId}`
+**URL:** `/b/{bucket}/queries/athena/{workgroupId}`
 
 ```xml
 <Athena :bucket :workgroupId>
@@ -84,7 +84,7 @@ interface ValueController<T> {
 
 ### Story 3: Landing Page (No Workgroup Selected)
 
-**URL:** `/bucket/{bucket}/athena`
+**URL:** `/b/{bucket}/queries/athena`
 
 ```xml
 <Athena :bucket>


catalog/app/containers/Bucket/Queries/Athena/spec.md
```diff
@@ -0,0 +1,155 @@
+# Athena State Management Specification
```
drernie, line 1: Should this actually be a CLAUDE.md file, so it is automatically loaded into Claude Code when looking into this folder?

drernie (dismissed):

catalog/app/containers/Bucket/Queries/Athena/spec.md
fiskus: I don't know. All CLAUDE.md are loaded automatically? Does it affect cost or prompt size? What if we have many CLAUDE.md in different directories, or it should be one?

If you already have answers to all these questions, could you show a master class?

nl0: > Should this actually be a CLAUDE.md file, so it is automatically loaded into Claude Code when looking into this folder?

i don't think so, because its purpose seems different -- it's a spec, not preferences/standards/tools stuff.

> All CLAUDE.md are loaded automatically?

not all, only starting from the cwd and up the hierarchy.
directory-level memories are read when either starting the CLI at that directory or when it  reads a file from that directory.
some context on the lookup mechanism: https://docs.anthropic.com/en/docs/claude-code/memory#how-claude-looks-up-memories

> Does it affect cost or prompt size?

yes, since it's autoloaded, but that's not something to be afraid of.

> What if we have many CLAUDE.md in different directories, or it should be one?

that really depends on the scope.
i'd say that we should include those only if there are any special instructions to be followed.
also, i believe we should try to be as vendor-neutral as possible: if we absolutely have to, we can add a CLAUDE.md and point to a spec from there (it supports linking).

---

my 2 cents: i would probably keep the specs in the meta repo, separate from the actual code, but we don't currently have established practices around that, so it's ok to keep it here (we can always move it later if we want to)

catalog/app/containers/Bucket/Queries/Athena/spec.md
```diff
@@ -0,0 +1,155 @@
...
+**`Value<T>`** - User selections and choices. Can be `null` when user deselects.
+
+**DataController** - Manages loading operations, handles `Loading`/`Error` states.
+**ValueController** - Manages user selections, cannot set `Loading`/`Error` directly.
+
+## XML User Stories
+
+### Story 1: View Execution Results
+
+**URL:** `/bucket/{bucket}/athena/workgroup/{workgroupId}/execution/{executionId}`
```
nl0: what is this URL?

catalog/app/containers/Bucket/Queries/Athena/spec.md
```diff
@@ -0,0 +1,155 @@
...
+
+```xml
+<Athena :bucket :workgroupId :executionId>
```
nl0, line 35: what are these XMLs? do they represent the UI structure?

nl0 (dismissed):
looks reasonable, tho i have some questions about the intermediary artifacts (specs)

catalog/app/containers/Bucket/Queries/Athena/spec.md
fiskus: This is a preudo-code that represents UI data and user stories

catalog/app/containers/Bucket/Queries/Athena/spec.md
fiskus: URL shows what data comes from URL, and not from somewhere else: bucket, workgroupId, executionId

nl0: from the catalog URL? it doesn't really match any catalog URL.
not sure it really matters, but it feels confusing

fiskus: Changed to real URL patterns

catalog/app/containers/Bucket/Queries/Athena/spec.md
drernie: FYI, the rest of the industry is standardizing on AGENTS.md https://agents.md
I predict Anthropic will fold within 3 months. Until then, we can ln -s :-)

nl0 (approved):

PR #4564, opened 2025-09-26
Create to revise

fiskus:
# Description

# TODO

- [ ] Unit tests
- [ ] Documentation


catalog/app/containers/Bucket/PackageDialog/State/workflow.ts
```diff
@@ -0,0 +1,93 @@
...
+function getWorkflowFallback(manifest: ManifestStatus, config: WorkflowsConfigStatus) {
+  if (config._tag !== 'ready') return undefined
+  if (manifest._tag !== 'ready') return undefined
+
+  const workflowId = manifest.manifest?.workflowId
+  if (workflowId) {
+    const found = config.config.workflows.find((w) => w.slug === workflowId)
+    if (found) return found
+  }
+  return config.config.workflows.find((w) => w.isDefault)
```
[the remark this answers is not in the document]
fiskus: We secured existence of config.config in this condition with types

catalog/app/containers/Bucket/PackageDialog/Inputs/Files/Input.tsx
```diff
@@ -1,8 +1,8 @@
+import type { ErrorObject } from 'ajv'
 import cx from 'classnames'
 import * as R from 'ramda'
 import * as React from 'react'
 import { useDropzone } from 'react-dropzone'
-import * as RF from 'react-final-form'
 import * as M from '@material-ui/core'
```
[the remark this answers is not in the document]
fiskus: :man_shrugging:

catalog/app/containers/Bucket/PackageDialog/State/manifest.ts
```diff
@@ -0,0 +1,39 @@
...
+    bucket: src?.bucket || '',
+    name: src?.name || '',
+    hashOrTag: src?.hash,
+    pause,
+  })
+  return React.useMemo(() => {
+    if (!open) return Idle
+    if (!src) return Ready()
+    return data.case({ Ok: Ready, Pending: () => Loading, Init: () => Idle, Err })
+  }, [src, open, data])
```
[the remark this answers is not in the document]
fiskus: Not sure what does it mean exactly. Seems like everything is memoized

catalog/app/containers/Bucket/PackageDialog/State/State.tsx
```diff
@@ -0,0 +1,156 @@
...
+  progress: UploadTotalProgress
+  onAddReadme: ReadmeHandler
+
+  // Copy
+
+  copy: CopyHandler
+}
+
+// TODO: split states for Create.tsx and Copy.tsx
+//       so, for example, Copy.tsx didn't include `files` and `entriesSchema` fetching
```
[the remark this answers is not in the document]
fiskus: That shows architecture intentions, but I don't have time to address them

catalog/app/containers/Bucket/PackageDialog/State/copy.ts
```diff
@@ -0,0 +1,103 @@
...
+              if (err.path === 'params.name') {
+                fields.name = new Error(err.message)
+              } else if (err.path === 'params.message') {
+                fields.message = new Error(err.message)
+              } else if (err.path === 'params.userMeta') {
+                fields.meta = new Error(err.message)
+              } else if (err.path === 'params.workflow') {
+                fields.workflow = new Error(err.message)
+              } else {
+                error = new Error(err.message)
```
nl0: this would probably look nicer as a `switch (err.path) { ...`

catalog/app/containers/Bucket/PackageDialog/State/create.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  const create = React.useCallback(
+    async (whenNoFiles?: 'allow' | 'add-readme') => {
+      setFormStatus({ _tag: 'submitting' })
+      try {
+        if (files.status._tag === 'error') {
+          throw {
+            _tag: 'error',
+            error: new Error(
+              'Files must be finished hashing and conform entries JSON Schema',
```
nl0: this sentence does not seem to make much sense

catalog/app/containers/Bucket/PackageDialog/State/files.ts
```diff
@@ -0,0 +1,243 @@
...
+        const index: number = Number(pointer[0] as string)
+        const logicalKey = entries[index].logical_key
+        if (!memo)
+          return {
+            [logicalKey]: error,
+          }
+        return {
+          ...memo,
+          [logicalKey]: error,
+        }
```
nl0:
```suggestion
        return {
          ...(memo || {}),
          [logicalKey]: error,
        }
```

catalog/app/containers/Bucket/PackageDialog/State/files.ts
```diff
@@ -0,0 +1,243 @@
...
+    if (typeof open === 'object') {
+      setInitial(mergeFiles(manifest, { added: open }))
+    }
+  }, [open, manifest])
+  const validate = React.useMemo(() => {
+    if (schema._tag === 'error') return () => [schema.error]
+    if (schema._tag !== 'ready') return () => [new Error('Schema is not ready')]
+    return mkMetaValidator(schema.schema)
+  }, [schema])
+  const status: FilesStatus = React.useMemo(() => {
```
nl0, line 172: some empty lines between definitions would help

catalog/app/containers/Bucket/PackageDialog/State/form.ts
```diff
@@ -0,0 +1,67 @@
...
+  error: Error,
+  fields?: {
+    files?: Error
+    message?: Error
+    name?: Error
+    userMeta?: Error
+    workflow?: Error
+  },
+) => ({ _tag: 'error' as const, error, fields })
+export const Success = (handle: PackageHandle) => ({ _tag: 'success' as const, handle })
```
nl0, line 38: using Effect's [Data](https://effect.website/docs/data-types/data/) would have saved you some boilerplate, while also providing some extras like equality checks and basic pattern matching

catalog/app/containers/Bucket/PackageDialog/State/manifest.ts
nl0: greptile is somewhat right. you memoize `data`, but `src` is not memoized here. it may be memoized on the calling site, but there's no guarantee. since you only need the boolean value here, consider passing `!src` to the  dependency tuple. even without that change, i don't see anything being refetched, just rerendered with identical values (tho with the newly created ones)

catalog/app/containers/Bucket/PackageDialog/State/meta.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (schema._tag !== 'ready') return () => [new Error('Schema is not ready')]
+    return mkMetaValidator(schema.schema)
+  }, [schema])
+  const status: MetaStatus = React.useMemo(() => {
+    if (form._tag !== 'error') return Ok
+    if (form.fields?.userMeta) return Err(form.fields.userMeta)
+    const errors = validate(meta || {})
+    return errors ? Err(errors) : Ok
+  }, [form, meta, validate])
+  return React.useMemo(() => ({ value, status, onChange: setMeta }), [status, value])
```
nl0, line 54: some newlines wouldn't hurt here either

catalog/app/containers/Bucket/PackageDialog/Create.tsx
```diff
@@ -0,0 +1,448 @@
...
+    name,
+    onAddReadme,
+    params,
+    progress,
+    setDst,
+    setSrc,
+    src,
+    workflow,
+    workflowsConfig,
+  } = state
```
nl0: why destructure it here in the body, not in the signature?

catalog/app/containers/Bucket/Bucket.tsx
```diff
@@ -96,6 +96,9 @@ export default function Bucket() {
             <Route path={paths.bucketPackageList} exact>
               <PackageList />
             </Route>
+            <Route path={paths.bucketPackageAdd} exact>
+              <PackageTree />
```
nl0: why do we need to render the component twice under different `Route`s?

catalog/app/constants/routes.ts
```diff
@@ -133,17 +133,25 @@ export const bucketPackageList = route(
...
 
+interface BucketPackageAddFilesOpts {
+  [logicalKey: string]: string // S3 url
+}
+
+export const bucketPackageAddFiles = route(
+  `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/add-files`,
+  (bucket: string, name: string, files: BucketPackageAddFilesOpts = {}) =>
+    `/b/${bucket}/packages/${name}/add/${mkSearch(files)}`,
+)
```
nl0: 1. "forward" and "backward" urls don't match -- one uses `/add-files` trailing segment, and the other uses `/add/`. should be the same.
2. i don't quite get why it's necessary to have a separate route for this. why can't you use a single route with optional params/segments? i suspect this would also cause some unnecessary re-renders and state loss while navigating between those, since the downstream component (PackageTree) will be recreated with fresh state.

nl0 (commented):
looks mostly reasonable, tho it's quite hard to digest such a volume.
aside from some style- and language-related nitpicks (inline),
i have a concern about the newly added route:
first, there's a bug (typo) in the url generation (urls are mismatched).
second, i don't quite understand the reason for introducing that new route, and also the implications of using two routes for rendering the same component (e.g. losing some transient state, unnecessary re-renders, etc)

catalog/app/constants/routes.ts
fiskus: 1. It's a first step to have a separate page for creating packages
2. The new URL contains files as a query string: `?file1=s3://file1&file2=s3://file2`. So, if we keep previous format ("`package-url?action=revisePackage`"), it would conflict with files

catalog/app/containers/Bucket/PackageDialog/Create.tsx
fiskus: I don't see a reason why not. That notation is more clear for me, but I'm changing it to the signature destructure, since it's not the first time you prefer to not have such notation.

catalog/app/containers/Bucket/PackageDialog/State/manifest.ts
fiskus: Adjusted the code, however I'd prefer to pass memoized objects in the hooks, and make it a rule with some exceptions, instead of handling unmemoized objects.

catalog/app/containers/Bucket/Bucket.tsx
fiskus: I don't understand the question. Component handles multiple routes already, and it handles `paths.bucketPackageAdd` as well.

catalog/app/containers/Bucket/PackageDialog/State/form.ts
fiskus: Ok

catalog/app/containers/Bucket/PackageDialog/State/create.ts
fiskus: Could you elaborate? What's wrong: meaning or grammar?

Changed to 'Files must complete hashing and comply with the entries JSON schema'

nl0: > Could you elaborate? What's wrong: meaning or grammar?

grammar

> Changed to 'Files must complete hashing and comply with the entries JSON schema'

that's better

@drernie suggestions?

catalog/app/containers/Bucket/Bucket.tsx
nl0: well, i meant it seems more appropriate to combine the route patterns and only render the `PackageTree` component under one `Route` component

fiskus: You mean combine all three route (`bucketPackageAddFiles`, `bucketPackageDetail`, `bucketPackageTree`) and render only one?
```
          <Route path={paths.bucketPackageDetail}>
            <PackageTree />
          </Route>
```

catalog/app/containers/Bucket/PackageDialog/State/copy.ts
```diff
@@ -44,16 +44,17 @@ function useCopy() {
...
+              switch (err.path) {
+                case 'params.name':
+                  fields.name = new Error(err.message)
+                case 'params.message':
+                  fields.message = new Error(err.message)
+                case 'params.userMeta':
+                  fields.meta = new Error(err.message)
+                case 'params.workflow':
+                  fields.workflow = new Error(err.message)
+                default:
```
nl0, line 60: this branch will be always executed, since there's no `break`s in the other branches, and `default` always matches

catalog/app/containers/Bucket/Bucket.tsx
nl0: oh, there's three... ok, i guess that's fine then
(but my suggestion was to merge them all, yep)

catalog/app/containers/Bucket/PackageDialog/State/copy.ts
nl0: pls fix this and i think it's good to go

fiskus: Fixed

catalog/app/constants/routes.ts
```diff
@@ -133,17 +133,25 @@ export const bucketPackageList = route(
...
 export type BucketPackageDetailArgs = Parameters<typeof bucketPackageDetail.url>
 
+interface BucketPackageAddFilesOpts {
+  [logicalKey: string]: string // S3 url
+}
+
+export const bucketPackageAddFiles = route(
+  `/b/:bucket/packages/:name(${PACKAGE_PATTERN})/add`,
+  (bucket: string, name: string, files: BucketPackageAddFilesOpts = {}) =>
+    `/b/${bucket}/packages/${name}/add/${mkSearch(files)}`,
```
nl0:
```suggestion
    `/b/${bucket}/packages/${name}/add${mkSearch(files)}`,
```

catalog/app/containers/Bucket/PackageDialog/State/copy.ts
```diff
@@ -0,0 +1,109 @@
...
+                  break
+                case 'params.userMeta':
+                  fields.meta = new Error(err.message)
+                  break
+                case 'params.workflow':
+                  fields.workflow = new Error(err.message)
+                  break
+                default:
+                  error = new Error(err.message)
+                  break
```
nl0:
```suggestion
```

default branch does not need a break

nl0 (approved):

PR #4570, opened 2025-09-30
Iceberg lambda

sir-sigurd:
# Description

# TODO

- [x] split shared
- [x] Unit tests
- [x] update hash when https://github.com/quiltdata/quilt/pull/4595 is merged
- [x] revert CI

### DIFF
--- a/.github/workflows/deploy-catalog.yaml
+++ b/.github/workflows/deploy-catalog.yaml
@@ -21,7 +21,7 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v5
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version-file: 'catalog/package.json'
           cache: 'npm'
--- a/.github/workflows/js-ci.yml
+++ b/.github/workflows/js-ci.yml
@@ -14,7 +14,7 @@ jobs:
         working-directory: catalog
     steps:
       - uses: actions/checkout@v5
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version-file: 'catalog/package.json'
           cache: 'npm'
@@ -38,7 +38,7 @@ jobs:
       NODE_OPTIONS: --max-old-space-size=4096
     steps:
       - uses: actions/checkout@v5
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version-file: 'catalog/package.json'
       - run: npx --package=markdownlint-cli markdownlint .
--- a/api/python/quilt3/formats.py
+++ b/api/python/quilt3/formats.py
@@ -194,9 +194,9 @@ class FormatRegistry:
     @classmethod
     def object_is_serializable(cls, obj):
         try:
-            format_handlers = cls.search(type(obj))
+            cls.search(type(obj))
             return True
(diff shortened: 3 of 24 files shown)


lambdas/iceberg/pytest.ini
```diff
@@ -0,0 +1,9 @@
+[pytest]
+env =
+    AWS_ACCESS_KEY_ID = test_key
```
nl0, line 3: pls use consistent whitespace style

lambdas/iceberg/src/t4_lambda_iceberg/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        pkg_name, pointer_name = key.removeprefix(quilt_shared.const.NAMED_PACKAGES_PREFIX).rsplit("/", 1)
+        if first_line:
+            return [
+                (
+                    query_maker.package_revision_add_single
+                    if pointer_name.isnumeric()
+                    else query_maker.package_tag_add_single
+                )(bucket=bucket, pkg_name=pkg_name, pointer=pointer_name, top_hash=first_line.decode())
+            ]
+        else:
```
nl0: since you are `return`ing, it doesn't make much sense to use all these `else`s

nl0: ?

sir-sigurd: replaced with ternary

lambdas/iceberg/src/t4_lambda_iceberg/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+                query_maker.package_entry_add_single(bucket=bucket, top_hash=top_hash),
+            ]
+            if first_line
+            else [
+                query_maker.package_manifest_delete_single(bucket=bucket, top_hash=top_hash),
+                query_maker.package_entry_delete_single(bucket=bucket, top_hash=top_hash),
+            ]
+        )
+    else:
+        raise ValueError(f"Unexpected key prefix: {key}")
```
nl0, line 84: i didn't mean those specific places, i meant this whole function: each branch has a return, so you don't have to use `else`/`elif`:
```suggestion
    if key.startswith(quilt_shared.const.MANIFESTS_PREFIX):
        top_hash = key.removeprefix(quilt_shared.const.MANIFESTS_PREFIX)
        return (
            [
                query_maker.package_manifest_add_single(bucket=bucket, top_hash=top_hash),
                query_maker.package_entry_add_single(bucket=bucket, top_hash=top_hash),
            ]
            if first_line
            else [
                query_maker.package_manifest_delete_single(bucket=bucket, top_hash=top_hash),
                query_maker.package_entry_delete_single(bucket=bucket, top_hash=top_hash),
            ]
        )

    raise ValueError(f"Unexpected key prefix: {key}")
```

and imo ternaries make this less readable than just early returns

sir-sigurd: > you don't have to use else/elif:

I don't see much profit in it

> and imo ternaries make this less readable than just early returns

I don't want to avoid `else` as well because either branches are equal and part of the normal flow

nl0: 🤷

PR #4572, opened 2025-10-01
Qurator: adjust styling for tool messages

nl0:
Make tool messages less prominent

- Tool icon
- Status icons
- Tool messages now use faint(er) color with hover effects
- Tool details are collapsed by default and expandable on click

Screenshot 2025-10-01 at 12 03 35

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -19,6 +19,7 @@ where verb is one of
 ## Changes
 
 - [Changed] Qurator: make tool messages less prominent ([#4572](https://github.com/quiltdata/quilt/pull/4572))
+- [Fixed] Qurator: Limit total search results contents context to 100k characters to avoid context window overflow ([#4573](https://github.com/quiltdata/quilt/pull/4573))
 - [Fixed] Empty directory upload on Bucket tab ([#4552](https://github.com/quiltdata/quilt/pull/4552))
 - [Changed] Improve handling of no results on packages tab ([#4539](https://github.com/quiltdata/quilt/pull/4539))
 - [Changed] Removed unnecessary blocks and buttons from deleted file page ([#4544](https://github.com/quiltdata/quilt/pull/4544))
--- a/catalog/app/containers/Search/AssistantContext.tsx
+++ b/catalog/app/containers/Search/AssistantContext.tsx
@@ -178,15 +178,16 @@ function useSearchContextModel(): MaybeEitherSearchContext {
   )
 }
 
-const MAX_CONTENT_LENGTH = 10_000
+const MAX_CONTENT_TOTAL_LENGTH = 100_000
 
-function truncateIndexedContent(hit: FirstPageHits[number]) {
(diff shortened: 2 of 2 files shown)


catalog/app/components/Assistant/UI/Chat/Chat.tsx
```diff
@@ -2,6 +2,7 @@ import cx from 'classnames'
 import * as Eff from 'effect'
 import * as React from 'react'
 import * as M from '@material-ui/core'
+import * as Icons from '@material-ui/icons'
```
fiskus, line 5: Please, use destructuring here and direct names for icons you are exporting

`import { CheckCircleOutline as IconCheckCircleOutline, ErrorOutline as IconErrorOutline } from '@material-ui/icons'`

nl0: i don't think it's a good idea -- it adds verbosity and repetition for no real benefit. bundle size effect seems to be the same, since it's getting tree-shaked ("Everything is in check (+97.16KB, -96.62KB, +0.0%)").
to determine the exact icons used throughout the codebase you can just grep for `Icons.`
i believe qualified (namespaced) imports are better in general, because the qualifiers (namespaces) add helpful context about the variables, and reduce maintenance burden (you don't have to add/remove individual imports).

fiskus: I'm ok with both formats. If you're going to use wildcard imports, then please change existing imports of icons

nl0: actually, mining the codebase for all the icons used (if that's actually something that will be done)
is easier with namespace imports, because it will give you the exact icon names in a single step (e.g. `... Icons.Error ...`), while doing so for individual import requires two steps: grep for `'@material-ui/icons'` import, and then see all the individual imports (because in some cases they will be split across multiple lines).

so, using individual imports seems like a lose-lose scenario (more effort, less value).

nl0: > I'm ok with both formats. If you're going to use wildcard imports, then please change existing imports of icons

in this PR?

fiskus: It makes more sense to do this in a standalone PR

nl0: k ill do that

nl0:
@fiskus ping

fiskus (approved):

catalog/app/components/Assistant/UI/Chat/Chat.tsx
fiskus: When are you planning to make a new PR?

nl0: probably today if i have time. maybe tomorrow

fiskus: Good

PR #4582, opened 2025-10-08
Package revisions comparison

fiskus:
Major concern, that we can keep as "known issue": we load manifests with all entries on frontend to compare

Screenshot from 2025-10-08 18-29-19
Screenshot from 2025-10-08 18-29-57

- [ ] ~Unit tests~ (Good candidate for that, but didn't have time)
- [ ] ~Open and Embed~ (Didn't check)
- [ ] ~Documentation~ (Nice to have, but didn't have time)
    - [ ] ~For end users~
    - [ ] ~For developers~

### DIFF
--- a/catalog/app/containers/Bucket/PackageCompare/RevisionsRange/useRevisionsList.ts
+++ b/catalog/app/containers/Bucket/PackageCompare/RevisionsRange/useRevisionsList.ts
@@ -1,13 +1,14 @@
 import * as React from 'react'
 
-import * as Model from 'model'
 import * as GQL from 'utils/GraphQL'
 
 import REVISIONS_LIST_QUERY from './gql/RevisionsList.generated'
 
-export type RevisionsListItem = Pick<
-  Model.GQLTypes.PackageRevision,
-  'hash' | 'message' | 'modified'
+type QueryData = GQL.DataForDoc<typeof REVISIONS_LIST_QUERY>
+type PackageData = NonNullable<QueryData['package']>
+export type RevisionsListItem = Omit<
+  NonNullable<PackageData['revisions']['page'][number]>,
+  '__typename'
 >
 
 type Result =


catalog/app/containers/Bucket/PackageCompare/Diff/Entries.tsx
```diff
@@ -0,0 +1,192 @@
...
+  | { _tag: 'unmodified'; entry: Model.PackageEntry }
+  | { _tag: 'added'; entry: Model.PackageEntry }
+  | { _tag: 'removed'; entry: Model.PackageEntry }
+  | { _tag: 'modified'; base: Model.PackageEntry; other: Model.PackageEntry }
+
+function isModified(base: Model.PackageEntry, other: Model.PackageEntry): boolean {
+  if (base.physicalKey !== other.physicalKey) return true
+  if (base.hash.value !== other.hash.value) return true
+  if (base.size !== other.size) return true
+  if (JSON.stringify(base.meta) !== JSON.stringify(other.meta)) return true
```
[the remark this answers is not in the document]
fiskus: Technically, that's not **false** positive

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
```diff
@@ -0,0 +1,350 @@
...
+
+  if (changes.length === 0) {
+    return <M.Typography className={classes.empty}>Nothing changed</M.Typography>
+  }
+
+  return (
+    <M.List dense>
+      {changes.map((c, index) => (
+        <SummaryItem change={c} key={index} />
+      ))}
```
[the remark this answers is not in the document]
fiskus: Good enough

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/RevisionsRange.tsx
```diff
@@ -0,0 +1,101 @@
...
+  onSwap,
+}: RevisionsProps) {
+  const classes = useStyles()
+  return (
+    <div className={classes.root}>
+      <div className={classes.range}>
+        <RevisionSelect revisions={revisions} value={base.hash} onChange={onBaseChange} />
+        <RevisionSelect
+          revisions={revisions}
+          value={other?.hash || ''}
```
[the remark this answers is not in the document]
fiskus: hash can't be empty, otherwise it's a wrong hash

catalog/app/containers/Bucket/PackageCompare/Diff/Preview.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+function Preview({ handle }: PreviewProps) {
+  return (
+    <Load handle={handle} options={{ context: CONTEXT.LISTING }}>
+      {(data: any) => (
+        <Display
+          data={data}
+          noDownload={undefined}
+          renderContents={(children: any) =>
+            (<PreviewContents children={children} />) as $TSFixMe
```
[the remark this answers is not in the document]
fiskus: Can't do anything here

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/RevisionsRange.tsx
```diff
@@ -0,0 +1,101 @@
...
+          onChange={onOtherChange}
+          other
+        />
+      </div>
+      <M.IconButton className={classes.swap} onClick={onSwap} disabled={!other}>
```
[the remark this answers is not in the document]
fiskus: Good enough

catalog/app/containers/Bucket/PackageCompare/PackageCompare.tsx
```diff
@@ -0,0 +1,181 @@
...
+    changeOther,
+    swap,
+
+    changesOnly,
+    toggleChangesOnly,
+  } = useRouter()
+
+  return (
+    <>
+      <MetaTitle>{[`${name} comparison`, bucket]}</MetaTitle>
```
nl0:
```suggestion
      <MetaTitle>{[`${name} revision comparison`, bucket]}</MetaTitle>
```

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+type WhatChanged =
+  | { _tag: 'meta'; keys: Change[] }
+  | {
+      _tag: 'modified'
+      logicalKey: string
+      hashChanged: boolean
+      sizeChanged: boolean
+      oldSize?: number
+      newSize?: number
```
nl0, line 25: you can probably consolidate this into a single field, smth like `sizeChanged: false | { oldSize, newSize }` -- this would also make invalid state impossible to represent (oldSize and/or newSize set while sizeChange is false)

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
```diff
@@ -0,0 +1,306 @@
...
+      )}
+    </span>
+  )
+}
+
+function getMetaChange([base, other]: [Revision, Revision]): Extract<
+  WhatChanged,
+  { _tag: 'meta' }
+> | null {
+  if (JSON.stringify(base.userMeta) === JSON.stringify(other.userMeta)) return null
```
nl0: it is possible for this to return a false-negative, since the keys are not guaranteed to be sorted:
```
> JSON.stringify({a:1,b:2})
'{"a":1,"b":2}'
> JSON.stringify({b:2,a:1})
'{"b":2,"a":1}'
```

so i'd suggest you always rely on the deep comparison here, and return null if the returned list of changes is empty (or consistently return an object with an empty change list)

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
```diff
@@ -0,0 +1,306 @@
...
+    if (!baseEntry) {
+      entryChanges.push({ _tag: 'added', logicalKey })
+    } else if (!otherEntry) {
+      entryChanges.push({ _tag: 'removed', logicalKey })
+    } else {
+      const hashChanged = baseEntry.hash.value !== otherEntry.hash.value
+      const sizeChanged = baseEntry.size !== otherEntry.size
+      const physicalKeyChanged = baseEntry.physicalKey !== otherEntry.physicalKey
+      const metaChanged =
+        JSON.stringify(baseEntry.meta) !== JSON.stringify(otherEntry.meta)
```
nl0: it would be more correct to use proper deep comparison here as well

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+      if (hashChanged || sizeChanged || physicalKeyChanged || metaChanged) {
```
nl0, line 172: `physicalKeyChanged` and `metaChanged` are not included in the resulting entry (related to our discussion from yesterday about the confusion around "Modified" status)

catalog/app/containers/Bucket/PackageCompare/Diff/Metadata.tsx
```diff
@@ -0,0 +1,123 @@
...
+import Revisioned from './Revisioned'
+
+type Change =
+  | { _tag: 'added'; hash: string; value: string }
+  | { _tag: 'removed'; hash: string; value: string }
+  | { _tag: 'unmodified'; value: string }
+
+// We believe showing braces frighten wet scientists
+function removeBraces(input: string) {
+  return input.replace(/}/g, '').replace(/{/g, '').replace(/]/g, '').replace(/\[/g, '')
```
nl0: won't this also remove the braces/brackets from inside the string values?

nl0: 🤔 maybe format the objects as yaml instead, and then diff the lines? (have to make sure the keys are sorted consistently though).
that would make it even less noisy (remove unnecessary quotes, for example), while keeping it far less confusing for people familiar with json, and also keeping the representation "correct"

catalog/app/containers/Bucket/PackageCompare/Diff/Metadata.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+function MetadataDiff({
+  revisions,
+  changesOnly: changesOnly = false,
+}: MetadataDiffProps) {
+  const colors = useColors()
+  const classes = useStyles()
+
+  const changes = React.useMemo(
+    () => getChanges(revisions, changesOnly),
```
nl0: maybe split diffing and filtering-out unmodified lines into separate `useMemo()`s for _some_ potential perf gain?

catalog/app/containers/Bucket/PackageCompare/Diff/Entries.tsx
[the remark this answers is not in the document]
nl0: > Technically, that's not false positive

why?

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/gql/RevisionsList.graphql
```diff
@@ -0,0 +1,13 @@
+query ($bucket: String!, $name: String!) {
+  package(bucket: $bucket, name: $name) {
+    bucket
+    name
+    revisions {
+      page(number: 1, perPage: 5) {
```
nl0: does this mean you only allow selecting one of the first 5 revisions?

nl0: you should either modify this to accept `page` and `perPage` arguments, or don't pass them to this query
(in `catalog/app/containers/Bucket/PackageCompare/RevisionsRange/useRevisionsList.ts:22-23`)

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/useRevisionsList.ts
```diff
@@ -0,0 +1,34 @@
...
+  | { _tag: 'loading' }
+  | { _tag: 'error'; error: Error }
+  | { _tag: 'ok'; revisions: ReadonlyArray<RevisionsListItem> }
+
+export default function useRevisionsList(bucket: string, name: string): Result {
+  const query = GQL.useQuery(REVISIONS_LIST_QUERY, {
+    bucket,
+    name,
+    page: 1,
+    perPage: 100, // Get enough revisions for the dropdown
```
nl0: the query does not accept these arguments. not sure why this doesn't trigger a type error

catalog/app/containers/Bucket/PackageCompare/router.ts
```diff
@@ -0,0 +1,80 @@
...
+  }, [bucket, name, push, baseHash, otherHash, urls, showAll])
+
+  const changesOnly = !showAll || showAll === 'false'
+  const toggleChangesOnly = React.useCallback(
+    (checked: boolean) => {
+      const route = checked
+        ? urls.bucketPackageCompare(bucket, name, baseHash, otherHash)
+        : urls.bucketPackageCompare(bucket, name, baseHash, otherHash, {
+            showAll: true,
+          })
```
nl0:
```suggestion
      const route = urls.bucketPackageCompare(bucket, name, baseHash, otherHash, { showAll: checked })
```

won't this work?

catalog/app/containers/Bucket/PackageCompare/useRevisionsPair.ts
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react'
+
+import * as Model from 'model'
+import * as GQL from 'utils/GraphQL'
+import type { PackageHandle } from 'utils/packageHandle'
+
+import REVISION_QUERY from './gql/Revision.generated'
+
+export type Revision = Pick<
```
nl0: this manual type construction is quite fragile and error-prone. prefer inferring the type from the query instead:
```ts
type QueryData = GQL.DataForDoc<typeof REVISION_QUERY>
type PackageData = NonNullable<QueryData['package']>
type Revision = NonNullable<PackageData['revision']>
```

it's somewhat verbose due to nullable prop chains, but it's guaranteed to be in sync with the gql doc

catalog/app/containers/Bucket/PackageCompare/PackageCompare.tsx
fiskus: I changed to "Comparing changes in ${name} revisions" in https://github.com/quiltdata/quilt/pull/4587, similar to how GitHub phrases it

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
fiskus: Yes, I intended to do this

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
fiskus: Thanks!
I thought, that if objects sorted differently, we must show that, because it changes the hash. But now I remember, we sort objects automatically before calculating checksum.

Changed in https://github.com/quiltdata/quilt/pull/4582/commits/73d33deb5c2498b1cd1c1c81442322f8898b84af

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
fiskus: Done

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
fiskus: I think, we agreed that there is no clear quick answer on what to write instead of "modified".

catalog/app/containers/Bucket/PackageCompare/Diff/Metadata.tsx
fiskus: But the second `useMemo` will be dependent on the result of first one? Why would it make a difference? Because, they will be calculated in a separate rendering ticks?

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/gql/RevisionsList.graphql
fiskus: Hardcoded 100 revisions in `".graphl"`

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/useRevisionsList.ts
fiskus: Hardcoded 100 revisions in graphql file

catalog/app/containers/Bucket/PackageCompare/router.ts
fiskus: `showAll` in the URL is opposite to the "Changes only" checkbox, so `{ showAll: !checked }`

Simple `const route = urls.bucketPackageCompare(bucket, name, baseHash, otherHash, { showAll: !checked })` always sets some value, but I wanted to remove `showAll` if it is false

catalog/app/containers/Bucket/PackageCompare/useRevisionsPair.ts
fiskus: Won't typescript throw an error if these types are not in sync?

Types are not only for machines, but for humans too. It could be a limitation of my setup, but it is much easier to me working with the former type, than with the latter:

Screenshot from 2025-10-10 15-56-24
Screenshot from 2025-10-10 15-57-01

catalog/app/containers/Bucket/PackageCompare/PackageCompare.tsx
fiskus: I resolve this, since you approved that PR

catalog/app/containers/Bucket/PackageCompare/router.ts
nl0: oh, i missed some details, sorry. i meant that `mkSearch` filters out nulls/undefineds, so you can do smth like `{ showAll: !checked || null }`

fiskus: Thanks! Changed here https://github.com/quiltdata/quilt/pull/4588

catalog/app/containers/Bucket/PackageCompare/useRevisionsPair.ts
nl0: > Won't typescript throw an error if these types are not in sync?

not necessarily, if they are still "compatible"

> It could be a limitation of my setup, but it is much easier to me working with the former type, than with the latter

well, it's related to the fact that you don't include `__typename`.
i can achieve the same result by omitting that explicitly, but technically that would be less correct, because that field is there, you're just not using it:

Screenshot 2025-10-10 at 16 57 09

catalog/app/containers/Bucket/PackageCompare/Diff/Metadata.tsx
fiskus: Good idea, implemented here https://github.com/quiltdata/quilt/pull/4588

catalog/app/containers/Bucket/PackageCompare/Diff/Metadata.tsx
nl0: only the second one would be recomputed when toggling `changesOnly`, the first one would not (and that seems to be more compute-intensive afaict, though i'm not sure it really is significant -- probably you could try comparing some manifests with huge metadata diffs, toggle `changesOnly`, and see if the UI freezes while the diff is being recomputed, and then compare that to the memoized version).

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
nl0: well, it seems like anything would be better (less confusing), then just showing nothing, e.g.
"Metadata changed", "Physical key changed"

catalog/app/containers/Bucket/PackageCompare/Diff/Metadata.tsx
fiskus: Thanks! Done in https://github.com/quiltdata/quilt/pull/4591

catalog/app/containers/Bucket/PackageCompare/useRevisionsPair.ts
fiskus: Done in https://github.com/quiltdata/quilt/pull/4591

catalog/app/containers/Bucket/PackageCompare/RevisionsRange/useRevisionsList.ts
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react'
+
+import * as Model from 'model'
+import * as GQL from 'utils/GraphQL'
+
+import REVISIONS_LIST_QUERY from './gql/RevisionsList.generated'
+
+export type RevisionsListItem = Pick<
+  Model.GQLTypes.PackageRevision,
```
nl0: same here re: manual vs inferred type

fiskus: Done

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
fiskus: WIP in a standalone PR

catalog/app/containers/Bucket/PackageCompare/Diff/Summary.tsx
fiskus: WIP in a standalone PR

PR #4595, opened 2025-10-14
py-shared: add athena/iceberg utils

sir-sigurd:
# Description

# TODO

- [x] Unit tests

### DIFF
--- a/py-shared/src/quilt_shared/athena.py
+++ b/py-shared/src/quilt_shared/athena.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
-import typing as T
 import time
-
+import typing as T
 
 if T.TYPE_CHECKING:
+    import logging
+
     from mypy_boto3_athena import AthenaClient
     from mypy_boto3_athena.type_defs import QueryExecutionTypeDef
-    import logging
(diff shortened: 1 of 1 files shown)


py-shared/src/quilt_shared/athena.py
```diff
@@ -0,0 +1,133 @@
...
+
+class AthenaQueryCancelledException(AthenaQueryBaseException):
+    state = "CANCELLED"
+
+
+# XXX: this is mostly copy-pasted from access_counts lambda, should be deduplicated
+class QueryRunner:
+    def __init__(self, *, logger: logging.Logger, athena: AthenaClient, database: str, workgroup: str):
+        # XXX: shouldn't we use its own logger?
+        self.logger = logger
```
nl0, line 41: you could probably use the dataclass decorator to avoid manually assigning all these fields

sir-sigurd: might make sense
though I'd prefer to have keyword only parameters, but it [requires python 3.10](https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY)

PR #4605, opened 2025-10-15
enable E501 in ruff

sir-sigurd:
<h2>Greptile Overview</h2>

Updated On: 2025-10-15 13:11:15 UTC

<h3>Summary</h3>

Enabled E501 (line-too-long) enforcement in ruff by removing it from the ignore list, ensuring all Python code adheres to the 119-character line limit.

- Removed E501 from `ruff.toml` ignore list to enforce line length checks
- Fixed the one existing violation in `athena.py` by moving an inline comment to a separate line

- This PR is safe to merge with no risk
- The changes are purely cosmetic - enabling a linter rule and fixing the single violation by reformatting a comment. No functional code was modified, and the change improves code quality enforcement going forward.
- No files require special attention

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| ruff.toml | 5/5 | Removed E501 (line-too-long) from ignored rules to enforce line length limit |
| py-shared/src/quilt_shared/athena.py | 5/5 | Split overly long comment line to comply with 119 character limit |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Dev as Developer
    participant Ruff as Ruff Linter
    participant Code as Codebase
    
    Dev->>Ruff: Remove E501 from ignore list
    Note over Ruff: E501 (line-too-long) now enforced
    Ruff->>Code: Scan for lines > 119 chars
    Code-->>Ruff: Found line 113 in athena.py (122 chars)
    Dev->>Code: Split comment to separate lines
    Note over Code: Line 113: comment moved<br/>Line 114: code remains
    Ruff->>Code: Re-scan all files
    Code-->>Ruff: All lines comply with 119 char limit
    Ruff-->>Dev: Linting passes ✓
```

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -24,7 +24,6 @@ select = [
 
 # Ignore rules that were disabled in pylint or would cause too many changes
 ignore = [
-    "E501",  # line-too-long (already handled by line-length setting)
     "E731",  # lambda assignment
     "B008",  # function calls in argument defaults
     "B904",  # raise without from
--- a/py-shared/src/quilt_shared/athena.py
+++ b/py-shared/src/quilt_shared/athena.py
@@ -110,7 +110,8 @@ class QueryRunner:
         results: list[QueryExecutionTypeDef | None] = [None] * len(query_list)
 
         remaining_queries = list(enumerate(query_list))
-        remaining_queries.reverse()  # Just to make unit tests more sane: we use pop() later, so keep the order the same.
+        # Just to make unit tests more sane: we use pop() later, so keep the order the same.
+        remaining_queries.reverse()
         pending_execution_ids = {}
 
         while remaining_queries or pending_execution_ids:


ruff.toml
```diff
@@ -24,7 +24,6 @@ select = [
 
 # Ignore rules that were disabled in pylint or would cause too many changes
 ignore = [
-    "E501",  # line-too-long (already handled by line-length setting)
```
nl0, line 27: i don't think we should do that, because that applies two similar, but not exactly the same checks. we already handle that by the formatting task, so if anything has to be reformatted due to line length, it will cause an error in the formatting ci job

sir-sigurd: > i don't think we should do that, because that applies two similar, but not exactly the same checks.

I don't understand that

nl0: i mean ruff formatting logic is not that strict about the line lengths compared to the linting logic,
~~e.g. it doesn't complain on that comment line you changed in this PR, otherwise the formatting job would fail~~
oh i see we don't have a repo-wide formatting check in CI.
but anyways, i don't think we want two CI jobs checking one thing -- it's enough to enforce this in the formatting job (which is commented-out currently)

nl0: has `ruff format` done that change, btw?

sir-sigurd: > has ruff format done that change, btw?

no

> but anyways, i don't think we want two CI jobs checking one thing

I don't see any problems with it and I don't want to have long lines with comments like I fixed in this PR (or with long string which AFAICT ruff won't reformat as well)

we have both black and pycodestyle in enterprise and I don't think it caused any issues
also prior to using ruff we had E501-like check in this repo

QuiltSimon:
Closing this on the maintainer's own objection rather than on age.

nl0's review thread from 2025-10-15 is still unresolved and reads: *"i don't think we should do that, because that applies two similar, but not exactly the same checks. we already handle that by the formatting task, so if anything has to be reformatted due to line length, it will cause an error in the formatting ci job."*

That's a rejection of the approach, not of the timing. And because `quilt/master` has `required_conversation_resolution` enabled, that single thread blocks this PR indefinitely regardless of CI state — so leaving it open costs review attention without any path to merging.

To be clear about what I did *not* conclude: **E501 is genuinely still ignored on master** (`ruff.toml:30`, `"E501",  # line-too-long (already handled by line-length setting)`). So the observation behind this PR remains accurate. nl0's position is that ruff's formatter already covers it — the disagreement is about whether the lint adds anything over the formatter, which is his call to make.

**Mechanical state**: 396 commits behind master, and three required contexts (`test-client (ubuntu-latest, 3.14)`, `test-client (windows-latest, 3.14)`, `test-lambda (iceberg)`) have never run on this head — they were added after the branch was cut. The green checks describe a CI configuration that no longer exists.

Reopen if nl0 changes his position on the formatter-vs-linter question. The branch survives closure and `refs/pull/4605/head` remains fetchable.

Closing as part of a reconciliation sweep.

PR #4657, opened 2025-12-04
`jest` → `vitest`

fiskus:
- [x] Unit tests
- [ ] ~~Open and Embed: Confirm that this change doesn't break Open variant and Embed widget~~
- [x]  Confirm that this change doesn't break PRODUCT mode


nl0 (approved):
nice

PR #4660, opened 2025-12-09
Catalog frontend tests: better isolation with `globals: false`

fiskus:
* Remove `globals: true` from vitest config
* Make `makeStyles` more deterministic
* Resolve other minor issues

- [x] Unit tests

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR improves test isolation by disabling Vitest's global API mode (`globals: false`), requiring explicit imports of test functions like `describe`, `it`, `expect`, `beforeEach`, and `afterEach`.

**Key changes:**
- Removed deprecated `makeStyles` mock that's no longer needed
- Added explicit `cleanup` calls in `afterEach` hooks for proper DOM cleanup between tests
- Replaced snapshot tests with targeted assertions using `getByText`, `getByTitle`, and other query methods
- Fixed console mocking to use proper lifecycle management with `beforeEach`/`afterEach` and `.mockRestore()`
- Switched from `id` attributes to `data-testid` for better test semantics
- Added missing import (`JsonPointer`) that was previously relying on global scope

All 56 test files updated consistently follow the same pattern for better test isolation and deterministic behavior.

- This PR is safe to merge with minimal risk
- The changes are purely test-related with no production code impact. All modifications follow consistent patterns across 56 files: explicit imports, proper cleanup lifecycle, and replacing fragile snapshot tests with targeted assertions. The removal of `globals: true` enforces better test isolation practices.
- No files require special attention

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| catalog/vitest.config.ts | 5/5 | Removed `globals: true` to enable strict test isolation, requiring explicit imports of test functions |
| catalog/app/components/BreadCrumbs/BreadCrumbs.spec.tsx | 5/5 | Added explicit test imports, cleanup lifecycle, replaced snapshots with targeted assertions |
| catalog/app/components/FileEditor/FileEditor.spec.tsx | 5/5 | Added cleanup, switched from id to data-testid attributes, replaced snapshots with assertions |
| catalog/app/utils/packageHandle.spec.ts | 5/5 | Improved console mocking with proper beforeEach/afterEach lifecycle and cleanup |
| catalog/app/components/JsonEditor/State.spec.js | 5/5 | Added missing import, replaced snapshots with explicit object/array assertions |
| catalog/app/containers/Bucket/File/Toolbar/Toolbar.spec.tsx | 5/5 | Added cleanup and afterEach, replaced snapshots with targeted DOM queries and assertions |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Dev as Developer
    participant Config as vitest.config.ts
    participant Tests as Test Files
    participant Vitest as Vitest Runner
    participant DOM as Test DOM

    Dev->>Config: Remove `globals: true`
    Note over Config: Disables global test API
    
    Dev->>Tests: Add explicit imports
    Note over Tests: import { describe, it, expect, afterEach }
    
    Dev->>Tests: Add afterEach(cleanup)
    Note over Tests: Ensure DOM cleanup between tests
    
    Dev->>Tests: Replace snapshots with assertions
    Note over Tests: Use getByText, getByTitle, etc.
    
    Dev->>Tests: Fix console mocking
    Note over Tests: beforeEach: mock, afterEach: restore
    
    Dev->>Tests: Remove makeStyles mock
    Note over Tests: No longer needed
    
    Vitest->>Tests: Run test suite
    Tests->>DOM: Render component
    Tests->>DOM: Query elements
    DOM-->>Tests: Return elements
    Tests->>Tests: Assert behavior
    Tests->>DOM: cleanup()
    DOM-->>Tests: Clear DOM
    
    Tests-->>Vitest: All tests pass
    Vitest-->>Dev: ✓ Better isolation achieved
```

### DIFF
--- a/catalog/app/containers/Bucket/Queries/Athena/model/requests.spec.ts
+++ b/catalog/app/containers/Bucket/Queries/Athena/model/requests.spec.ts
@@ -128,11 +128,15 @@ describe('containers/Bucket/Queries/Athena/model/requests', () => {
       const { result, waitForValueToChange, unmount } = renderHook(() =>
         requests.useCatalogNames('any'),
       )
-      expect(result.current.data).toBe(undefined)
 
-      await waitForValueToChange(() => result.current, { timeout: 5000 })
-      expect(result.current.data).toMatchObject({ list: ['bar', 'foo'] })
-      unmount()
+      try {
+        expect(result.current.data).toBe(undefined)
+
+        await waitForValueToChange(() => result.current, { timeout: 5000 })
+        expect(result.current.data).toMatchObject({ list: ['bar', 'foo'] })
+      } finally {
+        unmount()
+      }
     })
 
     it('return empty list', async () => {
@@ -222,10 +226,14 @@ describe('containers/Bucket/Queries/Athena/model/requests', () => {
         const { result, unmount, waitFor } = renderHook(() =>
           requests.useCatalogNames('any'),
         )
-        await waitFor(() => result.current.data instanceof Error)
-        expect(Log.error).toBeCalledWith(expect.any(Error))
-        expect(result.current.data).toBeInstanceOf(Error)
-        unmount()
+
+        try {
+          await waitFor(() => result.current.data instanceof Error)
+          expect(Log.error).toBeCalledWith(expect.any(Error))
+          expect(result.current.data).toBeInstanceOf(Error)
+        } finally {
+          unmount()
+        }
       })
     })
 
@@ -267,21 +275,25 @@ describe('containers/Bucket/Queries/Athena/model/requests', () => {
         (x: Parameters<typeof requests.useCatalogName>) => useWrapper(x),
         { initialProps: [undefined, null] },
       )
-      expect(result.current.value).toBe(undefined)
 
-      const error = new Error('Fail')
-      await act(async () => {
-        rerender([error, null])
-        await waitForNextUpdate()
-      })
-      expect(result.current.value).toBe(error)
+      try {
+        expect(result.current.value).toBe(undefined)
 
-      await act(async () => {
-        rerender([{ list: ['foo', 'bar'] }, null])
-        await waitForNextUpdate()
-      })
-      expect(result.current.value).toBe('foo')
-      unmount()
+        const error = new Error('Fail')
+        await act(async () => {
+          rerender([error, null])
+          await waitForNextUpdate()
+        })
+        expect(result.current.value).toBe(error)
+
+        await act(async () => {
+          rerender([{ list: ['foo', 'bar'] }, null])
+          await waitForNextUpdate()
+        })
+        expect(result.current.value).toBe('foo')
+      } finally {
+        unmount()
+      }
     })
 
     it('switch catalog when execution query loaded', async () => {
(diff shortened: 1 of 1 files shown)


catalog/app/components/Buttons/WithPopover.spec.tsx
```diff
@@ -1,11 +1,15 @@
...
 import * as Icons from '@material-ui/icons'
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, afterEach } from 'vitest'
 
 import WithPopover from './WithPopover'
 
 describe('components/Buttons/WithPopover', () => {
+  afterEach(() => {
+    cleanup()
+  })
```
nl0:
```suggestion
  afterEach(cleanup)
```

can it be done like that?

catalog/app/components/FileEditor/routes.spec.ts
```diff
@@ -93,9 +93,14 @@ describe('components/FileEditor/routes', () => {
...
   describe('useParams', () => {
     it('should throw error when no bucket', () => {
+      const restoreConsole = suppressErrorOutput()
+
       useParamsInternal.mockImplementationOnce(() => ({}))
       const { result } = renderHook(() => useParams())
+
       expect(result.error?.message).toBe('`bucket` must be defined')
+
+      restoreConsole()
```
nl0: this will not be called if the expectation fails, afaict, so you should probably wrap this into either try/catch/finally or before/after

catalog/app/containers/Bucket/Dir/Toolbar/__snapshots__/Toolbar.spec.tsx.snap
```diff
@@ -42,7 +42,7 @@ exports[`Toolbar > should render all buttons when all features are enabled 1`] =
 
 exports[`Toolbar > should render buttons for enabled features: add, organize 1`] = `
 <div
-  class="makeStyles-root-6"
+  class="Toolbar-root"
```
nl0: is this the same `-root` as on line 5? originally, they were different, now they are the same, and i'm not sure this is intentional

catalog/app/containers/Bucket/Queries/Athena/__snapshots__/Database.spec.tsx.snap
```diff
@@ -2,10 +2,10 @@
 
 exports[`containers/Bucket/Queries/Athena/Database > should disable selection if no spare values 1`] = `
 <div
-  class="makeStyles-root-17"
+  class="Database-root"
 >
   <div
-    class="MuiFormControl-root makeStyles-root-20 makeStyles-field-18"
+    class="MuiFormControl-root Database-root Database-field"
```
nl0: again, these `Database-root`s are supposed to be different classes

catalog/app/containers/Bucket/Queries/Athena/model/requests.spec.ts
```diff
@@ -120,13 +125,14 @@ describe('containers/Bucket/Queries/Athena/model/requests', () => {
...
-      const { result, waitForValueToChange } = renderHook(() =>
+      const { result, waitForValueToChange, unmount } = renderHook(() =>
         requests.useCatalogNames('any'),
       )
       expect(result.current.data).toBe(undefined)
 
-      await waitForValueToChange(() => result.current)
+      await waitForValueToChange(() => result.current, { timeout: 5000 })
       expect(result.current.data).toMatchObject({ list: ['bar', 'foo'] })
+      unmount()
```
nl0: this won't run if the expectation fails -- is this ok?

nl0:
> For the context, I'm not going to write any more `.toMatchSnapshot` tests. Do you still think, we need more accurate mock for `makeStyles`. I think, the simplest one is enough

i still think it's not an excuse to introduce incorrect behavior and inconsistent patterns.
if you are not going to add more snapshot tests, then why bother at all? and if bothered anyways, why not do it right instead of replacing inconvenient with incorrect?

catalog/app/containers/Bucket/Dir/Toolbar/__snapshots__/Toolbar.spec.tsx.snap
fiskus: There is only one component and only one `root` class. So, it should be the same class with the same styles.
It looks like, the snapshot now looks better and more correct.

catalog/app/containers/Bucket/Queries/Athena/__snapshots__/Database.spec.tsx.snap
fiskus: Yes, if snapshot supposed to be precise, it should be something like `Database-root`, and `Select-root`

catalog/app/containers/Bucket/Queries/Athena/model/requests.spec.ts
fiskus: I think, it's ok. And better than was before

fiskus:
I see that it can be seen as incorrect. But it's correct, in means, that I didn't test class names correctness. I used snapshot tests as smoke tests for shape. Probably, I can return empty strings, if I don't care about class names.

> if you are not going to add more snapshot tests, then why bother at all?

You mean, why mock `makeStyles`? Because, otherwise, it output different results with `globals: false`.

catalog/app/components/Buttons/WithPopover.spec.tsx
fiskus: Yes

PR #4839, opened 2026-04-21
Catalog: migrate bucket queries to role-scoped Bucket type

nl0:
Migrates user-facing catalog GraphQL bucket queries from the admin-bypass `bucketConfigs`/`bucketConfig` to the new role-scoped `buckets`/`bucket` + `Bucket` type (added server-side in the registry), unifies the urql GraphCache invalidation story, renames the stale `utils/BucketConfig` module, and stubs the dead `overviewUrl` / `linkedData` fields.

### Role-scoped bucket queries (user-facing)
- `utils/Buckets.graphql` (was `utils/BucketConfigList.graphql`) — navbar / search / AI-assistant / bucket previews now read `buckets` (role-scoped).
- `containers/Bucket/Overview/gql/Bucket.graphql` (was `BucketConfig.graphql`) — deep-link overview now reads `bucket(name)`.
- `components/Preview/loaders/Html/gql/BrowsableBucket.graphql` (was `BrowsableBucketConfig.graphql`) — HTML preview trust gate reads `bucket(name)`.
- `containers/NavBar/BucketCollaborators.graphql` — collaborators display reads `bucket(name)`.

### Module rename
`utils/BucketConfig.ts` → `utils/Buckets.ts`. Exports renamed:
- `useBucketConfigs` → `useBuckets` (module-internal)
- `useRelevantBucketConfigs` → `useRelevantBuckets`
- `useCurrentBucket`, `useIsInStack` — unchanged

All 13 consumers updated (tsx + jsx + js). Three files use named imports (`{ useRelevantBuckets }`) instead of namespace because they declare a local `Buckets` component/function that would clash with `import * as Buckets`; the rest use namespace.

### urql cache invalidation
New internal helper `invalidateRootField(cache, name)` (iterates `cache.inspectFields('Query')` + invalidates by `fieldKey`; the plain `cache.invalidate({__typename:'Query'}, field)` form silently no-ops in this setup).

Called from every mutation that can change a user's effective *managed-role* bucket set:
- `bucketRemove`
- `policyCreateManaged` / `policyUpdateManaged` / `policyDelete`
- `roleUpdateManaged` / `roleDelete`
- `admin.user.mutate.setRole`

Deliberately **not** called from:
- `bucketAdd` — a new bucket has no policy attachments yet; managed-role users' set is unchanged. Unmanaged-role users (who see all buckets regardless of policy) pick it up on the next natural refetch.
- Unmanaged policy / role mutations — the server grants unmanaged roles access to all buckets regardless of policy content, so Quilt-side bucket lists don't change.

`bucketUpdate` relies on entity-level invalidation of the `Bucket` by name (same key; different `__typename` from BucketConfig).

### Admin panel
Unchanged — `containers/Admin/Buckets/` and `containers/Admin/UsersAndRoles/gql/Buckets.graphql` (aliased `buckets: bucketConfigs`) still read the admin-bypass `bucketConfigs` so admins can manage every bucket in the stack regardless of their own role scope.

### overviewUrl
Dropped from the Overview query; stubbed to `undefined` with its conditional code paths dead-when-null. Full removal tracked as a follow-up.

### linkedData
Not included in the new `Bucket` type or `Buckets.graphql` — verified dead code path:
- `cfg.linkedData` is unset across **all** deployed stack variants (grep of `wt/dp` returns zero hits — no CloudFormation template, installer config, or stack variant sets it, including `open-quilt-bio.yaml`).
- `LinkedData.CatalogData` (the only consumer in the new query path) early-returns `null` when `cfg.linkedData` is unset, before `mkBucketRef` can read `bucket.linkedData`.
- `LinkedData.BucketData` (bucket-specific JSON-LD) uses its own `LinkedDataBuckets.graphql` against `bucketConfig(name)` — unchanged by this PR, still fetches `linkedData`.
- `LinkedData.PackageData` is mounted inside a commented-out `// TODO: restore linked data` block at `PackageRevisions.tsx:561-581` — dead.

Per-bucket `linkedData` values therefore never reach the DOM on any deployed stack. Full removal (both the cfg flag and the per-bucket field across registry + catalog + quilt3.admin) tracked as a follow-up.

### Drive-by cleanup
- Extracted `invalidateRootField` replaces a copy of the same inspectFields pattern previously inlined in `handlePackageCreation`.
- Trimmed narrative / task-referencing / urql-mechanics-explaining comments in `Provider.tsx`.
- Aligned local variable naming across consumers (mostly `const buckets = ...useRelevantBuckets()`).

## Consequences

- Admins operating in a scoped managed role see only role-scoped buckets in catalog listing AND deep-links (admin bypass no longer applies on catalog surfaces). **"Role is role"**: admin privilege unlocks admin-panel features, not cross-role bucket visibility.
- Out-of-scope catalog URLs return 404 / registry access-denied; S3 object reads for out-of-scope buckets return `Access denied`.
- Overview page's `overviewUrl`-driven branches are inert.

## Test plan

- [x] `npm run build` green (webpack + fork-ts-checker)
- [x] `npm run lint` clean (prettier + eslint)
- [x] `npm test` green (812 passed / 1 skipped)
- [x] End-to-end on a dev stack:
  - admin + managed role → role-scoped listing + `Access denied` on out-of-scope URLs
  - live role switching in the UI re-scopes subsequent calls on the same OAuth token (registry resolves `user + active role` at request time)
  - admin panel shows all buckets regardless of active role
  - **cache invalidation**: navbar bucket selector refreshes after add / edit / delete and after role-policy edits (the main bug this PR also fixes)
- [ ] Non-admin + managed role — skipped as low-value (same resolver path, covered by resolver-level tests in the registry change)

## Dependencies

Depends on the server-side registry change (new `buckets` / `bucket` GraphQL queries + `Bucket` type) landing and being deployed first. This PR will 500 on any `buckets { … }` query until that upgrade is live.

<h3>Greptile Summary</h3>

This PR migrates all user-facing catalog GraphQL bucket queries from the admin-bypass `bucketConfigs`/`bucketConfig` to the new role-scoped `buckets`/`bucket` endpoints, renames the `utils/BucketConfig` module to `utils/Buckets`, and wires `refetchBuckets()` (via a new `refetchRootField` helper) into every mutation that can change a user's effective managed-role bucket set to fix navbar cache invalidation.

The approach is well-reasoned: `refetchRootField` issues a `network-only` query rather than relying on `cache.invalidate({__typename:'Query'}, field)` which is documented to silently no-op in urql 2.x + graphcache 4.x. The new `Bucket` type entity key (`name`) co-exists cleanly with the `BucketConfig` key, and the downstream behavioral changes (out-of-scope buckets return null from `bucket(name)`, HTML preview treated as not-browsable, S3 signing restricted to role-scoped buckets) are all conservative and intentional per the \"Role is role\" design principle.

Safe to merge once the server-side registry change (new `buckets`/`bucket` queries + `Bucket` type) is deployed first.

No new P0/P1 bugs found. The cache invalidation strategy is well-documented and intentionally chose `refetchRootField` (network-only) over `cache.invalidate` on root fields because the latter silently no-ops in urql 2.x + graphcache 4.x. All null-return paths for the role-scoped `bucket(name)` query are handled safely (optional chaining, conservative fallbacks). The `linkedData` and `bucketUpdate` list invalidation concerns are already captured in prior review threads. All remaining considerations are P2 or lower.

catalog/app/utils/GraphQL/Provider.tsx — the `bucketUpdate` handler relies on entity-level cache cascade rather than an explicit list update (already flagged in a prior thread).

| Filename | Overview |
|----------|----------|
| catalog/app/utils/GraphQL/Provider.tsx | Core cache invalidation logic; adds refetchBuckets/refetchRootField helpers and wires them to all role-scope-changing mutations. bucketUpdate relies on entity-level cascade (not an explicit list update) which differs from the pattern in bucketAdd/bucketRemove. |
| catalog/app/utils/Buckets.ts | Renamed from BucketConfig.ts; useBuckets/useRelevantBuckets now query role-scoped `buckets`; useIsInStack and useCurrentBucket unchanged. |
| catalog/app/utils/Buckets.graphql | Renamed from BucketConfigList.graphql; now queries role-scoped `buckets { name title ... }` instead of admin-bypass `bucketConfigs`. |
| catalog/app/utils/AWS/useShouldSign.ts | Now uses role-scoped useIsInStack; S3 signing restricted to role-scoped buckets, causing Access Denied for out-of-scope buckets — intentional per "Role is role" policy. |
| catalog/app/utils/LinkedData.tsx | CatalogData now uses useRelevantBuckets (role-scoped, no linkedData field); BucketData still queries bucketConfig via LinkedDataBuckets.graphql (unchanged). CatalogData early-returns null when cfg.linkedData is unset, which is the case on all deployed stacks per PR description. |
| shared/graphql/schema.graphql | Adds new role-scoped Bucket type and buckets/bucket root queries; moves collaborators from BucketConfig to Bucket type. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[User-facing catalog request] --> B{Role-scoped check}
    B -->|bucket in role scope| C[bucket / buckets query returns data]
    B -->|bucket NOT in role scope| D[bucket returns null]

    C --> E[Navbar / Search / AI / Preview uses Bucket type]
    C --> F[Overview: inStack=true, shows full header]
    C --> G[Html loader: browsable per Bucket.browsable]
    C --> H[useShouldSign: signs S3 requests]

    D --> I[Overview: inStack=false, shows minimal header]
    D --> J[Html loader: browsable=false, restricted sandbox]
    D --> K[useShouldSign: does not sign, Access Denied]
    D --> L[Collaborators: empty list]

    M[Mutation triggers cache invalidation] --> N{Mutation type}
    N -->|bucketRemove| O[cache.updateQuery both lists + invalidate Bucket entity]
    N -->|bucketAdd| P[cache.updateQuery BUCKET_CONFIGS_QUERY only]
    N -->|bucketUpdate| Q[cache.invalidate Bucket entity cascades to list]
    N -->|policy/role/setRole mutations| R[refetchBuckets via network-only query]

    S[Admin panel] --> T[bucketConfigs admin-bypass unchanged]
```

<sub>Reviews (4): Last reviewed commit: ["catalog/CHANGELOG: update bucket-roles e..."](https://github.com/quiltdata/quilt/commit/8b57787a2c3076661ced8a8113d9ad30adc56533) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29086329)</sub>


nl0:
@greptileai check again pls

nl0:
Re greptile's P1 on `LinkedData.tsx` — intentional and verified dead, now documented in the PR body under a new `### linkedData` section. TL;DR:

- `cfg.linkedData` is unset across all deployed stack variants (`grep -r linkedData` on the deployment repo: zero hits). `CatalogData()` early-returns `null` on that check, before `mkBucketRef` ever reads `bucket.linkedData`.
- `BucketData` (bucket-specific JSON-LD) uses its own `LinkedDataBuckets.graphql` against `bucketConfig(name)` — unchanged by this PR, still fetches `linkedData`.
- `PackageData`'s only mount site is inside a commented-out `// TODO: restore linked data` block at `PackageRevisions.tsx:561-581`.

Per-bucket `linkedData` values never reach the DOM on any deployed stack — the new query correctly reflects that. Full removal (cfg flag + per-bucket field across registry + catalog + quilt3.admin) is tracked as a follow-up cleanup.

nl0:
@greptileai one more time pls

fiskus (dismissed):

fiskus (dismissed):

fiskus (approved):

catalog/app/utils/GraphQL/Provider.tsx
```diff
@@ -154,25 +185,42 @@ export default function GraphQLProvider({ children }: React.PropsWithChildren<{}
...
                 return
-              cache.invalidate({ __typename: 'Query' }, 'roles')
               cache.updateQuery(
                 { query: BUCKET_CONFIGS_QUERY },
                 R.evolve({ bucketConfigs: R.reject(R.propEq('name', vars.name)) }),
               )
+              cache.updateQuery(
+                { query: BUCKETS_QUERY },
+                R.evolve({ buckets: R.reject(R.propEq('name', vars.name)) }),
+              )
```
drernie, line 222: **`roles` invalidation dropped from `bucketAdd` / `bucketRemove`.**

Previously both mutations did `cache.invalidate({__typename:'Query'}, 'roles')`. Now gone. For `bucketRemove` this is arguably wrong: cached `Role.permissions` / `Policy.permissions` arrays may still reference the removed bucket until their own entities are invalidated. The `Bucket`-entity invalidate won't cascade into permission list shapes that embed the bucket by value. If the admin Roles or Policies panel is open when a bucket is deleted, stale rows can linger.

Consider restoring the `roles` (and possibly `policies`) invalidation here, or explicitly documenting why it's safe to drop.

drernie (commented):
Line-level notes from my review — see inline comments.

catalog/app/utils/GraphQL/Provider.tsx
```diff
@@ -14,19 +14,47 @@ import schema from 'model/graphql/schema.generated'
...
+// Force a fresh fetch of a root-level query through the supplied client
+// and let urql propagate the new data to all active subscribers. Use
+// when the post-mutation result can't be predicted client-side (e.g. a
+// role-policy edit that shifts which buckets fall into the caller's
+// scope). `cache.invalidate` on a root field is unreliable at notifying
+// active subscribers in urql 2.x + @urql/exchange-graphcache 4.x; a
+// `network-only` query reliably writes back through the cache exchange
+// and fans out. The microtask defer lets the current mutation's cache
+// write commit first.
+function refetchRootField(clientRef: React.RefObject<urql.Client>, query: RootQueryDoc) {
```
drernie, line 43: **`invalidateRootField` vs `refetchRootField` naming.**

`packages` uses invalidate (works because `PackagePushSuccess` writes back via the entity); `buckets` uses a `network-only` refetch. The asymmetry is intentional and well-explained in the comment above, but the helper pair reads as if they were interchangeable. Worth a one-liner near `invalidateRootField` noting that it's only safe when the mutation result rehydrates the field, otherwise use `refetchRootField`. Not blocking.

catalog/app/utils/GraphQL/Provider.tsx
```diff
@@ -356,6 +410,12 @@ export default function GraphQLProvider({ children }: React.PropsWithChildren<{}
                 cache.invalidate({ __typename: 'Query' }, 'admin')
                 cache.invalidate({ __typename: 'Query' }, 'roles')
               }
+              if (result.admin?.user?.mutate?.setRole?.__typename === 'User') {
+                // Over-invalidates when the edit targets another user;
+                // acceptable for a rare admin op, avoids a self-check
+                // against the current session user.
+                refetchBuckets()
+              }
```
drernie, line 418: The comment correctly flags over-invalidation when the admin edits another user's role. Worth double-checking this doesn't cause a user-visible flicker on the admin Users page (admin list refetch + `buckets` refetch in sequence on the same frame). Low risk given it's a rare admin op.

catalog/app/containers/Bucket/Overview/Overview.tsx
```diff
@@ -116,10 +116,12 @@ export default function Overview() {
...
   const s3 = AWS.S3.use()
-  const { bucketConfig } = GQL.useQueryS(BUCKET_CONFIG_QUERY, { bucket })
-  const inStack = !!bucketConfig
-  const overviewUrl = bucketConfig?.overviewUrl
-  const description = bucketConfig?.description
+  const { bucket: bucketData } = GQL.useQueryS(BUCKET_QUERY, { bucket })
+  const inStack = !!bucketData
+  // overviewUrl is dead; stubbed until the field + its Readmes/Imgs/
+  // Summarize code paths are removed.
+  const overviewUrl = undefined
```
drernie, line 123: Stubbing `overviewUrl = undefined` is fine for now, but the `Readmes` / `Imgs` / `Summarize` consumers that accept `overviewUrl` remain on the tree with dead branches. Please make sure the follow-up ticket to fully remove the field (and its consumers) actually exists before this merges — otherwise the dead code tends to stick around.

catalog/app/utils/Buckets.graphql
```diff
@@ -1,11 +1,9 @@
...
-  bucketConfigs {
+  buckets {
     name
     title
     iconUrl
     description
-    linkedData
-    overviewUrl
     tags
     relevanceScore
```
drernie, line 8: Nit: `relevanceScore` is selected unconditionally (same as before), so the registry must return it on every `Bucket`. Just flagging for the server-side reviewer.

catalog/app/utils/GraphQL/Provider.tsx
nl0: Confirmed — valid regression on the nav-away + nav-back path. The registry `bucketRemove` resolver cascade-deletes every `RolePolicyBucketPermission` row for the bucket, so server-side `Policy.permissions` / `ManagedRole.permissions` shrink. Client-side `cache.invalidate({__typename: 'Bucket', name})` doesn't touch those permission entities; the pre-PR `cache.invalidate('Query', 'roles')` only helped on nav-back (active subscribers never re-executed — known urql 2.2 + graphcache 4.4 limitation that motivated `refetchRootField` in the first place).

Will follow up with a small fix PR that strips the removed bucket from every cached `Policy.permissions` / `ManagedRole.permissions` via `cache.updateQuery`, mirroring the `policyDelete` / `roleDelete` cascade pattern already in this file. Addresses both the nav-back path and the panel-open path — strictly better than pre-PR.

On `bucketAdd` the drop was intentional: a newly added bucket has no policy attachments yet, so server-side permissions are unchanged.

catalog/app/utils/GraphQL/Provider.tsx
nl0: Good call — folding a short doc-comment cross-reference into the same follow-up PR.

catalog/app/utils/GraphQL/Provider.tsx
nl0: No visible flicker in the Playwright runs against the dev stack so far. The over-invalidation is rare (admin Users edit) and the adjacent comment block flags it; leaving as-is for now.

catalog/app/containers/Bucket/Overview/Overview.tsx
nl0: Tracked as a follow-up to fully remove `overviewUrl` (and `linkedData`, also dead) across registry + catalog + `quilt3.admin`. The `quilt3.admin.buckets` kwargs change is a breaking public-SDK change, so it needs a coordinated release — staged: deprecate fields in the registry schema, drop catalog consumers, then drop `quilt3.admin` kwargs and remove the registry fields. Will land as a separate PR.

catalog/app/utils/Buckets.graphql
nl0: Registry returns `relevanceScore` unconditionally on every `Bucket` — it's non-nullable in the schema (`Float!`). No action.

PR #4840, opened 2026-04-21
Qurator: wire in Platform MCP Server tools

nl0:
## Summary

Wires the Quilt Platform MCP Server into Qurator. Catalog users get the same ~25 platform tools (packages, search, S3, Athena, tabulator, `get_resource`) that external MCP clients see, with full lifecycle handling and a persistent status surface tied to the chat input.

## What changes for users

The catalog's prior in-app `catalog_global_getObject` is reshaped into `catalog_preview` (single `s3_uri` input), and the platform server's read surface (`platform__object_read`, `platform__s3_object_info`, `platform__package_browse`, etc.) is added alongside it. Two read paths, two purposes:

- **`catalog_preview`** — for content the model needs to *interpret*. Returns thumbnail-resized images (via the `/thumbnail` API gateway, ≤1024×768), native Bedrock `Document` blocks for ≤500 KiB PDFs / Office docs, and language-tagged text. Use when the goal is summarize / describe / extract.
- **`platform__object_read`** — for raw bytes, scripting, or any case where the model doesn't need to interpret content. Returns text decoded UTF-8 (capped at 16 KiB) or base64-encoded binary.

A `<reading-objects>` guidance message is added to the prompt context: call `platform__s3_object_info` first to inspect `ContentLength` / `ContentType`, then pick the right tool from the metadata.

### Trade-offs vs. the prior in-app tool

- **Two read paths to choose from.** The model now picks; the tool descriptions and the guidance message steer it. Risk: it picks the wrong one (e.g. raw-bytes a PDF instead of preview-rendering it). Mitigation: explicit when-to-use language on both tools, and the metadata-first guidance.
- **`catalog_preview` is catalog-only.** External MCP clients (non-catalog) only see the platform server, so they get `platform__object_read` but not the rich preview path. Native-document support on the platform server is a separate follow-up.
- **>500 KiB documents still can't go through Bedrock natively.** Same as before this PR. Not a new regression.
- **Image formats outside the thumbnail lambda's coverage** (`jpg|jpeg|png|gif|webp|bmp|tiff|tif|czi`) won't preview-resize and will fall through to the platform path. Same as before this PR.

### Other UX changes

- **Destructive tool calls execute without an in-chat confirmation step.** Unchanged in mechanism from before, but the platform server's surface includes more mutating tools than the prior in-app surface (S3 puts, package create, tabulator set/rename), so the blast radius is wider. A confirmation UX using MCP's `destructiveHint` / `readOnlyHint` annotations is a separate follow-up.
- **Connector status appears in the chat input's helper-text region** (severity-tinted, with inline reconnect / continue-without / acknowledge actions). No persistent banner. Detail in the UI section below.

## Architecture

```
Model/
├── Connectors/
│   ├── Connectors.ts       Backend interface, ConnectorState machine,
│   │                       lifecycle (manageConnector), service builder,
│   │                       contextContribution, React bridges
│   ├── Mcp.ts              MCP wire client + bearerPassthru adapter
│   ├── Connectors.spec.ts  30 lifecycle / service-primitive tests
│   ├── Mcp.spec.ts         15 wire-level tests
│   └── index.ts            barrel
└── Tool.ts                 (unchanged shape; Bedrock toolSpec built here)
```

A *connector* pairs a configuration (id, title, hint) with a `Backend` — the protocol contract (`initialize`, `listTools`, `callTool`, `ping`). The lifecycle (Connecting → Ready → Disconnected → Failed, heartbeat, reconnect budget) is generic in `Backend`; the platform-specific bits live in one factory call: `Mcp.bearerPassthru({ url, getToken })`. Adding a different backend (an OAuth-flavored MCP, an in-process tool source) is a new factory; the lifecycle doesn't move.

## Per-connector lifecycle

```
Connecting ──success──→ Ready
Connecting ──failure──→ Failed{acked:false}
Ready ──health-flip───→ Disconnected{retrying:true}
Disconnected ──recovery→ Ready
Disconnected ──exhaust→ Failed{acked:false}
Failed ──user retry───→ Connecting
Failed{!acked} ──ack──→ Failed{acked:true}
```

- **Heartbeat:** `ping` every 30 s with a 5 s timeout. Two consecutive transport failures (from heartbeat *or* tool calls — they share a `Health` ref) flip the state to `Disconnected`. The threshold-crossing watcher subscribes to the SubscriptionRef stream so cross-source crossings fire immediately.
- **Reconnect:** probe-driven. While Disconnected, the lifecycle pings at escalating cadence (5 s → 10 s → 20 s → 30 s cap); each successful ping triggers a bootstrap (cap: 3 attempts). The whole window is bounded by an outer 60 s timeout so the loop can't stay disconnected forever.
- **Read-only-tool retry-once:** tools annotated `readOnlyHint: true` retry once on transport error before counting toward the threshold. Destructive / unannotated tools fail immediately.
- **No silent degradation:** when the budget is exhausted, the conversation gates on `Failed{!acked}` until the user clicks reconnect or "continue without". `Failed{acked:true}` is sticky — the connector reads as `unavailable` in the prompt and the LLM stops trying its tools.
- **Force-reconnect from healthy states:** `connector.retry` from Ready or Disconnected degrades health to threshold so the lifecycle naturally falls into the reconnect probe loop. From Failed it drains the controls queue. From Connecting it's a no-op (bootstrap is already in flight).
- **Sole writer:** each lifecycle fiber owns one `SubscriptionRef<ConnectorState>`. Aggregate predicates (`isBlocked`, `awaitUnblocked`, `contextContribution`) merge across all per-connector refs. Lifecycle fibers are scope-bound and interrupted on Assistant unmount.

## Conversation integration

Two new state-machine elements:
- `State.AwaitingConnector { events, timestamp, waiter }` — only the waiter fiber handle, no per-connector snapshot.
- `Action.ConnectorReady`.

A small `advanceFromEvents` helper centralizes the gating decision at both entry points (`Idle.Ask` and `ToolUse.ToolResult` last-call):
1. Reads `connectors.isBlocked` synchronously (handlers don't observe asynchronously).
2. If blocked, forks `connectors.awaitUnblocked`; that fires `ConnectorReady` and the handler re-runs.
3. Otherwise forks the LLM request as before.

`llmRequest` folds `connectors.contextContribution` into the prompt context at request time. No React-side bridge.

## Tool dispatch

Each connector tool's executor goes through `connectors.byId[id].callTool(name, input)`:
1. Reads the connector's state — fast-fails with a `Tool.fail` text block if not Ready.
2. Fires the wire call. On a transport error from a `readOnlyHint` tool, the call is retried once.
3. Updates the shared `Health` ref on transport-level outcome (success → reset, failure → bump). Non-transport errors (server-side application errors, schema decode failures) don't disturb health.

## Wire client (`Mcp.ts`)

- `@effect/platform` `HttpClient` over `FetchHttpClient.layer`. Cancellation via fiber interrupt aborts the underlying fetch.
- `Eff.Schema` decode at every response boundary; failures map to `McpProtocolError`.
- Tagged errors via `_tag` discriminator: `McpAuthError`, `McpTransportError`, `McpProtocolError`, `McpRpcError`. The connector layer maps these onto an abstract `BackendError` (`Transport`/`Auth`/`Protocol`/`Application`); the wire-level tag is preserved on `cause` for telemetry but the lifecycle and UI never speak MCP vocabulary.
- Methods: `initialize` + `notifications/initialized`, `tools/list`, `tools/call`, `resources/read`, `ping`.
- `tools/list.inputSchema` is forwarded to Bedrock as-is. Server is authoritative on its own schema's draft and structure; we don't reshape.
- SSE handling for FastMCP `stateless_http=True` mode (`text/event-stream` responses with a single `message` event).
- Trace-context propagation explicitly disabled on each request (`HttpClient.withTracerPropagation(false)`). Effect's HttpClient auto-injects W3C `traceparent` + Zipkin `b3` headers from the active span; the platform MCP server's CORS allow-list rejects unknown headers (preflight returns 400 "Disallowed CORS headers"), and we don't have a tracing collector on the receiving end either.
- A small dedup in `Tool.make` lifts `description` out of the JSON schema body so Bedrock doesn't ship it twice (once on `toolSpec.description`, once in `inputSchema.json.description`). The MCP path already keeps these separate.

## UI

The user-facing surface is the chat input's helper-text region. When all connectors are `Ready`, the helper falls back to the default disclaimer. When any connector is non-Ready, the helper-text region carries:

- `<title>: connecting…` / `reconnecting…` (warning, no actions — already auto-progressing)
- `<title>: couldn't connect • reconnect • continue without` (error)
- `<title>: unavailable • reconnect` (warning, input re-enabled — acked is sticky)

Aggregate severity picks the worst category across connectors. Inline actions use the existing `MessageAction` style (clickable span, opacity transition on hover).

The chat panel is otherwise empty at the top — there's no persistent banner. The prior banner pushed history down and competed with the absolute-positioned hamburger menu; consolidating into the helper-text region makes the "input is disabled because…" relationship spatial.

The DevTools panel exposes a per-connector subsection: heading (`Connector: <title>`), hint (muted), state line + inline reconnect/acknowledge actions, error detail row, and a live `contextContribution` preview that hides when nothing is contributable.

## Auth

`Mcp.bearerPassthru({ getToken })` is the 1st-party adapter — caller resolves the bearer on every call (the adapter never manages refresh, expiry, or session state). Catalog wires `getToken` to re-read the redux session token via a memoized selector, so token rotation is handled without explicit plumbing. A future `Mcp.oauth(...)` factory (or any other backend) would be a sibling, not a fork of the lifecycle.

## Test plan

- [x] ESLint / Prettier — clean.
- [x] LSP workspace diagnostics — zero errors / warnings across all touched files.
- [x] Vitest under Node 20 — 68 Assistant tests pass:
  - **`Mcp.spec.ts`** (17) — wire-level `make()` integration, content mapping, parseSseToJson.
  - **`Connectors.spec.ts`** (36) — state predicates, lifecycle transitions, retry/ack from Failed, callTool gating, transport-error → threshold-crossing, TestClock-driven heartbeat / probe-driven reconnect, read-only retry-once, layer-level service primitives + multi-connector aggregation, stale-control drain on Failed entry, autoload of reference-grade resources.
  - **`Conversation.spec.ts`** (8) — Idle / AwaitingConnector / Abort / Discard transitions, ToolUse → blocked → AwaitingConnector path, AwaitingConnector waiter freshness across cycles.
  - **`navigation.spec.ts`** (7) — unchanged; snapshot still valid.
- [x] **Manual verification on the qur dev stack** (local catalog at :3000, deployed platform MCP server) — Ready, Disconnected → reconnect → Ready, Failed{!acked} → reconnect / continue-without, Failed{acked} unavailable; input gating; helper-text severity; DevTools panel.
- [x] **End-to-end smoke through 9 design job stories** on the qur dev stack — list/count, search packages, object search, package browse, read text, read image, athena (databases + tables), destructive write, catalog link.
- [x] **`catalog_preview` end-to-end** (local catalog → dev stack PMS) — model follows the metadata-first guidance: `platform__search_objects` → `platform__s3_object_info` → `catalog_preview` for an image; `catalog_preview` returns a metadata text block + Image content block; the model renders a description of the rendered image. Error fallback path also exercised (graceful text-block error when the thumbnail fetch fails).
- [ ] **End-to-end on nightly post-merge** — search/browse/read (incl. image and a PDF), one read-only and one destructive platform op, `catalog_preview` on text + image + small PDF.

## Depends on

- The Quilt Platform MCP Server's CORS + HTTP entry-point release — already deployed to the qur dev stack.
- A deployment-side change that routes `registry-<stack>.domain.com/mcp/platform/*` and sets `CORS_ORIGINS` on the MCP task — open, dev-stack-deployed; auto-deploys to nightly on merge.

<h3>Greptile Summary</h3>

This PR wires the Quilt Platform MCP Server into Qurator, adding ~25 platform tools (packages, search, S3, Athena, tabulator) to the catalog's chat UI. It introduces a complete connector abstraction layer (`Connectors.ts`, `Mcp.ts`) with lifecycle state machine, heartbeat, probe-driven reconnect, and an `AwaitingConnector` conversation state that gates LLM turns when connectors are unresolved. The `catalog_global_getObject` tool is reshaped into `catalog_preview` with a two-path read strategy guided by a new system message.

PR is safe to merge; no P0/P1 findings. All P2 issues are accounting subtleties and documentation gaps that don't break any current code path.

No critical or blocking bugs found. The four P2 comments cover: a nuanced health-accounting difference for read-only retries, a missing WEBP case in the thumbnail format switch that causes a graceful text fallback rather than a crash, a documentation gap on why HTTP 5xx errors are not retry-eligible, and a frozen-URL note for the dev override. All lifecycle, state-machine, auth, and wire-protocol paths look correct.

catalog/app/components/Assistant/Model/GlobalContext/preview.ts (WEBP format switch), catalog/app/components/Assistant/Model/Connectors/Connectors.ts (retry/health accounting), catalog/app/components/Assistant/Model/Assistant.tsx (fiber-leak on aborted render)

| Filename | Overview |
|----------|----------|
| catalog/app/components/Assistant/Model/Connectors/Connectors.ts | New file: full connector lifecycle (Connecting → Ready → Disconnected → Failed), heartbeat, probe-driven reconnect, callTool gating, health ref, React bridges. Well-structured; one minor retry-health accounting subtlety. |
| catalog/app/components/Assistant/Model/Connectors/Mcp.ts | New file: MCP wire client over Effect HttpClient; SSE parsing, error taxonomy, bearerPassthru adapter. Solid schema validation at every response boundary; tracerPropagation disabled to avoid CORS rejection. |
| catalog/app/components/Assistant/Model/Conversation.ts | Adds AwaitingConnector state + ConnectorReady action; centralizes gating in advanceFromEvents; merges connector contextContribution into every LLM prompt at request time. |
| catalog/app/components/Assistant/Model/Assistant.tsx | Wires platform connector config, allocates ConnectorsService via useConst with documented fiber-leak risk on aborted renders, provides it to ConversationActor layer. |
| catalog/app/components/Assistant/UI/Chat/Chat.tsx | Adds AwaitingConnectorState renderer, connector helper-text with severity coloring, per-connector state subscriptions via Actor.useState in a map (eslint suppressed, documented as stable). |
| catalog/app/components/Assistant/UI/Chat/DevTools.tsx | Adds ConnectorsPanel with per-connector state display, manual retry/ack actions, and live contextContribution preview. Effect stream cleanup is correct. |
| catalog/app/components/Assistant/Model/GlobalContext/preview.ts | Renames getObject → useCatalogPreview; adds single s3_uri input schema; keeps two-path read logic (image thumbnail vs. native Document block). Missing WEBP case in thumbnail format switch. |
| catalog/app/components/Assistant/UI/Chat/Input.tsx | Adds helperText/helperSeverity props; severity-tints FormHelperText via MUI JSS specificity trick. Clean change. |
| catalog/app/utils/Actor.ts | Implements the previously TODO'd actor listener interrupt on unmount; exports Actor interface. Correct use of useEffect cleanup. |
| catalog/app/containers/Auth/selectors.js | Adds token selector reading the bearer string directly; consistent with existing exp selector pattern; PFSCookieManager refactored to use it. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant UI as Chat UI
    participant Conv as ConversationActor
    participant CS as ConnectorsService
    participant LC as Lifecycle Fiber
    participant MCP as Platform MCP Server

    UI->>Conv: Action.Ask
    Conv->>CS: isBlocked
    alt blocked
        CS-->>Conv: true
        Conv->>CS: fork awaitUnblocked
        Conv-->>UI: State.AwaitingConnector
        LC-->>CS: connector flips Ready
        CS->>Conv: Action.ConnectorReady
    end
    Conv->>CS: contextContribution
    CS-->>Conv: tools + connector overview messages
    Conv->>MCP: LLM request via Bedrock
    MCP-->>Conv: LLMResponse with toolUses
    loop tool calls
        Conv->>CS: byId[connId].callTool
        CS->>LC: gate on Ready state
        LC->>MCP: POST bearer token
        MCP-->>LC: McpToolResult
        LC-->>Conv: Tool.Result
    end
    Conv->>MCP: LLM request with tool results
    MCP-->>Conv: final text response
    Conv-->>UI: State.Idle
```

1. `catalog/app/components/Assistant/UI/Chat/Chat.tsx`, line 519-533 ([link](https://github.com/quiltdata/quilt/blob/5a97edc2782bb1300f1b0623c47d238d479470f5/catalog/app/components/Assistant/UI/Chat/Chat.tsx#L519-L533)) 

   <a href="#">P1</a> **Conditional `Actor.useState` calls violate the Rules of Hooks**

   The comment acknowledges this but marks it safe because "`connectors.byId` is built once … and never re-keyed." That guarantee holds today, but the comment is the only guard. The ESLint rule is suppressed rather than satisfied. If `connectors` (the `ConnectorsService`) ever comes from a re-run of `useConnectors` (e.g., after a future hot-reload, StrictMode double-invoke, or a different connector list), the hook call-count will differ between renders and React will throw. `Object.values(connectors.byId).map(c => Actor.useState(c.state))` is the concrete violation line.

   A safe alternative is to lift the per-connector subscription into its own small component (`<ConnectorRow connector={c} />`) so each hook call is unconditionally inside its own render, or to aggregate state via the existing `useIsBlocked` + a single merged stream hook, avoiding the per-connector `useState` calls at the `Chat` level entirely.
</details>

<sub>Reviews (6): Last reviewed commit: ["Connectors.spec: restore precondition as..."](https://github.com/quiltdata/quilt/commit/34c862a074db117fdff235399edba441369beb67) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29100661)</sub>


nl0:
@greptileai have a look pls

nl0:
@greptileai one more time

nl0:
@greptileai come again pls

nl0:
@drernie @fiskus just point your agents at this

nl0:
@sir-sigurd you can point yours too =)

drernie:
Automated code review of [areas of concern](https://nightly.quilttest.com/b/quilt-dev/packages/proj/260408-qurator-ui/tree/latest/tools/07-code-review-concerns.md) identified two [concrete risks](https://nightly.quilttest.com/b/quilt-dev/packages/proj/260408-qurator-ui/tree/latest/tools/08-code-review-report.md):

1. `readOnlyHint` retry-once is broader than the concern asks for: all non-2xx
   HTTP responses are currently mapped to `McpTransportError`, so a read-only
   tool call can be retried after an HTTP 5xx response even if the server has
   already performed the operation.
2. The initial `Connecting` bootstrap has no timeout, and retry while
   `Connecting` queues a stale retry token that is drained later. A hung
   bootstrap can leave the connector blocked indefinitely.

These don't seem like showstoppers. But in case it helps, I created a quick child PR [4864](https://github.com/quiltdata/quilt/pull/4864) to fix it, with tests:

Feel free to reuse, ignore, or close.

drernie:
I also mooted what appeared to be the missing [Docs: CHANGELOG + Qurator.md](https://github.com/quiltdata/quilt/pull/4865/changes) as a separate PR.

sir-sigurd:
> ⚠️ **Disclaimer:** This review is AI-generated by a multi-agent run (12 focused passes + 6 verification passes, with skeptic/ground-truth checks against the actual code). I haven't personally reviewed any of it — I know almost nothing about frontend, so please weigh the findings accordingly: some are likely real, some may be off-base. Items marked "did not survive verification" at the end are flagged so reviewers don't chase them.

---

The architecture is sound — clean separation between lifecycle, wire client, and conversation gating; thoughtful state machine; thorough tests for the model layer. The findings below are mostly seams and rough edges, not structural issues.

## Tier 1 — worth addressing before merge

**1. Heartbeat bumps health on non-transport errors (inconsistency).**
`catalog/app/components/Assistant/Model/Connectors/Connectors.ts:510-514` increments `consecutiveFailures` on **any** ping error, while the `callTool` path at L399-405 correctly gates on `err.transient === true`. Net effect: a single auth error (token expiry, etc.) on a heartbeat counts toward the 2-failure threshold and triggers a reconnect that won't help. Fix: gate the heartbeat bump on `err.transient === true` for parity.

**2. `useActorLayer` listener fiber leaks on unmount.**
`catalog/app/utils/Actor.ts:189-231` — line 220 has `// TODO: stop/interrupt on unmount`. The listener is forked as `forkDaemon` with no cleanup. Each Assistant remount accumulates a daemon. Mirror the pattern from `Actor.useState` (L147-177) which does interrupt on unmount.

**3. `useConnectors` allocates the Effect Scope during render via `useConst`.**
`catalog/app/components/Assistant/Model/Assistant.tsx:104-129` — the comment in code itself acknowledges that an aborted render (Suspense unwind, Error Boundary) leaks the lifecycle fibers. The "AssistantProvider above Suspense" mitigation is organizational. Defer scope allocation into `useEffect`, or use a synchronous cleanup that joins the scope close.

**4. Read-guidance points at tools that may not exist.**
`catalog/app/components/Assistant/Model/GlobalContext/index.ts:7-25` injects the `<reading-objects>` block unconditionally. When the platform connector is `Failed{acked:true}`, `platform__s3_object_info` / `platform__object_read` are absent from the tool list — the model will keep trying them. Make the guidance conditional on the platform connector's readiness, or have it adapt to the actual present tools.

**5. JSON-RPC response `id` never validated against request `id`.**
`catalog/app/components/Assistant/Model/Connectors/Mcp.ts` `rpc()` (~L450-468) generates a UUID per request but never checks the response's `id`. With stateless HTTP this is benign in normal operation but is a JSON-RPC 2.0 spec violation and silent if a proxy ever reorders.

**6. Tool-name namespacing is a string convention with no guard.**
`Connectors.ts:445` — `${config.id}__${d.name}`. A config `id: "a__b"` + tool `c` collides with `id: "a"` + tool `b__c`. Either reject `__` in `config.id` at registration, or use a non-printable / UUID separator.

**7. `tools/list` ignores MCP pagination.**
`Mcp.ts:488-491` — no `cursor` is passed and the response schema doesn't decode `nextCursor`. Servers that return paginated results will silently expose only the first page.

**8. No feature flag / kill-switch for the platform connector.**
PR description: "depends on a deployment-side change… auto-deploys to nightly on merge." If the MCP endpoint is unhealthy at any point, every Assistant session enters Disconnected → Failed{\!acked} and gates the chat input. There is no config knob to disable the platform connector. Given the cross-deploy ordering, a runtime kill-switch (env / config flag in `usePlatformConnectorConfig`) is cheap insurance.

## Tier 2 — significant; fix soon

**9. No error-monitoring (Sentry) hook in the connector lifecycle.** The catalog uses `Sentry.captureException` elsewhere (`PFSCookieManager.tsx`, `CatalogSettings.tsx`). Connector failures are surfaced only in DevTools and the chat input helper-text — oncall has no signal when many users hit Failed.

**10. No state-transition logs.** `manageConnector` transitions and per-attempt bootstrap failures are silent. Without `Effect.log` at the transition points, supporting "stuck reconnecting" reports relies on the user opening DevTools.

**11. Heartbeat doesn't suspend on `document.hidden`.** `runHeartbeat` is a `while (true)` with a 30s sleep. With N connectors × M tabs, it adds up. A Page-Visibility-API gate is straightforward.

**12. Per-render cost on every connector state emit.**
- `useIsBlocked` re-evaluates the predicate on every state change with no dedup (Connectors.ts ~L922-944).
- `Chat.tsx:528` recomputes `helperSeverityFor(connectorStates)` unmemoized.
- `DevTools.tsx:290-308` recomputes `contextContribution` (which iterates connectors and serializes XML) on every state change. Reconnect probes (5s/10s/20s) emit several times per minute per connector.

**13. Hooks-in-loop in `Chat.tsx:513-516` (ESLint suppressed).** Safe today because connectors are static post-mount, but a footgun. Wrap in a `<ConnectorRow>` component so each connector gets its own hook-stack frame; suppression goes away.

**14. Accessibility: action spans aren't keyboard-accessible.** `MessageAction.tsx` uses `<span onClick=…>` without `role="button"`, `tabIndex`, or `onKeyDown`. Used for reconnect / acknowledge / continue-without and in DevTools. Same issue on the `ToolMessage` header click-to-expand. Use `<button>` or add the missing attributes.

**15. Destructive tools execute without an in-conversation confirmation.** PR description acknowledges this as a follow-up. The `destructiveHint` annotation is parsed but unused. New surface includes S3 puts, package create, tabulator set/rename — wider than what shipped before. Either gate destructive calls behind a UI confirmation or, at minimum, add a strong system-prompt instruction that the model must surface a plan before invoking destructive tools.

**16. Token rotation mid-call comes back as `Transport`, not `Auth`.** A 401 response during fetch maps to `McpTransportError` (transient: true). Two consequences: (a) read-only tools retry once and risk double-executing if the rotation actually succeeded server-side; (b) the lifecycle treats it as a transport blip rather than auth. Detect 401/403 explicitly in `Mcp.ts` and map to `McpAuthError`.

**17. Bedrock-bound JSON Schemas forwarded with no shape validation.** `Connectors.ts:350` casts `descriptor.inputSchema as unknown as JsonSchema7Root`. MCP servers can return draft-2020 features (`$ref`, `prefixItems`, `unevaluatedProperties`) that Bedrock will reject with an opaque message. A defensive sanity check (or surfacing the rejection back as a tool-load failure rather than a request-time failure) would help diagnose new MCP servers.

**18. No size cap on returned image / resource bodies.** `Mcp.ts:540` (`Buffer.from(img.data, 'base64')`) and the resource read path concatenate without bounds. A misbehaving server can OOM the tab.

## Tier 3 — polish / API design

**19. Backend interface is MCP-flavored.** `ping()` is mandatory, `readOnly` is essentially MCP semantics. An in-process or different-transport backend would have to no-op or duplicate intent. Model capabilities explicitly or split `TransportBackend` from `LocalBackend`.

**20. `Connectors.ts` is 944 lines** and mixes types, lifecycle internals, service builder, prompt-XML rendering, and React hooks. Natural seams: types → `Connectors.types.ts`; lifecycle → `Connectors.lifecycle.ts`; React bridges → `Connectors.react.tsx` (so the model layer stays UI-free).

**21. Lifecycle constants hardcoded** (HEARTBEAT_CADENCE 30s, HEALTH_THRESHOLD 2, probe cadences). Per-backend overrides would let an in-process or rate-limited backend tune them.

**22. `Disconnected{retrying}` is invisible to the model.** Only Ready and `Failed{acked}` render in the prompt; up to ~60s of probing pauses the conversation with no narrative for either user or model. Consider a brief Disconnected note in the overview.

**23. `bootstrap` is called on every reconnect.** Document or enforce idempotency on `Backend.initialize()` — a future stateful backend (OAuth) could accumulate sessions otherwise.

**24. Test gaps** worth filling: concurrent `callTool` from multiple tools, multi-connector partial-failure aggregate, SSE multi-event responses, token rotation between requests, retry while Connecting (no-op?), `Failed{acked} → retry → Failed` cycle, abort/discard during AwaitingConnector verifying waiter is interrupted, React-component tests for Chat helper-text + DevTools.

**25. Tool IDs hardcoded in guidance.** `READ_GUIDANCE` literally references `platform__s3_object_info` / `platform__object_read`. Brittle to renames or to a future second connector that exposes overlapping tools.

## Claims that did NOT survive verification (skip these)

- **"`@effect/platform` is unused; ~50–100 KB bundle savings."** Wrong: `Mcp.ts:22-24, 448` use `FetchHttpClient.layer` and `HttpClient.withTracerPropagation`.
- **"`AwaitingConnector` is missing handlers for `LLMError` / `LLMResponse`."** Those events are unreachable from that state by construction. By design, not a bug.
- **"`Tool.ts:84` strips all fields named `description` recursively."** It's a single-level destructure — only the top-level `description` is removed, exactly as intended.
- **"`isBlocked` / `awaitUnblocked` is a critical race."** The `SubscriptionRef.changes` stream re-emits the current value on subscribe (Connectors.ts:523-524 comment, L530); the read-then-fork pattern is closed by re-emit semantics. A code comment in `Conversation.ts` spelling that out would still help future readers.

---

**Net (per the agents):** the foundation is well-architected, but items **1, 2, 3, 4, and 8** are worth addressing before nightly rollout — they're either real bugs or rollout risk. Everything else is normal post-merge backlog.

fiskus:
> # Review: `feat/qurator-mcp` vs `origin/master`
> 
> 38 commits, ~4000 LOC, mostly net-new in `Assistant/Model/Connectors/`.
> 
> ## Overall: feature looks complete and well-built
> 
> - Tests, lint, and type-check all pass clean (one pre-existing nested-ternary
>   warning, unrelated TS errors in `react-table.d.ts` that pre-date this
>   branch).
> - **53 tests in `Connectors.spec.ts` + 8 in `Conversation.spec.ts` + 17 in
>   `Mcp.spec.ts`** — all green. Coverage is genuinely thorough: state
>   predicates, retry/ack drainage, heartbeat threshold, probe-cadence
>   escalation, exhaustion paths, autoload caps, and aggregate primitives.
> - The architecture is coherent: `Backend` is a transport-agnostic interface;
>   `Mcp.bearerPassthru` is the only concrete adapter today; lifecycle
>   (`manageConnector`) is generic in `Backend` and writes to its own
>   `SubscriptionRef`. Per-connector fibers are forked under the surrounding
>   `Scope`. Errors are abstracted to `Transport | Auth | Protocol |
>   Application` so UI doesn't speak MCP.
> - The bug fix in `Conversation.LLMResponse` (Conversation.ts:280) — merging
>   connector tools into the executor path, not just the LLM-request path — is
>   real and the regression test pins it correctly.
> 
> ## Real concerns
> 
> 1. **`useConnectors` can leak fibers on aborted render**
>    (Assistant.tsx:111-130). Acknowledged in comments, mitigated by mounting
>    `<AssistantProvider>` above any Suspense boundary, but still a footgun if
>    the tree changes. The proper fix (allocate in `useEffect`, expose Loading
>    state) is queued — fine to defer if the mitigation holds.
> 
> 2. **`HEALTH_THRESHOLD = 2`** (Connectors.ts:272) is aggressive. Two failed
>    pings (≤60s apart) plus tool-call transport bumps share the counter — a
>    single tab going to background or a brief network blip could flip
>    Ready→Disconnected. Worth watching in production telemetry; consider 3
>    or making it config.
> 
> 3. **No telemetry hooks on connector state transitions** — only
>    `Eff.Effect.logWarning`. For production debugging, Sentry breadcrumbs on
>    Failed/Disconnected entries would help correlate with user reports. Easy
>    to add later.
> 
> 4. **Tool name namespacing via `${id}__${name}`** (Connectors.ts:445). If a
>    connector id ever contains `__`, prefix routing breaks silently. Today
>    only `'platform'` is used, so cosmetic, but worth a runtime invariant.
> 
> ## Minor nits
> 
> - **`Mcp.ts:421`** — `resp.headers['content-type']` relies on
>   `@effect/platform`'s lowercase normalization. Probably correct but
>   undocumented; a `?.toLowerCase()` on the lookup key would be defensive.
> - **`Connectors.ts:350`** — `descriptor.inputSchema as unknown as
>   JsonSchema7Root` forwards the raw server schema. Fine in practice
>   (Bedrock validates), but the cast is unverified.
> - **Tool description deduplication** (Tool.ts:84) only applies to
>   Effect-Schema tools; MCP-authored schemas with a root-level `description`
>   would still ship it twice. Server-authored schemas rarely have one, so
>   probably moot — but the comment in Connectors.ts:343 implies the issue is
>   universally handled, which it isn't.
> - **`Chat.tsx:514-517`** uses `eslint-disable react-hooks/rules-of-hooks` to
>   call `Actor.useState` inside `.map`. Justified by the stability of
>   `connectors.byId`, but fragile if connectors ever become dynamic.
> - **`Input.tsx:97`** — nested ternary on `helperSeverity`. Same shape as
>   DevTools — easy to flatten to a lookup.
> - **`Conversation.ts:155 + 280`** — duplicate `Context.merge(reactCtx,
>   connectorsCtx)` of LLM-request and tool-execute paths. Minor DRY
>   opportunity.
> - **`SubscriptionRef.update` early-bail in `resetHealth`**
>   (Connectors.ts:307) relies on Effect's reference-equality skip. Correct
>   but subtle — a brief comment on *why* the equality check (avoid waking
>   subscribers) would help future readers.
> 
> ## Regression risks
> 
> I don't see any. The non-connector changes are:
> 
> - `Auth/selectors.js`: added `token` selector, used by Connectors and
>   `PFSCookieManager` (which moved off its private selector — equivalent
>   shape).
> - `PFSCookieManager.tsx`: deduplicates the token selector, behaviorally
>   identical.
> - `Tool.ts`: lifts `description` out of `inputSchema` body. Existing builtin
>   tools all relied on `jsonSchema.description` which is now in `description`
>   directly. Behavior preserved; payload to Bedrock is one field smaller.
> - `GlobalContext`: kept `catalog_preview` as the rich-render escape hatch
>   with new `READ_GUIDANCE` instructing the model when to use it vs.
>   `platform__object_read`. This is additive guidance; no existing tool
>   removed.
> 
> ## Verdict
> 
> Ready to land. The deferred fixes (scope leak, threshold tuning, telemetry)
> are reasonable to defer and not blockers.

catalog/app/components/Assistant/UI/Chat/Input.tsx
```diff
@@ -60,11 +69,36 @@ const darkTheme = createCustomAppTheme({ palette: { type: 'dark' } } as any)
...
+  helperText,
+  helperSeverity,
+  onSubmit,
+}: ChatInputProps) {
   const classes = useStyles()
+  const helperClass = helperSeverity
+    ? helperSeverity === 'error'
+      ? classes.hintError
+      : classes.hintWarning
+    : classes.hint
```
fiskus:
```suggestion
  const helperClassBySeverity = {
    error: classes.hintError,
    warning: classes.hintWarning,
  }
  const helperClass = helperSeverity
    ? helperClassBySeverity[helperSeverity]
    : classes.hint
```

fiskus (dismissed):

nl0:
Thanks @fiskus, @drernie, @sir-sigurd for the reviews. Here's the disposition.

## Landing in this PR

- **drernie's #4864** — pulling in the retry/bootstrap fix (split `transient` from `retryable`; bound initial `Connecting` bootstrap; `retry()` from `Connecting` is a true no-op). Will square the lifecycle-synthesized `retryable` flag so it matches `adaptError`'s explicit `false`.
- **drernie's #4865** — pulling in the CHANGELOG + Qurator.md docs, with the missing "acknowledge" control mentioned in the Connector Status section.
- **fiskus inline** (`Input.tsx` helper severity) — applying as a class-lookup, restructured so warning/error stack on top of the base hint class (specificity via `&$hint`).
- **fiskus minor** (`Mcp.ts:421`) — defensive `?.toLowerCase()` on the content-type lookup.
- **sir-sigurd Tier-1 #2** — `useActorLayer` listener fiber leak on unmount (mirror the cleanup pattern from `Actor.useState`).
- **sir-sigurd Tier-1 #5** — JSON-RPC response `id` echo-validation in `Mcp.ts` `rpc()`.

## Explicitly skipping

- **sir-sigurd Tier-1 #8 (kill-switch / feature flag for the platform connector)** — the platform connector *is* the integration layer between Qurator and the stack. Operating without a kill-switch is the intended posture; falling back to "Qurator drives the catalog only" isn't useful.

## Deferred — tracked in qhq-5d0.16

Everything else from the three reviews is captured in the post-merge follow-up epic, with reviewer attribution and pointers to existing siblings (`qhq-ekc` perf, `qhq-mfo` destructive UX, `qhq-5d0.5` scope leak, `qhq-5d0.8` merge dup, `qhq-5d0.10` PM truncation, `qhq-5d0.1` PMS preview parity).

High-level groups deferred:

- **Tier-1**: READ_GUIDANCE conditional on connector readiness (#4), `__` namespace guard (#6), `tools/list` pagination (#7).
- **Tier-2**: Sentry breadcrumbs (#9), state-transition logs (#10), `document.hidden` heartbeat gate (#11 → qhq-ekc), per-render memoization (#12 → qhq-ekc), hooks-in-loop component split (#13), a11y on action spans (#14), 401/403 → `McpAuthError` (#16), Bedrock JSON-Schema sanity check (#17), image/resource size cap (#18).
- **Tier-3**: Backend interface vs transport split (#19), `Connectors.ts` 944-line module split (#20), per-backend lifecycle constants (#21), Disconnected note in overview (#22), bootstrap idempotency contract (#23), test gaps (#24), tool-ID hardcoding in guidance (#25), `resetHealth` early-bail comment.

Greptile threads (the 04-21 P2 inline pair) are obsolete — both targeted `PlatformMcpContext/client.ts` which was deleted in the Connectors refactor (the rewrite uses `@effect/platform/HttpClient` + `Eff.Schema` envelope decoding).

drernie (dismissed):
LGTM

nl0:
Status update — pushed `95f2148c..a8c6673f` (10 commits ahead, all CI should re-run).

## Landed

| | Commit | Source |
|---|---|---|
| ✅ | `ead37665` Merge #4864 | drernie's retry/bootstrap fix |
| ✅ | `296c6734` Merge #4865 | drernie's CHANGELOG + Qurator.md docs |
| ✅ | `208ace11` `useActorLayer` interrupts listener fiber on unmount | sir-sigurd Tier-1 #2 |
| ✅ | `db43b2e0` Mcp: content-type lowercase + JSON-RPC id echo + 401/403→Auth | fiskus minor + sir-sigurd Tier-1 #5 + Tier-2 #16 |
| ✅ | `6e231266` Connectors: race heartbeat + probe sleeps against page-activity wake | sir-sigurd Tier-2 #11 (pulled forward from qhq-ekc) |
| ✅ | `a8c6673f` Chat/Input: helper severity classes via lookup | fiskus inline |

The wake-stream change is broader than originally planned for this session — it addresses the laptop-closed / tab-hidden case where browser timers freeze and the heartbeat sleep can fire arbitrarily late on resume, leaving Qurator gated for up to ~95s before reflecting reality. Now wakes immediately on `visibilitychange→visible` / `online`.

## Tests

`npm run test:only -- app/components/Assistant/` → **74 / 74 green** (was 65 pre-session). Added: HTTP id-mismatch (Mcp.spec.ts), 401→Auth mapping (Mcp.spec.ts), wake-cancels-heartbeat (Connectors.spec.ts). Plus the 4 from #4864.

## Still deferred → tracked in `qhq-5d0.16`

Updated the follow-up bead with done-marks on the items landed today. Notable remaining: Tier-1 #4 (READ_GUIDANCE conditional), Tier-2 #9-10 (Sentry/transition logs), #13-14 (a11y / hooks-in-loop), #17-18 (Bedrock schema sanity, image size cap), Tier-3 module restructure / test gaps. None are merge-blockers.

Greptile's two 04-21 inline P2s remain obsolete — they targeted the now-deleted `PlatformMcpContext/client.ts`.

catalog/app/components/Assistant/UI/Chat/Input.tsx
nl0: Took a slightly different shape — restructured the styles so warning/error stack on top of the base hint via `&$hint` parent selectors, then `cx(classes.hint, classes[severity])` does the lookup. Adding a third severity later is one rule, no new branch.

`a8c6673f`

nl0:
Cleanup pass — pushed `a8c6673f..5a97edc2` (2 commits).

**`d1a795a9`** — strip internal design-doc anchors and bead refs from comments. They were author navigation hints into a private design doc; not useful for catalog contributors. Trimmed a few "narrate the code" comments in the same pass. -41 lines net, no behavior change.

**`5a97edc2`** — multi-agent simplify pass:

- `Mcp.ts adaptError`: stop special-casing HTTP 401/403 in the connector adapter. The wire layer now raises `McpAuthError` directly on 401/403, so `adaptError` collapses back to a clean tag-table mapping. `McpAuthError` gains an optional `detail` field; existing "no session token" message preserved as the default.
- Extract `pingOrTimeout` helper — heartbeat and reconnect-probe shared the same `ping → timeoutFail → either` block.
- Add `bumpHealth` helper that caps `consecutiveFailures` at `HEALTH_THRESHOLD + 1` and skips emission when nothing changed. Prevents a sustained outage from spamming subscribers every tick.
- Expose `stateIsUnready(s) = s._tag !== 'Ready'` predicate; replaces two raw `_tag` literals in `Chat.tsx`.
- Comment trims (~22 lines) across `Connectors.ts`, `Mcp.ts`, `Tool.ts`, `Input.tsx`, `DevTools.tsx`, `Actor.ts`.

Tests: 74/74 Assistant green throughout.

Skipped (per agent triage): wake-stream → Context.Tag (over-engineering for one ambient stream), `contextContribution` memoization + `useIsBlocked` dedupe + `DevTools` debounce (all qhq-ekc perf-hardening territory), connector-id constant extraction (cross-file, deferred), `FetchHttpClient.layer` lift-out (cost is negligible — layer is a static `Layer.succeed`).

nl0:
@greptileai one more time pls

catalog/app/components/Assistant/Model/Connectors/Connectors.ts
```diff
@@ -0,0 +1,1016 @@
...
+        : Eff.pipe(
+            Eff.Stream.mergeAll(
+              all.map((r) => r.state.changes),
+              { concurrency: 'unbounded' },
+            ),
+            Eff.Stream.mapEffect(() => isBlocked),
+            Eff.Stream.filter((blocked) => !blocked),
+            Eff.Stream.take(1),
+            Eff.Stream.runDrain,
+          )
```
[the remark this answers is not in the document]
nl0: False alarm — same race sir-sigurd's [multi-agent review](https://github.com/quiltdata/quilt/pull/4840#issuecomment-4343064245) flagged as "did not survive verification". `SubscriptionRef.changes` replays the current value on subscribe, so the merged stream's first emission re-evaluates `isBlocked` against the post-flip state. Added a one-liner comment + a pin test in `5b985447`.

nl0:
@greptileai reassess

drernie (approved):

PR #4856, opened 2026-04-23
Catalog: drop unused linkedData / overviewUrl code paths

nl0:
## Summary

- Removes the JSON-LD stack (`utils/LinkedData.tsx` + all three mount points in Overview, OpenLanding, PackageRevisions/PackageTree comments).
- Removes the `overviewUrl` precomputed-S3-fetch fast paths from `bucketStats` / `bucketSummary` / `bucketReadmes` / `bucketImgs`; clients fall through to the registry `/search` path that already served every deployed bucket.
- Removes the `cfg.linkedData` stack config flag (from `Config.ts` + `config-schema.json`) and the `overviewUrl` / `linkedData` admin form fields.
- `BucketConfigSelection` GraphQL fragment no longer requests these fields; generated types regenerated via `npm run gql:generate`.

## Why

- `cfg.linkedData` was never set in any deployed stack — no variant, template, or installer config sets it. All three `LinkedData.*` components early-return null when the flag is unset, so per-bucket `linkedData` values never reached the DOM anywhere.
- No prod bucket has `overviewUrl` set; the precomputed S3 fetch branches were dead code in practice.

## Scope / out of scope

**This PR is catalog-only.** The GraphQL schema fields (`BucketConfig.{linkedData,overviewUrl}`, `BucketAdd/UpdateInput.*`) and the `quilt3.admin` SDK kwargs are NOT touched — they will be deprecated and removed in a follow-up. Because `BucketAddInput` / `BucketUpdateInput` are generated from the schema and require these fields, `editFormSpec.{overviewUrl,linkedData}` are kept as `() => null` placeholders (matches the existing `prefixes: () => null` idiom at the same site). Net effect: catalog sends `null` for these fields on every bucket edit; server ignores the change for buckets where nothing was set (i.e. all of them).

## Verified

Local smoke-tested against the roles dev stack via webpack-dev-server + Playwright:

- Admin → Buckets → Edit: Metadata card now shows only **Relevance score** and **Tags** — Overview URL and Structured data (JSON-LD) fields are gone.
- `BucketConfigSelection` GraphQL fragment on the wire no longer requests `overviewUrl` / `linkedData` (confirmed in network trace).
- `bucketUpdate` mutation payload contains `overviewUrl: null, linkedData: null` from the `() => null` stubs; server returns HTTP 200; tag edit persisted through the post-mutation refetch.
- No new console errors introduced by the diff.

<h3>Greptile Summary</h3>

This PR removes dead catalog code that was never active in any deployed stack: the JSON-LD (`LinkedData`) component tree, the `overviewUrl` precomputed-S3 fast paths in bucket requests, and the corresponding admin form fields and config schema entries. The `overviewUrl`/`linkedData` fields are stubbed as `() => null` in `editFormSpec` to keep the GraphQL mutation payload valid until the server-side schema drops them.

Safe to merge — purely removes code paths that were provably inert in all deployments.

All changes are consistent dead-code deletions. The return-type change in `bucketReadmes` (from `{forced, discovered}` to a flat array) is correctly reflected in the `Readmes` component. The `() => null` stubs satisfy the GraphQL schema contract explicitly. No logic regressions identified.

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Bucket/requests/requestsUntyped.js | Removes `overviewUrl` branches from `bucketStats`, `bucketSummary`, `bucketImgs`, and simplifies `bucketReadmes` to return a flat array instead of `{forced, discovered}`. |
| catalog/app/utils/LinkedData.tsx | File deleted; the entire JSON-LD component tree (`CatalogData`, `BucketData`, `PackageData`) is removed along with its GraphQL query and generated types. |
| catalog/app/containers/Admin/Buckets/Buckets.tsx | Removes `overviewUrl`/`linkedData` admin form fields; stubs both as `() => null` in `editFormSpec` to keep the `BucketUpdateInput` schema satisfied. |
| catalog/app/containers/Bucket/Overview/Overview.tsx | Drops `overviewUrl` prop threading and `LinkedData.BucketData` mount; updates `Readmes` Ok handler to consume the new flat-array return from `bucketReadmes`. |
| catalog/app/containers/Bucket/Overview/Header.tsx | Removes `overviewUrl` param and the Registry of Open Data on AWS (RODA) attribution block; `bucketStats` call now omits the unused `overviewUrl` arg. |
| catalog/app/containers/Admin/Buckets/gql/BucketConfigSelection.graphql | Fragment no longer selects `overviewUrl` or `linkedData`; generated TypeScript updated consistently. |
| catalog/app/utils/Config.ts | Removes `linkedData` from `ConfigJson` interface; `config-schema.json` updated in lock-step. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph BEFORE["Before — removed paths"]
        A1{overviewUrl set?} -->|yes| B1["Fetch stats.json / summary.json from S3 pre-render bucket"]
        B1 --> C1["Return pre-rendered data"]
        A1 -->|no| D1["Fall through to registry /search API"]
        E1{cfg.linkedData set?} -->|yes| F1["Render JSON-LD script tag via Helmet"]
        E1 -->|no| G1["Return null — always happened in prod"]
    end

    subgraph AFTER["After — current state"]
        H2["bucketStats / bucketSummary / bucketReadmes / bucketImgs"] --> I2["Registry /search API only"]
        J2["LinkedData components deleted"] --> K2["No structured data injected into DOM"]
        L2["Admin Metadata card"] --> M2["Relevance score + Tags only"]
    end
```

<sub>Reviews (2): Last reviewed commit: ["CHANGELOG: trim linkedData/overviewUrl e..."](https://github.com/quiltdata/quilt/commit/364d1bc881eb4ce07761d238a84e4341c255c557) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29474192)</sub>


drernie (dismissed):

catalog/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ where verb is one of
 
 ## Changes
 
+- [Removed] Drop unused `linkedData` (JSON-LD) and `overviewUrl` (precomputed summary/readme/image fetches) code paths from the catalog; the stack-level `cfg.linkedData` config flag and the `overviewUrl` / `linkedData` admin form fields are gone. GraphQL schema fields and quilt3.admin SDK kwargs remain for now; they will be deprecated and removed in a follow-up ([#4856](https://github.com/quiltdata/quilt/pull/4856))
```
fiskus: "they will be deprecated and removed in a follow-up" - not sure if it should belong to changelog and not to the PR description

fiskus (dismissed):

nl0:
@greptileai go again

fiskus (approved):

PR #4862, opened 2026-04-28
Refactor logo upload key + version-pin saved URL

fiskus:
## Summary

Two improvements to `useUploadFile` in `app/utils/CatalogSettings.tsx` (plus matching spec updates):

- **Use `path.extname`** instead of hand-rolled `file.name.split('.').pop()` for the upload key. Same behavior, less code (`c643af0a9`).
- **Capture `VersionId` from `putObject`** and serialize the saved URL via `s3paths.handleToS3Url({ bucket, key, version })` so the persisted `settings.logo.url` is pinned to the exact uploaded version (`252330828`).

The third commit (`d161ad10b` drop `workflow_dispatch` from `deploy-catalog.yaml`) is a no-op against the current base — that change has already been moved to `ci/deploy-catalog-workflow-dispatch`.

## Caveat — IAM

Version pinning produces URLs like `s3://bucket/catalog/logo.jpg?versionId=…`. Logo signs these via `s3.getSignedUrl('getObject', { …, VersionId })`. The signature is correct (verified — signatures differ when `VersionId` is included vs. omitted), but **fetching a specific version requires `s3:GetObjectVersion` on the service bucket**. With only `s3:GetObject`, S3 returns 403 even with a valid signature.

Current dev stack IAM appears to grant only `s3:GetObject` for the service bucket — observed 403 on the signed URL in `tf-dev-unstable`. To use this branch in prod, the service-bucket IAM policy needs `s3:GetObjectVersion` added (or this commit gets reverted).

## Test plan

- [ ] `npm run test:only -- CatalogSettings` passes (4 tests, including a "no versioning" case)
- [ ] On a stack with `s3:GetObjectVersion` granted, upload a logo and confirm it renders, then re-upload (same extension) and confirm older deployments with cached settings still resolve to the original version
- [ ] Confirm the IAM update is rolled out before merging this branch into `feat/logo-file-upload`

### DIFF
--- a/catalog/app/components/BucketIcon/BucketIcon.tsx
+++ b/catalog/app/components/BucketIcon/BucketIcon.tsx
@@ -11,10 +11,11 @@ const useStyles = M.makeStyles((t) => ({
   },
 }))
 
-interface BucketIconProps extends React.DetailedHTMLProps<
-  React.ImgHTMLAttributes<HTMLImageElement>,
-  HTMLImageElement
-> {
+interface BucketIconProps
+  extends React.DetailedHTMLProps<
+    React.ImgHTMLAttributes<HTMLImageElement>,
+    HTMLImageElement
+  > {
   alt: string
   className?: string
   classes?: {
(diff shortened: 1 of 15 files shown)


drernie:
Closeable?

PR #4902, opened 2026-05-15
docs(private-endpoints): lead with symptoms + egress decision

drernie:
**Draft.** Restructures `docs/advanced-features/private-endpoint-access.md` to address gaps surfaced during a recent support engagement, where the page was open during triage but didn't help diagnose a silent ECS-can't-reach-ECR deployment failure.

## Open questions

Tracked as `<!-- TODO -->` markers in the doc; not blocking review of the restructure but should be resolved before merge:

1. Platform-team confirmation that the runtime endpoint list is complete — grep misses transitive calls (e.g., via `quilt3` or third-party libraries). Ideally derived from the CFN template.
2. Exact CloudFormation event text when ECR is unreachable, for searchability.
3. Replace issue link with command-usage docs once `quiltx ecs reachability` ships.

## Summary

- **New "Before you begin" section** framing the page as proactive prerequisites for a planned private-endpoint deployment, with an explicit warning that a permissive-VPC POC stack will not redeploy unchanged into a locked-down VPC.
- **New "Network prerequisites" decision table** (IGW + public subnets / NAT or Transit Gateway / fully isolated VPC) with neutral tradeoff guidance — both NAT/TGW and pure VPC endpoints are supported deployment paths.
- **Reordered required-resources list** so egress (NAT/TGW or full endpoint set) is step 1 instead of step 3.
- **Expanded the "fully isolated VPC" guidance** into a split list:
  - *Startup endpoints* (ECR API, ECR DKR, S3, CloudWatch Logs) — without these the CloudFormation deploy hangs on ECS service creation because the `awslogs` log driver fails to initialize.
  - *Runtime endpoints* by component (ECS registry vs lambdas) — without these the task starts but specific Quilt features return errors.
- **Cross-link** to [quiltdata/quiltx#52](https://github.com/quiltdata/quiltx/issues/52) for the forthcoming `quiltx ecs reachability` diagnostic.

Source-verified against [`deployment/t4/template/containers.py:70`](https://github.com/quiltdata/deployment/blob/master/t4/template/containers.py#L70) (awslogs driver) and `boto3.client(...)` usage in `enterprise/registry/quilt_server/` and `quilt/lambdas/*/src/`.

### DIFF
--- a/docs/advanced-features/private-endpoint-access.md
+++ b/docs/advanced-features/private-endpoint-access.md
@@ -43,11 +43,18 @@ to pull Quilt's container images:
 | --- | --- |
 | Internet Gateway + public subnets | No egress changes needed. |
 | NAT Gateway or Transit Gateway with egress | No egress changes needed. |
-| Neither (fully isolated VPC) | You must create VPC interface endpoints for **every** AWS service Quilt uses (see step 4 below). This is the most restrictive option and requires the most setup. |
+| Neither (fully isolated VPC) | Create VPC interface endpoints for the AWS services Quilt uses (see step 4 below). |
 
-For most customers, **a NAT gateway is the simplest and recommended approach**
-when an Internet Gateway is not acceptable to your security policy.
-See [Amazon's guide on NAT gateways](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-nat-gateway.html#nat-gateway-creating).
+Both NAT/Transit Gateway egress and a complete VPC-endpoint set are
+supported. Choose whichever matches your security posture:
+
+- **NAT or Transit Gateway** is simpler to set up and maintain — one
+  egress path covers every current and future AWS service Quilt calls.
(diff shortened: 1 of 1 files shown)


drernie:
@sir-sigurd @nl0 
This is a crude guess, but it would be good to have clearer guidance.  Feel free to replace or rewrite.

QuiltSimon:
Reconciliation pass (see #engineering thread, 2026-08-29) — **parked with one named question.** The restructure is motivated right (a real support engagement where this page failed to diagnose ECS-can't-reach-ECR), CI is green, and the draft is mergeable — but it ships an endpoint list with the author's own open question standing: is the runtime endpoint list complete?

Named unblock: **@sir-sigurd or @nl0 confirm the endpoint list, ideally derived from the CFN template rather than grep** (the ask has been open since 2026-05-15). A hand-listed endpoint set that drifts is worse than none, so this shouldn't merge on review alone — but it also shouldn't rot: the TODO markers are cosmetic once the list is confirmed.

PR #4903, opened 2026-05-18
docs(entra): clarify protocol and IAM-team handoff

drernie:
## Summary

- State that Quilt uses OpenID Connect with the OAuth 2.0 authorization code flow, not SAML.
- Add a delegated-setup handoff table for customers whose IAM team registers the Entra application.
- Document the exact scopes requested by the Registry: `openid`, `email`, and `offline_access`.
- Remove the obsolete implicit-grant requirement and unused `profile` / `User.Read` permissions.
- Clarify that admin consent depends on organization policy and that group claims are needed only for mappings based on Entra group IDs.
- Point guest, multi-tenant, and high-group-membership deployments toward App roles.
- Keep the SSO permissions-mapping guide consistent with the Registry scopes.

Motivated by an enterprise rollout where the customer IAM intake form did not provide enough information to register the application.

Austin reported a separate advanced Entra claims-mapping/signing-key concern. That investigation is tracked in #5191 so the basic built-in groups-claim setup is not conflated with claims-mapping policies.

## Test plan

- [x] Run Markdown lint across the repository.
- [x] Run `git diff --check`.
- [x] Confirm the new relative documentation link resolves.
- [x] Verify the documented authorization flow and scopes against the Registry implementation.


Austin-s-h:
Additional Entra-side documentation about SSO...
1. If using group to role claim mapping in Quilt, an application-specific signing key is required in non single-tenant deployments. BOTH Sign + Verify keyCredentials + preferredTokenSigningKeyThumbprint) AND keep the claims policy scoped to "Groups assigned to the application".
2. If this key is not provided, groups based role assignment can work briefly until the signing key is requested.
3. If a user is part of many (150-200+) group claims, the response signature from the Entra side is different. Unclear if Quilt handles this. 


```
Entra rule: a token's claims are normally signed with Microsoft's **shared** signing key. The
moment you attach a **claims-mapping policy** (used to emit/transform custom claims — e.g. the
`groups` customization that was set up here), Entra requires the application to either:

1. be configured with an **application-specific (custom) token signing key** (upload a cert and
   set it as the token-signing key in the manifest, `keyCredentials` + `tokenEncryptionKeyId`/
   signing key id), **or**
2. set **`acceptMappedClaims: true`** in the app manifest's `api` section (tells Entra to accept
   mapped claims signed with the shared key — single-tenant only).

What happened here was:

1. **Baseline:** plain OIDC, no custom claims. Every user resolved to `default_role` (effectively
   "public"). Working.
2. **Change:** a **claims-mapping policy** was added to emit the `groups` claim so Quilt could map
   group membership → roles. It worked *briefly*.
3. **Enforcement:** Entra then enforced its rule — *a claims-mapping policy requires an
   application-specific signing key (or `acceptMappedClaims`)*. The app had **neither**
   (`acceptMappedClaims` is `null`, no custom key was ever uploaded) → **AADSTS50146** → no token.

So "worked the first day, then broke" = **Entra's signing-key enforcement kicking in after the
policy was attached**, not a lapsed certificate. The claims-mapping policy is still assigned to the
service principal, and removing the `groups` *claim* from token config does **not** remove the
*policy assignment* — which is the part that triggers the signing-key requirement.
```

PR #4924, opened 2026-05-28
refactor(catalog): add bucket context and S3 config factory

drernie:
## Current pain
Bucket-route components repeatedly read `useParams().bucket`, and S3 clients are created without a single bucket-config-aware path. That makes region/presign behavior easy to drift as more bucket-specific configuration is introduced.

## Changes
- Add `BucketContextProvider` and `useBucketContext()` for the bucket route subtree.
- Mount the provider at `containers/Bucket/Bucket.tsx`, reusing existing bucket existence/cache plumbing without adding a fetch.
- Replace direct bucket param reads under `containers/Bucket/*` with `useBucketContext()` where they read the route bucket.
- Extend the existing `useS3Factory()` / `useS3()` path to accept bucket config and preserve the old region-string/default behavior.
- Route bucket read-site S3 consumers through `useBucketContext().config`.
- Add hook/provider tests and update direct hook tests to mount the provider.

## Bucket context inventory
Direct bucket route reads were replaced in overview, dir/file, package tree/revisions/compare, queries, workflows, and package-dialog/file-picker paths. `Bucket.tsx` remains the route owner and provider mount point.

## S3 construction inventory
No separate S3 factory was added. Existing `utils/AWS/S3.jsx` remains the construction point; bucket subtree `AWS.S3.use()` call sites now pass bucket context config.

## Verification
- `npm run lint`
- `npm run test:only -- app/containers/Bucket/context.spec.tsx app/containers/Bucket/Queries/Athena/model/state.spec.tsx app/containers/Bucket/PackageDialog/State/schema.spec.ts app/containers/Bucket/PackageDialog/State/params.spec.ts app/containers/Bucket/Dir/Toolbar/CreatePackage/useSuccessors.spec.ts --no-cache` — 37 passed
- `NODE_OPTIONS="--localstorage-file=/tmp/quilt-catalog-vitest-localstorage.json" npm test` currently fails in unrelated suites under Node 26/npm 11:
  - `app/utils/spreadsheets/spreadsheets.spec.ts`: date expectations off by one day for ODS/FODS/CSV cases
  - `app/containers/Bucket/Queries/Athena/model/requests.spec.ts`: two `useQueryRun` async timing tests time out

## Pre-merge stack verification
Not run yet. Before merge: workflow-dispatch an interim catalog image for this branch, pin the immutable tag on the shared test stack, and record the test-stack URL/image tag here.

<h3>Greptile Summary</h3>

This PR centralizes bucket route reads into a `BucketContextProvider` / `useBucketContext()` hook pair, replacing scattered `useParams().bucket` calls across the bucket subtree and routing all S3 client construction through a bucket-config-aware `useS3(bucketConfig)` path.

- **New context module** (`context.tsx`): introduces `BucketContextProvider` (auto-reads `:bucket` from route params or accepts explicit `bucket` + `config` props) and `useBucketContext()` consumed by ~20 files previously doing direct `useParams` reads.
- **S3 factory extension** (`S3.jsx`): `useS3` / the internal factory now accept a `BucketConfig` object (or legacy region string) so that bucket-specific region is forwarded to the cached client; backward-compatible with callers that pass nothing.
- **Double-provider in `Bucket.tsx`**: an outer route-based provider lets `BucketInner` call `useBucketStrict()`, while a separate inner provider (mounted after `useBucketExistence` resolves) supplies the cached region config to the actual page components.

The refactor is mechanically correct and well-tested; the two-provider nesting in Bucket.tsx is a design choice that works today but leaves a latent trap for future components added in the wrong layer.

The migration is broad (30+ files) but each change is a straightforward substitution. The main fragility is that `config = {}` in `BucketContextProvider` creates a new object reference on every render, defeating the `useMemo` in `BucketContextValueProvider` and causing unnecessary context re-renders for all consumers of the outer provider. The double-provider pattern is functional but requires contributors to understand which layer they're adding code to.

catalog/app/containers/Bucket/context.tsx (config default stability) and catalog/app/containers/Bucket/Bucket.tsx (two-layer provider structure).

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Bucket/context.tsx | New file: implements BucketContextProvider and useBucketContext; the `config = {}` default creates an unstable object reference on each render, defeating the useMemo inside BucketContextValueProvider. |
| catalog/app/containers/Bucket/Bucket.tsx | Mounts two nested BucketContextProviders: an outer one (no config) for BucketInner and an inner region-aware one inside BucketReady; components between the two layers silently get the empty-config provider. |
| catalog/app/utils/AWS/S3.jsx | Extended useS3/useS3Factory to accept bucketConfig (string | object | undefined) with backward-compatible string-region path; cache key is still region only, which is correct given the current BucketConfig shape. JSDoc comment not updated. |
| catalog/app/containers/Bucket/Routes.ts | useBucketSafe now delegates to useBucketContext().name instead of useParams; contract changes from returning undefined to throwing if used outside the provider, but the only call site is useBucketStrict which re-validated anyway. |
| catalog/app/containers/Bucket/context.spec.tsx | New test file covering route-based and explicit-prop provider paths; covers the happy paths but lacks a test for the useBucketContext error when no provider is present. |
| catalog/app/containers/Bucket/Queries/Athena/model/state.tsx | Replaces direct useParams bucket read with useBucketContext; removes invariant that is now enforced by the provider. |
| catalog/app/containers/Bucket/Dir.tsx | Replaced useParams bucket read with useBucketContext and passes bucket config to AWS.S3.use(); straightforward migration. |

</details>

1. `catalog/app/utils/AWS/S3.jsx`, line 154-156 ([link](https://github.com/quiltdata/quilt/blob/c87027dd5f0f213b95b0d9a653556112bfbed806/catalog/app/utils/AWS/S3.jsx#L154-L156)) 

   <a href="#">P2</a> The JSDoc comment for `useS3Factory` still documents the old `(region?) => S3Client` signature. The factory returned now accepts a `bucketConfig` argument that can be a region string, a `BucketConfig` object, or `undefined`, so the comment should be updated to avoid misleading callers who consult it.

</details>

<sub>Reviews (1): Last reviewed commit: ["refactor(catalog): add bucket context an..."](https://github.com/quiltdata/quilt/commit/c87027dd5f0f213b95b0d9a653556112bfbed806) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34299174)</sub>

### DIFF
--- a/catalog/app/containers/Bucket/Bucket.tsx
+++ b/catalog/app/containers/Bucket/Bucket.tsx
@@ -6,7 +6,7 @@ import Layout, { Container } from 'components/Layout'
 import Placeholder from 'components/Placeholder'
 import { useBucketStrict } from 'containers/Bucket/Routes'
 import { NotFoundInTabs } from 'containers/NotFound'
-import { useBucketExistence } from 'utils/BucketCache'
+import { useBucketExistence, useGetCachedBucketRegion } from 'utils/BucketCache'
 import * as NamedRoutes from 'utils/NamedRoutes'
 import * as BucketPreferences from 'utils/BucketPreferences'
 import MetaTitle from 'utils/MetaTitle'
@@ -14,6 +14,7 @@ import * as RT from 'utils/reactTools'
 
 import * as AssistantContext from './AssistantContext'
 import * as BucketNav from './BucketNav'
+import { BucketContextProvider } from './context'
 import type { RouteMap } from './Routes'
 import * as Selection from './Selection'
 import { displayError } from './errors'
@@ -57,7 +58,7 @@ function BucketLayout({ bucket, children }: BucketLayoutProps) {
           </M.AppBar>
           <Container>
             {bucketExistenceData.case({
-              Ok: () => children,
+              Ok: () => <BucketReady bucket={bucket}>{children}</BucketReady>,
               Err: displayError(),
               _: () => <SuspensePlaceholder />,
             })}
@@ -68,7 +69,29 @@ function BucketLayout({ bucket, children }: BucketLayoutProps) {
   )
 }
 
+function BucketReady({ bucket, children }: BucketLayoutProps) {
+  const getCachedBucketRegion = useGetCachedBucketRegion()
+  const config = React.useMemo(
+    () => ({ region: getCachedBucketRegion(bucket) }),
+    [bucket, getCachedBucketRegion],
+  )
+
+  return (
+    <BucketContextProvider bucket={bucket} config={config}>
+      {children}
+    </BucketContextProvider>
+  )
+}
+
 export default function Bucket() {
+  return (
+    <BucketContextProvider>
+      <BucketInner />
+    </BucketContextProvider>
+  )
+}
+
+function BucketInner() {
   const bucket = useBucketStrict()
 
   const { paths } = NamedRoutes.use<RouteMap>()
--- a/catalog/app/containers/Bucket/Dir.tsx
+++ b/catalog/app/containers/Bucket/Dir.tsx
@@ -20,6 +20,7 @@ import * as Listing from './Listing'
 import * as FI from './PackageDialog/Inputs/Files/State'
 import * as Selection from './Selection'
 import Summary from './Summary'
+import { useBucketContext } from './context'
 import * as DirToolbar from './Dir/Toolbar'
 import { displayError } from './errors'
 import * as requests from './requests'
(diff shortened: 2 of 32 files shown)


drernie:
Addressed greptile review feedback:

- **context.tsx (P2, unstable default reference):** Fixed. Replaced the inline `config = {}` default with a module-level `const EMPTY_CONFIG: BucketConfig = {}` so the `useMemo([name, config])` dependency stays stable and consumers don't re-render unnecessarily. (123ea8a)

- **Bucket.tsx (P2, outer provider supplies `config: {}`):** No change. This is an observation about a possible future footgun (components between the outer and inner providers reading `config` for S3 construction); the comment confirms none do so today and proposes no concrete change. Noting it for awareness; nothing actionable now.

drernie:
Obsoleted by Data Products

PR #4921, opened 2026-05-27
Catalog: Glacier rehydration UX in the preview

fiskus:
## Summary

Replaces the dead-end "Object Archived — Preview not available" message with a rehydration flow for Glacier (`GLACIER`) / Deep Archive (`DEEP_ARCHIVE`) objects.

- Restore state is read from the `x-amz-restore` HEAD header. "Archived" now means *effectively* archived — a restored object previews normally and re-enables the toolbar.
- `RehydrateDialog`: pick a tier (Standard / Bulk / Expedited) and duration (1–90 days, validated). Expedited is offered only when the storage class resolves to `GLACIER` (hidden for `DEEP_ARCHIVE` and for unknown / HEAD-less classes, which fall back to Standard / Bulk). Errors show inline.
- `ArchivedMessage`: archived → restoring → restored. No background polling — while restoring it asks the user to refresh the page to check progress; the state is held optimistically after a `202`.
- Wired through all file views (standalone, packaged, embed).
- A per-bucket `ui.actions.restore` BucketPreference (default **on**) controls whether the Rehydrate button is shown, mirroring the other `ui.actions.*` toggles. The button is also hidden from unauthenticated users — restore is billable and mutates objects, so it's authenticated-only, like download/zip. Hiding it is UI-only — it neither grants nor removes access; the user's IAM role does.
- The `x-amz-restore` CORS expose-header (`s3-proxy/nginx.conf`) the browser HEAD needs to read restore state was already merged in #4953 — s3-proxy is a separately built/deployed component, so this PR stays pure catalog scope.

## Implementation notes

- The restore runs through a new `restoreObject` GraphQL mutation under the user's role credentials, replacing the old direct `s3.restoreObject` SDK call. Per-user IAM is preserved. Typed errors map to UI affordances via `interpretResult`: `RestoreAlreadyInProgress` → in-progress flip, `RestoreAccessDenied` → inline "ask your admin to enable rehydration" IAM hint, `InvalidObjectState` → calm "not archived / already restored" message, `GlacierExpeditedUnavailable` → suggest Standard/Bulk, `BucketNotFound` / `ObjectNotFound` → calm "no longer exists" copy.
- Restore state in listings, the version popover, and the gallery comes from S3's per-object `RestoreStatus` (requested via `OptionalObjectAttributes`), so a restored object reads as *not* archived there too — at no extra request cost, since the status rides along in the existing LIST call. The file-detail page additionally reads the live `x-amz-restore` HEAD header.

## Test plan

[...]

PR #4956, opened 2026-06-09
fix(catalog): admin Buckets crash after Add via front-page link

sir-sigurd:
# Description

Adding a bucket from the landing-page Add link (`+` icon in the bucket grid for admins with ≤8 buckets, or the "Add Bucket" button below it for admins with >2 buckets) crashed with "Unexpected Error" after the mutation succeeded — the bucket was actually added; only the UI errored. Any direct entry into `/admin/buckets?add=true` (bookmark, page refresh on the Add page, etc.) hits the same crash. Going through the admin Buckets list `+` button does not, because the list query seeds the cache first.

## Root cause

The `bucketAdd` Mutation updater in `catalog/app/utils/GraphQL/Provider.tsx` corrupted the urql graphcache whenever `bucketConfigs` had not been queried before the mutation fired — i.e. any navigation into `/admin/buckets?add=true` that did not first render the admin Buckets list.

Chain of events:

1. `cache.updateQuery({ query: BUCKET_CONFIGS_QUERY }, R.evolve({...}))` runs.
2. `readQuery` returns `null` (the query was never executed, nothing cached).
3. `R.evolve(transformations, null)` returns `{}` — `for..in null` does not iterate.
4. `updateQuery` only skips the write when the updater returns `null`; `{}` is not null, so it proceeds.
5. In production builds (only), `writeSelection` falls through with `bucketConfigs: undefined` → `ensureData(undefined) === null` → `writeLink('Query', 'bucketConfigs', null)` stores an explicit null link.
6. Subsequent schema-aware reads check `undefined`, not `null` — so they return `{bucketConfigs: null}` despite the NON_NULL schema.
7. `useQueryS` passes the truthy data through; `List.tsx` destructures `rows = null`; `R.sortBy(fn)(null)` calls `Array.prototype.slice.call(null, 0)` → `TypeError: can't convert null to object`.

Dev builds hide this because graphcache's dev branch warns and `continue`s on the undefined field instead of writing.

Latent since #2238 (June 2021), when the `bucketAdd` updater was first introduced. The landing-page Buckets grid used to seed `bucketConfigs` via `useRelevantBucketConfigs`, which incidentally masked the bug on the landing-page entry point. #4839 (April 2026) migrated that hook to `useRelevantBuckets` against the new role-scoped `buckets` field, so the landing page no longer seeds `bucketConfigs` — exposing the latent bug whenever an admin clicks Add from the landing page (or arrives at `/admin/buckets?add=true` by any other path that bypasses the admin list).

## Fix

Guard every mutation cache updater in `Provider.tsx` against uncached reads, via a shared `evolveCached(transformations)` helper (`R.unless(R.isNil, R.evolve(transformations))`). On an uncached read the updater returns null, `updateQuery` skips the write, and the next read of that query misses cache and refetches via Suspense — returning fresh post-mutation data.

Scope (grew during review with @nl0's sign-off):

- `bucketAdd` — the reported crash.
- `bucketRemove` — four writes, same class; the role-scoped `buckets` write is a genuine exposure on bookmarked `/admin/buckets` navigation.
- The policy / role create/update/delete cluster and the `USERS_QUERY` add/remove-user pair — the `roleDelete` / `policyDelete` / `roleUpdateManaged` strips of ad-hoc `{ policies { id roles { id } } }` shapes mirror the original bug (the admin Roles/Policies pages may never have seeded those queries).
- Two custom (non-`R.evolve`) updaters with the same hazard: `roleDelete`'s `DEFAULT_ROLE_QUERY` updater (dereferenced `data.defaultRole` on a possibly-null read) and the `tabulatorOpenQuery` updater (destructured `({ admin })` from it).

Intentionally untouched: `roleSetDefault`'s `DEFAULT_ROLE_QUERY` updater (ignores the cached read and always writes a complete result — valid even on a cold cache) and inner `R.evolve` calls inside `R.map(...)` (they transform list elements, never the root read).

Verified with a standalone repro against the actual `urql@2.2.3` + `@urql/exchange-graphcache@4.4.3` from `catalog/node_modules`: the null link only appears in `NODE_ENV=production`, only when the query was never seeded, and only when the updater returns a non-null value from a null read. With the guard in place, both seeded and unseeded cases work.

Remaining followup (not in this PR): the guard semantics are coupled to consumers reading through Suspense (`useQueryS`) — a future `cache-first` / `cache-only` reader of these queries would silently miss post-mutation updates skipped on a cold cache.

## TODO

<h3>Greptile Summary</h3>

Guards the `bucketAdd` graphcache updater against a null-read that corrupted `Query.bucketConfigs` in production builds when the user arrived at `/admin/buckets?add=true` via the landing-page link before the bucket list had ever been fetched. The fix — `R.unless(R.isNil, R.evolve({...}))` — is the same pattern already applied to `roleDelete` in #4751.

- **Root cause fixed**: `R.evolve(transform, null)` returns `{}` (not `null`), so `updateQuery` proceeded to write a null link for `bucketConfigs`; subsequent reads returned `{bucketConfigs: null}` which crashed `R.sortBy` in `List.tsx`.
- **Pattern gap**: `bucketRemove` still calls `cache.updateQuery` with bare `R.evolve` on both `BUCKET_CONFIGS_QUERY` and `BUCKETS_QUERY` — structurally identical exposure, though it requires an unusual navigation path to trigger.

Safe to merge — the targeted fix correctly prevents graphcache corruption on the reported landing-page path and matches the established roleDelete guard pattern.

The bucketAdd fix is correct and well-tested per the PR description. The only gap is that bucketRemove carries the structurally identical unguarded R.evolve calls on BUCKET_CONFIGS_QUERY and BUCKETS_QUERY, but the normal removal workflow loads the bucket list first, making accidental triggering unlikely today.

The bucketRemove updater block in catalog/app/utils/GraphQL/Provider.tsx (lines 227–261) would benefit from the same R.unless(R.isNil, ...) guard applied in this PR.

| Filename | Overview |
|----------|----------|
| catalog/app/utils/GraphQL/Provider.tsx | Fixes null-cache corruption in `bucketAdd` updater by wrapping `R.evolve` with `R.unless(R.isNil, ...)`; `bucketRemove` retains the same unguarded pattern for both `BUCKET_CONFIGS_QUERY` and `BUCKETS_QUERY`. |
| catalog/CHANGELOG.md | Prepends a clear [Fixed] entry describing the bug, the trigger path, and the cache effect. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["bucketAdd mutation succeeds\n(__typename === 'BucketAddSuccess')"] --> B["cache.updateQuery(BUCKET_CONFIGS_QUERY, updater)"]
    B --> C["cache.readQuery internally\nreturns data?"]
    C -->|"data is null\n(query never fetched)"| D["R.unless(R.isNil, ...) → returns null"]
    C -->|"data is {bucketConfigs:[...]}\n(query was cached)"| E["R.evolve appends new bucketConfig"]
    D --> F["updateQuery skips write\n(updater returned null)"]
    E --> G["Cache updated with new bucket"]
    F --> H["Next list read: cache miss\n→ Suspense refetch → full list returned ✓"]
    G --> I["List renders with new bucket ✓"]
```

1. `catalog/app/utils/GraphQL/Provider.tsx`, line 227-261 ([link](https://github.com/quiltdata/quilt/blob/b3e20a002ca30bcf86886fea4e52f74b61ccf431/catalog/app/utils/GraphQL/Provider.tsx#L227-L261)) 

   <a href="#">P2</a> **`bucketRemove` has the same unguarded `R.evolve` on `BUCKET_CONFIGS_QUERY` and `BUCKETS_QUERY`**

   The same null-read corruption path exists here: if `bucketConfigs` (or `buckets`) was never fetched before the mutation fires, `cache.readQuery` returns `null`, `R.evolve(_, null)` returns `{}`, and `updateQuery` writes a null link — identical to the `bucketAdd` bug this PR fixes. In normal admin-list navigation this is unlikely because the user must see the list to click Remove, but it is the same structural gap. Wrapping the two `R.evolve` calls with `R.unless(R.isNil, ...)` would make the updater consistent with the pattern now established for `bucketAdd` and `roleDelete`.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(catalog): admin Buckets crash after ..."](https://github.com/quiltdata/quilt/commit/b3e20a002ca30bcf86886fea4e52f74b61ccf431) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36479820)</sub>


sir-sigurd (commented):
The fix itself is correct and well-scoped for the reported crash: the `R.unless(R.isNil, …)` guard matches the existing `roleDelete` pattern, the null-skip semantics are right, and the bug reproduces end-to-end (`List.tsx` feeds a possibly-null `bucketConfigs` into `R.sortBy`, and `bucketConfigs` is NON_NULL in the schema). Worth noting on the upside: a cache already corrupted by a prior crash (`{bucketConfigs: null}`) self-heals on the next add. **Design**: limited surface — a one-line cache guard, no public API or migration involved. Nothing below blocks merge; the main item is a sibling gap with the same root cause, plus a couple of questions about the guarding strategy.

### P2

**`bucketRemove` carries the same unguarded pattern.** Lines 227–261 have four bare `cache.updateQuery(..., R.evolve(...))` calls with no `R.unless(R.isNil, …)`. The `bucketConfigs` one is low-risk — an admin has to load the list to click Remove, so that query is seeded. But the `buckets` one is a genuine exposure: `buckets` is role-scoped and may never be fetched in an admin session reached directly (e.g. a bookmarked `/admin/buckets`), so a removal would write a null `Query.buckets` link — the same corruption this PR fixes for `bucketConfigs`. The two permission-stripping calls (`policies`/`roles`) are bare too. Your call whether to guard these here or in a follow-up.

### Questions worth answering

**Spot-guarding vs. the root cause.** This guards one call site and points to `roleDelete` as the precedent — but `roleDelete` itself only guards one of its three `updateQuery` calls (line 379 is guarded; 367 and 392 are not), and roughly six other updaters carry the same bare pattern. The root cause you diagnosed lives in graphcache (an updater returning `{}` becomes a null link), so the next updater author re-arms it with no guardrail. Worth deciding: is a shared `updateQuery` wrapper that applies `R.unless(R.isNil, …)` by default — or invalidating the root field instead of relying on a Suspense refetch — a more durable fix than per-call-site guards?

**Correctness is coupled to the consumer's read policy.** Skipping the write and letting the list refetch only surfaces the new bucket because `List.tsx` reads through `useQueryS` (Suspense), which refetches on a cache miss. A future `cache-first`/`cache-only` reader of `bucketConfigs` would silently miss the just-added bucket, with no crash to signal it. Latent rather than a defect today, and related to the point above.

### Nits

- **No regression test.** There's a precise repro, but nothing in-repo fails without this change, and the production-only manifestation makes a silent regression plausible on a future urql/graphcache upgrade. Acknowledged that there's no test harness for these updaters today and `roleDelete` set the no-test precedent.
- **CHANGELOG entry has no PR link.** Every other entry carries a `([#NNNN](...))` reference; this one doesn't — add `([#4956](...))` for consistency.

---
Reviewed against b3e20a0

sir-sigurd:
Thanks for the review. Addressed in 8a6555e7.

**P2 — bucketRemove.** Guarded all four `R.evolve` calls there with `R.unless(R.isNil, ...)`. Agreed that the `bucketConfigs` one is low-risk in normal Remove navigation, but the `Query.buckets` one is a genuine bookmarked-`/admin/buckets` exposure (your wording), and the two permission-stripping calls match the same null-read pattern — once one of them is guarded the others are cheap, and shipping a PR that leaves a known sibling open felt wrong.

**Spot-guarding vs. root cause.** Agreed the root cause lives in graphcache and that this leaves ~6 other call sites unguarded plus two in `roleDelete`. A shared `updateQuery` wrapper (or `cache.invalidate(rootField)` for the non-optimistic alternative) is the durable fix; happy to do it as a followup once the immediate user-visible crashes are out of the way. Tracking that mentally as the next ticket — out of scope here to keep this PR scoped to the reported crash + the symmetric sibling.

**Cache-first/cache-only coupling.** Acknowledged. The guard is correct for every consumer today (all `useQueryS`, Suspense-refetching on cache miss), but a future cache-first/cache-only reader of `bucketConfigs` would silently miss the new bucket. That's the same problem the shared-wrapper refactor would solve once and for all — write-through optimistic update when the query is seeded, invalidate when it isn't. Filing as a followup along with the spot-guard sweep.

**Regression test.** No harness for these updaters today — `roleDelete` set the no-test precedent, and the standalone repro at `/tmp/claude/repro_final.mjs` is the closest thing. Worth adding alongside the shared-wrapper refactor.

**CHANGELOG link.** Fixed in 8a6555e7.

sir-sigurd (commented):
Follow-up round — this is likely the last one once the item below is resolved or consciously deferred.

The prior round's items are handled: `bucketRemove` is guarded, the shared `evolveCached` helper is in, and the CHANGELOG now carries the PR link (which also resolves Copilot's inline note about the missing link).

### P2

**The helper is half-applied, and the remaining gap is reachable.** `evolveCached` is wired into 6 of the ~16 `cache.updateQuery` call sites; the rest still use bare `R.evolve(...)` and keep the null-read exposure the helper exists to remove. The clearest case mirrors the fix you just made: `roleDelete` (line 369), `policyDelete` (309), and `roleUpdateManaged` (344) strip an ad-hoc `{ policies … }` / `{ roles … }` query that the admin Roles/Policies page may never have seeded — structurally identical to the `bucketRemove` permission-stripping calls now guarded. If that root field is uncached, the bare `R.evolve` writes a null `policies`/`roles` link and the next visit to that page reads `null`, the same crash class this PR fixes. So in the same file, `bucketRemove`'s strips are now safe while `roleDelete`'s are not — and `roleDelete` still guards only one of its three writes (line 379).

Remaining bare `R.evolve` call sites: 269, 279, 300, 309, 323, 333, 344, 369, 391, 429, 438. Mechanical to convert now that the helper exists.

Related: `roleDelete`'s `DEFAULT_ROLE_QUERY` updater (line 395) reads `data.defaultRole?.id` — if that query was never cached, `data` is `null` and this throws. Different shape (a custom function, not `R.evolve`), so `evolveCached` doesn't cover it; it'd need `data?.defaultRole`.

Your call on scope — finish here or in a follow-up.

---
Reviewed against 806cb9f

sir-sigurd:
@nl0 — could you weigh in on scope here?

The round-2 self-review flagged that `evolveCached` is half-applied: 6 of ~16 `cache.updateQuery` call sites in `catalog/app/utils/GraphQL/Provider.tsx` use the helper; the rest still use bare `R.evolve(...)`. The three sites most closely matching the bug pattern this PR fixes are:

- `roleDelete:369` — strips deleted Role from cached `{ policies { id roles { id } } }`
- `policyDelete:309` — strips deleted Policy from cached `{ roles { id ... on ManagedRole { policies { id } } } }`
- `roleUpdateManaged:344` — recomputes `{ policies { id roles { id } } }` to reflect the role's new policy set

All three target ad-hoc queries that the admin Roles/Policies pages may never have seeded; a mutation reached without that page first loading would write a null link, same crash class as the original `bucketAdd` bug. The remaining bare `R.evolve` sites (lines 269, 279, 300, 323, 333, 391, 429, 438) all have the same structural exposure but are less directly tied to a known crash path.

Separate item: `roleDelete:395` (`DEFAULT_ROLE_QUERY` updater) is a different shape — a custom updater function that reads `data.defaultRole?.id`, which throws on uncached `data === null`. Not covered by `evolveCached`; needs `data?.defaultRole?.id`.

**Question:** finish the sweep here (mechanical, ~11 lines + one defaultRole fix) or defer to a follow-up? Sergey doesn't have the historical context to make the call confidently.

nl0:
including the related fixes seems fine

sir-sigurd:
Sweep completed in 435c33c4 per @nl0's go-ahead:

- All 11 remaining bare `R.evolve` call sites converted to `evolveCached` (lines from the round-2 list).
- `roleDelete`'s `DEFAULT_ROLE_QUERY` updater now reads `data?.defaultRole?.id` and passes the null through (write skipped).
- One more custom updater with the same hazard, not on the round-2 list: the `tabulatorOpenQuery` updater destructured `({ admin })` from a possibly-null read — now returns null on a cold cache.

Intentionally untouched: `roleSetDefault`'s `DEFAULT_ROLE_QUERY` updater (ignores the cached read and always writes a complete result — valid even on a cold cache), and the inner `R.evolve` calls inside `R.map(...)` (they transform list elements, never the root read).

sir-sigurd (commented):
Notes from a self-review pass (round 3, closing): verified the sweep against `b397ddd` rather than taking the summary comment's word — all 11 call sites from the round-2 list are converted, the only bare `R.evolve` calls left are inner element-transformers inside `R.map(...)` plus the intentionally-unconditional `roleSetDefault` writer, the two custom updaters (`DEFAULT_ROLE_QUERY` in `roleDelete`, `tabulatorOpenQuery`) now return null on a cold cache so the write is skipped, and warm-cache behavior is unchanged (`R.unless(R.isNil, …)` is identity-equivalent to the previous code for non-nil reads). CHANGELOG matches the final scope and carries the PR link. Nothing new flagged; `test-catalog` is green on this head. The failing `codecov/project/auto-lambda` status is not required and not attributable here (no lambda code touched; the report compares 3 commits behind master).

Reviewed against b397ddd

nl0 (dismissed):

sir-sigurd:
@nl0 sorry — the conflict-resolution merge commit (CHANGELOG entry collided with #4921) dismissed your approval. The only change since you approved is that merge commit; could you re-approve?

nl0 (approved):

PR #4957, opened 2026-06-09
docs: document admin flag and union_roles behavior for SSO mappings

drernie:
Documents the SSO permission-mapping behavior introduced with `union_roles`
(Quilt Platform 1.69+), which neither the guide (`sso-permissions.md`) nor
the published schema (`sso-config-1.0.json`) currently covers.

Under `union_roles: true`, the `admin` flag is **tri-state** and is not
unioned the way `roles` are:

- omitted (or null) — the mapping does not vote on admin,
- `true` — the mapping grants admin,
- `false` — the mapping vetoes admin.

A user is made admin only if at least one matching mapping sets `admin: true`
and no matching mapping sets `admin: false`.

## `sso-permissions.md`

- Add a Note documenting the tri-state semantics and the "any `false` vetoes" rule.
- Add guidance: on broad catch-all mappings that privileged users also match, omit `admin` rather than setting `admin: false`.
- Clarify the worked example to reinforce why its second mapping omits `admin`.

## `sso-config-1.0.json`

The guide points users at this schema for field descriptions, but it predated `union_roles`:

- Add the missing `union_roles` property.
- Update `admin` to the tri-state nullable it actually is, with the union semantics described.
- Refresh the stale `mappings` description (no longer "Only the first matching mapping is used").
- Place `default_role` inside `properties` so it is actually validated — it was a no-op key at the schema root.

Docs/schema only; no Quilt Platform runtime behavior change. (Note: this does
change what validates against the published schema — moving `default_role`
into `properties` makes it actually validated, and `admin: null` flips from
invalid to valid — but nothing in this repo validates configs against the
schema file.)


sir-sigurd (commented):
P1 below: the contract text for first-match mode is wrong about omitted `admin`. **Design**: blocker until that's fixed — scope coherence and the schema edits themselves otherwise check out: nullable `admin`, the new `union_roles` property, and the `default_role` move into `properties` all verify against the registry implementation (the old root-level `default_role` was indeed an ignored unknown keyword; presence was already enforced by root `required`). Holding wording-level polish for a later pass — the items below would reshape the same paragraphs.

## P1

**P1 (reshape)** — `docs/advanced-features/sso-permissions.md:53-54` and the schema's `admin` description misdocument first-match mode. With `union_roles: false`, the registry still computes the admin flag from the single kept mapping and writes it unconditionally — a matched mapping that **omits** `admin` sets the user's admin flag to `false` on every login (enterprise `views/auth.py`: `admin = any(v is True ...) and not any(v is False ...)` over `matches[:1]`, then `user.is_admin = admin`). "Its `admin` value alone is used" / "its value is used directly" plus `"default": null` read as if omission were neutral in this mode; the old `"default": false` pinned the omitted case (and couldn't survive union mode, where `false` now means veto — but its replacement leaves the omitted case defined nowhere). State explicitly: in first-match mode, omitted/null `admin` is equivalent to `admin: false` — matched users are demoted.

## P2

**P2** — `docs/advanced-features/sso-permissions.md:42-53`: the page never states the underlying rule — for any user matching at least one mapping, the admin flag is recomputed and written on every login. Consequence: under `union_roles: true`, if **every** matching mapping omits `admin`, the user is demoted; omitting `admin` on the catch-all protects only users who also match an `admin: true` mapping. A manually-promoted admin who matches only the catch-all is demoted whether the catch-all omits `admin` or sets it `false`. Line 97's "Their admin flag is unchanged" covers only zero-match users — worth stating the matched-user rule right in the tri-state note.

**P2 (reshape)** — `docs/advanced-features/sso-permissions.md:88-94`: the walkthrough asserts the `admin@example.com` user is granted both roles and that the second mapping's omitted `admin` is what spares them — but mapping 2 only matches if the token carries a `groups` claim containing `rw`, which the example never grants that user. Without it, the user matches mapping 1 only and the new "remains admin because the second mapping **omits** `admin`" sentence describes a match that doesn't occur. One clause fixes it ("assuming their token also carries group `rw`"). The "granted both" claim predates this PR, but this PR rewrites the paragraph and adds reasoning that depends on the assumption.

**P2** — `docs/advanced-features/sso-permissions.md:39` + both new schema descriptions: no version gate. Pre-1.69 registries **silently accept** a config with `union_roles: true` and ignore the key (pydantic drops unknown fields; the upload mutation only rejects YAML/validation errors) — first-match stays in effect while the admin believes union+veto semantics are active. Verified against the pre-1.69 registry code and reproduced with pydantic v1. The sibling note already carries "(Quilt Platform 1.69+)"; the tri-state note and the schema's `admin`/`union_roles` descriptions should too, ideally with a sentence that older platforms ignore the key silently rather than rejecting it. Older registries also reject an explicit `admin: null` at upload (pydantic: "none is not an allowed value"), so the schema's null-acceptance is itself 1.69+.

*Alternatives: have you considered pinning the docs schema link to a release tag, or version-bumping the schema file instead of mutating 1.0?*

## Nits

- `docs/advanced-features/sso-permissions.md:42`: the guide bullet says "omitted"; the schema says "omitted (or null)". The schema now explicitly accepts `admin: null` (and YAML `admin:` with no value parses to null), so the guide bullet should say "omitted (or `null`)" too.
- `docs/advanced-features/sso-permissions.md:48`: the veto is stated as absolute, but line 18 of the same page exempts the config uploader (verified that the exemption holds in union mode — same code path). Worth cross-referencing so a reader applying the veto rule to the config-setter doesn't get a wrong prediction.
- `shared/schemas/sso-config-1.0.json`: the root has no `additionalProperties: false`, so a typo'd key — e.g. `union_role: true` — validates silently and the config falls back to first-match-wins; the registry ignores unknown keys too, so the typo is silent end-to-end. Pre-existing gap, costlier now that an optional behavior toggle exists. Caveat: tightening would newly reject configs that carry stray keys today — author's call.
- PR body: "Docs/schema only; no behavior change" is literally false for anything validating against the schema — the body itself says the `default_role` move makes it "actually validated", and `admin: null` flips from invalid to valid. Nothing in this repo validates against the file (the catalog SSO editor is a plain text field; the Python client passes the raw string through), so "no runtime behavior change" or "no platform behavior change" would be accurate.
- `shared/schemas/sso-config-1.0.json:48`: "if none of mappings are matched" is missing the article — "if none of the mappings match". Pre-existing, but the move promotes it from dead root-level text to a real property description surfaced by editors.

Reviewed against bf5749a

drernie:
Thanks — addressed in 01c2e21a6:

- **P1**: First-match mode now states *the default is no admin* — omitting/`null` `admin` grants no admin, and since the flag is written on every login a matched user is demoted if they previously had admin. Avoided framing it as "≡ `admin: false`" since `false` is a veto in union mode, not the same thing. Schema `admin` description updated to match.
- **P2 (recompute rule)**: tri-state note now states the admin flag is recomputed/written on every login for any user matching ≥1 mapping, and that omitting `admin` on a catch-all only protects users who *also* match an `admin: true` mapping.
- **P2 (example)**: clarified `admin@example.com` only receives both roles if the token also carries group `rw`.
- **P2 (version gate)**: added "(Quilt Platform 1.69+)" to the tri-state note and both schema descriptions, plus a note that pre-1.69 platforms silently ignore `union_roles` and reject explicit `admin: null`.
- **Nits**: "omitted (or `null`)" in the guide bullet; cross-referenced the config-setter veto exemption; `additionalProperties: false` at the schema root; grammar fix at `default_role`; corrected the PR-body "no behavior change" claim.

sir-sigurd (approved):
Approving — all points from my Jun 10 review are addressed in `01c2e21a`:

- **P1 (first-match mode)** — resolved. Both the guide and the schema now state the default is *no admin*: omitting/`null` `admin` grants none, and because the flag is rewritten on every login a matched user who would otherwise be admin is demoted. That closes the gap — the omitted case was previously defined nowhere.
- **P2s** — the matched-user recompute rule, the worked-example `groups: rw` assumption, and the pre-1.69 version gate (silent-ignore + `admin: null` rejection) are all in, in both files.
- **Nits** — "omitted (or `null`)" bullet, config-uploader veto exemption cross-ref, `additionalProperties: false` on the schema root, "none of the mappings match," and the PR-body "no runtime behavior change" correction — all done.

One deliberate call for the record: `additionalProperties: false` will now reject configs carrying stray/typo'd keys that validate today — right call given the new toggle, just noting it's a (minor) validation tightening, not pure docs.

Process nit for next time: when you push fixes addressing a review, **re-request review explicitly** (the ↻ next to the reviewer) so it re-enters the reviewer's queue — otherwise the review sits as a stale comment and the PR silently stalls. This one waited ~18 days for exactly that reason.

LGTM. 🚢

PR #4967, opened 2026-06-11
thumbnail: Pillow 13 compat — explicit I;16 conversion before PNG save

sir-sigurd:
# Description

Pillow 13 removes saving mode `I` (32-bit) images as PNG — previously an implicit, silent clip to 16-bit, deprecated in 12.x via python-pillow/Pillow#9023 because PNG has no 32-bit depth and the clip could silently lose data. The thumbnail lambda relies on exactly that implicit clip: `norm_img` outputs int32 (mode `I`), so on Pillow 13 every normalized greyscale thumbnail — multi-channel montages, Z-projections, greyscale CZIs — would 500 with `OSError: cannot write mode I as PNG`. Promoting the deprecation warning to an error kills 14 currently-passing test cases (including `c1_gray8.czi` — even uint8 input dies, since `norm_img` unconditionally casts to int32).

Fix: convert `I` → `I;16` explicitly in `handle_image` before the PNG save. The convert clamps out-of-range values exactly as the old implicit save did, so the output is **pixel-identical** to the old behavior: no golden files change (verified by exact comparison against the existing references). (The clamp is load-bearing — bicubic resize can overshoot the uint16 range, and 2D int32 sources skip `norm_img` entirely.) A `filterwarnings` regression guard (`"error:Saving I mode images as PNG"`) in pytest config keeps the deprecated path from creeping back while we're still on Pillow 12; on Pillow 13 the hard error takes over.

Considered and deferred: normalizing to uint8 in `norm_img` instead, which would additionally halve thumbnail payloads (these paths currently ship 16-bit greyscale PNGs that browsers quantize to 8-bit anyway), kill mode `I` entirely, and make `generate_thumbnail`'s resampler-fallback chain provably dead. That's a behavior change requiring regeneration of 14 goldens plus an `images/thumbs` package revision — split out as separate work so this compatibility fix stays zero-churn.

# TODO

- [x] Unit tests — regression guard via `filterwarnings`; all 14 affected cases pass under it (and fail without the fix)
- [x] Changelog entry

<h3>Greptile Summary</h3>

This PR adds an explicit `I` → `I;16` conversion in `handle_image` before saving to PNG, fixing a breaking change in Pillow 13 where the prior implicit 32-bit-to-16-bit clip was removed. A `filterwarnings = error` guard in pytest prevents the deprecated path from silently re-entering the pipeline while still on Pillow 12.

- **`__init__.py`**: After `generate_thumbnail` returns a mode-`I` image, `img.convert("I;16")` is called before `img.save(...)`. The conversion is lossless because `norm_img` already clamps int32 values to the uint16 range (0–65 535). Thumbnail resize continues in mode `I` as before; only the save step changes.
- **`pyproject.toml`**: Adds `filterwarnings = ["error:Saving I mode images as PNG"]` under `[tool.pytest.ini_options]` to catch any future regression while Pillow 12 is still in use.
- **`CHANGELOG.md`**: Entry added for the fix.

Safe to merge — the conversion is lossless, pixel output is identical to the old implicit behaviour, and the regression guard in pytest is wired correctly.

The fix is narrow and well-scoped: norm_img already clamps int32 values to [0, 65535] before the thumbnail pipeline, so converting from mode I to I;16 is provably lossless. The only call site for handle_image always passes thumbnail_format='PNG', so there is no risk of applying a 16-bit greyscale conversion before a format that does not support it. The filterwarnings guard is syntactically correct and will catch any future accidental regression on Pillow 12.

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py | Adds a 5-line I→I;16 conversion guard before PNG save; logically correct and lossless given norm_img's clamping guarantee. |
| lambdas/thumbnail/pyproject.toml | Adds filterwarnings = error for the Pillow I-mode PNG deprecation warning; correct pytest.ini_options syntax and pattern. |
| lambdas/thumbnail/CHANGELOG.md | Changelog entry added in the correct section with an appropriate description and PR link. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[handle_image called] --> B[read_image / format_aicsimage_to_prepped]
    B --> C[generate_thumbnail\nresize in mode I]
    C --> D{img.mode == 'I'?}
    D -- Yes --> E["img.convert('I;16')\n(NEW — explicit lossless clip)"]
    D -- No --> F[img.save as PNG]
    E --> F
    F --> G[Return info + bytes]

    style E fill:#d4edda,stroke:#28a745
```

<sub>Reviews (2): Last reviewed commit: ["thumbnail: fill in PR number in CHANGELO..."](https://github.com/quiltdata/quilt/commit/0105804781a7ce2908317eb8dbba72c6e0d186a4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36935034)</sub>

### DIFF
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,7 +17,7 @@ where verb is one of
 
 ## Changes
 
-- [Fixed] Convert normalized greyscale thumbnails (montages, Z-projections) to 16-bit explicitly before PNG save, instead of relying on the implicit 32-bit clip that Pillow 13 removes ([#4967](https://github.com/quiltdata/quilt/pull/4967))
+- [Fixed] Convert normalized greyscale thumbnails (montages, Z-projections) to 16-bit explicitly before PNG save, instead of relying on the implicit clip to 16-bit that Pillow 13 removes ([#4967](https://github.com/quiltdata/quilt/pull/4967))
 - [Fixed] 500 on image formats readable by bioio-imageio but missing from its declared extensions, e.g. `.jpeg` and `.webp` ([#4963](https://github.com/quiltdata/quilt/pull/4963))
 - [Fixed] Support float and 16-bit color images instead of failing with HTTP 500 ([#4961](https://github.com/quiltdata/quilt/pull/4961))
 - [Fixed] Rescale 16-bit greyscale images by their actual value range to avoid nearly black thumbnails for low-range (e.g. 12-bit) data ([#4960](https://github.com/quiltdata/quilt/pull/4960))
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,7 +17,7 @@ where verb is one of
 
 ## Changes
 
-- [Fixed] Convert normalized greyscale thumbnails (montages, Z-projections) to 16-bit explicitly before PNG save, instead of relying on the implicit 32-bit clip that Pillow 13 removes ([#????](https://github.com/quiltdata/quilt/pull/????))
+- [Fixed] Convert normalized greyscale thumbnails (montages, Z-projections) to 16-bit explicitly before PNG save, instead of relying on the implicit 32-bit clip that Pillow 13 removes ([#4967](https://github.com/quiltdata/quilt/pull/4967))
 - [Fixed] 500 on image formats readable by bioio-imageio but missing from its declared extensions, e.g. `.jpeg` and `.webp` ([#4963](https://github.com/quiltdata/quilt/pull/4963))
 - [Fixed] Support float and 16-bit color images instead of failing with HTTP 500 ([#4961](https://github.com/quiltdata/quilt/pull/4961))
 - [Fixed] Rescale 16-bit greyscale images by their actual value range to avoid nearly black thumbnails for low-range (e.g. 12-bit) data ([#4960](https://github.com/quiltdata/quilt/pull/4960))
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Fixed] Convert normalized greyscale thumbnails (montages, Z-projections) to 16-bit explicitly before PNG save, instead of relying on the implicit 32-bit clip that Pillow 13 removes ([#????](https://github.com/quiltdata/quilt/pull/????))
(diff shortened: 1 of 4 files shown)


sir-sigurd:
@greptileai

sir-sigurd (commented):

Checked: output behavior (the explicit `I` → `I;16` convert produces byte-identical PNGs to Pillow 12's implicit save path, including identical clamping of out-of-range values), edge cases (constant/NaN/negative inputs, resize overshoot, montage padding, paths that skip `norm_img`), Pillow 13 semantics (the convert is the upstream-recommended replacement; `I;16` save is not deprecated), pytest guard wiring (CI's rootdir resolution picks up the lambda's `pyproject.toml`; the filter prefix-matches Pillow 12.2's warning text and fires on an unconverted save; none of the suite's other warnings match), and a repo-wide sweep (thumbnail is the only lambda importing PIL; the Dockerfile installs from the same lockfile; nothing in the catalog assumes thumbnail bit depth) — nothing flagged there. Remaining items, all minor:

- **Comment accuracy** (`src/t4_lambda_thumbnail/__init__.py:381-383`): "lossless, since norm_img clamps values to the uint16 range" doesn't hold at the convert site — bicubic resize overshoots the uint16 range (measured ~67k / −800 on clamped input), and 2D int32 sources skip `norm_img` entirely. The convert clamps those exactly as the old implicit save did (verified byte-identical output), so the code is right; reword the justification to "clamps identically to the clip the implicit save applied, so output is unchanged". Also, Pillow 13 removes `I`-as-PNG saving outright (hard error) rather than "the implicit 16-bit clip" — the pyproject comment states this correctly; align. The PR description's "This is lossless — `norm_img` already clamps values to the uint16 range" repeats the same invariant and should be updated together.
- **CHANGELOG wording** (`CHANGELOG.md:20`): "the implicit 32-bit clip" vs the code comment's "16-bit clip" — same mechanism, contradictory numbers; "clip to 16-bit" is the accurate phrasing. Harmonize with however the code comment is reworded.
- **Test depth guard** (`tests/test_thumbnail.py`, golden comparisons): nothing asserts the saved mode-`I`-path PNG is 16-bit — today the full-range `I;16` goldens pin it via exact pixel equality, but a golden regeneration (sanctioned when resamplers change, and planned in the deferred uint8 follow-up) would silently bake in a depth regression; neither the warnings filter nor Pillow 13's hard error catches a wrong-mode convert. A one-line `Image.open(out).mode == "I;16"` assertion survives regeneration and forces the uint8 follow-up to flip it consciously.
- **Guard lifetime** (`pyproject.toml:41-45`): the filter entry has no expiry — once the pin bumps to Pillow 13 the warning ceases to exist upstream and the entry goes permanently inert; we don't want it lingering forever, so the bump PR should delete it (worth saying so in its comment).

(The PR description's `filterwarnings = error` phrasing was already updated to quote the actual single-message filter.)

Reviewed against 0105804

PR #4971, opened 2026-06-11
thumbnail: drop the dead resampler-fallback chain; unify uint16 rescale dispatch

sir-sigurd:
# Description

`generate_thumbnail` had a resampler-fallback chain — a `try/except` that, when the default resampler raises `"image has wrong mode"`, retries with every resampler (LANCZOS→…→NEAREST). It was added in #2888 (June 2022) *alongside* `_convert_I16_to_L`, to fix a real bug: a 16-bit `I;16` PNG that failed preview with "image has wrong mode" (the original "PNG file preview fails" report; the customer image was deliberately not committed, hence the synthetic `I16-mode.tiff` stand-in).

This chain is **dead code**. `Image.resize` with the default resampler rejects nothing; it's `thumbnail()`'s `reduce()` pre-shrink step (≥2× downscale) that rejects the `I;16*` family with "image has wrong mode", and `_convert_I16_to_L` always converts those to `L` *before* the resize — so nothing reaches the fallback.

Verified against **the actual file from the #2888 report** (recovered from `quilt-t4-staging`, `aneesh/tiff/image001.png`, a 16-bit `I;16` PNG):
- **Failed pre-#2888** — the old bare `Image.fromarray(img).thumbnail(size)` raises `ValueError: image has wrong mode` on it (confirmed against `871d2393^`).
- **Works now** — produces a valid 8-bit `L` thumbnail.
- **Fixed by the conversion, not the fallback** — instrumented: `image001.png` → `I;16` → `_convert_I16_to_L` → `L` → default resampler succeeds first try; the fallback is **never entered**. Disabling only the conversion is what forces the fallback (and it then degrades to NEAREST).
- A **corpus-wide sweep** (every local fixture + every readable package fixture: gray8/16/32-float, multi-channel CZIs, OME-TIFFs, montages, RGB/RGBA, float) — **zero** enter the fallback.

So this removes the chain, the mock-only test that exercised it, the orphaned `I16-mode-128-fallback.png` golden, and the now-unused `_mock`/`contextmanager` helper. With the fallback gone it also **unifies the contrast-stretch dispatch**: 16-bit greyscale is now selected by numpy dtype (`arr.dtype.kind == 'u' and arr.dtype.itemsize == 2`) in the same upfront block as float/color, replacing the post-`Image.fromarray` `img.mode == 'I;16'` guard and the `_convert_I16_to_L` wrapper (removed). The array is rescaled to uint8 *before* PIL builds the image, so it never reaches PIL as an `I;16*` mode at all.

**Coverage retained:** `I16-mode.tiff` (mode `I;16`, same as the original failing image) still runs end-to-end via the `I16-mode-128.png` row — so if the uint16 dispatch/rescale regresses, the I;16 array reaches PIL unconverted, `thumbnail()` 500s with `"image has wrong mode"`, and that row's status/pixel assertions fail loudly (verified). Plus the 7 `test_rescale_uint16_to_uint8_*` unit tests (repointed from `_convert_I16_to_L` to the array→array helper) and a new test that a byte-swapped (`I;16B`) array is rescaled to `L`.

**Byte order:** dispatching by dtype handles the byte-order variants (`I;16L/B/N`) by construction — the rescale runs on the numpy array before PIL sees it, and `kind == 'u' and itemsize == 2` matches big-endian uint16 (which `== np.uint16` does not). Bonus: this also fixes a pre-existing blind spot where big-endian uint16 *color* was missed by the old `arr.dtype == np.uint16` check. Behavior is identical for every input that occurs today — local + package goldens match, CPU and peak memory unchanged.

**Trade-off:** removes a degrade-instead-of-500 net for a hypothetical unconverted resampler-incompatible mode. No such mode is produced by any input today (everything is funneled to L/I/RGB/RGBA, all handled by the default resampler).

(Supersedes this PR's original scope — decoupling the fallback test's filename-string mock — which is moot once the test is deleted.)

# TODO

- [x] Unit tests — full suite passes; I;16 conversion path retains coverage and is confirmed to fail loudly on regression
- [x] Changelog entry — `[Changed]` line ([#4971]); no user-visible behavior change, matching the precedent of logging internal-only work

<h3>Greptile Summary</h3>

This PR removes the dead resampler-fallback chain in `generate_thumbnail` and unifies the uint16 contrast-stretch dispatch to operate on the numpy array (by `dtype.kind`/`itemsize`) before PIL ever sees the data, replacing the post-`fromarray` `I;16` mode guard and the `_convert_I16_to_L` wrapper.

- **Dead code removed**: the `try/except ValueError("image has wrong mode")` fallback-resampler loop and the `_convert_I16_to_L` wrapper are deleted; the PR demonstrates through instrumentation that `_convert_I16_to_L` always runs before `thumbnail()`, so the fallback is never entered.
- **Dispatch unified**: `arr.dtype.kind == "u" and arr.dtype.itemsize == 2` replaces `arr.ndim == 3 and arr.dtype == np.uint16` (color) + `img.mode == "I;16"` (greyscale), correctly covering big-endian uint16 (`">u2"`, which `== np.uint16` misses) for both greyscale and color inputs.
- **Test coverage retained and expanded**: the `I16-mode.tiff` golden now runs end-to-end against the main path; 7 `_rescale_uint16_to_uint8` unit tests are re-pointed from the removed wrapper; a new `test_generate_thumbnail_handles_byte_swapped_uint16` test verifies the big-endian fix.

Safe to merge — removes unreachable code and tightens the uint16 dispatch; every current input is covered by the new path, the golden fixtures match, and the added byte-swapped test verifies the bonus big-endian fix.

The fallback chain removal is well-supported by the author's instrumentation and the preserved end-to-end golden for I16-mode.tiff. The dtype-based dispatch correctly covers native and big-endian uint16 for both greyscale and color arrays, and PIL always receives a uint8 array so thumbnail() cannot raise image has wrong mode. No existing behavior changes; the only new behavior is the big-endian uint16 fix, which now has a dedicated unit test.

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py | Removes _convert_I16_to_L wrapper and the try/except fallback-resampler chain; extends the uint16 dispatch condition to cover all uint16 dtypes (including big-endian) by kind/itemsize, so PIL always receives a uint8 array. Logic is correct and well-commented. |
| lambdas/thumbnail/tests/test_thumbnail.py | Removes the mock-based fallback test and _mock helper; re-points 7 unit tests from _convert_I16_to_L to _rescale_uint16_to_uint8; adds test_generate_thumbnail_handles_byte_swapped_uint16 for the big-endian fix; I16-mode.tiff now runs end-to-end on the main path. |
| lambdas/thumbnail/tests/data/I16-mode-128-fallback.png | Deleted golden image for the removed fallback-resampler test; no longer needed since the fallback path is gone and I16-mode-128.png covers the end-to-end I;16 thumbnail. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["generate_thumbnail(arr, size)"] --> B{"arr.dtype.kind == 'f'?"}
    B -- yes --> C["rescale = _rescale_float_to_uint8"]
    B -- no --> D{"kind=='u' and itemsize==2?"}
    D -- yes --> E["rescale = _rescale_uint16_to_uint8"]
    D -- no --> F["rescale = None"]
    C --> G{"rescale is not None?"}
    E --> G
    F --> G
    G -- no --> H["arr unchanged"]
    G -- yes --> I{"4-channel RGBA?"}
    I -- yes --> J["dstack rescale RGB + alpha_to_uint8"]
    I -- no --> K["arr = rescale(arr)"]
    J --> L["Image.fromarray(arr)"]
    K --> L
    H --> L
    L --> M["img.thumbnail(size)"]
    M --> N["return img"]
```

<sub>Reviews (5): Last reviewed commit: ["thumbnail: dispatch the uint16 rescale b..."](https://github.com/quiltdata/quilt/commit/0306187f3564288af890f796b492270994ae4969) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36972981)</sub>


sir-sigurd:
@greptileai

sir-sigurd:
@greptileai — note: this PR's scope changed substantially since the last pass. It now *removes* the dead resampler-fallback chain from generate_thumbnail (rather than reorganizing its test). Please re-review the latest commit (`bff91354`).

sir-sigurd:
@greptileai — re-review please; latest is `8ae3c0ab` (fallback chain removed, plus an explanatory comment on why the I;16→L conversion is required, with a Pillow docs link).

sir-sigurd (commented):

The dead-code removal is well-supported — the fallback genuinely can't fire on any input the readers produce today, confirmed by running Pillow 12.2 directly (every mode `Image.fromarray` yields here — L, I, I;16, F, RGB, RGBA — survives the default resampler once `_convert_I16_to_L` has run). The retained `I16-mode-128.png` row exercises the conversion end-to-end against an exact-pixel golden, so a conversion regression now surfaces as a loud failure. Checked: behavior on removal, untrusted-input error handling, test-coverage retention, PIL mode reachability — notes below.

**P2 — the removal narrows the safety margin to an unenforced invariant**

With the fallback gone, the I;16→L conversion is guarded only by the exact-string `img.mode == 'I;16'`. A big-endian uint16 array produces mode `I;16B`, which skips that guard and then raises `"image has wrong mode"` at `img.thumbnail()` on any real downscale — the exact failure #2888's fallback was added to catch. Verified in Pillow 12.2: `Image.fromarray(<big-endian uint16>)` → `I;16B`; `I;16B.thumbnail((128,128))` raises; NEAREST would have succeeded.

Not reachable today — tifffile/ome-tiff/czi/imageio all normalize byte order to native on read, so only `I;16` reaches the guard. But that's now the only thing between a byte-swapped uint16 input and a 500, and it's an invariant that lives in the readers, isn't enforced here, and isn't audited across every bioio backend. Two cheap ways to make the removal self-contained: broaden the guard to `img.mode.startswith('I;16')`, or add a one-line comment recording the "readers normalize byte order; only I;16 reaches here" assumption next to the guard.

(Aside: the description's "the only mode the default resampler rejects is I;16" is slightly off — `Image.resize` with the default resampler rejects nothing; it's `thumbnail()`'s reduce step that rejects the `I;16*` family, and only at ≥2× downscale. The conclusion holds; worth tightening the wording, since that imprecise model is what justifies the narrow guard.)

**P3 — minor losses**

- The removed `raise ValueError(f"...mode {img.mode}")` named the offending mode; a future failure now surfaces only PIL's generic "image has wrong mode" (CloudWatch retains the traceback, so low stakes).
- The description says the `I16-mode-128.png` row "raises 'image has wrong mode' if `_convert_I16_to_L` regresses" — true and verified, but it's the row's exact-pixel assertion that fails, not an assertion on that string. Minor over-claim.

**Nits**

- CHANGELOG: skip is recorded as "no user-visible behavior change," but this CHANGELOG does log no-behavior-change internal work (e.g. #4968 "output unchanged"). A `[Removed]` line would be consistent with that precedent — your call.
- `lambdas/thumbnail/tests/test_thumbnail.py:98` — the `# BUG: lambda doesn't preserve source format.` header now stacks directly above the new `# I;16 input exercises…` comment and the surviving row, reading as a double-comment. Not wrong (the BUG note is a general header), just slightly awkward.

Reviewed against bff9135

sir-sigurd:
@greptileai — re-review please; reworked since the last pass. Final state (`0306187f`): removes the dead resampler-fallback chain **and** unifies the uint16 rescale dispatch by numpy dtype (replacing the post-`fromarray` I;16 mode guard + the `_convert_I16_to_L` wrapper). Behavior-identical (local + package goldens, CPU/memory unchanged).

sir-sigurd (commented):
Notes from an AI-assisted self-review pass (follow-up round; likely the last).

The prior pass's main concern — the narrow `img.mode == 'I;16'` guard missing byte-order variants — is resolved: selecting the rescale by numpy dtype (`kind == 'u' and itemsize == 2`) handles `I;16L/B/N` by construction. Re-confirmed the dead-code removal (the resampler fallback is genuinely unreachable — in Pillow 12.2, `I;16`/`I;16B` are the only `Image.fromarray`-producible modes the default resampler's `reduce()` step rejects, and uint16 is now rescaled to L/RGB/RGBA before the image is built), behavior on every native-order input class (pixel- and mode-identical), untrusted-input error handling, and the byte-order numpy semantics. Nothing flagged there.

Three minor notes:

**P3 (reshape) — `generate_thumbnail` comment dropped the color rationale.** The reworded comment motivates the float16-construct, mode-F-PNG, and I;16-`reduce()` cases — all greyscale. But the float and uint16 branches also rescale *color* arrays, and that's required because PIL won't build a color image from non-uint8 (it raises `TypeError`). A maintainer reading only the new comment could conclude color uint16/float can skip the rescale. Re-add a clause noting PIL only builds color images from uint8.

**P3 — big-endian uint16 *color* has no test.** `test_generate_thumbnail_handles_byte_swapped_uint16` pins byte-swapped *greyscale* → L. The description calls big-endian *color* a separate bonus fix (the old `== np.uint16` check missed `>u2` color → `TypeError` at `Image.fromarray`; matching on dtype now routes it through the rescaler). A color-specific regression would pass every test today. A `(H, W, 3)` or `(H, W, 4)` `>u2` case would close the gap.

**Nit — CHANGELOG.** Skipped as "no user-visible behavior change," but this changelog logs no-user-visible internal work (#4968 "output unchanged", "Switch to uv", "Regenerate requirements.txt"). A `[Changed]` line for removing the dead fallback chain would match that precedent — `[Changed]` rather than `[Fixed]`, since the big-endian path isn't reachable today.

Reviewed against 0306187

PR #4974, opened 2026-06-12
thumbnail: fix norm_img on constant and NaN greyscale planes

sir-sigurd:
# Description

`norm_img` (greyscale normalization for the n-dim montage / Z-projection path) clipped percentiles with `da.percentile` on a 2-D dask array, then rescaled via `(x - min) / max`. Two degenerate inputs misbehaved:

- **Constant plane** → `0/0` → `NaN`, and the `int32(NaN)` cast is platform/numpy-version-dependent (`0` on arm64, `INT_MIN` on the x86-64 Lambda runtime).
- **Any NaN** → `da.percentile` returns `NaN` bounds → the whole tile blanks.

Both are reachable through the public API: float scientific images with masked/invalid (`NaN`) pixels, and blank/uniform channels in multi-channel microscopy. These are long-standing bugs in `norm_img`, not regressions from the recent thumbnail work; there's no separate GitHub issue (tracked internally).

This rewrites `norm_img` to normalize on the *computed* plane in NumPy: a finite-aware percentile clip + non-finite filtering (extracted into `_finite_clip_range`, shared with `_rescale_float_to_uint8` so their filtering/clip/fallback can't diverge), `constant → black`, `NaN → black` per pixel, `±inf → saturate` to the range ends.

Output is **bit-identical** to the previous implementation for finite input the old code rendered well-definedly — verified against the golden thumbnails, no existing golden changed (one new content-pin golden is added, see Tests) — and stays 16-bit (`I;16`). (The one finite-input change: a sparse / near-constant plane, where the percentiles collapse but min ≠ max, now renders visibly via the min/max fallback instead of the old `0/0`→black — `test_norm_img_sparse_stays_visible`.) It also drops the `da.percentile`-on-2-D call, which raises `NotImplementedError` on a multi-chunk plane under the deployed dask `2026.1.1`. That last part is **defensive hardening**: bioio's readers currently emit single-chunk planes (verified even for a mosaic CZI and a 6184×7712 pyramid), but that isn't a documented contract.

## Performance

CPU is neutral-to-better — one two-quantile `np.percentile` plus in-place arithmetic replaces the per-channel `da.percentile`/`da.min`/`da.max` reductions. Peak memory on multi-channel montages is **lower**, not higher: ~602 MiB vs ~1036 MiB on a 10-channel montage (`handle_image` at 1024×768), because each plane is computed eagerly and only its int32 result is retained, rather than holding the full lazy `da.percentile` graph until the final `.compute()`. (`da.concatenate` is a barrier needing all N planes resident regardless, so the old lazy path couldn't stream below that anyway.)

## Tests

- Unit tests pin the constant / `NaN` / `±inf` / sparse / color-passthrough behavior, plus a multi-chunk guard — only a synthetic dask array can reach that path; no real file produces a multi-chunk plane through the lambda's readers.
- An end-to-end test runs a committed OME-TIFF fixture (blank + `NaN` channels) through `handle_image` and asserts no non-finite warning — verified to fail on the old code. The same fixture is also added as a content-pinned golden row in `test_generate_thumbnail` (new golden `blank-and-nan-channels-256.png`), so a decode-time NaN-mangling regression is caught.
- Unit tests pin the `_finite_clip_range` collapsed-percentiles-with-NaN contract at both call sites (`norm_img` and `_rescale_float_to_uint8`): the fallback ranges over the finite values, not the NaN-poisoned `arr.min`/`max`.

## Follow-up (separate PR)

Add a mosaic CZI fixture to the `images/pylibczirw` test package for `reconstruct_mosaic` coverage — no mosaic-origin image is tested today. It will be `--large-files`-gated (the smallest true mosaic is ~30 MB) and needs a test-data package revision.

# TODO

- [x] Unit tests

<h3>Greptile Summary</h3>

This PR rewrites `norm_img` to fix two degenerate cases in the n-dim montage / Z-projection greyscale normalisation path: constant planes previously produced `0/0 → NaN` and a platform-dependent `int32(NaN)` cast, while any NaN pixel caused `da.percentile` to return NaN bounds and blank the entire tile. It also extracts the percentile-clip logic into a shared `_finite_clip_range` helper so `norm_img` and `_rescale_float_to_uint8` can no longer diverge in their clip/fallback semantics.

- **`norm_img` rewrite**: computes eagerly in NumPy on the materialised plane, uses `_finite_clip_range` (finite-aware percentile with min/max fallback), renders constant planes and all-NaN planes as black, clips NaN pixels to black per-pixel, saturates ±inf to the range ends, and returns int32 — described as bit-identical to the old path for purely finite input.
- **`_finite_clip_range` helper**: factors out the `np.percentile(finite, (0.01, 99.99))` + min/max fallback that was duplicated in `_rescale_float_to_uint8`, keeping both code paths' clipping behaviour in lock-step.
- **Tests**: new unit tests cover constant, all-NaN, per-pixel NaN, ±inf, multi-chunk, sparse, and colour-passthrough paths, plus an end-to-end test through `handle_image` with a constructed OME-TIFF.

Safe to merge — the rewrite fixes two real rendering bugs with no change to correct-input output, is well-tested, and refactors only the thumbnail normalisation path.

The logic is correct across all handled cases (constant, all-NaN, per-pixel NaN, ±inf, multi-chunk). The NaN-zeroing step comes after the arr == imax guard so the two don't interfere, _clip_range is always called with a non-empty finite array, and the eager materialisation via np.asarray is intentional and safe for the single-chunk planes this path receives. No regressions for finite input, and the _rescale_float_to_uint8 refactor is a pure deduplication with identical semantics.

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py | Rewrites norm_img to fix 0/0→NaN and da.percentile-on-NaN bugs; extracts _clip_range shared helper; logic is correct and well-commented |
| lambdas/thumbnail/tests/test_thumbnail.py | Adds comprehensive unit and end-to-end tests for all degenerate norm_img inputs; warnings module already imported at top level |
| lambdas/thumbnail/CHANGELOG.md | Adds accurate changelog entry for the fix |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["norm_img(img: da.Array)"] --> B{3-D shape?}
    B -- Yes --> C["Return img unchanged\n(colour plane)"]
    B -- No --> D["np.asarray(img).astype(float64)"]
    D --> E["finite = arr where isfinite"]
    E --> F{finite.size == 0?}
    F -- Yes --> G["Return zeros (black)"]
    F -- No --> H["_clip_range(finite)\n→ lo, hi"]
    H --> I{hi == lo?}
    I -- Yes --> J["Return zeros (black)\nconstant plane"]
    I -- No --> K["np.clip(arr, lo, hi)"]
    K --> L["arr -= lo\narr /= hi-lo\narr *= 65536"]
    L --> M["arr == 65536 → 65535\n±inf saturated by clip"]
    M --> N["arr[isnan] = 0\nNaN → black"]
    N --> O["da.from_array(arr.astype(int32))"]
```

<sub>Reviews (2): Last reviewed commit: ["thumbnail: add CHANGELOG entry for the n..."](https://github.com/quiltdata/quilt/commit/6933fce6b486d2cfbc1f076460b112ddf81d8047) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37176062)</sub>


sir-sigurd:
@greptileai please review

sir-sigurd (commented):

This pass led with the requested question — does the rewrite affect CPU or memory? — and ran the usual correctness/consistency sweep alongside. Net: the fix is correct and the new branches are well covered. CPU is neutral-to-slightly-better; memory is fine on the single-plane (Z-projection) path, with one montage-path point worth a sanity check (below). Everything here is a nit or a design question — nothing blocking. Checked: behavior preservation / API contract, scope, and an adversarial pass against the approach.

### Performance (CPU / memory)

**CPU — neutral to better.** The single-plane path replaces two separate `da.percentile` calls plus `da.min`/`da.max` (four dask reductions + scheduler overhead) with one two-quantile `np.percentile` and in-place arithmetic. For the common finite plane the finite mask is computed once and the work is strictly less than before.

**Memory, single greyscale plane (Z-projection) — comparable to before.** Peak is the float64 working copy (8 bytes/px) plus the int32 result (4 bytes/px) held briefly together. The float64 (vs the uint8 path's float32-when-possible) is the deliberate cost of the bit-identical-output guarantee, so that's expected.

**Memory, multi-channel montage — one thing worth confirming.** The old `norm_img` was lazy: each channel's normalized plane stayed a deferred dask graph and the whole montage computed once at the final `.compute()`, which gave dask the option to evaluate and free planes one at a time while assembling the output. The new `norm_img` materializes each plane eagerly (`np.asarray`) and returns a concrete array, so all N normalized planes are held in the `projections` list at once before `da.concatenate`. For the typical 2–6 small channels this is a non-issue. But the montage is assembled at full resolution before the resize (`handle_image` computes it, then `generate_thumbnail` shrinks), so for many large channels the resident set is now ~N full-res planes rather than potentially ~1 — and OOM is the documented failure mode of this lambda. I couldn't confirm from inspection alone whether the old lazy path actually streamed (it's dask-scheduler-dependent), so this isn't necessarily a regression — but it's the one place the memory profile genuinely changed, and pushing the largest realistic multi-channel montage through it would settle it.

### Nits

**P3 (reshape)** — `norm_img` computes `np.isfinite(arr)` twice on the non-finite path (`__init__.py:190`): once for `.all()`, once for the boolean index. The sibling `_rescale_float_to_uint8` already computes the mask once, reuses it, and frees it — this is the one piece of the finite handling the shared-helper refactor didn't unify. It only costs on the non-finite branch (the common finite plane short-circuits to a single call), so the impact is small, but it's a free consistency win. (Copilot flagged the same.)

**P3** — `tifffile` is imported in `test_handle_image_blank_and_nan_channels_through_public_path` (`tests/test_thumbnail.py:577`) but isn't a declared test dependency; it currently arrives transitively via `bioio-tifffile`. If that edge ever changes the test fails at import rather than skipping. `pytest.importorskip("tifffile")` or adding it to the test group fixes it. (Greptile and Copilot both flagged this. Note `dask` is imported the same undeclared-transitive way at the top of this file, so it isn't a new pattern — tidy both or neither.)

### Description accuracy

**P3** — the description says output is "bit-identical to the previous implementation for finite input." That holds for finite planes the old code rendered well-definedly, but not for an all-finite *sparse / near-constant* plane where the percentiles collapse while min ≠ max: the old code hit the same `0/0 → NaN → int32` path and rendered garbage/black, while the new min/max fallback now renders the sparse pixels visible (the `test_norm_img_sparse_stays_visible` case). That's a beneficial change, but it is a finite input whose output changed — worth tightening the claim to "…for finite input that previously rendered well-definedly," or naming the sparse case alongside constant/NaN.

### Design question (your call)

**P3 (reshape)** — a constant plane renders **black** here, but the same constant plane renders at its **grey level** through `_rescale_float_to_uint8`. The shared-helper docstring notes this as a deliberate difference, and black is defensible for a montage cell that carries no contrast — but it's worth a beat: a legitimately uniform bright-field cell goes black in the montage while staying visible via the other path.
*Alternatives: would keeping the level (scaling `lo` into the 16-bit range) be better here, so a uniform plane stays visible and the two paths agree?*

Reviewed against 6933fce

sir-sigurd:
Thanks — addressed in `80848e30`.

**Nits (fixed):**
- **`np.isfinite` twice** → `norm_img` now computes the mask once and frees it, mirroring `_rescale_float_to_uint8`.
- **`tifffile` undeclared dep** → rather than declare it or `importorskip`, I **prebuilt the fixture** (`tests/data/blank-and-nan-channels.ome.tiff`, ~50 KB) and the end-to-end test now just reads it, like the other `tests/data` fixtures. No `tifffile` import remains. (`dask` stays imported at module scope — but as you noted it's guaranteed: importing the module under test imports `dask`, so it can't fail independently. The asymmetry is why I didn't guard it.)
- **Description "bit-identical" claim** → tightened to "finite input the old code rendered well-definedly," and called out the sparse/near-constant case explicitly (it now renders visibly via the min/max fallback instead of `0/0`→black).

**Memory (montage) — measured, and it's not a regression — it's lower.** Peak RSS on the 10-channel montage (`s_1_t_1_c_10_z_1.ome.tiff`, `handle_image` at 1024×768):

| | peak RSS |
|---|---|
| OLD (`da.percentile` lazy) | 1036 MiB |
| NEW (NumPy eager) | 602 MiB |

The old lazy path held the full `da.percentile` graph (4 reductions × 10 channels) live at the single `.compute()`; computing each plane eagerly and keeping only the int32 result peaks lower. And as you suspected the streaming couldn't help anyway — `da.concatenate` is a barrier that needs all N planes resident regardless, so neither path drops below ~N.

**Constant→black (design question):** I'd keep **black**, on these grounds — a *truly* constant plane is a blank/empty channel (black is the right read), while a *near*-constant one (a real bright-field cell has some variation) hits the new min/max fallback and stays visible. "Keep level" is also genuinely ambiguous here: `norm_img` is dtype-polymorphic (uint8/uint16/int/float planes), unlike the float-only `_rescale_float_to_uint8` that can assume the `[0, 1]` convention — so there's no single correct level to map a constant to across dtypes. Happy to switch to keep-level if you'd rather the two paths agree on the degenerate case; it'd be a small change.

sir-sigurd (commented):
Follow-up notes from an AI-assisted self-review pass:

Likely the last substantive pass — everything new is minor. The three actionable items from the previous notes are handled: the finite mask is computed once (now inside the shared helper), the end-to-end test reads a committed fixture instead of importing `tifffile`, and the bit-identical comment is qualified for the sparse case. The new `_finite_clip_range` consolidation is behavior-preserving — verified against the full suite (all passing) and a numerical sweep (no out-of-range int cast; NaN never reaches the int32 cast). Checked: scope coherence and an adversarial pass over the consolidated design.

The two design questions from the previous notes — the montage path's eager per-channel materialization (memory) and constant-plane→black vs the float path's keep-level — are unchanged by this refactor; still open for your call, not re-listed here.

New, all minor:

**P3** — `_finite_clip_range` documents the `None` return (no finite values) but not that a non-`None` `(lo, hi)` can still have `hi == lo`: a genuinely constant finite plane survives the percentile-collapse min/max fallback, and both callers must (and do) re-check `hi == lo` afterward. A maintainer who trusts a non-`None` return as a usable range and drops that check would silently reintroduce the `0/0 → NaN` case this PR fixes. One sentence in the docstring closes the gap.
*Alternatives: would a constant-case sentinel (or a 3-tuple flagging it) make the caller's obligation impossible to skip?*

**P3 (reshape)** — `_rescale_float_to_uint8` still opens with `if not arr.size: return arr.astype(np.uint8)`, but that branch is now dead: an empty array reaches `_finite_clip_range`, which returns `None` (`np.isfinite` over an empty array is vacuously all-True → empty finite → `None`), and the `None` branch already returns an equivalent empty `uint8` array. `norm_img` has no such guard. Dropping it removes the dead branch and the asymmetry between the two callers.

**P3** — the PR description still calls the helper `_clip_range`; it was renamed to `_finite_clip_range`, and it now also owns the non-finite *filtering*, not just the clip/fallback. Worth updating the description so a reader grepping the name finds it and sees the actual scope.

**P3 (reshape)** — the return annotation `-> "Tuple[float, float] | None"` is quoted, but it needn't be: the target is Python 3.13 and there's no `from __future__ import annotations`, so the unquoted form evaluates fine. Every other annotation in the file is unquoted; `Optional[Tuple[float, float]]` would also match the file's existing `from typing import ...` style.

**P3** — `_rescale_uint16_to_uint8` still carries its own inline percentile-collapse → min/max fallback, so that fallback now lives in two shapes once `_finite_clip_range` owns it for the float paths. The uint16 exclusion is deliberate (it uses the histogram percentile for memory and has no non-finite values), but the "shared so they can't diverge" framing now covers two of the three rescale paths — worth a word so the next editor knows the uint16 path is intentionally separate.

**P3** — the fixture regen recipe in the test comment reproduces the committed file's pixels exactly but not its bytes (tifffile writes a random OME UUID per file). A one-line note ("pixels reproduce exactly; bytes differ — tifffile injects a random OME-UUID") saves a future maintainer from thinking a regenerated fixture drifted.

Reviewed against ce583d8

sir-sigurd:
Thanks — all six addressed in `192f39e8`:

- **`_finite_clip_range` `hi == lo` caveat** → docstring now states a non-`None` `(lo, hi)` can still be constant (percentiles *and* min/max collapse), so callers must re-check; both do.
- **Dead `if not arr.size` guard in `_rescale_float_to_uint8`** → dropped. An empty array now flows through `_finite_clip_range` → `None` → the existing black branch, yielding an equivalent empty `uint8` array (verified for `(0,)`, `(0,3)`, `(5,0,3)`); the two callers are symmetric now.
- **Return annotation** → unquoted to `Optional[Tuple[float, float]]` (matches the file's `from typing import` style).
- **uint16 path** → added a note at its fallback that it intentionally keeps its own clip/fallback (histogram percentile for bounded memory, no non-finite values), so the "can't diverge" framing is scoped to the two float paths.
- **PR description** → `_clip_range` → `_finite_clip_range`, and noted it now owns the non-finite filtering too.
- **Fixture comment** → noted the regen recipe reproduces pixels exactly but not bytes (tifffile injects a random OME-UUID).

Behavior-preserving — full suite + all 30 goldens green; `test-lambda (thumbnail)` passing on CI.

The two design questions are unchanged: montage memory is settled (eager peaks **lower** — 602 vs 1036 MiB on the 10-channel montage), and constant→black is still your call (holding it, with the dtype-polymorphism rationale).

sir-sigurd (commented):
Follow-up note from an AI-assisted self-review pass — likely the last.

Recording one pre-existing numerical edge for the record (not introduced by this PR — the old `(x - min) / max` math overflowed the same way):

**P3** — in `norm_img`, `arr /= hi - lo` can divide by `inf`. If `_finite_clip_range` returns extreme native-float64 bounds — pixels near `+1.79e308` and `−1.79e308`, with enough mass in both tails to survive the 0.01/99.99 percentile clip — then `hi - lo` overflows to `inf`, every finite pixel maps to `0`, and the former max goes `inf/inf → NaN → 0`. A genuinely contrast-bearing plane then renders all-black, and the `hi == lo` guard doesn't catch it (the bounds differ).

Only native float64 input can reach it: float32 data has a finite span in float64, so `hi - lo` stays finite. Real scientific planes don't carry ±1.8e308 pixels, so this is theoretical — noting it as a known edge rather than something to fix here. A finite-span check on `hi - lo` would close it if we ever want to.

Reviewed against b44320d

sir-sigurd (commented):

A from-scratch pass over the PR's final state, at the reviewer's request — the earlier notes tracked the work incrementally; nothing below repeats them. More rounds coming — the finer polish pass is held until the P2 below is resolved (its fix moves the code that pass would comment on), and the addressed items will get a re-look.

For the record, re-verified independently of the earlier notes: the claimed bugs reproduce on the old code (constant plane → 0/0 warnings; one NaN pixel blanks the whole tile; multi-chunk 2-D raises `NotImplementedError` on the locked dask 2026.1.1). The bit-identity claim held under two separate empirical checks — old-vs-new across 12 array families plus a 300-array fuzz of the percentile/min-max identities, and a 2000-span output-range fuzz (nothing escapes `[0, 65535]`, truncation and the `== imax` clamp included). The committed fixture's pixels match the regen recipe exactly; all three bot comments are resolved at HEAD; every line of `_finite_clip_range`/`norm_img` is hit by the new tests. Checked with nothing to flag: behavior preservation and call-site compatibility, scope coherence, security surface, the lazy/eager dask boundary, and refactor losses against the old code.

### P2

**(reshape)** `src/t4_lambda_thumbnail/__init__.py:173` — `np.percentile(finite, ...)` flatten-copies its input. `overwrite_input=(finite is not arr)` lets it partition the compacted copy in place: `finite` is a discarded private array whose element order stops mattering after ranging (only `min()`/`max()` follow), and the guard keeps the all-finite path — where `finite is arr` and element order must survive — untouched. Measured: 1061 → 844 MB process peak ranging a 400 MB NaN-bearing plane. The NaN path is this PR's target, OOM is this lambda's documented failure mode, and both callers get the reduction through the shared helper.

### P3

- `__init__.py:175` — the one interaction the helper exists to unify — collapsed percentiles *with* non-finite pixels present — is untested in both consumers: changing `finite.min()/max()` to `arr.min()/max()` passes the entire suite, golden-image tests included (reproduced twice, independently). A regression there bypasses the callers' `hi == lo` guard silently (`NaN != NaN`). One sparse-plane-with-NaN case pins it, at helper level or per consumer.
- `tests/test_thumbnail.py:592` — the end-to-end test asserts warning-free + `I;16` only, and it is the sole test that pushes NaN/constant data *through decode*: a bioio/tifffile NaN-mangling regression would pass everything. Cheap content pin matching the existing pattern: add `blank-and-nan-channels.ome.tiff` as a row in the main parametrized golden test (one line plus one committed golden).
- **(reshape)** `__init__.py:201` — `.astype(np.float64)` copies unconditionally; `copy=False` halves the materialization-phase footprint for float64 sources. Safe, not just cheaper: dask's `finalize` deep-copies compute results so the returned buffer is owned — verified on the locked 2026.1.1, including deliberately-aliasing constructions (`np.shares_memory` false in every case).
- **(reshape)** `__init__.py:207,213,227` — `da.from_array` with the default name hashes the full plane to build its key (SHA1 fallback; no xxhash in the venv): ~8 ms at 2k×2k up to ~200 ms at 10k×10k, per channel, buying nothing — each plane is unique and computed once. `name=False` skips the hash; verified the montage assembly computes identically with it.
- `__init__.py:195` — follow-up-scoped note: the montage path now ranges over an unbounded float64 plane via exact `np.percentile`, while #4968 bounded the uint16 single-plane path to ~2N with the histogram — for the same uint16 planes, the montage path is now the more OOM-prone of the two. A straight swap here would break the bit-identical guarantee (the histogram percentile is sub-ULP inexact, and the ×65536 quantization can surface that), so it needs its own pass — worth folding into the declared mosaic follow-up.
- `__init__.py:207` — observational: downstream of `norm_img`, `da.pad`/`da.concatenate` now shuffle already-materialized planes; laziness still earns its keep upstream (plane slicing, Z-`max`) and on the non-montage branches, but the montage assembly could move to plain NumPy in a follow-up.
- `__init__.py:161` — the helper docstring's "Shared by norm_img and _rescale_float_to_uint8 so their clip/fallback … can't diverge" reads as the complete list of synced sites; `_rescale_uint16_to_uint8` deliberately duplicates the constants and fallback (its side documents the opt-out, the helper side doesn't point back). One back-pointer sentence keeps an in-place constants edit from stranding the uint16 copy.
- **(reshape)** `__init__.py:219` — "the old 0/0 left black" is wrong on the production platform (x86-64 rendered INT32_MIN garbage, not black) and sits eight lines below the correct "platform- and numpy-version-dependent" statement. One word: "left black" → "left undefined".
- **(reshape)** `tests/test_thumbnail.py:611` — the comment narrows what the filter enforces: `simplefilter("error")` escalates every warning category (the file's idiom, and defensibly intended), not just `RuntimeWarning`. Reword to "any warning (e.g. the 0/0 RuntimeWarning) fails".

Reviewed against b44320d

sir-sigurd:
Thanks for the thorough from-scratch pass — addressed in `7f3ae8a4`.

**P2 — `overwrite_input`:** `_finite_clip_range` now passes `overwrite_input=(finite is not arr)`, so the compacted non-finite copy is partitioned in place while the all-finite path (where element order must survive for the caller's rescale) is untouched. Confirmed your measurement direction — it's the NaN path, this PR's OOM-prone target, that benefits, and both callers get it via the helper.

**P3s fixed:**
- `astype(np.float64, copy=False)` — float64 sources skip the copy; safe since the dask compute result is owned (in-place math below stays private).
- `da.from_array(..., name=False)` at all three sites — skips per-plane content hashing that buys nothing (unique, computed once).
- `_finite_clip_range` docstring now points back to the uint16 path's deliberate opt-out.
- "the old 0/0 left **black**" → "left **undefined**" (x86-64 rendered INT32_MIN, not 0).
- test comment: `simplefilter("error")` escalates **any** warning, not only `RuntimeWarning`.
- **Coverage gap (the helper's whole reason for existing):** added `test_norm_img_sparse_plane_with_nan_ranges_over_finite_values` — collapsed percentiles *with* NaN present. Verified it discriminates: mutating `finite.min()/max()` → `arr.min()/max()` blanks the plane (`out.max()` 65535 → 0), failing the assert.
- **Content pin:** added `blank-and-nan-channels.ome.tiff` as a row in `test_generate_thumbnail` (committed golden), so a decode-time NaN-mangling regression is caught — the warning-only end-to-end test wouldn't.

**Deferred (noted, not in this PR):**
- The `hi - lo == inf` overflow on pathological native-float64 bounds (±1.8e308 with mass in both tails) — theoretical and pre-existing (the old `(x-min)/max` overflowed the same way); a finite-span check would close it if we ever want to.
- The montage path ranges an unbounded float64 plane via exact `np.percentile`, so for uint16 planes it's now more OOM-prone than the #4968 histogram single-plane path. A straight swap breaks the bit-identical guarantee (the histogram percentile is sub-ULP inexact), so it needs its own pass — fits the norm_img-unification follow-up.
- Moving the now-materialized montage assembly (`da.pad`/`da.concatenate`) to plain NumPy — same follow-up.

Full suite + 30 goldens green throughout; CI passing on the latest.

sir-sigurd (commented):
Notes from a fresh AI-assisted self-review pass — a full re-review at `7f3ae8a4`, likely the last.

The three optimizations added since the previous pass were re-examined independently and hold up: `astype(copy=False)` avoids copying the peak float64 plane (a real memory win on this path), `name=False` skips per-plane graph-key hashing, and the `overwrite_input=finite is not arr` guard partitions only the throwaway compacted copy while leaving the all-finite path's element order intact. The fix's premise was reconfirmed at the source level — on the x86-64 Lambda runtime `int32(NaN)` yields INT_MIN (Intel `cvttsd2si`; numpy#21166), so constant planes genuinely rendered garbage in production, not black. Checked clean this round: behavior preservation and bit-identity for well-defined finite input, the lazy/eager dask boundary, edge/failure modes, cross-path consistency, iteration residue, and the docstrings/CHANGELOG.

Three minor items, all additive:

- **`_rescale_float_to_uint8` lacks a collapse-percentile-with-NaN test of its own.** The interaction `_finite_clip_range` exists to get right (range over *finite* values, not `arr.min/max` which would be NaN) is pinned at the helper through the new `norm_img` test, but the float consumer has no binding at its own call site — an inline or special-case edit there would regress silently. One small test mirroring `test_norm_img_sparse_plane_with_nan_ranges_over_finite_values` closes it.
- **Pre-existing: `dask` is used directly but not declared.** The lambda calls dask's public API (`da.from_array`, `da.concatenate`, `da.pad`) yet `dask` isn't in `pyproject.toml` — it arrives transitively via `bioio~=3.0`, pinned only in `uv.lock`. A `bioio` bump could drift the resolved dask version with nothing anchoring it at the source level. Not introduced here (this PR drops the 2-D `da.percentile` call, reducing reliance), but a `dask[array]` direct dependency would be worth adding while in this file.
- **Description nit (optional):** the Tests section doesn't mention the new content-pin golden (`blank-and-nan-channels-256.png`) and its OME-TIFF fixture, so "no golden changes" could read as "no new goldens." A one-line note would clear that up.

Reviewed against 7f3ae8a4

sir-sigurd:
Thanks — all three addressed in `c1415220`:

- **Float-path collapse-with-NaN test** → added `test_rescale_float_to_uint8_sparse_with_nan`, mirroring the `norm_img` case, so the "range over finite values, not NaN-poisoned `arr.min/max`" contract is now pinned at `_rescale_float_to_uint8`'s own call site too.
- **`dask` undeclared** → added `dask[array]~=2026.1` to `pyproject.toml` dependencies. Resolved version is unchanged (`2026.1.1`); the lock diff is just the two lines recording it as direct, and `--frozen` stays consistent.
- **Description** → Tests section now mentions the new content-pin golden (`blank-and-nan-channels-256.png`) and clarifies "no *existing* golden changed" vs. the one new golden added.

Full suite **88 passed**, all 30 network goldens still pixel-identical; CI green on the latest.

sir-sigurd:
**Design question — constant plane → black vs keep-level: keeping black.** Rationale, in priority order:

1. **Behavior-preserving.** Verified the old code rendered a constant plane black on *both* platforms — ARM (`int32(NaN)=0`) and x86-64 Lambda (`int32(NaN)=INT_MIN`, which clamps to 0 in the `I→I;16` PNG conversion). So the visible output was already black; keep-level would be a deliberate *appearance change*, which is out of scope for a fix whose invariant is "no visible change for well-defined inputs."

2. **No dtype-agnostic "level" exists in this path.** `norm_img` is dtype-polymorphic (casts every input to float64 up front). For a constant there's no single right level: uint16 `5000` → `5000`? float `0.5` → `0.5×65535=32768`? The two uint8 paths *can* keep level only because they're type-specialized, and they already disagree on how (`lo×255` vs `>>8`) — adding a third convention here is more divergence, not less. Full agreement is blocked anyway until the paths share an output depth (they're 16-bit vs 8-bit today).

3. **Matches mainstream library convention.** Checked the tools whose model matches ours (linear map of `[data_min, data_max] → display range`): matplotlib `Normalize` maps a constant to **0/black** and documents it ("returns 0 if `vmin == vmax`"); OpenCV `normalize(NORM_MINMAX)` likewise yields 0. scikit-image's clip-to-out_range is the lone keep-level variant, and it depends on a known output range. The microscopy viewers don't translate to our model (napari errors on `min==max` contrast limits; ImageJ is a windowing model, not normalize-to-fill).

4. **Practically tiny.** Only a *bit-exact-flat* plane goes black — a noisy/sparse one hits the min/max fallback and stays visible. A perfectly flat montage channel is almost always an empty/disabled acquisition channel, for which black ("no contrast here") is the honest rendering.

Reconciling all three rescale paths' edge semantics (and output depth) is the norm_img-unification follow-up — tracked, not this PR. Considering this design question resolved.

Refs: [matplotlib `Normalize`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Normalize.html), [OpenCV `normalize`](https://docs.opencv.org/master/d2/de8/group__core__array.html#ga87eef7ee3970f86906d69a92cbf064bd).

PR #4975, opened 2026-06-12
es_ingest: log Elasticsearch bulk indexing errors

sir-sigurd:
# Description

When the `es_ingest` lambda got a bulk Elasticsearch response with item-level errors, it raised a bare `BulkDocumentError` with no detail — operators had no way to tell which documents failed or why (the long-standing TODO at `bulk()`).

This widens the bulk `filter_path` from `errors` to `errors,items.*.error` so per-item error info comes back. On failure the items are aggregated by `(operation, error type, reason)` — failures are usually systematic (one bad mapping affecting many docs), so this bounds the log/message size — each distinct error is logged via `logger.error`, and a summary (`N document(s) failed in bulk request (M distinct error(s))`) is attached to `BulkDocumentError`.

Successful items carry no `error` field and so are pruned from the response, keeping the common no-error case compact. `BulkDocumentError` is still raised, so SQS retry/DLQ behavior is unchanged.

## Design decision: don't log per-document `_id`

Failure logs carry the error **type, reason, and count**, but not the document `_id`. `_id` exists on successful items too, so adding it to `filter_path` would un-prune every success and echo full item metadata back on *every* healthy bulk call. Given the deployed load, that's the wrong trade-off:

- es_ingest runs on a **160 MB** lambda (Lambda CPU scales with memory) and processes **one S3 object per invocation of up to 10,000 docs / 8 MB** (`BATCH_MAX_DOCS` / `BATCH_MAX_BYTES`), at up to 20× concurrency — the minimal `filter_path` is what keeps the common no-error response tiny (`{"errors": false}`).
- A fully-successful 10k-doc batch would echo back **~1.5 MB** of `_id`s to receive and JSON-parse on that 160 MB lambda, on every successful invocation; the compact response is ~17 bytes.
- The entry `_id` is `entry:{manifest_hash}:sha256(logical_key)` — opaque and non-reversible, so it wouldn't even identify the offending file. `error.reason` (which names the field/value) is the actionable signal.

So failures log as `Bulk <op> failed (xN): <type>: <reason>` plus the exception summary, and the handler logs the failing batch's `s3://bucket/key` — an O(1) correlator that (unlike `_id`) names the object to fetch and inspect, and which stays in S3 on failure for retry/DLQ. That's enough to diagnose the systematic failures that dominate, without taxing the hot path.

# TODO

- [x] Unit tests
- [x] [Changelog](../tree/master/docs/CHANGELOG.md) entry

<h3>Greptile Summary</h3>

This PR resolves a long-standing TODO by widening the Elasticsearch bulk `filter_path` from `"errors"` to `"errors,items.*.error"`, enabling per-item error details to be captured and logged on failure. Errors are aggregated by `(operation, type, reason)` before logging, keeping output compact for the common case of systematic failures.

- **Error aggregation and logging**: on bulk failure, a `Counter` collapses identical `(op, type, reason)` tuples into a single `logger.error` line with a repeat count, then attaches a human-readable summary (`N document(s) failed, M distinct error(s)`) to the raised `BulkDocumentError`.
- **Fallback path**: when `errors: true` but no per-item error details are returned (unexpected for ES 6.8), the raw response is logged and a descriptive detail string is attached to the exception so the failure remains diagnosable.
- **Tests**: three focused new/updated tests cover the happy path, the aggregation path, and the no-details fallback; `filter_path` is now asserted exactly (not `mocker.ANY`) in the error and ok tests.

Safe to merge; the change is additive (broader filter_path + aggregated logging) with no impact on the existing retry/DLQ path or SQS behavior.

The logic is correct, the fallback for unexpected ES responses is handled, and BulkDocumentError is still raised so downstream retry behavior is unchanged. The one observation is that the log loop iterating failures.items() has no explicit upper bound — ES reason strings can carry per-field or per-value detail, making it possible for a large batch to produce many distinct tuples and a corresponding burst of logger.error calls. This is unlikely in practice for the systematic failures that dominate, but it could matter in edge cases with diverse per-document reasons.

lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py — specifically the for … in failures.items() loop

| Filename | Overview |
|----------|----------|
| lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py | Core change: widens filter_path to "errors,items.*.error", aggregates per-item errors via Counter, and attaches a detail string to BulkDocumentError. Logic is correct; one minor note on unbounded logger.error loop for highly-distinct failures. |
| lambdas/es_ingest/tests/test_main.py | Good new test coverage: test_bulk_ok verifies the updated filter_path, test_bulk_error checks Counter aggregation and exception message, test_bulk_error_without_details covers the fallback path. All assertions are specific and meaningful. |
| lambdas/es_ingest/CHANGELOG.md | Changelog entry added correctly at the top of the Changes section. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[es.bulk called\nfilter_path: errors,items.*.error] --> B{TransportError?}
    B -- 429 --> C[sleep_until_timeout\nraise TooManyRequestsError]
    B -- other --> D[re-raise]
    B -- no --> E{resp errors == True?}
    E -- no --> F[Return normally\ncompact response ~17 bytes]
    E -- yes --> G[Build Counter\ngroup by op + type + reason]
    G --> H{any failures in Counter?}
    H -- yes --> I[logger.error per\ndistinct group\nop, count, type, reason]
    I --> J[detail = N doc\nfailed M distinct errors]
    H -- no --> K[logger.error raw resp\ndetail = no per-item details]
    J --> L[raise BulkDocumentError with detail]
    K --> L
```

<sub>Reviews (1): Last reviewed commit: ["es\_ingest: guard against errors flag wit..."](https://github.com/quiltdata/quilt/commit/b947156e2a75011fb98e5419c64b11ac044328d1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37615128)</sub>

### DIFF
--- a/lambdas/es_ingest/CHANGELOG.md
+++ b/lambdas/es_ingest/CHANGELOG.md
@@ -17,7 +17,7 @@ where verb is one of
 
 ## Changes
 
-- [Added] Log Elasticsearch bulk indexing errors ([#PR](https://github.com/quiltdata/quilt/pull/PR))
+- [Added] Log Elasticsearch bulk indexing errors ([#4975](https://github.com/quiltdata/quilt/pull/4975))
 - [Changed] Upgrade to Python 3.13 ([#4656](https://github.com/quiltdata/quilt/pull/4656))
 - [Changed] Switch to EventBridge S3 events ([#4471](https://github.com/quiltdata/quilt/pull/4471))
 - [Added] Bootstrap the change log ([#4422](https://github.com/quiltdata/quilt/pull/4422))
--- a/lambdas/es_ingest/CHANGELOG.md
+++ b/lambdas/es_ingest/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Added] Log Elasticsearch bulk indexing errors ([#PR](https://github.com/quiltdata/quilt/pull/PR))
 - [Changed] Upgrade to Python 3.13 ([#4656](https://github.com/quiltdata/quilt/pull/4656))
 - [Changed] Switch to EventBridge S3 events ([#4471](https://github.com/quiltdata/quilt/pull/4471))
 - [Added] Bootstrap the change log ([#4422](https://github.com/quiltdata/quilt/pull/4422))
--- a/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
+++ b/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
@@ -112,5 +112,11 @@ def handler(event, context):
         params["VersionId"] = version_id
 
     data = s3_client.get_object(**params)["Body"].read()
-    bulk(context, es, data)
+    try:
+        bulk(context, es, data)
+    except BulkDocumentError:
+        # Surface which batch failed. The object isn't deleted (below), so it
+        # stays in S3 for retry/DLQ and can be fetched for inspection.
+        logger.error("Bulk indexing failed for s3://%s/%s", bucket, key)
+        raise
     s3_client.delete_object(**params)
--- a/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
+++ b/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
@@ -1,3 +1,4 @@
+import collections
 import json
 import os
 import random
@@ -38,7 +39,10 @@ def bulk(context, es, data: bytes):
     try:
         resp = es.bulk(
             data,
-            filter_path="errors",
+            # Return per-item errors so failures can be diagnosed. Successful
+            # items have no `error` field, so they're pruned from the response,
+            # keeping the (common) no-error case compact.
+            filter_path="errors,items.*.error",
             # wait as much as possible because it's better die trying than just die
             # leave a second to avoid lambda timeout
             request_timeout=context.get_remaining_time_in_millis() / 1000 - 1,
@@ -65,9 +69,21 @@ def bulk(context, es, data: bytes):
         logger.warning("Sleeping for %s seconds to avoid ES overload", time_to_sleep)
         time.sleep(time_to_sleep)
     if resp["errors"]:
-        # TODO: log errors from items.*.error?
+        # Failures are usually systematic (e.g. a single bad mapping affecting
+        # many documents), so aggregate by error to bound the log/message size.
+        failures = collections.Counter(
+            (op, error.get("type"), error.get("reason"))
+            for item in resp.get("items", [])
+            for op, details in item.items()
+            if (error := details.get("error"))
+        )
+        for (op, error_type, reason), count in failures.items():
+            logger.error("Bulk %s failed (x%s): %s: %s", op, count, error_type, reason)
         # TODO: ignore index_not_found_exception for delete operations?
-        raise BulkDocumentError
+        raise BulkDocumentError(
+            f"{sum(failures.values())} document(s) failed to index "
+            f"({len(failures)} distinct error(s))"
+        )
 
 
 def handler(event, context):
--- a/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
+++ b/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
@@ -80,9 +80,11 @@ def bulk(context, es, data: bytes):
         )
         if failures:
             for (op, error_type, reason), count in failures.items():
-                logger.error("Bulk %s failed (x%s): %s: %s", op, count, error_type, reason)
+                # %r so a user-controlled `reason` (ES echoes field names/values)
+                # can't forge log lines via embedded newlines.
+                logger.error("Bulk %s failed (x%s): %r: %r", op, count, error_type, reason)
             detail = (
-                f"{failures.total()} document(s) failed to index "
+                f"{failures.total()} document(s) failed in bulk request "
                 f"({len(failures)} distinct error(s))"
             )
         else:
--- a/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
+++ b/lambdas/es_ingest/src/t4_lambda_es_ingest/__init__.py
@@ -78,13 +78,22 @@ def bulk(context, es, data: bytes):
             for op, details in item.items()
             if (error := details.get("error"))
         )
-        for (op, error_type, reason), count in failures.items():
-            logger.error("Bulk %s failed (x%s): %s: %s", op, count, error_type, reason)
+        if failures:
+            for (op, error_type, reason), count in failures.items():
+                logger.error("Bulk %s failed (x%s): %s: %s", op, count, error_type, reason)
+            detail = (
+                f"{failures.total()} document(s) failed to index "
+                f"({len(failures)} distinct error(s))"
+            )
+        else:
+            # `errors` is set but no per-item error details came back. This
+            # shouldn't happen for ES 6.8 (failed items always carry an `error`),
+            # but log the raw response so the failure stays diagnosable instead
+            # of raising with no detail.
+            logger.error("Bulk reported errors but no per-item error details were found: %s", resp)
+            detail = "bulk reported errors but no per-item error details were found"
         # TODO: ignore index_not_found_exception for delete operations?
-        raise BulkDocumentError(
-            f"{failures.total()} document(s) failed to index "
-            f"({len(failures)} distinct error(s))"
-        )
+        raise BulkDocumentError(detail)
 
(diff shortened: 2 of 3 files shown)


sir-sigurd (commented):

Sanity-checked the overall approach, security, edge cases, ES 6.8 response handling, performance, and the tests — no blockers. The `filter_path="errors,items.*.error"` compact-no-error-response behavior is confirmed for ES 6.8: filter_path prunes empty objects/arrays, so a fully-successful bulk returns `{"errors": false}` with no `items` (per elastic/elasticsearch#63842). The change is a contained logging addition; SQS retry/DLQ behavior is unchanged; CHANGELOG entry present.

**On the open `_id` question**

`reason` isn't a reliable document locator — for most error types it names the offending field/value, not the document `_id`, so the logs say *what kind* of error and *how many*, not *which* document. Since failures are usually systematic, a per-doc `_id` adds little for the common case. If it's worth capturing, a middle option avoids un-pruning every success: widen `filter_path` to include `items.*._id` and attach a *sample* `_id` per error group rather than every id.

**Nits**

- Tests don't exercise the count-collapse path: both mocked errors in `test_bulk_error` are distinct, so every `count` is 1 — the `(x{count})` aggregation and the `({len(failures)} distinct error(s))` suffix are never asserted. Add a repeated-error case asserting `(x2)` and the distinct-count suffix.
- The "bounds the log/message size" rationale holds for systematic failures, not universally: the Counter key includes the free-text `reason`, so for error types whose top-level `reason` embeds per-document values, distinct reasons scale with doc count. (ES often puts per-doc specifics in `caused_by`, which isn't captured, so the common case is fine.) Either soften the comment or cap the number of distinct groups logged.
- `sum(failures.values())` can be `failures.total()` (Counter API, 3.10+; runtime is 3.13) — purpose-built, reads cleaner.
- `resp.get("items", [])` (defensive) vs `resp["errors"]` (direct) reads as an inconsistency; the items-pruned-on-success rationale is in the comment up at `filter_path`. A word at the `.get` site, or leave it — minor.

Reviewed against c9285cd

sir-sigurd:
Follow-up self-review pass — likely the last round.

The prior round's nits are addressed in `6bbeb4bf`: the count-collapse path is now tested (`(x2)` plus the distinct-error-count message), `failures.total()` replaces `sum(failures.values())`, and the aggregation comment no longer overpromises a hard bound. The ES 6.8 compact-no-error-response behavior is now firmly confirmed — `filter_path` doesn't preserve empty objects/arrays (per the ES maintainers in elastic/elasticsearch#63842; it's a Jackson-library property, stable across 6.8/7.x), so an all-success bulk is `{"errors": false}` with `items` omitted.

Two low-priority notes remain (both optional — I'd set them aside earlier, raising now only for completeness):

- The failure count and message come from the filtered `items`, not the authoritative top-level `errors` flag. If a response ever set `errors: true` without a matching per-item `error` object (unreachable under the documented 6.8 contract — failed items always carry an `error` object), the code would raise `BulkDocumentError("0 document(s) failed to index (0 distinct error(s))")` with no log lines, silently defeating the diagnostic purpose. Cheap insurance: guard the raise on `if failures:` with a raw-response fallback log otherwise. This path is also untested.
- Logged `reason` strings can carry document field values — ES parse-error reasons sometimes echo the offending value, so user metadata can land in CloudWatch (a different access boundary than ES), a new path vs. the prior silent failure. Fine as a deliberate trade-off; only worth hardening (`%r`/newline-strip) if the indexed metadata is ever sensitive.

Reviewed against 6bbeb4bf

sir-sigurd (commented):
Notes from an AI-assisted self-review pass — likely the last round.

Checked the overall approach, ES 6.8 response handling, edge cases, security/log hygiene, Python idioms, and the tests — no blockers. The per-item response handling matches the ES 6.8 bulk contract (`filter_path` prunes successful items, so an all-success bulk stays `{"errors": false}`), and the three tests pin the new aggregation/fallback behavior (they fail if the change is reverted).

**One design question worth noting (not blocking)**

The aggregation keeps a single invocation's logs compact, but for the systematic failures it's optimized for, the whole batch raises → SQS redelivers the same S3 object → `bulk()` re-runs → the same aggregated groups get logged again on every retry until the message reaches the DLQ. Retry/DLQ behavior is unchanged by design, so this is mostly pre-existing — but the per-retry log volume is new (it was zero before this PR). Open question for a future change: should a systematic per-item error (vs. a transient 429) short-circuit to the DLQ rather than burning the full retry budget?

**Low-priority observations (all optional)**

- The `BulkDocumentError` summary hardcodes `"{N} document(s) failed to index"`, but the count aggregates every operation — including `delete` (the test fixture itself includes a delete failure, so it reads "3 document(s) failed to index"). The per-line logs name the real operation correctly; only the rolled-up summary is imprecise. Operation-neutral wording ("failed in bulk request") plus the matching test-string update would fix it. *(Same point Copilot raised inline.)*

- On failure, neither the per-group logs nor the exception names the source `s3://{bucket}/{key}`. It's the cheapest correlator for finding which batch failed, it's O(1) (unlike a per-document `_id`), and it's already in scope in `handler` — but it currently lands in CloudWatch only at DEBUG (`logger.debug("Invoked with event...")`), which is below the default WARNING level. Worth surfacing bucket/key on the error path.

- `reason`/`type` are logged with `%s`. ES 6.8 echoes user-controlled field names verbatim into the top-level `reason` (e.g. `strict_dynamic_mapping_exception`, `mapper_parsing_exception`), and a package author can embed a newline via an escaped JSON key — so a crafted failing document can split or forge CloudWatch log lines. Logging with `%r` escapes control chars at zero cost and makes any injected content unambiguous. This extends the earlier `%r` note: the concern isn't only metadata sensitivity — it's log integrity, which holds regardless.

**On the bot comments**

- Greptile's "unbounded log loop": the loop is bounded by batch size — `BATCH_MAX_DOCS` caps each S3 object at 10,000 docs and the lambda processes one object per invocation, so the worst case is ~10k `logger.error` calls, not unbounded. The cap-vs-soften tradeoff was settled last round (softened the comment rather than add a hard cap); the worst case is a bounded, rare tail, so leaving it uncapped stands.
- Copilot's summary-wording note is the first observation above.

Reviewed against b947156e

sir-sigurd:
Thanks @copilot and @greptileai — dispositions below (all in `beafe8f2` and `131071e2`):

**Applied**
- **Summary wording** (Copilot): the `BulkDocumentError` message aggregates all operations, not just `index`, so it now reads "N document(s) failed **in bulk request**" instead of "failed to index". Tests updated.
- **Log integrity**: error `type`/`reason` are now logged with `%r`, so a user-controlled `reason` (ES echoes field names/values) can't forge CloudWatch log lines via embedded newlines.
- **Failed-batch correlator**: on failure the handler now logs the batch's `s3://bucket/key`. It's O(1) (unlike a per-document `_id`) and the object isn't deleted on failure, so it stays in S3 to fetch and inspect.

**Not adopting**
- **Greptile P2 "unbounded log loop":** the loop is bounded, not unbounded — `BATCH_MAX_DOCS` caps each batch at 10,000 docs and the lambda processes one S3 object per invocation, so the worst case is ~10k `logger.error` calls in a rare all-distinct batch. Capping individual lines was considered and set aside in favor of the softened rationale; a hard cap isn't warranted for a bounded, rare tail.

**Follow-up (separate change)**
- Systematic per-item failures re-log on every SQS retry until the message reaches the DLQ. Retry/DLQ behavior is unchanged by this PR (and the prior behavior logged nothing); short-circuiting systematic errors straight to the DLQ instead of burning the retry budget is worth considering separately.

sir-sigurd (commented):
Follow-up self-review pass — likely the last round.

The three notes from the prior pass are addressed in `beafe8f2` and `131071e2`, each with test coverage:

- The summary message is now operation-neutral (`"{N} document(s) failed in bulk request"`) instead of hardcoding "to index", so a delete/update failure isn't mislabeled.
- On `BulkDocumentError` the handler now logs the source `s3://{bucket}/{key}` and re-raises, so the failing batch is identifiable on the error path (the object stays in S3 for retry/DLQ).
- `reason`/`type` are logged with `%r`, so a user-controlled `reason` can't forge CloudWatch log lines via embedded newlines.

No new issues in the new commits: the source-logging line uses `%s` for the object key, but the keys this lambda consumes are machine-generated batch keys (`{random}/object`), not user-controlled, so the `%r` hardening is correctly scoped to the user-controlled `reason`. Tests pass.

The retry-amplification point from the prior pass (systematic failures re-log on every SQS retry until DLQ) is unchanged and remains a separate, optional future consideration.

Reviewed against 131071e2

PR #4976, opened 2026-06-12
thumbnail: stream montage normalization to bound multi-channel memory

sir-sigurd:
# Description

The n-dim **montage / Z-projection** path normalizes each channel plane with `norm_img`. The recent eager rewrite (#4974) computes every plane up front, so the montage holds **all N normalized planes at once** before assembling them. Real-lambda benchmarking showed this raises peak memory on large multi-channel images and **OOMs a couple of channels sooner** than the previous lazy `da.percentile` path — a regression on exactly the OOM-prone case (OOM is this lambda's documented failure mode).

This restores the lazy path's memory behavior **without giving up the eager correctness/perf fixes**:
- The per-plane normalization stays **eager NumPy** (`_normalize_plane`) — finite-aware percentiles, constant/NaN handling, and the uint16 histogram+LUT rescale all stay exact and bit-identical.
- `norm_img` now wraps it in **`dask.delayed` / `da.from_delayed`**, and the montage goes back to `da.pad`/`da.concatenate`. The planes are deferred tasks in the montage graph rather than concrete arrays, so dask **releases each plane once it has been copied into the montage** and they don't stay resident through the downstream resize. (Returning a concrete array — `da.from_array` — instead bakes all N planes into the graph, and `handle_image` holds the montage through `generate_thumbnail`'s resize, so all N linger there.)

Net: eager's speed **+** the lazy path's OOM resistance + the correctness fixes.

## Output is unchanged

Bit-identical — verified against all montage goldens, including the `--large-files` ones (actk, 3d-cell-viewer, variance-cfe, pipeline-4, multi-scene CZIs) where the streaming actually engages. Ruff + pyflakes clean.

## Real-lambda benchmark (thumb.dev, final image, 2048 MB, 5 cold runs each)

Three normalization variants on the same images (uint16 and float32 montages, 4000×4000 channels): the original lazy `da.percentile` path, the eager rewrite #4974 merged to master, and this PR.

| montage | lazy `da.percentile` (pre-#4974) | eager (#4974, current master) | **this PR (lazy streaming)** |
|---|---|---|---|
| uint16 8ch | ok 5/5 · 15.9 s | ok 5/5 · 14.8 s | **ok 5/5 · 12.3 s** |
| uint16 10ch | ok 5/5 · 19.1 s | **OOM 5/5** | **ok 5/5 · 14.3 s** |
| float 6ch | ok 5/5 · 18.5 s | ok 5/5 · 17.6 s | **ok 5/5 · 16.2 s** |
| float 8ch | **OOM 2/5** | **OOM 5/5** | **ok 5/5 · 18.2 s** |

- **#4974's eager rewrite (current master) regressed multi-channel memory** — it OOMs at the cliffs (uint16 10ch, float 8ch) where the older lazy path coped. **This PR fixes both** and is the only variant that succeeds 5/5 everywhere (at float 8ch even more robustly than the old lazy path, which is flaky right at the 2048 MB edge).
- **~15-25% faster than the old lazy path** (eager NumPy beats per-plane `da.percentile`), and as fast as #4974's eager rewrite — `dask.delayed` changes only scheduling, not the math.

Method: `aws lambda invoke --log-type Tail` → `REPORT` line; cold start forced before every run (config bump + `wait function-updated`); same function, image swapped per version.

### What the win actually is (instrumented)

The instantaneous peak *while the montage is assembled* is similar lazy vs eager — the final `da.concatenate` holds all N planes momentarily either way. The difference is **afterward**: staged `tracemalloc` over the full `handle_image` shows eager keeps all N planes resident from `format` through the resize (254 MB → 493 MB held), while lazy drops to just the montage output after `.compute()` (242 MB) — so the planes don't coexist with the resize. The 5×-cold A/B above holds the histogram/LUT **constant** in both arms, so it isolates this lazy-release effect as the OOM-vs-ok difference at 10ch. (The uint16 histogram/LUT is kept and lowers the *per-plane* float64 transient, but it's not the lever that flips the OOM here.)

# TODO

- [x] Unit tests

### DIFF
--- a/.github/workflows/deploy-lambdas.yaml
+++ b/.github/workflows/deploy-lambdas.yaml
@@ -4,14 +4,12 @@ on:
   push:
     branches:
       - master
-      - thumbnail-montage-bounded-memory  # TEMP: build branch image for dev-stack benchmark; revert before merge
     paths:
       - '.github/workflows/deploy-lambdas.yaml'
       - 'lambdas/**'
 
 jobs:
   deploy-lambda-s3:
-    if: github.ref == 'refs/heads/master'  # TEMP: skip on the benchmark branch; revert before merge
     strategy:
       fail-fast: false
       matrix:
@@ -69,9 +67,9 @@ jobs:
       fail-fast: false
       matrix:
         path:
-          # TEMP: thumbnail only for the dev-stack benchmark; restore indexer +
-          # tabular_preview before merge
+          - indexer
           - thumbnail
+          - tabular_preview
     runs-on: ubuntu-latest
     # These permissions are needed to interact with GitHub's OIDC Token endpoint.
     permissions:
@@ -89,5 +87,10 @@ jobs:
           aws-region: us-east-1
       - name: Push Docker image to Prod ECR
         run: ./lambdas/scripts/upload_ecr.sh 730278974607 "quiltdata/lambdas/${{ matrix.path }}:${{ github.sha }}"
-      # TEMP: GovCloud push removed for the branch build (dev-stack pulls from
-      # Prod only); restore before merge.
+      - name: Configure AWS credentials from GovCloud account
+        uses: aws-actions/configure-aws-credentials@v6
+        with:
+          role-to-assume: arn:aws-us-gov:iam::313325871032:role/github/GitHub-Quilt
+          aws-region: us-gov-east-1
+      - name: Push Docker image to GovCloud ECR
+        run: ./lambdas/scripts/upload_ecr.sh 313325871032 "quiltdata/lambdas/${{ matrix.path }}:${{ github.sha }}"
--- a/.github/workflows/deploy-lambdas.yaml
+++ b/.github/workflows/deploy-lambdas.yaml
@@ -4,12 +4,14 @@ on:
   push:
     branches:
       - master
+      - thumbnail-montage-bounded-memory  # TEMP: build branch image for dev-stack benchmark; revert before merge
     paths:
       - '.github/workflows/deploy-lambdas.yaml'
       - 'lambdas/**'
 
 jobs:
   deploy-lambda-s3:
+    if: github.ref == 'refs/heads/master'  # TEMP: skip on the benchmark branch; revert before merge
     strategy:
       fail-fast: false
       matrix:
@@ -67,9 +69,9 @@ jobs:
       fail-fast: false
       matrix:
         path:
-          - indexer
+          # TEMP: thumbnail only for the dev-stack benchmark; restore indexer +
+          # tabular_preview before merge
           - thumbnail
-          - tabular_preview
     runs-on: ubuntu-latest
     # These permissions are needed to interact with GitHub's OIDC Token endpoint.
     permissions:
@@ -87,10 +89,5 @@ jobs:
           aws-region: us-east-1
       - name: Push Docker image to Prod ECR
         run: ./lambdas/scripts/upload_ecr.sh 730278974607 "quiltdata/lambdas/${{ matrix.path }}:${{ github.sha }}"
-      - name: Configure AWS credentials from GovCloud account
-        uses: aws-actions/configure-aws-credentials@v6
-        with:
-          role-to-assume: arn:aws-us-gov:iam::313325871032:role/github/GitHub-Quilt
-          aws-region: us-gov-east-1
-      - name: Push Docker image to GovCloud ECR
-        run: ./lambdas/scripts/upload_ecr.sh 313325871032 "quiltdata/lambdas/${{ matrix.path }}:${{ github.sha }}"
+      # TEMP: GovCloud push removed for the branch build (dev-stack pulls from
+      # Prod only); restore before merge.
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Changed] Lower peak memory of the normalized greyscale montage/projection path for unsigned 16-bit images via a histogram percentile and lookup-table rescale, instead of a float64 copy of the whole plane (generated thumbnails unchanged) ([#4976](https://github.com/quiltdata/quilt/pull/4976))
 - [Fixed] Normalized greyscale thumbnails (multi-channel montages, Z-projections) with a constant channel or NaN pixels now render deterministically instead of coming out blank or garbled; other thumbnails are unchanged ([#4974](https://github.com/quiltdata/quilt/pull/4974))
 - [Changed] Simplify 16-bit greyscale handling: rescale to 8-bit by array dtype before resizing and drop the now-unreachable resampler fallback (generated thumbnails unchanged) ([#4971](https://github.com/quiltdata/quilt/pull/4971))
 - [Changed] Compute the 16-bit rescale percentiles via a histogram instead of `np.percentile`, lowering peak memory on large images (output unchanged) ([#4968](https://github.com/quiltdata/quilt/pull/4968))
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,7 +17,7 @@ where verb is one of
 
 ## Changes
 
-- [Changed] Lower peak memory of the normalized greyscale montage/projection path for unsigned 16-bit images via a histogram percentile and lookup-table rescale, instead of a float64 copy of the whole plane (generated thumbnails unchanged) ([#4976](https://github.com/quiltdata/quilt/pull/4976))
+- [Changed] Normalize greyscale montage/projection planes lazily so the montage streams — dask computes and frees one channel at a time instead of holding all of them at once — keeping peak memory bounded on large multi-channel images and running faster; generated thumbnails unchanged ([#4976](https://github.com/quiltdata/quilt/pull/4976))
 - [Fixed] Normalized greyscale thumbnails (multi-channel montages, Z-projections) with a constant channel or NaN pixels now render deterministically instead of coming out blank or garbled; other thumbnails are unchanged ([#4974](https://github.com/quiltdata/quilt/pull/4974))
 - [Changed] Simplify 16-bit greyscale handling: rescale to 8-bit by array dtype before resizing and drop the now-unreachable resampler fallback (generated thumbnails unchanged) ([#4971](https://github.com/quiltdata/quilt/pull/4971))
 - [Changed] Compute the 16-bit rescale percentiles via a histogram instead of `np.percentile`, lowering peak memory on large images (output unchanged) ([#4968](https://github.com/quiltdata/quilt/pull/4968))
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,7 +17,7 @@ where verb is one of
 
 ## Changes
 
-- [Changed] Normalize greyscale montage/projection planes lazily so the montage streams — dask computes and frees one channel at a time instead of holding all of them at once — keeping peak memory bounded on large multi-channel images and running faster; generated thumbnails unchanged ([#4976](https://github.com/quiltdata/quilt/pull/4976))
+- [Changed] Normalize greyscale montage/projection planes lazily so dask releases each plane once it is copied into the montage instead of keeping every channel resident through the downstream resize — lowering peak memory on large multi-channel images and running faster; generated thumbnails unchanged ([#4976](https://github.com/quiltdata/quilt/pull/4976))
 - [Fixed] Normalized greyscale thumbnails (multi-channel montages, Z-projections) with a constant channel or NaN pixels now render deterministically instead of coming out blank or garbled; other thumbnails are unchanged ([#4974](https://github.com/quiltdata/quilt/pull/4974))
 - [Changed] Simplify 16-bit greyscale handling: rescale to 8-bit by array dtype before resizing and drop the now-unreachable resampler fallback (generated thumbnails unchanged) ([#4971](https://github.com/quiltdata/quilt/pull/4971))
 - [Changed] Compute the 16-bit rescale percentiles via a histogram instead of `np.percentile`, lowering peak memory on large images (output unchanged) ([#4968](https://github.com/quiltdata/quilt/pull/4968))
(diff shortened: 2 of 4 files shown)


sir-sigurd:
@greptileai please review

sir-sigurd (commented):
Notes from an AI-assisted self-review pass.

Checked: the streaming / bounded-memory mechanism (instrumented + measured), numerical bit-identical equivalence (uint8 / uint16 / byte-swapped `>u2`, 20k+ random trials), refactor and comment drift, Python/NumPy dtype traps, and security. One thing is worth a look before merge; the rest are nits. (Leaving the temporary `deploy-lambdas.yaml` scaffolding aside — that revert is already understood.)

## The measured mechanism doesn't match the description

I instrumented the montage assembly and ran it. Findings:

- **All N normalized planes are resident at peak — not freed one at a time.** Every per-channel plane feeds a single final `da.concatenate` (materialized in one `img.compute()` at `__init__.py:511`), so none can be released until the whole montage is built. Measured 10/10 planes alive at peak for a 10-channel montage; peak ≈ 2× the montage (all planes + the concatenation buffer), under both the threaded and synchronous scheduler. So the description's "dask computes and frees one plane at a time … keeping peak memory to ~the montage plus a single plane instead of all N", the matching CHANGELOG line ("frees one channel at a time"), and the code comments in `norm_img` / `_format_n_dim_ndarray` describe behavior that doesn't occur.

- **What actually lowers peak is the uint16 histogram+LUT path**, not the lazy wrapping. Same 10-channel montage: 332 MB through the LUT path vs 609 MB through the float path (the float64 copy + percentile copy per plane is the difference). That's the reverse of the description's note that the histogram/LUT "alone is not an end-to-end win" — it *is* the win.

- **`dask.delayed` buys nothing here.** Eager assembly (the previous structure) + the LUT path peaks at the same ~322 MB as the lazy version. So the lazy graph adds indirection without lowering peak.

Two ways to reconcile:
1. Keep `dask.delayed`, but rewrite the description, CHANGELOG, and the `norm_img` / `_format_n_dim_ndarray` comments to credit the LUT path and drop the "frees one plane at a time / N→1" framing.
2. Drop the `dask.delayed` wrapping and add the LUT path on top of the existing eager structure — same memory, less indirection, and it sidesteps the threaded scheduler computing several planes' float64 transients at once on the float path (which is a potential regression versus the old sequential eager path for float-typed montages — the uint path is unaffected).

Caveat: these are workstation `tracemalloc` numbers, isolated from the decode/resize pipeline, so the real-lambda RSS you measured is authoritative for the end-to-end OOM outcome. But "all N planes resident" and "eager == lazy peak" are properties of the graph and hold regardless of where they're run.

## Nits

- `_percentile_uint16`'s comment (`__init__.py:542`) still says "The sole caller (`_rescale_uint16_to_uint8`) guards emptiness" — it's now reached via `_uint16_clip_range` from both `_rescale_uint16_to_uint8` and `_normalize_plane`. The emptiness contract still holds (both callers guard `arr.size` first), but the comment misstates who the callers are.
- `_finite_clip_range` and `_uint16_clip_range` docstrings say "Shared by `norm_img` …" — `norm_img` no longer calls them; `_normalize_plane` does.
- `_uint16_clip_range` / `_percentile_uint16` also serve uint8 and byte-swapped `>u2`, not just native uint16, so the names read narrower than the contract (Copilot noted the same). Either rename, or note the real contract ("unsigned, ≤2 bytes").
- `da.from_delayed(delayed(_normalize_plane)(img), shape=img.shape, dtype=np.int32)` trusts `_normalize_plane`'s return shape/dtype; a future change to either would silently produce wrong graph metadata rather than raise.
- `_normalize_plane`'s float branch does `arr = arr.astype(np.float64, copy=False)` then mutates in place — safe today (dask hands it a freshly-computed block) but, now that this is a standalone function, the `copy=False` aliasing trap is easier to trip from a future direct caller or test.
- The empty-array concern raised in the bot review is already handled on this revision (the `if not arr.size` guard in `_normalize_plane` plus `test_norm_img_empty_plane_renders_black`).

Reviewed against 81a5a00

sir-sigurd:
Thanks — you're right that the "frees one plane at a time / montage + 1 plane" framing was wrong, and I dug into *why* lazy still helps end-to-end. Staged `tracemalloc` over the full `handle_image` (10ch), eager (`da.from_array`) vs lazy (`da.from_delayed`):

| stage | eager | lazy |
|---|---|---|
| after format | 254 MB (all N planes materialized) | 2 MB (deferred graph) |
| after compute | current **493** / peak 494 | current **242** / peak 480 |
| after resize | 493 held | 242 |

So you're correct the **assembly peak is ~equal** (494 vs 480 — the concatenate holds all N momentarily). What I missed, and what this shows: **eager keeps the N concrete planes resident from `format` through the resize**, because `da.from_array` bakes them into the montage graph and `handle_image` holds the montage (`img`) until `generate_thumbnail` returns. Lazy releases each plane once it's copied into the montage, so only the output (~half) carries into the resize. That lingering ~N-planes is what tips eager over 2048 on large multi-channel images — not a lower concatenate peak.

On **"`dask.delayed` buys nothing"**: my 5×-cold real-lambda A/B holds the histogram/LUT **constant** in both arms — `26971d0c` (eager **+LUT**) vs `8aa8845d` (lazy **+LUT**) — and eager OOMs 5/5 at 10ch while lazy is ok 5/5. Since the LUT is identical in both, the lazy release is the only difference, so it *is* the lever here (your isolated `_format_n_dim_ndarray` measurement has no downstream resize, which is why it read equal — and you noted the real-lambda RSS is authoritative). The histogram/LUT is kept and does lower the per-plane float64 transient, but it's not what flips the OOM.

I've rewritten the description, CHANGELOG, and the `norm_img` / `_format_n_dim_ndarray` comments to describe **release-before-resize** instead of one-plane-at-a-time, and fixed the nits: comment drift (`_finite_clip_range` / `_uint16_clip_range` / `_percentile_uint16` are reached via `_normalize_plane`, not `norm_img`), the `from_delayed` shape/dtype contract, the uint8/`>u2` reach of the uint16 helpers, and the float-path `copy=False` aliasing assumption. (The bot empty-array note was already handled — `_normalize_plane`'s `if not arr.size` + `test_norm_img_empty_plane_renders_black`.)

A final real-lambda benchmark is planned before merge; it'll also isolate the LUT-vs-float-path contribution so the attribution is fully pinned.

sir-sigurd:
Followed up on the float-montage concern (lazy + threaded scheduler possibly computing concurrent float64 transients). Float32 multi-channel montages, 4000², 2048 MB, 5× cold each:

| | 4ch | 6ch | 8ch |
|---|---|---|---|
| OLD-lazy | ok 5/5 · 14.6s | ok 5/5 · 18.6s | **ok 5/5 · 24.2s** |
| eager (#4976 `da.from_array`) | ok 5/5 · 12.2s | ok 5/5 · 16.0s | **OOM 5/5** |
| lazy-delayed | ok 5/5 · 12.0s | ok 5/5 · 15.1s | **ok 5/5 · 18.1s** |

The worry didn't materialize: lazy-delayed **matches OLD-lazy's OOM resistance on float montages** (8ch ok 5/5, where eager OOMs 5/5) and is ~18-25% faster. Vs the OLD-lazy baseline — which was *also* threaded (`da.percentile`), so it had the same concurrency — lazy-delayed is not worse anywhere in 4-8ch. (Your concern compared against the eager *sequential* path; but eager OOMs at 8ch anyway, and OLD-lazy is the deployed baseline.)

So across **both uint16 and float** montages: lazy-delayed = OLD-lazy's OOM resistance + ~20-30% faster + the correctness/histogram fixes, and it fixes eager's 8ch OOM in both. A fresh-build re-confirm on the merged SHA is provenance-only (the deployed `8aa8845d` is behaviorally identical — later commits are comments + the `np.arange(imax)` no-op).

sir-sigurd:
**Final benchmark — on the merged-final image (`16d0e31d`), 5× cold, 2048 MB.** Confirms the result on the actual final SHA (not just the earlier `8aa8845d`, which is behaviorally identical):

| montage | OLD-lazy (deployed) | eager #4976 | **this PR (final)** |
|---|---|---|---|
| uint16 8ch | ok 5/5 · 15.9s | ok 5/5 · 14.8s | ok 5/5 · **12.3s** |
| uint16 10ch | ok 5/5 · 19.1s | **OOM 5/5** | ok 5/5 · **14.3s** |
| float 6ch | ok 5/5 · 18.5s | ok 5/5 · 17.6s | ok 5/5 · **16.2s** |
| float 8ch | **OOM 2/5** | **OOM 5/5** | ok 5/5 · **18.2s** |

This PR is the only version that succeeds 5/5 everywhere — it fixes eager's OOM at both cliffs (uint16 10ch, float 8ch), and at float 8ch it's actually *more* robust than the currently-deployed lazy code (which is flaky right at the 2048 edge there). ~15-25% faster than the deployed path across the board, output bit-identical.

sir-sigurd (commented):
Notes from an AI-assisted self-review pass (against `16d0e31d`) — likely the last round before the final benchmark and merge.

The earlier blocker (the description vs. measured-mechanism mismatch) is resolved. The description, CHANGELOG, and the `norm_img` / `_format_n_dim_ndarray` comments now describe release-before-resize, and the mechanism holds on a trace: `handle_image` runs `img = generate_thumbnail(img.compute(), size)`, so the montage dask collection stays bound through `generate_thumbnail` — in the eager (`da.from_array`) case its graph pins all N concrete planes across the resize, while `da.from_delayed` holds only deferred tasks.

Checked, nothing flagged beyond the notes below:

- **Bit-identical equivalence is solid.** Brute-forced the unsigned path against the float64 reference across all 65536 LUT values × 400 `(lo, hi)` pairs (0 mismatches) and 130k+ random / bimodal / skewed trials (realized histogram-vs-`np.percentile` difference always ~7.3e-12, **zero int32 output flips**). uint8 / `>u2`, the top-bin clamp, NaN handling, and the empty-plane guard all hold.
- Also checked the normalization math, refactor fidelity (no dropped behavior or comments beyond the nits below), whether `dask.delayed` earns its keep (the simpler incremental-assembly attempt in `166e1388` regressed +145 MB, so yes), the security / DoS surface (net reduced — fixed-size LUT and block-bounded histogram replace the full-plane float64 copy), and possible simplifications (none — the duplicated guards are intrinsic to the two ranging paths).

## Before merge

- **Revert the `deploy-lambdas.yaml` scaffolding.** Current state on this branch: the ECR build matrix is `thumbnail`-only (indexer + tabular_preview wouldn't be built/pushed on master), the GovCloud ECR push is deleted, and `deploy-lambda-ecr` has no `if: master` guard (only `deploy-lambda-s3` does), so a branch push lands a build in prod ECR. All marked "revert before merge" — recording it as the gate.

## Nits

1. `_percentile_uint16`'s comment (around `__init__.py:585`) still says the sub-ULP percentile difference "vanishes in the uint8 rescale" — but the function now also feeds the **16-bit** rescale through `_uint16_clip_range` → `_normalize_plane`. The equivalence still holds (measured above); the comment's scope is just incomplete now.
2. `_normalize_plane`'s docstring (`:230`) still says "norm_img wraps this in dask.delayed so the montage assembly **streams**" — the one-plane-at-a-time framing the correction commit removed from `norm_img` and `_format_n_dim_ndarray`. This sibling was missed.
3. `norm_img`'s docstring (`:293`) no longer states *what* the normalization produces (I;16 int32); the functional summary moved down into `_normalize_plane`, so the entry point reads as mechanics-only.
4. The montage memory bound rests on dask's threaded scheduler releasing planes before the resize and not over-parallelizing the float path — observed in the benchmark, not enforced in code. It is **not** a regression (the deployed `da.percentile` path ran the same concurrency; #4974's sequential eager path is unmerged and OOMs anyway), but `img.compute(scheduler="synchronous")` or a `num_workers` cap would make the bound explicit, and channel/size combos past float 8ch @ 4000² are untested.
5. `test_norm_img_uint_path_bit_identical_to_float_reference` pins 4 hand-picked cases at seed 0. A property-style sweep over many sizes and distributions would lock the equivalence in CI rather than relying on fixed arrays (it passes such a sweep run externally).
6. `_uint16_clip_range` / `_percentile_uint16` are named for uint16 but also serve uint8 and byte-swapped `>u2` (Copilot's inline note too). The docstrings state the real contract; the names still read narrower. (Carried over from the prior notes.)
7. `da.from_delayed(..., shape=img.shape, dtype=np.int32)` (`:308`) trusts `_normalize_plane`'s return shape/dtype — a future change silently corrupts the montage graph instead of raising. The warning comment is there; nothing enforces it. (Carried over.)

## Worth a look (not blocking)

8. Since the win is purely release-before-resize, a simpler variant — keep eager assembly but explicitly free the montage collection before the resize in `handle_image` — might capture most of it without the `delayed` layer. But eager-assembly variants already regressed once (`166e1388`), and the lazy path measured a slightly lower assembly peak (480 vs 494 MB), so it would need its own benchmark before it's worth disturbing a working, measured approach.

Reviewed against 16d0e31d

sir-sigurd:
Thanks — addressed in `6c998484` (comments/tests only; output unchanged, so the final benchmark above still stands):

- **Nit 1** — `_percentile_uint16` comment now says the sub-ULP difference vanishes in both the uint8 *and* the 16-bit rescale (it feeds `_normalize_plane` now too).
- **Nit 2** — `_normalize_plane` docstring's leftover "montage assembly streams" → release-each-plane framing (the sibling the correction commit missed).
- **Nit 3** — `norm_img` docstring restores the functional summary (16-bit int32 / I;16) at the entry point.
- **Nit 5** — added `test_norm_img_uint_path_bit_identical_sweep`: 60 seeded random unsigned planes (uint8/uint16/`>u2`, varied size + distribution) asserted bit-identical to the float64 reference — locks your brute-force result in CI rather than relying on 4 fixed cases. (0 flips, as expected.)

Not changing, with reasons:
- **Nit 4 (explicit scheduler / `num_workers` cap)** — agreed it's *observed not enforced*, but the float A/B above confirms it's not a regression vs the deployed (also-threaded) path, and forcing `scheduler="synchronous"` would invalidate the just-completed final benchmark and could cost the plane-parallelism that's part of the ~15-25% speedup. Leaving as-is; if we ever want the bound *guaranteed* config-independent, it's a one-liner + its own benchmark — noted as a follow-up rather than folded in here.
- **Nit 6 (uint16 names vs uint8/`>u2`)** — docstrings state the real contract; renaming `_percentile_uint16` is pre-existing scope. Left.
- **Nit 7 (`from_delayed` shape/dtype trust)** — can't assert without computing the plane eagerly (which defeats the laziness this PR adds), so the warning comment is the most we can do.
- **Nit 8 (eager-assembly + explicit free)** — the eager-assembly variant already regressed once (`166e1388`, +145 MB) and lazy measured a lower assembly peak, so not worth disturbing a working, measured approach.

Merge gate acknowledged: I'll revert the `deploy-lambdas.yaml` scaffolding (thumbnail-only matrix, dropped GovCloud push, no `if: master` on the ECR job) immediately before merge.

PR #4982, opened 2026-06-15
Catalog: Tabulator tables in the Athena query selector

fiskus:
## Summary

Surfaces Tabulator tables alongside saved queries in the Athena **Query** tab, so users can preview a table without hand-writing SQL. Selecting a table runs a bounded `SELECT * … LIMIT 100` preview against the bucket's Tabulator catalog and redirects to the execution.

## Changes

- **Query selector**: the `QuerySelect` now supports grouped options with `ListSubheader`s ("Saved queries" / "Tables"); ungrouped usage (ElasticSearch) renders flat as before.
- **Tabulator tables source**: new `TabulatorTables` GraphQL query reads `bucketConfig.tabulatorTables`, exposed through the Athena model as `tables`. A null `bucketConfig` (not found / no access) is treated as "no tables".
  - NOTE: sourced from admin `bucketConfig` for now — the role-scoped `Bucket.tabulatorTables` field isn't ready yet.
- **Table preview**: `submitTablePreview` resolves the `<stack>-tabulator` catalog from the workgroup's catalog list and starts a `LIMIT 100` execution. Missing catalog / workgroup surface as a visible error.
- **Editor folding**: the Athena query editor is now collapsed by default (`M.Collapse`) with a toggle button; selecting a saved query opens it. A preview that fails to start keeps its error alert visible even while the editor is folded.

## Test plan

- [ ] `npm run test:only -- Athena QuerySelect` (catalog)
- [ ] `npx tsc --noEmit` (catalog)
- [ ] Manual: select a saved query → editor opens; select a Tabulator table → preview runs and results render; bucket without tables shows no "Tables" group.


fiskus:
Closing as superseded.

This prototype put Tabulator tables into the Athena query selector with an auto-run `SELECT * … LIMIT 100` preview and a collapsed editor. It was demoed in a standup; the team then pivoted to a tables-first presentation.

The shipped approach is #4995 (Bucket Overview v2): Tabulator tables surface on the Overview (gated on the `overviewV2` bucket preference) and as chips in the Queries → Athena tab (gated on the catalog-level `beta` setting). Key behavior change vs this prototype — no auto-run; the chip/Query button prefills a runnable query and the user presses Run (avoids surprise Athena cost and per-table fan-out), and it reuses the existing `useTabulatorTables`/`bucketConfig` config instead of a new GraphQL field.

PR #4984, opened 2026-06-15
thumbnail tests: add mosaic CZI fixture and gate large files before download

sir-sigurd:
# Description

Adds a real mosaic (whole-slide) acquisition to the thumbnail test set so the multi-tile-origin image category is covered end-to-end. `OverViewScan.czi` — the canonical mosaic from the AICS test resources — is added to the `images/pylibczirw` test-data package, and its golden thumbnail to `images/thumbs`; both package top-hashes in `test_thumbnail.py` are bumped to the new revisions.

**What this actually covers.** bioio-czi's pylibCZIrw backend stitches mosaic tiles in C++ and exposes a single pre-stitched greyscale plane with no `M` dimension (`mosaic_tile_dims → None`; `_get_stitched_dask_mosaic` is a passthrough to `xarray_dask_data`). So the lambda takes the same single-plane path as the existing `c1_gray*` fixtures — not a distinct mosaic code path. The fixture's value is the **pinned golden**:

- guards against an upstream mosaic-stitching regression that would silently break thumbnails for the whole whole-slide/tiled-acquisition category, and
- guards against a future bioio that re-exposes mosaic tiles — `_format_n_dim_ndarray` would silently mis-slice such an array under its 5-D TCZYX assumption, and the golden's dims/pixel assertions would catch the divergence.

`OverViewScan.czi` is 82 MB, so it runs only under `pytest --large-files`.

**Opportunistic fix (surfaced in review).** Adding a fixture over the size gate exposed that `test_handle_image` called `Package.install` (which downloads the file) *before* checking `src_entry.size`, so large fixtures were downloaded and then skipped. This was already happening for 8 pre-existing fixtures — the gate dates to #4612; `OverViewScan.czi` just made it visible. Reordered to `browse` (manifest only) → size-gate → `install`, so the 9 currently-gated large files are no longer fetched on a normal `pytest` run.

Spun off from #4974.

# TODO

- [x] Unit tests

<h3>Greptile Summary</h3>

This PR adds an `OverViewScan.czi` mosaic fixture to the CZI thumbnail test suite and fixes a pre-existing bug where large files were downloaded before the size gate was checked.

- **Gate-ordering fix**: `browse()` (manifest only) is now called before the size check, so the 9 currently-gated large fixtures are no longer fetched on a normal `pytest` run; `install()` only fires when the file passes the gate.
- **New mosaic fixture**: `OverViewScan.czi` (82 MB) is added behind the `--large-files` flag with a clear comment explaining that bioio-czi exposes this as a single pre-stitched greyscale plane; package top-hashes for `CZI_PKG` and `THUMBS_PKG` are bumped accordingly.

Safe to merge — the gate reorder is a straightforward correctness fix with no logic changes to the thumbnail path itself, and the new fixture is properly guarded behind --large-files.

The only functional change is moving browse() + size-check before install(), which closes a gap where large files were downloaded and then immediately skipped. The new OverViewScan.czi fixture exercises a well-understood code path and is gated so it does not slow normal runs. No production code is touched.

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/tests/test_thumbnail.py | Gate reordered to browse→size-check→install so large files are no longer downloaded unnecessarily; new OverViewScan.czi mosaic fixture added and both package hashes bumped. |

</details>

<sub>Reviews (3): Last reviewed commit: ["thumbnail tests: trim the OverViewScan f..."](https://github.com/quiltdata/quilt/commit/b577949e01a2fa03f20232f7c643d8f139c9d579) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37649016)</sub>

### DIFF
--- a/lambdas/thumbnail/tests/test_thumbnail.py
+++ b/lambdas/thumbnail/tests/test_thumbnail.py
@@ -946,7 +946,7 @@ def test_handle_image(pytestconfig, pkg_ref, lk):
         top_hash=top_hash,
     )
     src_entry = src_pkg[lk]
-    # Gate on size before install() downloads the file (browse() read only the manifest).
+    # Gate on size before install() downloads the file (browse() reads only the manifest).
     if not pytestconfig.getoption("large_files") and src_entry.size > 20 * 1024 * 1024:
         pytest.skip("Skipping large file test; use --large-files to enable")
     quilt3.Package.install(
--- a/lambdas/thumbnail/tests/test_thumbnail.py
+++ b/lambdas/thumbnail/tests/test_thumbnail.py
@@ -787,8 +787,8 @@ def test_norm_img_leaves_color_planes_unchanged():
 TEST_DATA_REGISTRY = "s3://quilt-test-public-data"
 TIFF_PKG = "images/bioio-tifffile", "5fa99558a167d6430defbfa4033808c7e7004b847e94a213292c2c776ef43ac5"
 OME_TIFF_PKG = "images/bioio-ome-tiff", "6dbddd093e0a92cfc1cc5957ad7a7177ba98a0fee5d99ffaea58e30b7c46e182"
-CZI_PKG = "images/pylibczirw", "552c9290ffa24738a578c494b7fc9f95cc03e3d12d701bc0bd944f5c1c558b2c"
-THUMBS_PKG = "images/thumbs", "5ae720b35c5abc296ad5e708d3dea402bc4269c6e6f1920aa48cbc4be21ba443"
+CZI_PKG = "images/pylibczirw", "617551541881add8011f55de0c3936a90fc2188a40b6ef47c7e6ab20c3d2c8bf"
+THUMBS_PKG = "images/thumbs", "6244534f2034cf1166107a8da3915cb469761cf342d85eb653623d9f92390474"
 SIZE = (1024, 768)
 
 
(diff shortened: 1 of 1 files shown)


sir-sigurd:
@greptileai please review

sir-sigurd (commented):

Verified clean — no action needed here:

- The two package top-hash bumps are purely additive: the CZI package gains only `OverViewScan.czi` and the thumbs package only its `.png` golden; no existing entry's content changed or was dropped, and both `latest` tags advanced to match.
- The fixture passes bit-for-bit under `--large-files`, so the golden corresponds to current code.
- Security (a pinned content-addressed top-hash can't be substituted), scope vs. description, the Python (tuple shape/unpacking, the hashes, the size gate), and the comment's two factual claims — pylibCZIrw pre-stitches to a single plane, and the lambda has no mosaic-specific branch — all hold.

Three minor notes, all optional:

1. `--large-files-gated (82 MB)` reads as if `--large-files-gated` were a pytest option; the actual flag is `--large-files`. Reword as prose, e.g. "gated behind `--large-files` (82 MB)".
2. The parenthetical "`_format_n_dim_ndarray`'s 5-D TCZYX assumption would not handle" a re-exposed-M array is slightly imprecise: the function wouldn't raise on a 6-D array, it would silently slice `[0, 0, 0, :, :]` and produce a wrong plane. The golden still catches the divergence (the `dims`/pixel assertions fail), so the guard works as intended — the wording just reads as if the code itself rejects the case. Optional tightening.
3. Pre-existing, not this PR's job: there's no recorded recipe for regenerating the `images/thumbs` goldens and re-pushing the test-data packages when bioio changes its output — it lives only in the package commits. Every existing fixture shares this; this one adds another. A short README or script under `lambdas/thumbnail/tests/` would help the next bioio bump.

Reviewed against 12d8cfb6

sir-sigurd:
Addressed review feedback in d33c4cc6:

- **Gate ordering (Copilot):** reordered `test_handle_image` to `browse` + check `src_entry.size` **before** `Package.install`, so gated fixtures aren't downloaded just to skip. Verified `OverViewScan.czi` now skips in ~1s without `--large-files` (previously it pulled the full 82 MB first), and still passes with `--large-files`. This was already happening for 8 pre-existing large fixtures (the gate dates to #4612); `OverViewScan.czi` just surfaced it — the reorder now spares all 9 currently-gated files the download.
- **Comment wording (self-review):** named the actual flag (`--large-files`, not "--large-files-gated"), and clarified that a re-exposed mosaic dimension would be *silently mis-sliced* (caught by the golden's dims/pixel assertions) rather than rejected outright.

Out of scope (pre-existing, shared by every fixture): there's no recorded recipe for regenerating the `images/thumbs` goldens / re-pushing the test-data packages on a bioio bump — it lives only in the package commits. Can file a follow-up to add a short README or script under `lambdas/thumbnail/tests/`.

sir-sigurd:
Follow-up self-review pass — likely the last round.

The gate-before-download reorder now runs the size check against the manifest (`browse()`) before `install()` downloads; CI covers that path via the small fixtures. The earlier comment notes are resolved in the new commits, and the reorder spares every currently-gated large fixture the wasted download, not just the new one.

Reviewed against b577949e

sir-sigurd:
@greptileai please re-review the latest commits (gate-ordering fix and comment trims since the initial review).

PR #4992, opened 2026-06-16
thumbnail: render BGR color CZIs correctly (bump bioio-czi to 2.7)

sir-sigurd:
# Description

`bioio-czi` 2.7.0 made BGR-pixel-type CZIs (`Bgr24`/`Bgr48`/`Bgr96Float`) readable, but it returns their color samples in raw on-disk **BGR** order without converting to RGB. The thumbnail lambda's color path passed those samples through unchanged, so a color CZI rendered with red and blue swapped. This bumps `bioio-czi` and converts BGR→RGB so color CZIs preview with correct colors.

## Changes

- Bump `bioio-czi` `~=2.4` → `~=2.7` (lockfile 2.7.2; `bioio`/`bioio-base` relock 3.0→3.3, within their existing `~=3.0` floors). `pylibczirw` stays 5.1.1.
- Reverse the sample axis (BGR→RGB) for CZI color output in `format_aicsimage_to_prepped`, gated on the CZI reader plus a Samples axis (CZI color is always 3-sample `Bgr*`).
- Un-xfail `c1_bgr48.czi` and `rgb-image.czi` (readable since 2.7.0); their goldens were added to the `images/thumbs` test package. `c1_bgr24.czi` (CZI metadata error) and `c1_bgr96float.czi` (pixel type unimplemented in pylibczirw) remain xfail.
- Add independent channel-order oracles (caught even though the golden is self-generated): `test_handle_image_bgr_czi_channel_order` asserts the real wooden-dice photo renders warm-toned (R > B), and `test_handle_image_multichannel_bgr_czi_channel_order` covers the multi-channel montage branch via a synthetic 2-channel `Bgr24` fixture (`tests/data/multichannel-bgr.czi`).

## Validation

- Full thumbnail suite green; the 7 existing greyscale-CZI and the non-CZI RGB goldens are byte-for-byte unchanged (no regression from the bump).
- `rgb-image.czi` output cross-checked against an independent, non-libCZI parser (Gohlke's `czifile`): same dice, same colors, same orientation/arrangement.

## Impact

Production rendering change for color CZIs only — previously these failed to preview. Ships via the normal deployment route.

# TODO

- [x] Unit tests
- [x] Security: Confirm that this change meets security best practices and does not violate the security model
- [x] Changelog entry

<h3>Greptile Summary</h3>

[...]

PR #4994, opened 2026-06-16
thumbnail tests: golden-regeneration-proof channel-order oracle for the non-CZI color path

sir-sigurd:
# Description

The thumbnail lambda's color goldens (`test_generate_thumbnail`, `test_handle_image`) are generated by the lambda itself and committed as the expected output. A byte-golden does catch a channel-swap as a *failure*, but the tempting fix — regenerating the golden — would then enshrine the swap silently. That is the exact path by which the BGR R/B swap nearly shipped (#4992): the relevant color goldens didn't exist (they were `xfail`), and the bug was caught only by a human eyeballing the output.

#4992 closed that gap for the **CZI** path with an independent channel-order oracle (`test_handle_image_bgr_czi_channel_order`). This PR adds the equivalent guard for the **non-CZI** color path, which had none:

- `penguin.jpg` — bioio-imageio 8-bit RGB (PIL is a genuinely different decoder here)
- `sat_rgb.tiff` — bioio-tifffile 8-bit RGBA
- `float-rgb.tiff` — bioio-tifffile float16, through `_rescale_float_to_uint8`

`test_color_path_channel_order` renders each fixture through `handle_image`, independently decodes the source, coarsens both into a grid of block means, and asserts that on clearly-colored blocks the thumbnail's regional R-vs-B lean matches the source. An R/B swap flips every sign → agreement collapses to ~0% (validated by mutation). Because the assertion is against an independent decode, not a stored golden, a golden regeneration can't fake it; the synthetic-array rescale unit tests can't catch a swap at all (their min/max checks are invariant under channel permutation).

Scope notes:
- Test-only; no behavior change. No CHANGELOG entry (not end-user-significant).
- Orientation was deliberately left out: the single-plane color path has no transposition logic, so an orientation regression there is implausible and would be caught by the byte goldens anyway; the montage layout path is covered separately (`test_handle_image_multichannel_bgr_czi_channel_order`).
- For the TIFF fixtures, tifffile is what bioio-tifffile wraps, so there the oracle guards the lambda's own channel handling rather than the decode; `penguin.jpg` (PIL vs bioio-imageio) is the genuinely cross-library case.

I also independently audited every existing color golden (local + the S3 `images/thumbs` package) against an independent decode — all render with correct channel order and orientation; no enshrined bug was found. This test makes that conclusion durable against future regenerations.

# TODO

- [x] Unit tests

<h3>Greptile Summary</h3>

This PR adds `test_color_path_channel_order`, a parametrized channel-order oracle for the three non-CZI color paths in the thumbnail lambda (JPEG via bioio-imageio, 8-bit RGBA TIFF, and float16 TIFF). It is the non-CZI sibling of the existing `test_handle_image_bgr_czi_channel_order`, closing the gap that nearly let a BGR R/B swap ship undetected (#4992).

- Each fixture is independently decoded (PIL for JPEG, raw `tifffile` for TIFFs), coarsened into a 12×12 grid of block means, and compared against the lambda's own thumbnail output by sign of the R−B difference. Because the comparison is against an independent decode rather than a stored golden, golden regeneration cannot silently enshrine a swap.
- Test-only change; all three fixture files already exist on disk; no production code is modified.

Pure test addition with no changes to production code; all fixture files are pre-existing and already covered by the byte-golden suite.

The new test logic is sound: the coarse-grid approach tolerates resize and scale differences between decoder outputs, the sign-only comparison is invariant to the float-vs-uint8 scale mismatch in the float fixture, the `colored.sum() >= 5` pre-condition provides a clear sanity guard, and the 90% agreement threshold definitively distinguishes correct channel order from a full R/B swap. All three fixture files exist in the data directory and are independently verified by the existing golden tests. No production paths are touched.

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/tests/test_thumbnail.py | Adds `test_color_path_channel_order` (parametrized over 3 local fixtures), helper `_coarse_means`, and a module-level `_COLOR_ORACLE_FIXTURES` list. Test-only; no production code changed. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["fixture file on disk\n(penguin.jpg / sat_rgb.tiff / float-rgb.tiff)"] --> B["Independent reference decode\n(PIL / tifffile / tifffile.astype float64)"]
    A --> C["handle_image()\n→ PNG bytes"]
    C --> D["Image.open().convert('RGB')\n→ uint8 HxWx3"]
    B --> E["_coarse_means(ref)\n12×12×3 block means"]
    D --> F["_coarse_means(out)\n12×12×3 block means"]
    E --> G["rb_ref = R_mean − B_mean\ncolored = |rb_ref| > 5% dynamic range"]
    F --> H["rb_out = R_mean − B_mean"]
    G --> I["assert colored.sum() ≥ 5"]
    G --> J["agree = sign(rb_ref) == sign(rb_out)\nover colored blocks"]
    H --> J
    J --> K["assert agree > 0.90\n(R/B swap → agree ≈ 0%)"]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A["fixture file on disk\n(penguin.jpg / sat_rgb.tiff / float-rgb.tiff)"] --> B["Independent reference decode\n(PIL / tifffile / tifffile.astype float64)"]
    A --> C["handle_image()\n→ PNG bytes"]
    C --> D["Image.open().convert('RGB')\n→ uint8 HxWx3"]
    B --> E["_coarse_means(ref)\n12×12×3 block means"]
    D --> F["_coarse_means(out)\n12×12×3 block means"]
    E --> G["rb_ref = R_mean − B_mean\ncolored = |rb_ref| > 5% dynamic range"]
    F --> H["rb_out = R_mean − B_mean"]
    G --> I["assert colored.sum() ≥ 5"]
    G --> J["agree = sign(rb_ref) == sign(rb_out)\nover colored blocks"]
    H --> J
    J --> K["assert agree > 0.90\n(R/B swap → agree ≈ 0%)"]
```

</a>

<sub>Reviews (2): Last reviewed commit: ["thumbnail tests: add golden-regeneration..."](https://github.com/quiltdata/quilt/commit/12ede8b3427fb233b069f6895467c903afdd1ff1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37925091)</sub>

### DIFF
--- a/lambdas/thumbnail/tests/test_thumbnail.py
+++ b/lambdas/thumbnail/tests/test_thumbnail.py
@@ -12,6 +12,7 @@ import dask.array as da
 import numpy as np
 import pytest
 import responses
+import tifffile
 from bioio import BioImage
 from PIL import Image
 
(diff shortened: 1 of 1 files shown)


sir-sigurd:
@greptileai

sir-sigurd (commented):

The independent channel-order check behaves as intended: it passes on all three fixtures, and forcing an R/B swap on the rendered output collapses block-agreement from ~100% to 0% on every fixture — so it fails when the bug it guards is present. The float path's sign-preservation argument (one joint `lo`/`hi` → positive-slope affine → preserves the sign of R−B) is sound, and leaving green unguarded is appropriate because the real swap mode (BGR reversal) is R↔B.

Checked: blind-spot pass, scope vs. description, Python/numpy correctness, the check's validity (mutation-confirmed above), flakiness/determinism, consistency with the CZI sibling tests, and simplification/docs. No blockers — the items below are all polish.

### Nits

**`_coarse_means` assumes a ≥12px, exactly-3D image.** For a fixture narrower or shorter than the 12×12 grid, `np.linspace(0, h, grid + 1, dtype=int)` produces a zero-width block, and `mean(0)` over it returns NaN with a `RuntimeWarning: Mean of empty slice`. The suite's `filterwarnings` only promotes the Pillow "Saving I mode" warning to an error, so that one passes through as noise rather than failing the run. A 2D or 4D decode would hit the `h, w, c = rgb.shape` unpack with an opaque `ValueError`. All three current fixtures are large and 3-channel, so it's latent — but an `assert rgb.ndim == 3` (or a one-line note on the shape the helper expects) would make the contract explicit for the next fixture.

**(reshape) Test name drops the oracle-family prefix.** The two sibling checks are `test_handle_image_bgr_czi_channel_order` and `test_handle_image_multichannel_bgr_czi_channel_order`; this one is `test_color_path_channel_order`. It does call `handle_image`, so `test_handle_image_color_channel_order` would be accurate and keep the three greppable as one family — which the docstring already treats them as.

### Notes (no code change)

**Fixture list vs. the color goldens.** The three fixtures here are the same color inputs that have byte goldens in `test_generate_thumbnail`, and the two lists aren't linked. Duplicating them is the right call (explicit test data reads better than deriving one list from the other), so no change is needed — just flagging that a future color reader added with a golden but no entry here would silently skip the channel-order check. A one-line cross-reference comment would pre-empt that if it's worth doing.

**The check asserts R-vs-B only.** Green is never compared, so a swap touching only green wouldn't be caught. The actual bug this guards is R↔B and is caught; this is just to note that "channel-order" is really R/B-order — a half-sentence to that effect would set expectations.

**Description audit claim.** The "independently audited every existing color golden (local + S3)" note is useful context but isn't reproducible from the diff — the durable guarantee is what this committed test checks (the three fixtures).

Reviewed against 12ede8b3

PR #4998, opened 2026-06-16
thumbnail: contrast-stretch wide/signed integer images instead of failing

sir-sigurd:
# Description

`generate_thumbnail` only rescaled `float` and `uint16` to uint8; every other integer fell through to PIL unchanged. As a result, signed and wider-than-16-bit integer images either failed with **HTTP 500** (all color, and all 64-bit — PIL has no matching mode) or rendered as a **clamped/dark 16-bit** `I` image (`int8`/`int16`/`int32`, `uint32` greyscale). Float of any width was already fine.

This adds `_rescale_int_to_uint8` (float64 contrast-stretch, sharing `_finite_clip_range` / `_saturate_to_uint8` with the float path) and routes `uint32`/`uint64`, `int8`/`int16`/`int64`, and 3-D `int32` **color** through it. `_alpha_to_uint8` is generalized to scale wide-unsigned alpha by its high byte (`>>24`, `>>56`) rather than only `uint16`'s `>>8`.

**The normalized montage passes through via an explicit flag, not a value scan.** `norm_img`'s greyscale montage / Z-projection is already contrast-stretched into the 16-bit range; the producer (`format_aicsimage_to_prepped` → `handle_image`) marks it `normalized` and `generate_thumbnail` passes it straight to a 16-bit `I;16` PNG without inferring anything from the pixels — so a normalizer output-range change can't silently re-route it, and the montage isn't scanned (`min`/`max`) on the hot path. Raw arrays (including raw `int32`) carry `normalized=False` and are contrast-stretched by dtype, so there's no residual int32 limitation.

Verified the full test suite (including the real microscopy/CZI fixtures from `s3://quilt-test-public-data`) is unchanged; the normalized montage's 16-bit passthrough is pinned byte-identical.

# TODO

- [x] Unit tests
- [x] Security: Confirm that this change meets security best practices and does not violate the security model — no new input surface; same image-decode path
- [ ] Open and Embed: N/A — thumbnail-lambda-only change
- [ ] Documentation
    - Preview.md's "Integer images deeper than 16 bits…" warning (docs/Catalog/Preview.md) becomes stale, but the lambda fix isn't live until the thumbnail lambda is redeployed — the doc is updated with the release, same process as the BGR-CZI fix (#4992)
- [x] Changelog entry

<h3>Greptile Summary</h3>

This PR fixes `generate_thumbnail` for signed and wide-unsigned integer images (`uint32`/`uint64`, `int8`/`int16`/`int64`, and 3-D `int32` color) that previously either crashed with HTTP 500 or rendered as clamped/dark 16-bit output. It adds a `_rescale_int_to_uint8` contrast-stretch function (sharing `_finite_clip_range` and `_saturate_to_uint8` with the float path), threads a `normalized` flag from `_format_n_dim_ndarray` → `format_aicsimage_to_prepped` → `generate_thumbnail` so that already-normalized montage data is never re-stretched, and generalizes `_alpha_to_uint8` to correctly scale wide-unsigned and signed integer alpha channels.

- **New `_rescale_int_to_uint8`**: uses float64 for types wider than 2 bytes (avoiding float32's 24-bit mantissa losing low-contrast bands at high offsets) and delegates range computation to the shared `_finite_clip_range`.
- **`normalized` flag propagation**: cleanly separates the \"greyscale montage already in 16-bit range\" case (pass-through) from raw integer image data (contrast-stretch), driven by the producer rather than value inspection.
- **`_alpha_to_uint8` generalization**: signed alpha scales the positive range to full opacity via float arithmetic (avoiding the arithmetic right-shift that gave ~50% transparency for max-value int32), and unsigned alpha uses `>> (8*(itemsize-1))` for uint32/uint64.

Safe to merge — the change adds a new contrast-stretch path for dtypes that previously failed outright or rendered incorrectly, and the normalized flag cleanly separates montage pass-through from raw data rescaling without any value inspection.

The dispatch logic is correctly ordered (uint16 is caught before the new wide-integer branch, uint8 falls through to no-op, normalized montages are protected by an explicit flag from the producer). The float32/float64 threshold at itemsize≤2 matches the mantissa limits precisely. Signed-integer alpha is now scaled via float arithmetic rather than arithmetic right-shift, giving the correct full-opacity mapping. The test suite is thorough, covering empty arrays, constant images, sparse label masks, outlier pixels, high-offset low-contrast bands, and the normalized passthrough. No existing behavior changes.

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py | Core implementation: adds _rescale_int_to_uint8 for wide/signed integer contrast-stretch, propagates the normalized flag through _format_n_dim_ndarray → format_aicsimage_to_prepped → generate_thumbnail, and generalizes _alpha_to_uint8. Logic is correct — float64 threshold for types >2 bytes, proper signed-alpha scaling via float, unsigned shift by (itemsize-1)*8. Dispatch ordering is safe: uint16 is caught before the new iu branch. |
| lambdas/thumbnail/tests/test_thumbnail.py | Comprehensive test coverage added: unit tests for _rescale_int_to_uint8 (wide-unsigned, signed, high-offset low-contrast, constant, empty, sparse, outlier clipping), integration tests for wide-int greyscale/color/RGBA in generate_thumbnail, normalized passthrough, raw int32 stretch, and _alpha_to_uint8 for uint32 and all signed types. |
| lambdas/thumbnail/tests/data/rgb-uint32.tiff | New binary fixture for the uint32 RGB oracle test in _COLOR_ORACLE_FIXTURES; no code concerns. |
| lambdas/thumbnail/CHANGELOG.md | Changelog entry added for the fix; accurate description of what was broken and what is now fixed. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A3["_format_n_dim_ndarray"] -->|"array, not has_samples"| A2["format_aicsimage_to_prepped"]
    A2 -->|"array, normalized"| A["handle_image"]
    A -->|"arr.compute() + normalized flag"| B["generate_thumbnail"]

    B --> C{"normalized=True?"}
    C -->|Yes| D["rescale = None\npass-through to mode-I"]
    C -->|No| E{"dtype dispatch"}
    E -->|"kind == f"| F["_rescale_float_to_uint8"]
    E -->|"kind == u, itemsize == 2"| G["_rescale_uint16_to_uint8"]
    E -->|"kind in iu, != uint8"| H["_rescale_int_to_uint8 NEW"]
    E -->|"uint8"| I["rescale = None"]

    H --> J{"RGBA shape?"}
    F --> J
    G --> J
    J -->|Yes| K["rescale color channels\n+ _alpha_to_uint8"]
    J -->|No| L["rescale full array"]
    K --> M["Image.fromarray"]
    L --> M
    D --> M
    I --> M
    M --> N["img.thumbnail"]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A3["_format_n_dim_ndarray"] -->|"array, not has_samples"| A2["format_aicsimage_to_prepped"]
    A2 -->|"array, normalized"| A["handle_image"]
    A -->|"arr.compute() + normalized flag"| B["generate_thumbnail"]

    B --> C{"normalized=True?"}
    C -->|Yes| D["rescale = None\npass-through to mode-I"]
    C -->|No| E{"dtype dispatch"}
    E -->|"kind == f"| F["_rescale_float_to_uint8"]
    E -->|"kind == u, itemsize == 2"| G["_rescale_uint16_to_uint8"]
    E -->|"kind in iu, != uint8"| H["_rescale_int_to_uint8 NEW"]
    E -->|"uint8"| I["rescale = None"]

    H --> J{"RGBA shape?"}
    F --> J
    G --> J
    J -->|Yes| K["rescale color channels\n+ _alpha_to_uint8"]
    J -->|No| L["rescale full array"]
    K --> M["Image.fromarray"]
    L --> M
    D --> M
    I --> M
    M --> N["img.thumbnail"]
```

</a>

<sub>Reviews (3): Last reviewed commit: ["thumbnail: thread an explicit &quot;normalize..."](https://github.com/quiltdata/quilt/commit/bee853eda6933992b335bf76c349b7468a574e2f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37970362)</sub>


sir-sigurd:
@greptileai please review

sir-sigurd (commented):

Checked scope coherence (single goal; the alpha generalization is a required consequence of widening the dtype routing, not a ride-along; the fix is exercised, not inert), behavior preservation across reader paths, whether the int64/uint64/wide-alpha generality is justified (it is — one uniform rule, not speculative knobs), security (no new surface; the float64-copy memory cost is equivalent to the pre-existing float path), numerical correctness of the contrast stretch (verified clean — float64 precision, operation order, rounding/saturation, and every asserted test value reproduce by hand under percentile ranging), edge/failure modes, and peak memory. Nothing here blocks merge; the items below are design points and small notes for follow-up.

### Worth a look

**P2 — the 2-D int32 "indistinguishable" rationale is overstated** (`src/t4_lambda_thumbnail/__init__.py:706-712`)

The normalized montage/Z-projection is in fact *structurally* distinguishable from a raw int32 plane: `_normalize_plane` always emits values in `[0, 65535]` with a 5px zero-padded border, whereas a raw int32 label/segmentation mask (a real input class for this lambda) can be negative or `> 65535` and renders as a near-uniform white/black blob after the `I → I;16` clamp. This is not a regression — 2-D int32 already passed through before this change — and deferring the fix is reasonable here because it protects the byte-identical montage output. But the comment claims more than is true: prefer "hard to distinguish robustly" over "indistinguishable," and consider a follow-up that routes out-of-`[0, 65536)` values to the stretch path.

*Alternatives: could a range check (values outside `[0, 65536)` → stretch) separate raw masks from montage output without churning the montage goldens?*

### Smaller notes (none blocking)

**P3 (reshape) — signed-integer alpha is mishandled** (`__init__.py:689`)
`alpha >> (8 * (itemsize - 1))` is an arithmetic (sign-extending) shift. For signed RGBA the positive-max value maps to 127 (half opacity), and int8 (`>> 0`) is identity so negatives wrap. The widened dtype routing now admits signed RGBA (int16/32/64), but the comment only describes the unsigned convention and no test covers it. Not a regression (signed RGBA returned HTTP 500 before) and no known reader emits it — the color channels still render, only the alpha is imperfect. Either special-case `kind == "i"` or document signed RGBA as out of scope and pin the chosen behavior with a test.

**P3 (reshape) — big-endian int32 routing is byte-order-asymmetric vs. its sibling check** (`__init__.py:706`, comment at `:700-701`)
`arr.dtype == np.int32` is byte-order-strict (`np.dtype('>i4') == np.int32` is `False`), so a raw big-endian 2-D int32 takes the 8-bit stretch while the uint16 check (`kind == "u" and itemsize == 2`) is deliberately byte-order-agnostic — and the header comment advertises that parity as the design principle. Harmless today (`norm_img` only ever emits native-order int32, and a raw big-endian int32 actually renders *better* as an 8-bit stretch). The keeper: tighten the comment to note the `== np.int32` check is intentionally exact — it must match `norm_img`'s native output, not all int32 — so a future maintainer doesn't "fix" it to `kind`/`itemsize`.

**P3 (reshape) — full float64 working copy; optional int8/int16 float32 win** (`__init__.py:661`)
`out = arr.astype(np.float64)` is a full 8-bytes/pixel copy (≈9× the input for int8, ≈2.3× for int32/uint32), in a file otherwise architected to avoid full-plane float copies via LUT/histogram. This is largely unavoidable for ranges wider than 16 bits (no LUT past 65536 bins) and is consistent with the existing float path — an accepted tradeoff. The one cheap win: int8/int16 are exactly representable in float32, so a `float32 if itemsize <= 2` selector would halve their working copy. Note the guard must be `<= 2`, not the float path's `<= 4` — int32 genuinely needs float64 (float32 collapses a high-offset band from 256 levels to 3). Weigh the saving (rarest input) against carrying a second, slightly different dtype guard.

**P3 (reshape) — `np.isfinite` is dead work on integer input** (`__init__.py:153`, `_finite_clip_range`)
Integer arrays are always finite (the function's own comment says so), yet `np.isfinite(arr)` still allocates a full bool mask and scans it. A `kind in "iu"` early-out would skip it, as `_uint16_clip_range` already avoids the equivalent. Negligible next to the float64 copy; micro-note.

**P3 — 3-D integer stacks with `shape[2] ∉ {3, 4}`** (`__init__.py:713-727`)
These now do rescale work and then fail at `Image.fromarray` ("Cannot handle this data type"). Pre-existing for uint8 (already failed); this change widens the dtype set reaching it but the outcome is unchanged (a 500 either way). Low likelihood from these readers — flagging as a known edge, not a new defect.

Reviewed against 996ad550

sir-sigurd (commented):
Notes from a second AI-assisted self-review pass on the rework — likely the last round unless the design question below changes the shape of things.

The two new commits resolve the substantive items from the first pass: signed alpha now maps the max positive value to 255 (negatives clamp transparent) with a test, the 2-D int32 passthrough is range-gated so out-of-range raw int32 stretches, `_finite_clip_range` skips the finite mask for integers, and int8/int16 use float32. The earlier Greptile and Copilot comments all sit on those superseded commits and are resolved at the head. Re-checked behavior preservation / API contract, the abstraction surface, an adversarial pass, edge cases, peak memory, security, and the contrast-stretch numerics: montage byte-identity still holds (`_normalize_plane` always emits `[0, 65535]`, so the range gate always passes it through unchanged), uint16 alpha is byte-identical, and the signed-alpha and float32/float64 selector are numerically correct (verified including int64 max → 255). No blockers — the open items are a stale description, one design question, and two small notes. I'm holding the line-level polish pass this round: the two code items below would both touch the same range-gate line, so nits there would just go stale.

### Worth a look

**P2 — the description no longer matches the shipped behavior** (the "2-D int32 stays a passthrough…" paragraph)
It still says a raw 2-D int32 greyscale "shares that path and stays clamped-16-bit: a documented, accepted limitation" and is "indistinguishable from a raw int32 plane." The code now range-gates the passthrough (`arr.min() >= 0 and arr.max() <= 65535`), so a raw 2-D int32 outside `[0, 65535]` (negative, or a label mask > 65535) is contrast-stretched, not clamped — matching the in-code comment and `test_generate_thumbnail_out_of_range_int32_greyscale_is_stretched`. Only an in-range 2-D int32 still passes through, which is the residual limitation. Update the paragraph, and re-check the `docs/Catalog/Preview.md` TODO against the narrowed limitation.

**P2 (reshape) — the passthrough decision re-derives the montage by scanning pixel values** (`__init__.py:728`)
`arr.ndim == 2 and arr.dtype == np.int32 and arr.min() >= 0 and arr.max() <= 65535` infers "this is the normalized montage, pass it through" from the values at runtime. Two consequences: (1) the montage's byte-identical output now rests on the invariant that `_normalize_plane` always emits `[0, 65535]`, enforced only by a comment — a future change to the normalizer's output range would silently send the montage into the stretch path and churn every montage golden, with no localized failure to catch it; (2) `arr.min()` and `arr.max()` are two full-array passes that now run on the lambda's primary output, the montage (a quick measurement put them at ~10% on top of the fromarray-plus-resize step; no extra memory). Both fall out of re-deriving from the data a fact the producer already knows.
*Alternatives: could `_normalize_plane`/`handle_image` carry the "already normalized" fact (a sentinel or an explicit flag) so the passthrough keys on provenance instead of a value-range coincidence?*

### Smaller notes (none blocking)

**P3 (reshape) — empty 2-D int32 isn't guarded at the gate** (`__init__.py:728`)
`arr.min()` on a zero-size array raises `ValueError`, where every other path in the file guards emptiness (`_finite_clip_range`, the `_rescale_*` helpers, `_normalize_plane`). A zero-size plane can't realistically reach here — the montage always pads 5px and single planes go through `_normalize_plane` — so this is a consistency gap rather than a reachable bug; a one-token `arr.size and …` prefix closes it and keeps the int32 paths uniform on empty input.

**P3 (reshape) — the dedicated signed-alpha branch handles an input "no reader emits"** (`__init__.py:692`)
The branch is correct and reachable (signed RGBA now routes through the rescale path), and it directly answers the first round's note. But by its own comment no reader produces signed alpha, so a dedicated float32-scale branch plus a 4-dtype test is calibrated for a format that doesn't exist — worth deciding whether that's proportionate or whether documenting signed RGBA as out of scope would do. Defensible as-is.
*Alternatives: could signed alpha fold into the float branch (cast → scale by `iinfo.max` → saturate), leaving one non-unsigned path, or is the integer-range distinction necessary here?*

Reviewed against 710236e8

sir-sigurd:
@greptileai please re-review — the int32 passthrough is now driven by an explicit `normalized` flag from the producer (no value scan), and signed-integer alpha is handled. Latest commit: bee853ed.

sir-sigurd (commented):
Notes from a third AI-assisted self-review pass — this looks like the last round; what's left is four cosmetic items plus one optional guard.

The provenance-flag rework resolves last round's main point cleanly: the montage is now identified by an explicit `normalized` flag threaded from the producer rather than by scanning pixel values, which drops the two full-array passes on the montage path, removes the empty-array crash, and also fixes in-range low-contrast raw int32 (now stretched rather than passed through dark). Re-checked the flag's correctness on every reader path (including that the middle-timepoint reslice preserves the S axis, so `not has_samples` stays a faithful proxy for "norm_img normalized this"), behavior preservation / API contract, edge cases, security, and a polish pass. The montage's 16-bit byte-identical output still holds (`test_norm_img_path_saves_16bit_png`, via `handle_image`), no other callers are affected by the signature/tuple-return changes, and the full suite passes. No blockers; the items below are all minor.

### Smaller notes

**P3 — one residual stale phrase in the description.** The "2-D int32 accepted limitation" paragraph is gone, but the closing "Verified…" line still says "the montage `int32`-2D passthrough is pinned byte-identical." The passthrough is keyed on the flag now, not on being int32-2D (a raw int32-2D plane is always stretched). Reads truer as "the normalized montage 16-bit passthrough is pinned byte-identical."

**P3 — `generate_thumbnail` trusts `normalized=True` without re-checking the array** (`__init__.py:~732`). `normalized=True` sets `rescale = None` and the array goes straight to `Image.fromarray`. It's safe today — the only producer, `norm_img`, always emits a 2-D int32 plane when `not has_samples` — but the guarantee moved from self-validating data to a flag spread across four functions. A future `norm_img` change, or a new caller that sets the flag, would reach `fromarray` unvalidated (a 500 or a wrong mode) with nothing localizing the failure.
*Alternatives: could a cheap `arr.ndim == 2 and arr.dtype == np.int32` assert under `normalized` fail loud if the producer contract ever drifts?*

**P3 (reshape) — a test's parametrization is organized around the removed value-gate** (`test_generate_thumbnail_raw_int32_greyscale_is_stretched`). Its three cases (`in-range-low`, `above-65535`, `negative`) and their comments ("would render dark / would clamp if passed through") describe the old value-gate's failure modes; under the flag design all three take the identical path with identical assertions, and dtype breadth is covered elsewhere. Collapsing to one case (the `negative` one doubles as a signed-mask example) or relabeling around "raw int32 always stretches regardless of range" drops the leftover.

**P3 (reshape) — comment lists int32 oddly** (`__init__.py:~739`). "Wider/signed integers… (uint32/64, int8/16/64, int32)" calls out int32 separately from int8/16/64, leftover from when it was special-cased; `int8/16/32/64` reads cleaner.

Reviewed against bee853ed

PR #4999, opened 2026-06-17
thumbnail: emit 8-bit thumbnails everywhere (drop the 16-bit normalized path)

sir-sigurd:
# Description

Makes the thumbnail lambda emit **8-bit** PNGs everywhere. The normalized
greyscale path (multi-channel montages, Z-projections, greyscale microscopy)
was the lambda's only 16-bit output — `norm_img` contrast-stretched each plane
to the full 16-bit range as `int32` (PIL mode `I`, saved `I;16`) — while every
other path already produced `uint8`.

The 16-bit depth was never a designed contract: it originated as the 2019
normalization codomain, and #2888 already shipped an `I;16`→8-bit downconvert
elsewhere. No consumer reads the extra bits — the catalog renders thumbnails
through a plain ``, browsers can't display more than 8 bits, and the
per-image contrast stretch already destroys absolute pixel meaning. 8-bit
output is smaller and unifies the rescale code.

## Changes

- `norm_img` / `_normalize_plane` now emit `uint8` (mode `L`).
- The 16-bit LUT (`_norm_uint_to_int32`) becomes `_lut_uint_to_uint8`, **shared
  with `_rescale_uint16_to_uint8`** so the raw and montage greyscale paths map
  identical pixels identically; `_normalize_plane`'s float path shares the
  affine-stretch tail (`_stretch_to_uint8` → `_saturate_to_uint8`) with
  `_rescale_float_to_uint8`, kept at `float64` to range the montage's float
  planes at full precision.
- `generate_thumbnail`'s `normalized` branch passes `uint8` straight through to
  a mode-`L` PNG; the `I`→`I;16` conversion in `handle_image` and the now-moot
  "Saving I mode images as PNG" pytest warning filter are removed.
- Regenerated the 4 normalized in-repo goldens; updated the `norm_img` unit
  tests and oracle to `uint8`. Regenerated the `images/thumbs` S3 test-data
  package and bumped `THUMBS_PKG`.

## Note on resampling

Beyond the depth change, the normalized path now **resizes in 8-bit (mode `L`)
like every other path**, instead of the higher-precision mode-`I` resize it
uniquely used. This is *not* a pure bit-depth reduction: the old mode-`I` resize
was actually more faithful to the ideal downscaled signal. But the output is
8-bit regardless, so the difference (≤~4% at ≤0.6% of pixels, all at
high-gradient edges) is sub-perceptual on a lossy contrast-stretched preview,
and the montage path now eats the same edge-quantization the rest of the lambda
already does. Resizing in float and quantizing last would avoid the loss but
would re-introduce a non-`uint8` intermediate and de-unify the path — a
conscious reject. Pre-resize, the new `uint8` planes are bit-identical to the
old `/257` (verified), so the regenerated goldens reflect only the depth +
resize-path change, not a normalization-math change.

# TODO

- [x] Unit tests
- [ ] Open and Embed: Confirm that this change doesn't break Open variant and Embed widget — N/A (lambda output depth only)
- [x] Documentation
    - [x] run `optipng` on any new PNGs

<h3>Greptile Summary</h3>

This PR unifies the thumbnail lambda's output depth by making the normalized greyscale path (multi-channel montages, Z-projections) emit 8-bit PNGs (mode `L`) instead of 16-bit `I;16`, matching every other existing path. The internal `_norm_uint_to_int32` LUT is renamed to `_lut_uint_to_uint8` and now shared between `_rescale_uint16_to_uint8` and `_normalize_plane`, ensuring raw and montage greyscale paths apply identical per-pixel arithmetic.

- **Depth change**: `_normalize_plane` / `norm_img` now produce `uint8` (mode `L`); the `I`→`I;16` conversion in `handle_image` and the associated Pillow deprecation warning filter are removed.
- **Shared LUT**: `_lut_uint_to_uint8` replaces both the old `_norm_uint_to_int32` and the previously inline LUT in `_rescale_uint16_to_uint8`, so both paths are guaranteed to map identical pixels identically.
- **Tests & goldens**: Four in-repo PNG goldens regenerated to 8-bit; all `norm_img` unit tests and the reference oracle updated from `int32`/`I;16` to `uint8`/`L`; a new `test_norm_img_path_saves_8bit_png` pins the output depth explicitly.

Safe to merge; the only non-code change flagged is an unfilled placeholder PR link in the CHANGELOG entry.

The implementation is correct and internally consistent: the LUT, float, and empty-plane branches all converge on uint8, generate_thumbnail correctly builds a mode-L image from the uint8 output, and the removed I→I;16 conversion in handle_image was the only consumer of the old 16-bit path. The test suite is thorough — bit-identity sweeps, golden comparisons, and a dedicated depth-pin test cover the changed path well. The only gap is the unfilled placeholder in the CHANGELOG entry.

CHANGELOG.md — the new entry has a placeholder [#XXXX] link that should be [#4999].

| Filename | Overview |
|----------|----------|
| lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py | Core depth change: _norm_uint_to_int32 renamed to _lut_uint_to_uint8 and now emits uint8; _normalize_plane and norm_img updated to uint8/mode-L throughout; handle_image's I→I;16 conversion removed; logic is correct and internally consistent |
| lambdas/thumbnail/tests/test_thumbnail.py | All norm_img tests, oracles, and depth assertions updated from int32/I;16 to uint8/L; test_norm_img_path_saves_8bit_png now pins the 8-bit output; coverage is thorough |
| lambdas/thumbnail/CHANGELOG.md | New [Changed] entry added; placeholder [#XXXX] link was not replaced with the actual PR number (#4999) |
| lambdas/thumbnail/pyproject.toml | Removes the pytest filterwarning that guarded against saving mode-I images as PNG — no longer needed since mode-I is no longer emitted |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[handle_image] --> B[format_aicsimage_to_prepped]
    B --> C{n-dim reader?}
    C -- Yes --> D[_format_n_dim_ndarray]
    C -- No --> E[raw dask_data\nnormalized=False]
    D --> F{greyscale?}
    F -- Yes --> G[norm_img\ndask.delayed]
    F -- No / color --> H[pass through\nnormalized=False]
    G --> I[_normalize_plane\nuint8 output]
    I --> J{arr dtype}
    J -- uint 16-bit --> K[_lut_uint_to_uint8\n65536-entry LUT]
    J -- float / other --> L[float64 arithmetic\n+ _saturate_to_uint8]
    K --> M[uint8 plane]
    L --> M
    E --> N[generate_thumbnail]
    H --> N
    M --> N
    N --> O{normalized=True?}
    O -- Yes --> P[Image.fromarray mode L\n8-bit PNG]
    O -- No --> Q[rescale by dtype to uint8\nImage.fromarray]
    P --> R[PNG bytes]
    Q --> R
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A[handle_image] --> B[format_aicsimage_to_prepped]
    B --> C{n-dim reader?}
    C -- Yes --> D[_format_n_dim_ndarray]
    C -- No --> E[raw dask_data\nnormalized=False]
    D --> F{greyscale?}
    F -- Yes --> G[norm_img\ndask.delayed]
    F -- No / color --> H[pass through\nnormalized=False]
    G --> I[_normalize_plane\nuint8 output]
    I --> J{arr dtype}
    J -- uint 16-bit --> K[_lut_uint_to_uint8\n65536-entry LUT]
    J -- float / other --> L[float64 arithmetic\n+ _saturate_to_uint8]
    K --> M[uint8 plane]
    L --> M
    E --> N[generate_thumbnail]
    H --> N
    M --> N
    N --> O{normalized=True?}
    O -- Yes --> P[Image.fromarray mode L\n8-bit PNG]
    O -- No --> Q[rescale by dtype to uint8\nImage.fromarray]
    P --> R[PNG bytes]
    Q --> R
```

</a>

<sub>Reviews (1): Last reviewed commit: ["thumbnail: emit 8-bit thumbnails everywh..."](https://github.com/quiltdata/quilt/commit/b1473d8bea163f1896e909cf4ec676178fedb928) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38097319)</sub>


sir-sigurd (commented):

Checked the rescale math, the output contract and its downstream consumers, the refactor equivalence, edge/failure paths, and the golden regeneration. The core held up under verification: the now-shared `_lut_uint_to_uint8` is byte-identical to the old inline LUT in `_rescale_uint16_to_uint8` (the raw uint16 path's output is unchanged), the `_normalize_plane` float branch matches the test oracle over thousands of randomized cases, NaN→0 / +inf→255 / −inf→0 / constant→black all carry over, and no consumer reads more than 8 bits (the catalog renders the PNG bytes through a plain ``; the assistant preview path reads them as an opaque blob). The four in-repo goldens were re-rendered from the live path (pixel-identical, optipng applied), so the size drop is just the depth change. Nothing below is a defect — all are optional nits, with #1 and #2 the ones worth acting on.

### Nits

1. **Resize tradeoff isn't noted at the point of change** — `src/t4_lambda_thumbnail/__init__.py:707`, `generate_thumbnail`'s `if normalized:` branch. The deliberate decision that this path now resizes at 8-bit (it used to resize the higher-precision mode-`I` array) lives only in the CHANGELOG and the PR description, both of which disappear from view at the code. A one-line comment at the branch would keep the rationale where the next change to it happens.

2. **The S3 thumbnails aren't depth-pinned** — `tests/test_thumbnail.py:~1170`, `test_handle_image`. The golden comparison asserts pixel values but not dtype (`assert_equal` ignores dtype when the values coincide). Output depth is asserted explicitly only by `test_norm_img_path_saves_8bit_png`, which exercises the single in-repo `cell.tiff`; the S3 montage thumbs would catch a wrong 16-bit regeneration only because the values would differ, not by an explicit check. An `assert actual.reader.data.dtype == np.uint8` (or a mode-`L` assert) in `test_handle_image` is cheap insurance.

3. **A now-load-bearing invariant is implicit** — `src/t4_lambda_thumbnail/__init__.py:~286`, the `# XXX: do we need to cast to uint8?` line in `norm_img`'s color-passthrough. With the `I`→`I;16` conversion gone, `Image.fromarray` choosing mode `L` silently relies on "normalized ⇒ uint8 2-D". That holds today — a color plane is passed through only with `normalized=False` — but nothing documents or asserts it, so a future change to `norm_img`'s return shape could pick the wrong PNG mode without a test failing. Optional comment or assert; not a defect.

4. **(reshape) `_stretch_to_uint8` is defined after its callers** — `src/t4_lambda_thumbnail/__init__.py:649`. A top-to-bottom reader hits all three call sites before the definition. Defensible as-is — the current spot groups the terminal primitives (`_stretch_to_uint8` → `_saturate_to_uint8` → `_alpha_to_uint8`) together — so this is a take-it-or-leave-it readability call.

5. **(reshape) Pre-existing: the constant-plane `>> 8` keep-level** — `src/t4_lambda_thumbnail/__init__.py:595-597`, `_rescale_uint16_to_uint8`. A constant sub-full-range uint16 plane (e.g. an all-`4095` 12-bit image) renders near-black via `>> 8`, where the float and wide-integer siblings keep the level via a clamped affine map. This branch is **pre-existing — untouched by this PR** (the diff only replaced the non-constant inline LUT), and it's an edge of an edge (a fully uniform image). Noting it only because the surrounding function changed here; opportunistic cleanup at most.

Greptile's one flag — the `[#XXXX]` CHANGELOG placeholder — was already fixed to `#4999` and no longer applies at the current head.

Reviewed against 0008d877

sir-sigurd (commented):
Notes from a second AI-assisted self-review pass (re-run from scratch at the current head). Likely the last — verified clean again, no defects; everything below is optional polish.

Spot-checks that held: the raw uint16 path is byte-identical to before (2,980 random `(lo, hi)` pairs, zero mismatches through the shared `_stretch_to_uint8`); float `±inf`/NaN handling carries over; no path produces mode `I`/`F` now, so dropping the `I` → `I;16` conversion is safe; no repo consumer reads more than 8 bits (catalog ``, assistant `arrayBuffer`); the 8-bit resize diverges ~1 LSB worst-case from the old mode-`I` resize. The prior pass's two acted-on nits are fixed, and the function-ordering nit is withdrawn (grouping the terminal `uint8` primitives together reads better).

### Nits (#1–2 worth acting on)

1. **Test oracle is less independent than its wording suggests** — `_norm_float_reference` (`tests/test_thumbnail.py:833`) and the comment at `src/t4_lambda_thumbnail/__init__.py:256`. It's independent of the implementation only in range derivation (`np.percentile` vs the histogram); the per-pixel arithmetic is the same `_stretch_to_uint8` op, so the bit-identity sweep pins histogram-vs-`np.percentile` agreement, not the stretch itself — a bug there moves oracle and code together. The float-branch "bit-identical to `_norm_float_reference`" comment is tautological for finite input. Worth tightening both comments; a finite-float sweep wouldn't add coverage.

2. **An invariant the code now quietly depends on is unasserted** — `norm_img`'s `# XXX: do we need to cast to uint8?` (`src/t4_lambda_thumbnail/__init__.py:286`) and the `if normalized:` branch at `:706`. With the `I` → `I;16` conversion gone, the mode-`L` passthrough relies on "normalized ⇒ 2-D uint8". Holds today (color → `normalized=False` → rescaled downstream), but a future color or non-uint8 normalized producer would pick the wrong PNG mode silently. A one-line assert or comment is cheap insurance. (Same item the prior pass raised; still open.)

3. **(reshape) Constant-plane policy fork is replicated, not centralized** — `__init__.py:234`/`:253`/`:593`. Montage→black vs raw→keep-level lives in several near-identical `if hi == lo:` guards; a later change that "finishes the unification" could silently collapse the deliberate divergence. Defensible as-is.

   *Alternatives: would naming the fork (a policy-tagged fill helper) make it grep-able rather than ambient?*

4. **Description drift** — the "Changes" section credits `_saturate_to_uint8` as the shared piece, but the helper actually extracted and shared is `_stretch_to_uint8`. One-line description fix.

5. **CHANGELOG leads with the weaker rationale** — `CHANGELOG.md:20`: "browsers can't show more than 8 bits" is disputable by an external caller; the stronger argument (the per-image contrast stretch already voids absolute pixel meaning) is already in the PR body. Minor.

Reviewed against 7d247ad

sir-sigurd (commented):
Notes from a third AI-assisted self-review pass, re-run from scratch at the current head. Likely the last — the core re-verified clean; both items below are optional.

Re-checked end to end and confirmed: the raw uint16 path stays byte-identical to before the rescale merge (5,000+ random `(lo, hi)` pairs through the shared `_stretch_to_uint8`, zero mismatches), `_rescale_finite_to_uint8` reproduces the old float and integer helpers exactly (working-dtype choice, constant-level convention, saturate tail), the merged `generate_thumbnail` routes every dtype the same way as before, the dask `dtype=np.uint8` contract holds on all four `_normalize_plane` branches, and a full-repo sweep found no consumer that reads more than 8 bits (catalog ``, assistant blob). The 8-bit-resize tradeoff measures ≤1/255. Peak memory is neutral-to-better — the uint8 montage is a quarter of the old int32 one. Greptile's one flag (the `[#XXXX]` CHANGELOG placeholder) is already fixed.

### Nits

1. **The description names a function this PR deletes** — the "## Changes" section says the float path shares the tail "with `_rescale_float_to_uint8`", but a later commit here merged `_rescale_float_to_uint8` + `_rescale_int_to_uint8` into `_rescale_finite_to_uint8`; neither old name exists at head. One-line description fix (the auto-generated summary above it is stale for the same reason, but that block is bot-owned).

2. **`_rescale_finite_to_uint8`'s working-dtype line silently downcasts any float wider than float64** — `src/t4_lambda_thumbnail/__init__.py:622`. `work = np.float32 if arr.dtype.itemsize <= (4 if is_float else 2) else np.float64` sends a longdouble/float128 plane (itemsize 16 on the Lambda's Linux x86_64) to float64, where the old `_rescale_float_to_uint8` kept its own precision. The comment asserts "the only wider float here is float64", but nothing enforces it. Unreachable today — no bioio reader emits longdouble — and float64 is ample for an 8-bit output regardless, so this is a comment-only invariant worth a one-line guard or a less absolute comment, not a defect.

Reviewed against 7917af7c

PR #5107, opened 2026-07-07
catalog: raise PDF/PPTX preview resolution to 2048x1536

Austin-s-h:
## Summary

Bumps the PDF/PPTX thumbnail preview request from `w1024h768` to `w2048h1536` for a noticeably sharper on-screen preview. The thumbnail lambda's `SUPPORTED_SIZES` already includes `(2048, 1536)`, so no lambda change is required — this is a frontend-only change.

Alongside the size bump, this hardens the PDF loader/renderer:

- **Fix an infinite re-render loop** in the loader by pre-signing the S3 URL with `useMemoEq` keyed on the handle (`bucket`/`key`/`version`) so the `Data.use` input is stable across renders.
- **Preserve the retry button on transient errors**: `loadPdf` re-throws the raw error so `useErrorHandling` wraps it into `PreviewError.Unexpected` with `retry: data.fetch` attached, instead of baking in `retry: null`.
- **Stabilize multi-page navigation** in the renderer via `signRef` / `firstPageBlobRef` so paging doesn't churn the `Data.use` input.

Adds `Pdf.spec.ts` covering `detect()`, request parameters (including the new `w2048h1536` size), glacier→Archived / forbidden→Forbidden mapping, and the unexpected-error rethrow path.

## Note

Sibling of #5108 (replace pdf2image with pypdfium2 for PDF thumbnails), which edits `lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py`. A trivial merge-order conflict is expected and resolvable by taking both hunks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

This PR raises the PDF/PPTX thumbnail resolution from `w1024h768` to `w2048h1536` (already in the lambda's `SUPPORTED_SIZES`) and hardens both the loader and renderer against re-render instability.

- **Loader (`Pdf.js`)**: pre-signs the S3 URL with `useMemoEq` keyed on `[sign, bucket, key, version]` primitives so the `Data.use` params carry a stable string instead of a function reference, breaking a previous infinite re-render loop; non-Forbidden errors are re-thrown raw so `useErrorHandling` can attach `retry: data.fetch`.
- **Renderer (`Pdf.jsx`)**: replaces `sign` and `firstPageBlob` directly in `Data.use` params with `useRef` wrappers (`signRef`, `firstPageBlobRef`) so `R.equals` sees the same object identity across renders and page navigation doesn't trigger spurious re-fetches.
[...]

PR #4859, opened 2026-04-26
preview: update consumers for shared FCS and PDF changes

Austin-s-h:
## Summary

Update consumer lambdas to use the new shared FCS parsing layer from [#4858](https://github.com/quiltdata/quilt/pull/4858), plus PDF preview refactor sharing the same lambda base.

## Affected lambdas

- `lambdas/indexer/` — index FCS metadata
- `lambdas/preview/` — render FCS preview with new `vegaLite` field
- `lambdas/tabular_preview/` — minor adjustments
- `lambdas/thumbnail/` — PDF refactor + FCS-aware thumbnails

## Stack relationship

This PR is **stacked on top of [#4858](https://github.com/quiltdata/quilt/pull/4858)** (rebase-based). Until #4858 lands on master, this PR's diff against `master` will include the shared-layer changes from #4858; review the diff in that lens, or review #4858 first.

| PR | Layer |
|---|---|
| [#4858](https://github.com/quiltdata/quilt/pull/4858) | Shared lambda parsing + `vegaLite` spec |
| **#4859 (this)** | Consumer lambdas updated to use shared layer |
| [#4860](https://github.com/quiltdata/quilt/pull/4860) | Catalog frontend Vega chart rendering |

## Validation

- `cd lambdas/<consumer> && uv run pytest`
- End-to-end FCS file preview smoke (deployment-side)

## Deployment context

Tracked in [deployment#2395](https://github.com/quiltdata/deployment/pull/2395); rolled out via [deployment#2394](https://github.com/quiltdata/deployment/pull/2394).

PR #5108, opened 2026-07-07
feat(thumbnail): render PDF thumbnails with pypdfium2 instead of pdf2image

Austin-s-h:
## Summary

Replaces the `pdf2image` Python dependency in the thumbnail lambda with `pypdfium2` native bindings, extracted into a dedicated `pdf_thumbnail.py` helper.

- **Drops the `pdf2image` library** — the only dependency change is swapping `pdf2image` for `pypdfium2>=4,<5` in `pyproject.toml`; all other pins and the `t4-lambda-shared` source are unchanged.
- **Poppler is now optional, not required.** When `pdftoppm`/`pdfinfo` are present on `PATH` they're still used as a fast path; when they aren't, `pypdfium2` renders natively. This is a compatibility improvement (native fallback), not a removal of the Poppler code path — the subprocess route is retained.
- **Fixes a file-handle/mmap leak**: the `PdfDocument` and page are closed in `try/finally` on every render and page-count.
- **Configurable render DPI** via `PDF_PREVIEW_DPI` (default 300, clamped to 72–600); pages are resized with LANCZOS, and `pdf_render_dpi` / `pdf_resize_filter` are reported in the response info.
- Guards `pptx_to_pdf` against a missing `libreoffice` binary with a clean `PDFThumbError`.

Adds `tests/test_pdf_thumbnail.py` (11 tests, all passing locally via `uv run pytest`).

## Note

Touches `lambdas/thumbnail/src/t4_lambda_thumbnail/__init__.py` alongside sibling PR #5107 (raise PDF preview resolution to 2048x1536). A trivial merge-order conflict is expected and resolvable by taking both hunks.

<h3>Greptile Summary</h3>

This PR replaces the `pdf2image`/Poppler subprocess dependency in the thumbnail lambda with `pypdfium2` native bindings, extracted into a new `pdf_thumbnail.py` helper. A `pdftoppm`/`pdfinfo` fast path is preserved when those binaries are present on `PATH`, with `pypdfium2` as the fallback.

- **Dependency swap**: `pdf2image` removed from `pyproject.toml` and `uv.lock`; `pypdfium2>=4,<5` (locked to 4.30.0) added — no Poppler system binary required in the base Lambda image.
- **Resource management**: `PdfDocument` and `PdfPage` are now closed in `try/finally` blocks on every render and page-count, fixing the per-render file-handle/mmap leak mentioned in the PR description. The intermediate `PdfBitmap` from `render()` is not yet wrapped in its own `try/finally`.
- **New features**: Configurable render DPI via `PDF_PREVIEW_DPI` (clamped 72–600, default 300); `pdf_render_dpi` and `pdf_resize_filter` added to response info. `pptx_to_pdf` now raises `PDFThumbError` on missing `libreoffice` rather than letting `FileNotFoundError` propagate.

Safe to merge; the core rendering logic is sound, error handling is improved over the previous implementation, and the 11 new tests cover all key branches.

The migration from pdf2image to pypdfium2 is clean and well-tested. The only items worth addressing are the PdfBitmap not being explicitly closed after to_pil(), and shutil.which() being probed on every request instead of once at module load. Neither affects correctness in CPython reference-counting GC, but worth tightening before shipping to production.

lambdas/thumbnail/src/t4_lambda_thumbnail/pdf_thumbnail.py — bitmap lifecycle and repeated shutil.which() calls

### DIFF
--- a/lambdas/thumbnail/CHANGELOG.md
+++ b/lambdas/thumbnail/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Changed] Render PDF thumbnails with `pypdfium2` (native bindings) instead of `pdf2image`/Poppler: simpler dependencies, cross-platform, configurable render DPI (`PDF_PREVIEW_DPI`), and a try/finally cleanup that closes the document and page to fix a per-render file-handle/mmap leak
 - [Changed] Stream the source download to disk instead of into memory, lowering peak memory on large images ([#5000](https://github.com/quiltdata/quilt/pull/5000))
 - [Changed] Thumbnails are now 8-bit everywhere: normalized greyscale montages and Z-projections (multi-channel microscopy stacks, greyscale CZIs) are emitted as 8-bit PNGs (mode `L`) instead of 16-bit (`I;16`). These previews are already contrast-stretched per image, so the extra depth carried no recoverable pixel meaning; output is smaller and the affected thumbnails now resize at 8-bit like every other path (slightly coarser tonal steps) ([#4999](https://github.com/quiltdata/quilt/pull/4999))
 - [Fixed] Signed and wider-than-16-bit integer images (`uint32`/`uint64`, `int8`/`int16`/`int64`, and 32-bit-integer color) now preview — contrast-stretched to 8-bit — instead of failing with HTTP 500 (color and 64-bit integers) or rendering as a clamped/dark 16-bit image ([#4998](https://github.com/quiltdata/quilt/pull/4998))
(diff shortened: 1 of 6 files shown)


QuiltSimon:
Reconciliation pass (see #engineering thread, 2026-08-29) — **parked on one adoption decision.** Author is no longer in the org; this is part of the deliberately harvested preview set (#5105/#5107/#5108/#5109 — #5110 already resolved via #5189 + #5205).

Named unblock: **a lambda/preview owner adopts or declines the set** (sir-sigurd on merge history, or drernie who engaged on #4859). If adopted: this one is CONFLICTING and rewrites the same files as #5105 — land #5105 (+10/-1, reviewed, effectively mergeable) first, then rebase this on top. The pypdfium2-vs-pdf2image swap deserves its own review pass at that point; nothing here has run in CI (`loffice`/poppler-marked tests are excluded).

PR #5152, opened 2026-07-26
Enable ruff B905 and make zip() strictness explicit

sir-sigurd:
# Description

Drops `B905` from `ruff.toml`'s ignore list and adds an explicit `strict=True` at all 14 flagged `zip()` calls. The rule was ignored when `target-version` moved to `py310` — dropping Python 3.9 made it newly applicable — and deferred as "clean up separately".

Every flagged site pairs iterables that are equal-length by construction, so `strict=True` is a latent-bug detector rather than a behavior change on any reachable path. Two shapes cover all 14:

- **Built in lockstep in the same loop** — `_calculate_missing_hashes`'s `physical_keys`/`sizes`, `_push`'s `entries`/`file_list`, `verify`'s expected-hash lists, `access_counts`'s query list, and the two tests.
- **A results accumulator pre-allocated as `[None] * len(input)` and only ever mutated in place** — all six `data_transfer.py` sites, plus the two that consume `calculate_multipart_checksum` and `copy_file_list_fn` results. Tenacity retries replay the same list object, so a retry can't desync the lengths.

The one site with real committed side effects at the zip is `_push`: `copy_file_list_fn` has already uploaded bytes by then, and entries needing a copy are `_set` on the new package *only* inside that loop. Silent truncation there means a push that reports success while publishing a manifest missing files it just uploaded, so this is where `strict=True` earns the most.

Two sites keep their existing `assert len(a) == len(b)`. Those are eager — they fire before any transfer task is dispatched — whereas `strict=` is lazy and only raises once the shorter iterable is exhausted. Complementary, so the asserts stay.

Note that `verify()` gets less protection than the rest: it returns `False` on the first hash mismatch, so its two loops usually exit before the length check can fire.

Not done by autofix — `ruff check --unsafe-fixes` writes `strict=False`, which silences the rule without adding the check.

Removing the ignore enables the rule everywhere: only `api/python/pyproject.toml` defines `[tool.ruff]`, and it does `extend = "../../ruff.toml"`. CI's two ruff jobs (`lint-quilt3` in `api/python`, `lint-rest` over the rest) together cover the same file set as a root-level run.

No CHANGELOG entry: no reachable call path changes behavior.

# TODO

- [x] Unit tests — no new tests; the existing `api/python`, `py-shared`, and `access_counts` suites pass unchanged

<h3>Greptile Summary</h3>

Makes iterable-length assumptions explicit across Python code and tests.
- Enables Ruff rule B905 globally by removing it from the ignore list.
- Adds `strict=True` to 14 existing `zip()` calls in package transfer, checksum, verification, analytics, and test code.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| api/python/quilt3/data_transfer.py | Adds strict length validation to lockstep transfer and checksum task/result iteration. |
| api/python/quilt3/packages.py | Adds strict length validation when associating package entries, transfer results, and checksum results. |
| lambdas/access_counts/src/t4_lambda_access_counts/index.py | Requires Athena query metadata and execution IDs to have matching lengths before copying result files. |
| api/python/tests/test_formats.py | Makes the equal-length object and metadata test fixture invariant explicit. |
| py-shared/tests/test_athena.py | Makes the equal-length Athena query and stubbed execution-ID fixture invariant explicit. |
| ruff.toml | Enables Ruff B905 so future zip calls must state their strictness explicitly. |

<sub>Reviews (2): Last reviewed commit: ["Drop the CopyFileListFn docstring"](https://github.com/quiltdata/quilt/commit/02523705145a3bec698b4dcac152e164fcfdb2c7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47335273)</sub>

### DIFF
--- a/api/python/quilt3/data_transfer.py
+++ b/api/python/quilt3/data_transfer.py
@@ -604,7 +604,7 @@ def _copy_file_list_internal(file_list, results, message, callback, exceptions_t
 
     assert len(file_list) == len(results)
 
-    total_size = sum(size for (_, _, size), result in zip(file_list, results) if result is None)
+    total_size = sum(size for (_, _, size), result in zip(file_list, results, strict=True) if result is None)
 
     lock = Lock()
     futures = deque()
@@ -669,7 +669,7 @@ def _copy_file_list_internal(file_list, results, message, callback, exceptions_t
                     _copy_remote_file(ctx, size, src.bucket, src.path, src.version_id, dest.bucket, dest.path)
 
         try:
-            for idx, (args, result) in enumerate(zip(file_list, results)):
+            for idx, (args, result) in enumerate(zip(file_list, results, strict=True)):
                 if result is not None:
                     continue
                 run_task(idx, worker, idx, *args)
@@ -1019,7 +1019,9 @@ def _calculate_checksum_internal(
     results: list[str | Exception | None],
 ) -> list[str | Exception]:
     total_size = sum(
-        task.size for task, result in zip(tasks, results) if result is None or isinstance(result, Exception)
+        task.size
+        for task, result in zip(tasks, results, strict=True)
+        if result is None or isinstance(result, Exception)
     )
     stopped = False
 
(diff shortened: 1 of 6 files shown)


sir-sigurd:
@greptileai please re-review — dropped the `CopyFileListFn` docstring as an unrelated ride-along. The diff is now just the `ruff.toml` line plus the 14 `strict=True` additions.

PR #5161, opened 2026-07-28
Catalog: make the object delete a soft delete

sir-sigurd:
# Description

The catalog's object delete now writes an S3 delete marker instead of destroying a specific object version.

`DeleteDialog` does not use the handles its callers pass in — it re-resolves each one through `useFilesListing`, and that chain calls `headObject`, which returns a `VersionId` on a versioned bucket whether or not one was requested. A concrete version therefore reached `s3.deleteObject` from **all three** delete entrypoints: the file viewer, the row action in a directory listing, and bulk "Delete selected". Each of them permanently destroyed a version. Not sending `VersionId` at that single call makes the action object-scoped everywhere at once.

This is a conformance fix, not a change of intent. The 1.63.0 release notes and the sandbox guide both tell customers the delete "only adds a delete marker to the latest version, so prior versions will still be available from packages" and that deleted files "remain recoverable". The implementation never behaved that way.

Permanent erasure of a specific version becomes an admin operation, performed by explicitly attaching a custom policy granting `s3:DeleteObjectVersion`. The prefix-permissions example policy no longer includes that action by default and presents it as an opt-in instead, and the FAQ recipe for purging a package's objects now states the prerequisite. A paired registry-side change drops the action from the policies generated for managed read-write roles, so the two must ship in the same release — otherwise the catalog delete starts failing with `AccessDenied` for those users.

The cross-account bucket-policy example is deliberately left as is: cross-account access requires both the bucket policy and the caller's identity policy to allow an action, so with the identity side locked down, trimming the bucket policy would grant nothing a managed role could exercise while risking a service path no test covers.

The delete control is deliberately **not** hidden on pinned-version views: viewing an object always is a version view, so gating on that would remove the control everywhere. Instead the action is scoped to the object and the dialog says so.

The confirmation dialog previously showed only a title and each object's version id. That id named a version the operation no longer targets — on a pinned view, one that outlives the delete — so it is replaced with copy stating what the delete covers and how recoverability follows the bucket's versioning state, per the [S3 DeleteObject contract](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html). A pinned view additionally says so, since that is where the mismatch between screen and effect is sharpest.

Deleting an object whose current version is already a delete marker only stacks another marker, so that control is now hidden. Only the pinned view was affected — the latest view's version-less head already returns the delete-marker response and hid it. This is kept out of `notAvailable`, which also gates download, so a pinned version that still exists stays downloadable.

**Open question for @nl0** — the sign-off asked for "don't offer a no-op delete — show that state instead". The first half is done. For the second, the latest view already renders the Deleted preview error, but on `?version=<older surviving version>` that version's content is legitimately viewable, so the existing treatment can't just be reused. Which do you want there: the button gone only, an indicator that the object is currently delete-marked alongside the version's content, or a disabled control that explains itself? Happy to follow up separately rather than hold this.

Exposure is limited to buckets that opt in via `ui.actions.deleteObject`, which defaults to `false`.

# TODO

- [x] Unit tests
- [x] Documentation
    - [x] Markdown somewhere in docs/**/*.md that explains the feature to end users

<h3>Greptile Summary</h3>

The PR changes Catalog object deletion from version-scoped erasure to object-scoped S3 deletion.
- Omits `VersionId` so versioned buckets receive a delete marker.
- Updates the confirmation dialog to distinguish recoverable versioned deletion from permanent unversioned deletion.
- Hides deletion when the current object state is already a delete marker while preserving access to pinned surviving versions.
- Documents the separate `s3:DeleteObjectVersion` permission required for permanent version erasure.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Bucket/requests/object.ts | Removes `VersionId` from the S3 DeleteObject request, making Catalog deletion object-scoped. |
| catalog/app/containers/Bucket/Toolbar/DeleteDialog.tsx | Clearly warns that deletion is permanent when the bucket does not retain versions, resolving the prior recoverability issue. |
| catalog/app/containers/Bucket/File/File.jsx | Determines the current object's delete-marker state independently from the pinned version being displayed. |
| catalog/app/containers/Bucket/File/Toolbar/useFeatures.ts | Suppresses redundant object deletion when the current state is already deleted without disabling pinned-version downloads. |
| catalog/app/containers/Bucket/File/Toolbar/Organize/Context.tsx | Passes pinned-version context into the object-scoped deletion confirmation dialog. |

<sub>Reviews (3): Last reviewed commit: ["catalog: don&#39;t offer delete when the obj..."](https://github.com/quiltdata/quilt/commit/bcf8fa61b006f622d4063c90d84bad7db26d102e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47910495)</sub>


sir-sigurd:
@greptileai the cross-account bucket-policy edit has been reverted since your last pass; please re-review at head.

sir-sigurd:
@greptileai please re-review at head — since your last pass (8592cdd) this adds the confirmation-dialog copy, hides the delete control when the object is already delete-marked, and trims the changelog entry.

catalog/app/containers/Bucket/Toolbar/DeleteDialog.tsx
```diff
@@ -105,6 +105,13 @@ export default function DeleteDialog({ close, handles }: DeleteDialogProps) {
     <>
       <M.DialogTitle>{title}</M.DialogTitle>
       <M.DialogContent>
+        {!isComplete && (
+          <M.DialogContentText>
+            Adds a delete marker, so these objects stop appearing in listings and
+            search. Prior versions are retained and stay available in each object's
+            version history.
```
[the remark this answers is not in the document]
sir-sigurd: Agreed, and fixed in 7cfdad4a — this was raised independently by the operator too. The dialog no longer promises recoverability unconditionally; it now forks on the bucket state per the [DeleteObject contract](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html): "Where the bucket keeps versions, a delete marker hides the file and earlier versions stay available; where it does not, the deletion is permanent."

Two details worth recording, since both bit earlier drafts. Wording is "earlier versions" rather than "all versions": on a versioning-suspended bucket a plain delete removes a current `null` version and replaces it with a marker, so that content is lost while earlier versions survive. And the fork is stated in prose rather than detected in code — a `handle.version` check looked attractive but directory handles resolve via `ListObjectsV2`, which returns no version ids, so bulk and directory deletes on a versioned bucket would have been misreported as permanent.

nl0:
> Which do you want there: the button gone only, an indicator that the object is currently delete-marked alongside the version's content, or a disabled control that explains itself? Happy to follow up separately rather than hold this.

i'd prefer "a disabled control that explains itself" i guess

catalog/app/containers/Bucket/Toolbar/DeleteDialog.tsx
```diff
@@ -105,6 +111,20 @@ export default function DeleteDialog({ close, handles }: DeleteDialogProps) {
     <>
       <M.DialogTitle>{title}</M.DialogTitle>
       <M.DialogContent>
+        {!isComplete && versionPinned && (
+          <M.DialogContentText color="textPrimary">
+            <strong>You are viewing a specific version of this file.</strong>
```
nl0, line 116:
```suggestion
            <strong>You are viewing a historic version of this file.</strong>
```

maybe smth like this?

nl0 (dismissed):

catalog/app/containers/Bucket/Toolbar/DeleteDialog.tsx
sir-sigurd: I'd rather keep "specific" — "historic" is reachable-false, in two ways.

That line is driven by nothing more than `?version=` being present in the URL, and `VersionInfo` links *every* dropdown row, latest included (the `(latest)` suffix is only a label). So clicking the latest row — or opening an older shared link that pinned what was then current — gives a pinned view of the **current** version.

And `objectVersions` maps `id: v.VersionId` verbatim, so on an unversioned bucket the single `ListObjectVersions` row yields `?version=null`. There the line can appear on a view where the delete is permanent.

I did look at computing it properly: the page holds both the object's current version id and the pinned one, so comparing them gives older-vs-latest, and the line could then be suppressed entirely when they match (no mismatch to warn about). I decided against it — the current-version id is captured at page load, so a new version uploaded while the dialog is open makes the comparison conclude "latest" and drop the warning exactly when it started being needed. A false negative on a destructive-action warning seemed worse than a flat adjective, and it would reintroduce the version-derived branch this PR otherwise avoids (directory handles come from `ListObjectsV2`, which returns no version ids at all).

Happy to revisit as a follow-up if the wording still bothers you — "older" rather than "historic" if we do, since we'd only say it where it's true. I'll pick that up with the disabled-control change.

sir-sigurd:
On "a disabled control that explains itself" — noted, and I'll do it that way. What's on this branch hides the control instead, which was written before your answer; converting it needs a disabled state the Organize menu doesn't have yet, so I'd rather not bolt it on here. Tracking it as a follow-up together with the wording thread above.

PR #5162, opened 2026-07-30
pkgevents: drop s3:TestEvent silently, report other failures per message

sir-sigurd:
# Description

The `pkgevents` lambda crashed with `KeyError: 'Records'` on `s3:TestEvent`, which S3 delivers whenever a bucket's notification configuration is created — i.e. on every bucket add. Because the S3 events were pulled lazily inside `itertools.chain.from_iterable`, that crash aborted the *whole* SQS batch: real `package-revision` events co-batched with the test event were retried and then dead-lettered, so EventBridge subscribers silently missed those package revisions.

`s3:TestEvent` messages are now dropped silently: they arrive on every bucket add, so a warning there would be 100% noise. `indexer` already guards against them and the inline S3→EventBridge lambda in the CloudFormation template was fixed separately, so `pkgevents` was the last S3-notification consumer without the guard.

Any other message without `Records` (including bodies that aren't valid JSON objects at all — S3 is the expected producer on this queue, not the only possible one) — plus per-event processing errors and entries rejected by `put_events` — is now reported via `ReportBatchItemFailures`, so only the affected messages are retried and eventually dead-lettered while the rest of the batch completes. Skipping such a message would delete it permanently; failing it keeps the payload recoverable in the DLQ, which also makes the previous log-the-whole-body warning redundant. When an error can't be attributed to specific messages (e.g. the `put_events` call itself fails), the handler still raises and fails the whole batch, because anything missing from `batchItemFailures` is treated as processed and deleted. Retries are message-grained, so a message whose events straddle a `put_events` batch boundary can republish a few events on retry — accepted, and strictly less duplication than today's whole-batch retry.

The new return value is inert until the SQS event source mapping sets `FunctionResponseTypes=["ReportBatchItemFailures"]`; a paired deployment-side change adds that flag, so the two changes can ship in either order.

Observed on an internal deployment: 30 errors / 2443 invocations over 7 days, with a matching `s3:TestEvent` message sitting in the lambda's DLQ. Not a regression — the crashing line is unchanged since #2074 (2021).

# TODO

- [x] Unit tests

<h3>Greptile Summary</h3>

Updates the pkgevents Lambda to isolate SQS message failures while silently discarding S3 test notifications.
- Parses and processes each SQS message independently.
- Reports malformed messages, processing errors, and EventBridge entry failures through `batchItemFailures`.
- Adds coverage for test events, malformed payloads, per-message processing failures, EventBridge partial failures, and batching.

This follow-up appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| lambdas/pkgevents/src/t4_lambda_pkgevents/__init__.py | Replaces whole-batch failure handling with per-message parsing, EventBridge failure attribution, and an SQS partial-batch response. |
| lambdas/pkgevents/tests/test_index.py | Expands tests to cover S3 test events, malformed messages, processing exceptions, partial EventBridge failures, and API-sized batches. |
| lambdas/pkgevents/CHANGELOG.md | Documents the corrected S3 test-event and partial-batch behavior. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant S3
  participant SQS
  participant Lambda as pkgevents
  participant EB as EventBridge
  S3->>SQS: Notification message
  SQS->>Lambda: Batched records
  alt s3:TestEvent
    Lambda-->>Lambda: Drop silently
  else Valid package event
    Lambda->>EB: PutEvents (up to 10 entries)
    EB-->>Lambda: Per-entry results
  else Invalid or processing failure
    Lambda-->>Lambda: Record messageId as failed
  end
  Lambda-->>SQS: batchItemFailures
```

<sub>Reviews (5): Last reviewed commit: ["pkgevents: drop EventsQueue for buffer-t..."](https://github.com/quiltdata/quilt/commit/c42537e6a4440dafc7533653fb1b57d1e27d7352) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48693644)</sub>

**Context used:**

- Knowledge Base — [Package events and push lambdas (pkgevents, pkgpush)](https://app.greptile.com/quilt-bio/-/custom-context/knowledge-base/quiltdata/quilt/-/docs/lambdas-package-events-push.md)

### DIFF
--- a/lambdas/pkgevents/CHANGELOG.md
+++ b/lambdas/pkgevents/CHANGELOG.md
@@ -17,6 +17,7 @@ where verb is one of
 
 ## Changes
 
+- [Fixed] Skip SQS messages without S3 records, e.g. `s3:TestEvent`, instead of failing the whole batch and dead-lettering the real package events in it ([#5162](https://github.com/quiltdata/quilt/pull/5162))
 - [Fixed] Process package pointers from year 2026+ ([#4683](https://github.com/quiltdata/quilt/pull/4683))
 - [Changed] Migrate to proper package structure ([#4647](https://github.com/quiltdata/quilt/pull/4647))
 - [Changed] Switch to uv ([#4647](https://github.com/quiltdata/quilt/pull/4647))
--- a/lambdas/pkgevents/CHANGELOG.md
+++ b/lambdas/pkgevents/CHANGELOG.md
@@ -17,7 +17,7 @@ where verb is one of
 
 ## Changes
 
-- [Fixed] Skip SQS messages without S3 records, e.g. `s3:TestEvent`, instead of failing the whole batch and dead-lettering the real package events in it ([#5162](https://github.com/quiltdata/quilt/pull/5162))
+- [Fixed] Drop `s3:TestEvent` messages and report other failures per message via `ReportBatchItemFailures`, instead of failing the whole SQS batch and dead-lettering the real package events in it ([#5162](https://github.com/quiltdata/quilt/pull/5162))
 - [Fixed] Process package pointers from year 2026+ ([#4683](https://github.com/quiltdata/quilt/pull/4683))
 - [Changed] Migrate to proper package structure ([#4647](https://github.com/quiltdata/quilt/pull/4647))
 - [Changed] Switch to uv ([#4647](https://github.com/quiltdata/quilt/pull/4647))
--- a/lambdas/pkgevents/src/t4_lambda_pkgevents/__init__.py
+++ b/lambdas/pkgevents/src/t4_lambda_pkgevents/__init__.py
@@ -1,4 +1,3 @@
-import itertools
 import json
 import re
 import time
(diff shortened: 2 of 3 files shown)


sir-sigurd:
@greptileai review

sir-sigurd:
@greptileai review

sir-sigurd:
@greptileai review

sir-sigurd:
**Live verification** (2026-07-31, disposable dev stack): lambda code built from this PR's head `c42537e6a` with the CI zip recipe, SQS event source mapping carrying `FunctionResponseTypes=["ReportBatchItemFailures"]` (ships in a paired deployment-side change). All messages injected through the real SQS queue.

| Case | Result |
|---|---|
| `s3:TestEvent` alone | ✅ consumed silently — one 2.4 ms invocation, zero app log lines, message deleted, DLQ empty |
| Pointer-created event | ✅ correct `package-revision` (handle + topHash) published to EventBridge, message deleted |
| Co-batched poison (non-JSON body) + `s3:TestEvent` + good event | ✅ one invocation: per-message `[ERROR]` naming only the poison's messageId; good event still published; good + test messages deleted; poison retained |
| Poison redrive | ✅ redelivered on the queue's 240 s visibility cadence, dead-lettered after exactly `maxReceiveCount` (15) receives with 15 logged exceptions, payload intact in the DLQ for forensics |
| Flag with the *old* handler code (upgrade transient) | not exercised live; per AWS docs a null response under the flag is a complete success and a thrown exception still fails the whole batch — i.e. exactly today's behavior |

PR #5176, opened 2026-08-12
Fix invalid Material Icons ligature names (no rendered change)

sir-sigurd:
# Description

The ligature names below are not names the Material Icons font carries. The font matches the longest ligature it has for the leading characters and passes the remainder through, so each of these already drew the correct glyph — followed by the leftover characters, which the font maps to zero-ink glyphs that still carry a full em of advance each.

**Nothing renders differently after this PR.** MUI v4's `Icon` root is `width: 1em; height: 1em; overflow: hidden`, so that stray advance is clipped and never affects ink or layout. Measured in a running catalog build against a real stack: `palette_outlined` in the Perspective toolbar is an 18×18 box holding 180px of content, 162px clipped; `bar_charts` in the Analytics header is 24×24 holding 48px; sibling icons on the same pages clip 0. So this is latent-bug hygiene, not a visible fix. It is still worth landing: MUI's own source annotates that `overflow: hidden` as a Chrome bug workaround, "To remove at some point" — drop the clip and each of these starts reserving its stray advance, pushing whatever sits beside it.

| where | was | now | stray tail |
|---|---|---|---|
| `JsonEditor/Note.tsx`, `Preview/renderers/Perspective`, `Admin/UsersAndRoles/BucketsPermissions.tsx` | `info_outlined` | `info_outline` | `d` |
| `Bucket/FileProperties.tsx` | `warning_outline` | `warning` | `_outline` |
| `Preview/renderers/Perspective` | `palette_outlined` | `palette` | `_outlined` |
| `Bucket/File/Analytics.tsx` | `bar_charts` | `bar_chart` | `s` |
| `Bucket/Meta.tsx` | `label_outlined` | `label_outline` | `d` |

Two traps worth recording, since this is easy to reintroduce. Checking `@material-ui/icons` proves nothing: it ships `InfoOutlined`, `PaletteOutlined` and `WarningOutlined` as SVG components, none of which are ligature names. And the `_outline` suffix is not a general pattern — `error_outline` works because it is its own icon in the base set, not an outlined variant of `error`.

These names were never valid, so this is not font drift. None of them appears in the upstream icon set at either end of its history — not in the 2018 release (932 icons), not today (2,234) — while `info_outline`, `error_outline` and `label_outline` are present in both. The catalog has loaded the same `icon?family=Material+Icons` link since 2017, and the set has only grown over that span (`bar_chart` is absent from the 2018 list and present now), with legacy `_outline` names retained rather than pruned. Checked at those two points rather than continuously.

Coverage: extracted every statically written ligature name in `catalog/app` — bare `<M.Icon>` and `<OutlinedIcon>` children, quoted names inside child expressions such as ternaries, and `icon="…"` string props, that last spelling being the one an earlier pass missed (`Section`'s `icon`) — and checked each against the GSUB ligature tables of both fonts `catalog/app/index.html` loads (MaterialIcons v145, 2,234 ligatures; MaterialIconsOutlined v110, 2,195). On master that is 101 distinct names across 268 sites, of which these five, at seven sites, were the only invalid ones; on this branch the same extraction returns zero. A further thirteen sites assemble the name at runtime (`{icon}` props, `` {`chevron_${direction}`} ``) and are out of reach of any static check.

# TODO

- [ ] Unit tests — none here; the extraction above is mechanical and would make a good guard as a test asserting every statically written ligature name in `catalog/app` is present in the ligature table of a font the catalog loads. Left out to keep this diff to the names themselves. Worth weighing on its merits: these were introduced independently between 2019 and 2023, each wrong on first write and none ever visible, which is the argument for a mechanical guard rather than closer review.
- [x] [Changelog](../tree/master/catalog/CHANGELOG.md) entry — skipped per the template's "not significant to end users": there is no user-visible change. (Note the template links `docs/CHANGELOG.md`, which is the quilt3 Python changelog; catalog changes belong in `catalog/CHANGELOG.md`.)

<h3>Greptile Summary</h3>

The PR replaces eight invalid Material Icons ligature literals with valid names while preserving the glyphs currently shown by the clipped MUI icon container.
- Corrects direct `<M.Icon>` children in JSON editor, Perspective, permissions, file properties, and navigation views.
- Corrects icon names passed through `Section` string props in Analytics and S3 object tags.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/components/JsonEditor/Note.tsx | Replaces the invalid `info_outlined` ligature with `info_outline`. |
| catalog/app/components/Preview/renderers/Perspective/Perspective.tsx | Corrects the informational and palette toolbar icon ligatures. |
| catalog/app/containers/Admin/UsersAndRoles/BucketsPermissions.tsx | Replaces the invalid informational icon ligature with its valid equivalent. |
| catalog/app/containers/Bucket/File/Analytics.tsx | Corrects the Analytics section icon passed through the `Section` string prop. |
| catalog/app/containers/Bucket/FileProperties.tsx | Replaces the invalid warning icon ligature with `warning`. |
| catalog/app/containers/Bucket/Meta.tsx | Corrects the object-tags section icon passed through the `Section` string prop. |
| catalog/app/containers/NavBar/NavMenu.tsx | Replaces the invalid bookmarks icon ligature with `bookmarks`. |

<sub>Reviews (2): Last reviewed commit: ["Fix two more invalid ligatures passed as..."](https://github.com/quiltdata/quilt/commit/726e2c3e0d5c411889523bb055cd80b5b4bd5e6f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52427933)</sub>


sir-sigurd:
@greptileai please re-review — new commit since your last pass (`afee2127e` → `726e2c3e0`) fixes two further invalid ligatures reached through `Section`'s `icon` string prop (`bar_charts`, `label_outlined`), and the description has been corrected: these names already rendered the right glyph, and MUI's clipped 1em icon box means the fix produces no rendered change.

fiskus:
It seems that practically "info_outlined" became "info_outline" (correct) + "d" (invisible, because doesn't exists in icons font).

fiskus (approved):

PR #5183, opened 2026-08-20
Catalog: don't offer admin controls the registry refuses for service users

sir-sigurd:
# Description

The admin Users table offered controls that the registry refuses. A service user — the stack-managed `_canary` account behind a deployment's canaries — has its email, role and active flag re-forced on every deploy, and the registry rejects admin attempts to change them; but the table still rendered a working Enabled toggle, a click-to-edit email and a Delete action for that row.

Those three are now gated on `isService`, a field the users query has fetched since it was introduced without ever being used in UI logic. The role and admin controls keep following the server-computed `isRoleAssignmentDisabled` / `isAdminAssignmentDisabled` flags, which account for service users in a paired registry-side change. Because that widens a flag whose explanatory copy was SSO-specific, `RoleSelect` now takes the reason for non-assignability, so a service user's roles dialog no longer cites an SSO role mapping that a stack may not have.

Without this gating, each attempt lands on the "unexpected operation error" path — a raw exception message inside a toast for the Enabled toggle, a generic "Something went wrong" for the others, and a Sentry event for what is in fact expected policy. Pre-gating is the pattern this page already uses for refusals that are predictable from data it has fetched: the same mechanism hides Delete for your own row and puts the roles dialog into read-only mode for SSO-managed users.

# TODO

- [x] n/a Unit tests — `UsersAndRoles/` has no test files, so there is no existing harness to extend; verified with `tsc --noEmit` and `oxlint` instead
- [x] Security: Confirm that this change meets security best practices and does not violate the security model — this only removes affordances; the registry remains the enforcement point and is unchanged by this PR
- [x] Open: Confirm that this change doesn't break the Open variant — gating is per-row on a field the query already returns, and rows where `isService` is false behave exactly as before
- [x] n/a Documentation
- [x] [Changelog](../tree/master/catalog/CHANGELOG.md) entry — in `catalog/CHANGELOG.md`, which is where catalog-only changes are recorded rather than `docs/CHANGELOG.md`

<h3>Greptile Summary</h3>

The PR prevents administrators from attempting unsupported changes to stack-managed service users and gives the read-only roles dialog service-specific explanatory text.
- Disables active-state changes and removes email editing and deletion for service users.
- Distinguishes service-managed roles from SSO-managed roles in helper text.
- Adds the catalog changelog entry with the correct pull request link.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/CHANGELOG.md | Adds the service-user UI fix entry and correctly links it to pull request #5183. |
| catalog/app/containers/Admin/UsersAndRoles/RoleSelect.tsx | Supports reason-specific explanatory text when role assignment is disabled. |
| catalog/app/containers/Admin/UsersAndRoles/Users.tsx | Removes unsupported active, email, and deletion controls for service-user rows. |

<sub>Reviews (3): Last reviewed commit: ["Catalog: lift RoleSelect&#39;s non-assignabl..."](https://github.com/quiltdata/quilt/commit/91ba05cc9240c928aca6c2e307722ba8ddfb5c6f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55100839)</sub>


sir-sigurd:
@greptileai review

sir-sigurd:
@greptileai review

PR #5214, opened 2026-08-25
tabular_preview: make a zero-row preview a loadable Arrow file

fiskus:
# Description

A tabular preview with no rows fails to render — the catalog shows "Could not
render tabular data" where an empty table belongs.

`write_data_as_arrow` iterates `table.to_batches(...)`, which yields nothing for
a 0-row table, so it writes an IPC file carrying a schema and no record batches.
That file is valid — pyarrow reads it back as a 0-row table — but perspective's
`arrow_loader.cpp` calls `Table::FromRecordBatches` without a schema, so Arrow
has nothing to infer from and rejects it. Fixed upstream in perspective 3.2.0
([finos/perspective#2854](https://github.com/finos/perspective/pull/2854)); we
are below that.

Reached by a header-only file (an empty query export, a filter that matched
nothing), a file whose every data row was skipped as invalid, or a first batch
already too large for `max_size`. The fix writes one empty batch when nothing
else was written.

Surfaced while building preview test samples for #4671.

# TODO

- [x] Unit tests — `test_preview_csv_no_rows`; both cases fail without the fix
- [x] Security: no new surface — no new input handling, no new dependency
- [x] Open and Embed: server-side change
- [x] [Changelog](../tree/master/lambdas/tabular_preview/CHANGELOG.md) entry

<h3>Greptile Summary</h3>

The PR makes zero-row tabular previews loadable by emitting one schema-bearing empty Arrow record batch when no data batches were written.
- Adds the empty-batch fallback to Arrow serialization.
- Covers header-only and entirely malformed CSV inputs.
- Documents the Catalog rendering fix in the changelog.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py | Adds a schema-preserving empty record batch when Arrow serialization otherwise produces no batches. |
| lambdas/tabular_preview/tests/test_index.py | Adds coverage for header-only CSV and inputs where every malformed row is skipped. |
| lambdas/tabular_preview/CHANGELOG.md | Records the zero-row preview compatibility fix. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A[Tabular input] --> B[Convert table to record batches]
  B --> C{Any batch written?}
  C -->|Yes| D[Finish Arrow IPC file]
  C -->|No| E[Write schema-bearing empty batch]
  E --> D
  D --> F[Compress response]
  F --> G[Catalog renders table]
```

<sub>Reviews (2): Last reviewed commit: ["Drop the pinned version from the comment"](https://github.com/quiltdata/quilt/commit/eb778beda0cfad3262399f19b8cf88f7a5e0d347) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56643061)</sub>

**Context used:**

- Knowledge Base — [Data preview processing](https://app.greptile.com/quilt-bio/-/custom-context/knowledge-base/quiltdata/quilt/-/docs/data-preview-processing.md)

### DIFF
--- a/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
+++ b/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
@@ -121,7 +121,7 @@ def write_data_as_arrow(data, schema, max_size):
             if not writer.stats.num_record_batches:
                 # perspective's arrow_loader.cpp calls Table::FromRecordBatches
                 # without a schema, so it rejects a batch-less file. Fixed in
-                # 3.2.0 by finos/perspective#2854; the catalog pins 1.9.4.
+                # 3.2.0 by finos/perspective#2854.
                 writer.write(pyarrow.RecordBatch.from_pylist([], schema=schema))
 
     return memoryview(buf.getvalue()), truncated
--- a/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
+++ b/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
@@ -104,6 +104,7 @@ def write_data_as_arrow(data, schema, max_size):
         data = data.to_batches(OUT_BATCH_SIZE)
 
     truncated = False
+    written = False
     buf = pyarrow.BufferOutputStream()
     with pyarrow.CompressedOutputStream(buf, "gzip") as sink:
         with pyarrow.ipc.new_file(sink, schema) as writer:
@@ -117,6 +118,18 @@ def write_data_as_arrow(data, schema, max_size):
                     truncated = True
                     break
                 writer.write(batch)
+                written = True
+
+            if not written:
+                # A file carrying a schema and no record batches is not loadable by
+                # the consumer: perspective builds its table from the batches and
+                # rejects a batch-less file outright ("Must pass at least one record
+                # batch or an explicit Schema"), so the preview fails instead of
+                # showing an empty table. Reachable whenever nothing gets written --
+                # a header-only CSV, one whose every data row was skipped as invalid,
+                # or a first batch already too large for max_size. Emit one empty
+                # batch so the schema arrives in a form that loads.
+                writer.write(pyarrow.RecordBatch.from_pylist([], schema=schema))
 
     return memoryview(buf.getvalue()), truncated
 
--- a/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
+++ b/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
@@ -119,12 +119,7 @@ def write_data_as_arrow(data, schema, max_size):
                 writer.write(batch)
 
             if not writer.stats.num_record_batches:
-                # A file carrying a schema and no record batches is not loadable:
-                # perspective builds its table from the batches and rejects one with
-                # none ("Must pass at least one record batch or an explicit Schema"),
-                # so the preview fails instead of showing an empty table. Reached by
-                # a header-only file, one whose every data row was skipped as
-                # invalid, or a first batch already too large for max_size.
+                # perspective rejects an IPC file with a schema and no batches
                 writer.write(pyarrow.RecordBatch.from_pylist([], schema=schema))
 
     return memoryview(buf.getvalue()), truncated
--- a/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
+++ b/lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
@@ -104,7 +104,6 @@ def write_data_as_arrow(data, schema, max_size):
         data = data.to_batches(OUT_BATCH_SIZE)
 
     truncated = False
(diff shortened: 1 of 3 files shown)


fiskus:
@greptileai please re-review

sir-sigurd (commented):
Reviewed at `eb778bed`. The premise checks out: `arrow_loader.cpp:65` in v1.9.4 really is `arrow::Table::FromRecordBatches(batches)` with no schema, so a batch-less file aborts; and a header-only CSV infers an all-`null` schema, which 1.9.4's `convert_type` maps to `DTYPE_STR`, so the empty batch does render. Two things I would change before merge, plus coverage notes.

**`src/t4_lambda_tabular_preview/__init__.py:121` — the truncated-first-batch case now renders a silently empty table.** When the loop breaks on the very first batch, the fallback fires and the response becomes a successful zero-row table with `truncated: true`. For h5ad that is worse than before: `preview_h5ad` is the only caller passing a real `max_size`, `showLoadMore` returns false for h5ad (`catalog/app/components/Preview/loaders/Tabular.tsx:214`), and the toolbar shows only "Partial preview" because `size` is 0 — so the user gets an empty gene table with no way to ask for more. Previously they got "Could not render tabular data": wrong, but not *silently* wrong. Writing the empty batch only when nothing was dropped would keep both cases honest. Reachability is low — a 100-row batch crosses the 100 KB `small` budget at roughly 125 float64 columns, so it takes a small matrix with an unusually wide `var` — but it is invisible when it does happen.

**`__init__.py:122` — the comment names the wrong release.** finos/perspective#2854 merged 2024-11-18 and shipped in **v3.1.6** later the same day; v3.2.0 is three weeks after that. The PR description says 3.2.0 as well. The comment exists to date the workaround's removal, so both are worth correcting.

**`__init__.py:125` — the fallback write skips the guards every other write honors.** Not reachable in practice (`preview_csv` passes `max_size=None`, and through h5ad you would need ~2000 `var` columns to cross 100 KB at ~48 B/column), so I would not hold the PR for it — but it is the one write that can push a payload past `max_size` without setting `truncated`, and that asymmetry is easy to trip over later.

Coverage, none of it blocking:

- `tests/test_index.py:143` — `test_preview_h5ad[True]` goes through the same zero-row path but asserts only `len(df) == 0`, which pyarrow satisfies from a batch-less file too. Revert the six added lines and it still passes, so the h5ad half of the fix is unpinned.
- `tests/test_index.py:67` — the third cause named in the description ("a first batch already too large for `max_size`") is untested, and it is the path where the unguarded write lives. Worth noting that for CSV `max_size` is `None`, so that break can only come from the 4 MB `MAX_OUT` guard — the `max_size` form is h5ad-only.
- `tests/test_index.py:76` — the new test drops the `csv`/`tsv` parametrization and the full-headers assertion its sibling carries. `Content-Type` is load-bearing here: `Tabular.tsx:118` picks the Arrow branch off it, and a regression there would leave the suite green.
- `tests/test_index.py:98` — `num_record_batches == 1` is a pyarrow-side proxy for a constraint that lives in perspective. pyarrow reads a batch-less file happily, which is exactly how this escaped; what actually holds is the `null` → `DTYPE_STR` mapping above, and nothing in either suite pins it.

One scope note on the changelog: the entry reads as though every zero-row preview now loads, but `write_pandas_as_csv` is untouched — an empty `.jsonl` still yields a bare `\n`, which perspective's CSV path rejects for the same reason. Worth either narrowing the entry to the Arrow path or fixing the sibling.

lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
```diff
@@ -118,6 +118,12 @@ def write_data_as_arrow(data, schema, max_size):
                     break
                 writer.write(batch)
 
+            if not writer.stats.num_record_batches:
```
sir-sigurd, line 121: Anchoring the main point from my review: when the loop above breaks on the *first* batch, `truncated` is already `True`, and this fallback then turns that into a successful zero-row table. h5ad is the only caller that passes a real `max_size`, and it gets no "Load more" (`catalog/app/components/Preview/loaders/Tabular.tsx:214`) and only a "Partial preview" label — so the user sees an empty gene table where they used to see an error. Gating the write on `not truncated` keeps both cases honest.

lambdas/tabular_preview/src/t4_lambda_tabular_preview/__init__.py
```diff
@@ -118,6 +118,12 @@ def write_data_as_arrow(data, schema, max_size):
...
+                # perspective's arrow_loader.cpp calls Table::FromRecordBatches
+                # without a schema, so it rejects a batch-less file. Fixed in
+                # 3.2.0 by finos/perspective#2854.
```
sir-sigurd, line 124: finos/perspective#2854 shipped in **v3.1.6** (published 2024-11-18, hours after the merge), not 3.2.0 — v3.2.0 is three weeks later. Same correction applies to the PR description.

PR #5219, opened 2026-08-26
catalog: seed the front door with varied example queries and real package handles

fiskus:
The chips under the front door's search bar are the bar's seed queries: clicking one types its label into the field, so the label *is* the query. Three fixes.

**Package handles reach the chip verbatim.** `packageKeyword()` stripped the handle to its last segment and turned hyphens into spaces, so `research/2026-08-24-trunk-based-development-practice` arrived as "2026 08 24 trunk based development practice" — not a name any search matches, and not what the recent-packages tile below shows for the same package. The handle now goes in whole, set in the mono face per the Mono Identity Rule. `label` stays a plain string (it is the chip's key and what `onSelect` types); the span to set apart rides alongside it as `code`.

**The prompts share the row round-robin.** `useExampleQueries(limit)` passed the chip count as the package count, so five recent revisions took every slot and the tag, bucket-title and fallback prompts could never render. `shareOut` gives each source a turn before any takes a second slot and hands a dry source's turns to the rest — so all three kinds show while there is room, and one still fills the row when it is the only one with anything to say.

**Glyphs are sized, and on the face the rest of the chrome wears.** `MuiChip` sizes its icon only for `size="small"`, so an `M.Icon` rendered at its own 24px default beside a 13px label in a 32px pill. Now 18px — `MuiChip`'s own `iconSmall` value — on the outlined face the rail and content bar already use; that font is already loaded in `index.html`, so no new asset. They stay bare marks rather than gaining a disc: `icon` and `avatar` share their margins, but only `avatar` carries a box, and a ground-bearing circle is the identity avatar's shape.

Scoped to the chips — the tiles below them are still on the filled face. Real handles are longer than the mangled ones, so the row still wraps; a `maxWidth` would truncate for free (`MuiChip`'s label already ellipsises) if that reads better than the wrap.

### Test plan

- `npm run typecheck` and `npm run lint:app` — clean
- `npx vitest run app/website/pages/Landing/FrontDoor` — 75 passed

Before:
Screenshot_2026-08-26_12-16-47

After:
Screenshot_2026-08-26_13-32-58

<h3>Greptile Summary</h3>

The PR updates the front-door example-query chips to preserve complete package handles, distribute slots among package, tag, and bucket prompts, and align chip icon styling with the surrounding interface.
- Adds round-robin allocation across query sources with generic fallback prompts.
- Renders package handles in monospace while preserving the complete query submitted by a chip.
- Adds focused tests for allocation, rendering, selection, and icon styling.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/website/pages/Landing/FrontDoor/useExampleQueries.ts | Builds package, tag, and bucket prompts and allocates the available chip slots among those sources. |
| catalog/app/website/pages/Landing/FrontDoor/ExampleQueries.tsx | Adds outlined, explicitly sized icons and monospace rendering for package-handle substrings without changing the selected query string. |
| catalog/app/website/pages/Landing/FrontDoor/useExampleQueries.spec.ts | Covers verbatim package handles, source sharing, single-source fallback behavior, and the allocation helper. |
| catalog/app/website/pages/Landing/FrontDoor/ExampleQueries.spec.tsx | Covers themed rendering, outlined icons, handle styling, and full-query selection behavior. |

<sub>Reviews (3): Last reviewed commit: ["catalog: seed the front door with varied..."](https://github.com/quiltdata/quilt/commit/18736586a4b1ea8eb7261dadf0dc852788dc493a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57026512)</sub>


fiskus:
@greptileai please re-review

fiskus:
@greptileai please re-review

PR #5220, opened 2026-08-26
catalog: stop the volume row's access readout from covering its tags

fiskus:
On the volumes list, a bucket's tag chips are drawn over by the "Shared with <n>+" readout on the right of the row. It hits every tagged row, not just heavily tagged ones.

`ListItemSecondaryAction` is absolutely positioned, so the row can only make room for it by reserving a fixed right-hand gutter — `paddingRight: t.spacing(14)`, 112px. The readout is `Shared with ` plus a count plus an optional `+`, which is wider than that, so it hangs left over whatever is at the row's content edge. The tags are what's there.

The readout now sits in the row's flow as an ordinary flex child, so the two cannot overlap whatever the count reads, and the reserved gutter is gone. `DataProductRow` had the same slot with the same fixed gutter and gets the same treatment.

While in there: the readout was spacing itself twice. `root` is a flex row with `gap: t.spacing(0.5)`, and the label *also* ended in `&nbsp;` — so the glyph sat 4px from "Shared with" while the count sat ~8px from it. Label and count are now one text run separated by the sentence's own word space, with the flex gap left to do only what it is for. `whiteSpace: nowrap` moves onto the readout itself so it cannot break between the words in either consumer; both wrappers are `flexShrink: 0`, so it will not be squeezed.

### Test plan

- `npx vitest run app/containers/Home` — 38 passed
- New `BucketRows.spec.tsx` (the file had no coverage): asserts the row renders no `.MuiListItemSecondaryAction-root` while both the readout and the tags are present. Confirmed it fails against the pre-fix component and passes after — jsdom does no layout, so the mechanism is what can be pinned, not the overlap itself.

Before:
Screenshot_2026-08-26_14-32-24

After:
Screenshot_2026-08-26_14-31-52

<h3>Greptile Summary</h3>

The PR moves bucket collaborator and data-product access readouts into normal flex flow so they no longer overlap adjacent tags, and consolidates collaborator label/count spacing. It also adds a regression test that verifies the bucket readout no longer uses Material UI’s absolutely positioned secondary-action slot.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Home/BucketGrid/BucketRows.tsx | Replaces fixed-gutter secondary actions with non-shrinking in-flow access elements for bucket and data-product rows. |
| catalog/app/containers/Home/BucketGrid/Collaborators.tsx | Consolidates the collaborator label and count into one non-wrapping text run while preserving count emphasis. |
| catalog/app/containers/Home/BucketGrid/BucketRows.spec.tsx | Adds coverage confirming that tags and the access readout render without the absolute secondary-action wrapper. |
| catalog/CHANGELOG.md | Documents the corrected volume-row overlap and spacing behavior. |

<sub>Reviews (2): Last reviewed commit: ["catalog: stop the volume row&#39;s access re..."](https://github.com/quiltdata/quilt/commit/36b8a790f5bd4f738650c3d44e7f255d1da79025) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57046413)</sub>

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Fixed] Volumes list: the "Shared with" readout no longer overlaps a bucket's tags, and the extra space inside it is gone ([#5220](https://github.com/quiltdata/quilt/pull/5220))
 - [Fixed] Front door: the example-query chips show real package handles, a mix of prompts rather than five recent packages, and icons that match the rest of the app ([#5219](https://github.com/quiltdata/quilt/pull/5219))
 - [Changed] Bucket identity tints go from six to fifteen, so a wall of same-prefix volumes reads as distinct objects rather than a few colors repeating. Same scheme — a pale ground with a dark ink of its own hue, initials clearing AA — in two tiers so a hue family can contribute twice; the accent, semantic and error hues stay excluded ([#5210](https://github.com/quiltdata/quilt/pull/5210))
 - [Fixed] Admin Buckets and role bucket-permission lists show the same per-bucket mark the volumes landing does — a bucket with no custom icon wears its tinted initials disc instead of an anonymous glyph stub, so one bucket reads as one object on every screen ([#5210](https://github.com/quiltdata/quilt/pull/5210))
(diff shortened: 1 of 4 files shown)


fiskus:
@greptileai please re-review

PR #5221, opened 2026-08-26
catalog: drop the white ground behind the metadata order selectors

fiskus:
In the search sidebar's package-level metadata section, the two "Order by" controls sit on white patches of their own rather than on the panel's ground.

`Filters/Select` paints `background: t.palette.background.paper` on its root. Those two controls are its only callers, so the rule goes with the component rather than being overridden at the call site — which also takes the now-empty `useStyles` and the `cx` import with it.

### Test plan

- `npx vitest run app/containers/Search app/components/Filters` — 112 passed

<h3>Greptile Summary</h3>

This PR removes the Material UI paper background from the metadata-order selectors so they inherit the search sidebar panel ground.
- Removes the component-local `makeStyles` background rule.
- Removes the now-unused `classnames` dependency and forwards the caller’s `className` directly.

The PR appears safe to merge with no actionable defects identified.

The change is limited to removing an unwanted component background and redundant class merging while preserving caller classes and all selection behavior.

| Filename | Overview |
|----------|----------|
| catalog/app/components/Filters/Select.tsx | Removes the root background style and obsolete class merging without changing the selector’s value or event behavior. |

<sub>Reviews (1): Last reviewed commit: [aabd8c7](https://github.com/quiltdata/quilt/commit/aabd8c729c364ccae8c2215038f5279f349294da) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57067406)</sub>

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Fixed] Search sidebar: the metadata "Order by" selectors no longer sit on a white patch of their own ([#5221](https://github.com/quiltdata/quilt/pull/5221))
 - [Fixed] Volumes list: the "Shared with" readout no longer overlaps a bucket's tags, and the extra space inside it is gone ([#5220](https://github.com/quiltdata/quilt/pull/5220))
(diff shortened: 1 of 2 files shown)


fiskus:
Superseded by #5222, which rewrites this control and needs the same deletion — folded it in there so the two do not conflict on the same lines. Closing this in favour of that one.

PR #5222, opened 2026-08-26
catalog: sort the package metadata list from one control, next to the list

fiskus:
The search sidebar offered the metadata ordering as two selects — `Order by [Name ▾] [A → Z ▾]` — right-aligned above the "Find metadata" field. Three things came out of that: the control that orders the list sat apart from the list, behind a text field and the "Some metadata not displayed" paragraph; it was the only right-aligned element in a left-aligned panel; and at caption size with `disableUnderline` the pair read as grey text with chevrons rather than as a control.

It is now a single `Sort by:` select carrying the four combinations — `Name A → Z`, `Name Z → A`, `Type A → Z`, `Type Z → A` — in the panel's own alignment, directly above the list it orders.

`Filters/Select` also painted `background: t.palette.background.paper` on its root, so the control sat on a white patch rather than the panel's ground. These two are its only callers, so the rule goes with the component — **this supersedes #5221**, which did only that; close it in favour of this one, since both touch the same lines and would conflict.

### On folding the two axes

`model.ts` argued against exactly this:

> `// What the list is ordered by, and which way: two axes, so two controls. Folded`
> `// into one four-entry list they read as four unrelated modes rather than a choice`
> `// plus a direction.`

The counter-argument is on the same page. The result list's own "Sort by" is `PRESET_ORDERINGS` — `Best match`, `Most recent first`, `Least recent first`, `A → Z`, `Z → A` — which is field and direction folded into one flat list of five, and reads fine. The volumes list does the same. DESIGN.md asks for one control vocabulary ("a control means one thing on every screen"), and the facet panel was the exception. Four options is one fewer than the control it now matches.

The direction labels keep their arrows, which the old comment noted are shared with `PRESET_ORDERINGS` on purpose.

### Unchanged

The `mo` param still serializes as `<by>:<direction>` and still falls back per axis, so existing links keep working and `mo=type:sideways` still orders by type. `FACET_ORDER_BY` / `FACET_ORDER_DIRECTIONS` stay as the two axes the param validates against; `FACET_ORDERINGS` is the panel's view of their combinations.

### Test plan

- `npx vitest run app/containers/Search app/components/Filters` — 113 passed
- The model spec's ordering test now pins that the offered list is exactly every combination of the two axes, each once, and that the default is among them — so adding an axis without extending the list fails.

Before:
Screenshot_2026-08-26_15-26-27

After:
Screenshot_2026-08-26_15-26-32

<h3>Greptile Summary</h3>

The PR consolidates package-metadata ordering into one adjacent “Sort by” control while preserving the existing URL representation and per-axis fallback behavior.
- Adds four explicit metadata-ordering combinations to the search model.
- Replaces the separate field and direction controls with one combined select.
- Removes the shared select’s paper-colored background and updates ordering tests and the changelog.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/components/Filters/Select.tsx | Removes the paper background from the shared select root; no eligible blocking regression was established. |
| catalog/app/containers/Search/Layout/PackageFilters.tsx | Replaces two metadata-ordering selects with one combined control whose values remain aligned with model parsing and serialization. |
| catalog/app/containers/Search/model.ts | Adds the complete four-entry ordering preset list without changing the URL format or fallback semantics. |
| catalog/app/containers/Search/model.spec.ts | Verifies that presets cover every ordering-axis combination exactly once and include the default. |
| catalog/CHANGELOG.md | Documents the consolidated metadata sorting control. |

<sub>Reviews (2): Last reviewed commit: ["catalog: sort the package metadata list ..."](https://github.com/quiltdata/quilt/commit/bc5c5484ea0199c7f104bd55ec1e9154a009dfba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57070845)</sub>

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Changed] Search sidebar: the package metadata list is sorted from one "Sort by" control, sitting directly above the list ([#5222](https://github.com/quiltdata/quilt/pull/5222))
 - [Fixed] Volumes list: the "Shared with" readout no longer overlaps a bucket's tags, and the extra space inside it is gone ([#5220](https://github.com/quiltdata/quilt/pull/5220))
 - [Fixed] Front door: the example-query chips show real package handles, a mix of prompts rather than five recent packages, and icons that match the rest of the app ([#5219](https://github.com/quiltdata/quilt/pull/5219))
 - [Changed] Bucket identity tints go from six to fifteen, so a wall of same-prefix volumes reads as distinct objects rather than a few colors repeating. Same scheme — a pale ground with a dark ink of its own hue, initials clearing AA — in two tiers so a hue family can contribute twice; the accent, semantic and error hues stay excluded ([#5210](https://github.com/quiltdata/quilt/pull/5210))
--- a/catalog/app/components/Filters/Select.tsx
+++ b/catalog/app/components/Filters/Select.tsx
@@ -1,12 +1,5 @@
-import cx from 'classnames'
 import * as React from 'react'
 import * as M from '@material-ui/core'
-
-const useStyles = M.makeStyles((t) => ({
-  root: {
-    background: t.palette.background.paper,
-  },
-}))
 
 interface SelectFilterProps<T> {
   className?: string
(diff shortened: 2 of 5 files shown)


fiskus:
@greptileai please re-review

PR #5224, opened 2026-08-26
catalog: pin the admin form's actions with position: sticky

fiskus:
On Admin → Buckets, the sticky Cancel/Add bar ends short of the cards above it.

`StickyActions` wrapped itself in its own `M.Container maxWidth="lg"`, centred on the viewport with `left: 50%` + `translateX(-50%)`. The app's shared container is `maxWidth={false}` with `disableGutters` — "every view is effectively full-width now" (`components/Layout/Container.tsx`) — so on any viewport wider than `lg` the bar sat in a 1280px box while the cards ran to the page edge.

### Why `position: sticky` and not a wider box

Holding the bar in place took a scroll listener, a resize observer, two debounced `DOMRect`s, a viewport-geometry predicate and a height placeholder — about ninety lines to reproduce what `position: sticky; bottom: 0` does in one declaration. It also travels within its own container, so it takes that container's width without being told it.

The alternative of hardcoding the inset (`calc(100% - 280px - 24px)`) doesn't hold: the rail is not always a column. Per `Layout.tsx`, "under 960px there is no room for a 256px column beside the content, so the rail becomes an overlay". That needs a media query mirroring the shell's breakpoint, and pins three constants this component doesn't own — the rail's width, that breakpoint, and Admin's `spacing(0, 3)`. Sticky needs none of them.

### The one thing CSS can't express

Whether the bar is *currently* pinned, which decides the shadow — a shadow means "this floats above the plane and will leave" (the Overlay-Only Rule), and once it lands it shouldn't have one. An `IntersectionObserver` on a 1px marker just past the bar's resting place answers exactly that: out of view for as long as the bar is pinned. 1px rather than 0 because a zero-area target's intersection is unreliable; pulled back with `marginTop: -1` so it occupies nothing.

### Also gone

The anchor refs the measurement needed (`parentRef` and its five call sites), and `use-debounce` / `use-resize-observer` from this module. The pinned state was an `M.Paper elevation={8}`; it's now the same shadow and radius on the bar itself, so there's one element instead of two swapping places. The slide-in keyframes are dropped — shadow and padding transition instead.

Net: **-92 / +47**.

### Test plan

- `npx vitest run app/containers/Admin` — 30 passed

**Needs an eyeball.** None of this is testable in jsdom, which computes no geometry and has no `IntersectionObserver`. Worth checking on Add a bucket and on the edit page's per-card bars:

- the bar spans the cards' width, at a wide viewport and below 960px
- it pins while scrolling and lands at the end of the form without jumping
- the shadow appears while pinned and goes when it lands
- the per-card bars on the edit page still pin within their own card

<h3>Greptile Summary</h3>

The PR replaces JavaScript-driven fixed positioning for Admin bucket form actions with CSS sticky positioning and uses a sentinel to control the floating visual treatment.
- Makes action bars inherit the width of their form or card container.
- Removes scroll, resize, and geometry-measurement state.
- Adds a changelog entry for the alignment fix.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Admin/Buckets/Buckets.tsx | Reworks Admin bucket action bars around CSS sticky positioning and IntersectionObserver-based presentation state; no follow-up-eligible blocking issue was established. |
| catalog/CHANGELOG.md | Documents the corrected alignment of the Admin bucket action bar. |

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;master&#39; into dx/sticky-act..."](https://github.com/quiltdata/quilt/commit/1d890714954b191315baf9992fd0baa18451206b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57097269)</sub>

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Fixed] Admin buckets: the sticky Cancel/Add bar lines up with the form above it ([#5224](https://github.com/quiltdata/quilt/pull/5224))
 - [Removed] The unbaked data-products GraphQL contract (added in [#5203](https://github.com/quiltdata/quilt/pull/5203), never served by a registry) is out of the schema; the `data-products` preview UI is unaffected and keeps reading fixture data ([#5223](https://github.com/quiltdata/quilt/pull/5223))
 - [Changed] Search sidebar: the package metadata list is sorted from one "Sort by" control, sitting directly above the list ([#5222](https://github.com/quiltdata/quilt/pull/5222))
 - [Fixed] Volumes list: the "Shared with" readout no longer overlaps a bucket's tags, and the extra space inside it is gone ([#5220](https://github.com/quiltdata/quilt/pull/5220))
--- a/catalog/app/containers/Admin/Buckets/Buckets.tsx
+++ b/catalog/app/containers/Admin/Buckets/Buckets.tsx
@@ -6,8 +6,6 @@ import * as R from 'ramda'
 import * as React from 'react'
 import * as RF from 'react-final-form'
 import * as RRDom from 'react-router-dom'
-import { useDebounce } from 'use-debounce'
-import useResizeObserver from 'use-resize-observer'
 import * as M from '@material-ui/core'
 import * as Lab from '@material-ui/lab'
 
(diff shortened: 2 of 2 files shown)


fiskus:
@greptileai please re-review

PR #5226, opened 2026-08-26
catalog: bucket header lays out on explicit responsive tiers

QuiltSimon:
## What

Replaces the bucket header's emergent flex-wrap with three designed responsive states, per the "deliberate breakpoints" pass done in impeccable live mode:

- **Wide**: one calm grid row — title | stats | settings. The readout strip never wraps; the title truncates first (full name in a tooltip).
- **Medium (≤1000px)**: title + settings hold row one; the stats strip gets its own full-width row.
- **Narrow (≤640px)**: readouts settle into a two-column grid with the create-package action spanning full width below.

The admin settings sprocket moves off the bucket title to the card's far edge, behind a `palette.divider` hairline, `text.secondary` warming on hover, with an arrowed tooltip — config set apart from the bucket's readout.

## Why

At intermediate widths the old single flex row had no gap and no designed intermediate state: the sprocket overlapped the title, "Create package" got crushed, and wrap points were emergent rather than chosen.

## Verified

- `npm run typecheck` clean
- `Header.spec.tsx` 9/9 green
- Exercised live against the running dev server across all three tiers (variants cycled + accepted in impeccable live mode)

## Related

Sibling PR (same design pass, package top bar): dx/pkg-topbar-tiers — will cross-link once open.

<h3>Greptile Summary</h3>

The PR replaces the bucket header’s flex wrapping with three explicit responsive grid tiers and moves the administrator settings action into a separated edge column.

- Adds wide, medium, and narrow layouts for the title, readouts, and actions.
- Truncates long bucket names while retaining the full value in a tooltip.
- Adds a changelog entry for the redesigned header.

The responsive redesign needs adjustment before merging because the desktop sidebar can leave the wide, non-wrapping header active when its actual container is too narrow.

The new viewport-based breakpoint does not account for the persistent sidebar and nested horizontal insets, allowing the intrinsic-width stats and settings tracks to overflow and be clipped at realistic intermediate desktop widths.

**Files Needing Attention:** catalog/app/containers/Bucket/Header.tsx

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Bucket/Header.tsx | Introduces the explicit responsive grid and settings treatment, but the viewport breakpoint can leave the non-wrapping wide tier active inside an undersized desktop content column. |
| catalog/CHANGELOG.md | Accurately documents the intended responsive header behavior and settings-control relocation. |

<sub>Reviews (1): Last reviewed commit: ["catalog: bucket header lays out on expli..."](https://github.com/quiltdata/quilt/commit/e6f7c452f55ca5de475e8befb97987df78214a90) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57297401)</sub>

**Context used:**


QuiltSimon:
Sibling PR from the same design pass (package top bar): #5227

QuiltSimon:

Review ran against b0732d0 (10 findings, adversarially verified); fixes landed in 9ab9734. Eight fixed, two dismissed in writing:

1. **FIXED — tier cutoffs measured the viewport, but the header's width is viewport − shell chrome (256px rail + paddings).** The never-wrap tier engaged at 1001–1300px, exactly where it *cannot* fit (~650–950px of card), pushing Create package and the sprocket past the card edge. The one-row tier now starts at ≥1300px (≈950px of card, always fits) and everything below uses the stacked tiers, which tolerate any width. Cutoffs use `t.breakpoints.down(N)` per house convention — these were the only raw `@media(max-width)` literals in catalog against 101 uses of the helper.

2. **FIXED — non-admins got a phantom 24px gutter.** The grid declared the settings track unconditionally while the control renders only for admins; columnGap applies between tracks regardless. The settings column now exists only via a `withSettings` modifier class applied with the control.

3. **FIXED — readouts overflowed their cells on phones.** `repeat(2, minmax(0,1fr))` forced ~128px columns that StatsItem can't shrink into. Now `repeat(auto-fit, minmax(140px,1fr))` — wide readouts wrap to fewer columns instead of colliding.

4. **FIXED — truncated titles were unreadable on touch/keyboard.** The title keeps ellipsis + hover tooltip where hover exists and wraps naturally at ≤640px, where screens are mostly touch.

5. **FIXED — `'& > div'` / `'& > button'` structural reach-ins.** Stats' tier rules moved into its own `makeStyles` sheet and the create button carries a real class — no more coupling to Stats' root element or MUI Button's rendered tag.

6. **FIXED — the settings control had no accessible name.** MUI v4 Tooltip only sets `aria-describedby` while open and `M.Icon` is `aria-hidden`; the IconButton now carries `aria-label="Bucket settings"`.

7. **FIXED — long-press tooltip could ride into navigation on touch.** `disableTouchListener` on the tooltip; the label reaches touch AT via the aria-label instead.

8. **FIXED — the spec never rendered the branch this PR rewrote.** `isAdmin` mock parameterized; both settings branches now covered (11/11 green, was 9 tests).

9. **DISMISSED — dead `flexWrap`/`justifyContent` in `useStatsStyles`:** no longer applicable; the restructure in (5) made `useStatsStyles` the single owner of those declarations, so nothing is shadowed anymore.

10. **DISMISSED (accepted trade-off) — 26px hit target on the sprocket.** Real, but `size="small"` matches the deliberate quiet-corner design and the app's stated a11y posture is best-effort with keyboard focus as the floor (which the control has). Bumping the whole affordance to 44px is a product-level density decision — the same one already deferred app-wide (small icon buttons throughout toolbars).

Re-verified after fixes: `npm run typecheck` clean, `Header.spec.tsx` 11/11, webpack dev build clean.

PR #5227, opened 2026-08-26
catalog: package top bar degrades on two explicit tiers

QuiltSimon:
## What

Replaces the package-page top bar's emergent flex crush with two designed responsive states, from the same impeccable-live "deliberate breakpoints" pass as #5226:

- **Wide**: one grid row — crumbs | action cluster. Crumbs truncate/wrap first; the cluster (selected-items, Revise package, push, get, overflow) never squashes.
- **Stacked (≤1100px viewport)**: the cluster moves onto its own full-width row below the crumbs, **right-anchored** — keeping the wide tier's home edge so the often icon-collapsed strip doesn't sit parked at the left with dead space — wrapping with even 8px gaps so it never overflows. Crumbs lose their `calc(100% - 160px)` width caps and take the full row.

An `:empty` guard keeps bare `<TopBar crumbs={...} />` renders (loading/error states) from leaving a phantom second row.

## Why

At medium widths the non-shrinking action cluster crushed the breadcrumbs; below that, the stacked strip sat left-parked with no designed intermediate state. Three alternatives were cycled in live mode; the right-anchored tier won.

## Verified

- `npm run typecheck` clean
- Exercised live against the running dev server across both tiers (three variants cycled, this one accepted)
- `TopBar` is shared by the dir view, file view, and bare loading/error renders — all covered by the `:empty` guard

## Related

Sibling PR from the same design pass (bucket header): #5226

<h3>Greptile Summary</h3>

The PR replaces the package-page top bar's flex layout with explicit wide and stacked responsive grid tiers.
- Keeps the action cluster in a non-shrinking wide column and moves it below the breadcrumbs at widths up to 1100px.
- Allows stacked actions to wrap with consistent gaps and right alignment.
- Adds an empty-content guard for bare loading and error top bars.
- Documents the responsive-layout change in the Catalog changelog.

The PR appears safe to merge after addressing two non-blocking polish issues: phantom spacing in bare wide top bars and the unresolved changelog link.

The responsive behavior is otherwise localized and coherent, but the explicit wide grid retains a gap when actions are hidden, and the changelog points to a placeholder rather than PR 5227.

**Files Needing Attention:** catalog/app/containers/Bucket/PackageTree/PackageTree.tsx, catalog/CHANGELOG.md

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Bucket/PackageTree/PackageTree.tsx | Introduces explicit responsive grid tiers; the empty wide state still reserves the grid's column gap. |
| catalog/CHANGELOG.md | Adds the package-top-bar entry but leaves its pull-request number and URL as placeholders. |

<sub>Reviews (1): Last reviewed commit: ["catalog: package top bar degrades on two..."](https://github.com/quiltdata/quilt/commit/b093ec59b40fee87204cd81dd93c4b04fbf0a9d2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57298143)</sub>


QuiltSimon:

Review ran against b0732d0; fixes landed in de4f712. Four findings, three fixed, one dismissed in writing:

1. **FIXED — hardcoded `@media (max-width: 1100px)` vs house convention.** Both queries now use `t.breakpoints.down(1100)`, the pattern `Workflows/Detail.tsx` and `Workflows/List.tsx` already use for this exact threshold. (Verified: MUI v4 `down(number)` emits `max-width: 1099.95px` — the 0.05px boundary shift is intentional de-duplication with the existing convention.)

2. **FIXED — stacked-tier margin reset was insertion-order fragile.** `.content > *` ties with `.button` at (0,1,0) specificity, and JSS inserts the button sheets later, so the reset silently lost — the stacked tier ran with 16px effective gaps and stray ±3px vertical margins instead of the even 8px the comment claimed. Now `&& > *` (doubled class, (0,2,0)) wins regardless of sheet order.

3. **FIXED — mixed margin/gap spacing model made the crumbs seam uneven between views.** Dir view's first action carries `marginLeft: 8` (→ 24px seam), file view's doesn't (→ 16px). A `&& > :first-child { marginLeft: 0 }` normalizes both to the grid's 16px columnGap; intra-cluster margins between actions are untouched.

4. **DISMISSED (follow-up, not this PR) — the same crumbs+actions bar exists unshared in `Dir.tsx` and `File.jsx`.** Real duplication with the same crush failure, but no existing shared mechanism was ignored — this PR's `TopBar` is a self-contained private component that lifts cleanly to `components/` later. One-view scope matches the PR title; extraction is a separate change.

Also re-verified after fixes: `npm run typecheck` clean, webpack dev build clean.

PR #5230, opened 2026-08-26
Catalog: replace layout-property animations with compositor-friendly motion

QuiltSimon:
## What

From an impeccable design audit: four catalog sites animated layout properties (margin / width / max-height), which forces reflow on every animation frame. Each is replaced with compositor-friendly motion (or the layout transition is dropped where no compositor-safe equivalent exists), preserving the interactive behavior:

- **`components/Collaborators/Avatars.tsx`** — the avatar-stack hover spread animated `margin` on the nested wrappers and the "N+" label. The overlap margins are now static and the spread animates `transform: translateX` instead: the wrappers are nested, so the per-level 10px shift compounds with depth, reproducing the cumulative effect of the old `-8px → 2px` margin animation, and the label's `-8px` counter-shift matches its old `12px → 4px` margin move (guarded with `:not($moreIconized)` since the iconized label never moved). The `row-reverse` z-index workaround is untouched.
- **`components/Dialog/PopoverOptions.tsx`** — the Tabs root transitioned `width` (live, not dead code: `GetOptions` sets per-tab widths of 320px/640px via the tab `className`). A container's layout width cannot animate off the main thread, and scaling would distort the contents — so the transition is dropped and the tab switch is covered by the panel's existing 150ms opacity fade. Per-tab widths still apply; only the resize is instant now.
- **`containers/Auth/StrenghtenPasswordField.tsx`** — the strength meter's `::after` fill animated `width` (10% → 100%). It now spans full width and animates `transform: scaleX(0.1 … 1)` with `transform-origin: left`; same fill fractions, same colors.
- **`containers/Search/List/Hit.tsx`** — the preview expand/collapse transitioned `max-height` on the contents (`106px ↔ 80vh`) and inner image (`20vh ↔ 80vh`). Both transitions are dropped: they animated layout across the whole card, and the state change is already signaled by the (compositor-safe) expand-icon rotation and overlay fade. Expand/collapse behavior itself is unchanged.

The transitions this PR touches honor `prefers-reduced-motion: reduce` (states still apply; the motion is skipped).

## Why

Layout-property animations re-run layout and paint on the main thread every frame (layout thrash); transform/opacity run on the compositor. Flagged by an impeccable design audit; `catalog/DESIGN.md` remains binding (no new decoration, 150–250ms standard easing kept where motion remains).

## Verified

- `cd catalog && npm run typecheck` — clean on this branch.
- `npm run test:only -- app/containers/Search/List/Hit.spec.tsx app/containers/Search/List/index.spec.tsx` — 17/17 pass on this branch.
- Changelog entry added; PR number will be baked in a follow-up commit.

<h3>Greptile Summary</h3>

The PR replaces layout-triggering Catalog animations with compositor-friendly transforms where equivalent behavior is available and removes width or max-height transitions where it is not. It also adds reduced-motion handling for the newly transformed avatar and password-strength animations.

- Reworks nested collaborator-avatar spreading from animated margins to cumulative transforms.
- Makes the password-strength fill use `scaleX` while preserving its existing fill fractions.
- Drops animated popover width and preview max-height changes while retaining their final layout states.
- Documents the animation-performance change in the Catalog changelog.

The PR appears safe to merge, with no concrete changed-code regression identified.

The transformed avatar and strength-meter states preserve the prior final geometry, while the removed width and max-height transitions leave their state-dependent layouts and content visibility intact.

| Filename | Overview |
|----------|----------|
| catalog/app/components/Collaborators/Avatars.tsx | Replaces nested margin animation with equivalent cumulative transforms and skips the new motion for reduced-motion users. |
| catalog/app/components/Dialog/PopoverOptions.tsx | Removes the Tabs container’s width transition while preserving per-tab widths and the existing panel presentation. |
| catalog/app/containers/Auth/StrenghtenPasswordField.tsx | Replaces percentage-width strength fills with equivalent left-origin scale transforms and disables the transformed fill transition under reduced motion. |
| catalog/app/containers/Search/List/Hit.tsx | Removes max-height transitions without changing collapsed or expanded dimensions, overflow, or interaction state. |
| catalog/CHANGELOG.md | Adds an entry describing the compositor-friendly animation changes; its PR-number placeholder is explicitly scheduled for a follow-up. |

<sub>Reviews (1): Last reviewed commit: ["Catalog: replace layout-property animati..."](https://github.com/quiltdata/quilt/commit/8390d9969eb9c7af6ccfa5edb5453992597d996c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57301211)</sub>


QuiltSimon:

Review ran against 8390d99 (8 finder angles, adversarially verified — one candidate refuted: the suspected transform stacking-context paint inversion doesn't happen, MuiAvatar is `position: relative`). Fixes landed in 0c3fd08. Ten findings → six fixed, three resolved by one revert, one dismissed in writing:

1. **FIXED — the comment's justifying fade never played.** `TabPanel` was unkeyed, so React reused the DOM node across tab switches and the `$show` animation ran only on mount — the popover snapped 320↔640px with no motion at all. Now `key={activeIndex}` remounts the panel and the cross-fade the comment describes actually happens.

2. **FIXED — `@keyframes activate` was orphaned in the wrong JSS sheet.** JSS resolves `$keyframe` refs per StyleSheet; the underline animation had *never played* (and logged a JSS warning on every mount). Keyframes moved into `useTabStyles` next to their reference.

3. **FIXED — reduced-motion guard desynced the strength meter.** The `:after`-scoped `transition: none` killed the fill's color fade while the track underneath kept fading 300ms — the two halves of one meter disagreed in hue mid-transition. The reduce branch now keeps the color cross-fade (color isn't motion) and drops only the transform tween.

4. **FIXED — hardcoded `.3s ease` literals.** The meter now uses `t.transitions.create(...)` per the 20-call-site house convention.

5. **FIXED — search-hit veil faded over snapped content.** With the max-height transition gone, the `fadeOver` opacity transition was animating over geometry that had already jumped. The veil now snaps with the layout; the 150ms chevron rotate (compositor-safe, an affordance not geometry) stays.

6. **FIXED by revert — `Avatars` is a dead component.** Zero call sites (its last consumer was replaced by a text readout in #5189); the margin→transform rewrite polished unreachable UI, and carried two latent defects the review confirmed (hover spread escaping the root's gradient box; reduce-mode teleport). Reverted to master, changelog claim dropped. Deleting the component outright is a separate cleanup — flagged below.

7. **FIXED — `overflow: hidden` left unexplained.** Comment now states what the clip is for post-transition-removal.

8. **DISMISSED (follow-up, not this PR) — reduced-motion is hand-patched per site while ~50 declarations across 35 files stay unguarded, with two competing conventions already in-tree.** Real and systemic, but a global policy (a `@media (prefers-reduced-motion)` rule in `global-styles.tsx` or a `reducedMotion` mixin next to `mixins.lineClamp`, plus a DESIGN.md doctrine entry) is its own change with app-wide blast radius — wrong to smuggle into a four-file audit PR. Follow-ups worth filing: the global reduce policy, and deleting the dead `Avatars` component + its barrel export.

Re-verified after fixes: `npm run typecheck` clean, webpack dev build clean. The changelog entry was rewritten to match what the PR now actually does.

PR #5233, opened 2026-08-27
catalog: contain a malformed search URL on the search page, and name the filter

fiskus:
# Description

Split out of #4463. Independent of #5232.

A malformed filter value in a search URL currently takes down the entire catalog. This contains it to the search page and says what went wrong.

## The failure

`Predicates.Datetime`, `Predicates.Number` and `Predicates.KeywordEnum` call `JSON.parse` on a raw filter value taken off the URL. `Filter.fromURLSearchParams` (`model.ts:522`) calls them with no guard, and that runs inside the `useMemo` in `useUrlState` — i.e. in `SearchUIModel.Provider`'s own render pass, which is *above* every error boundary search has. So the throw unwinds to `Errors.ErrorBoundary` in `app.tsx` and replaces the whole catalog with the app-level screen.

That screen renders a fixed `'Something went wrong'` for any non-credentials error — it never shows `error.message`. So the user gets a blank-slate failure page, and the only description of the real fault is a `SyntaxError` in the console naming a character offset into a string nobody printed.

The realistic trigger is not adversarial input. A valid date filter serializes to `modified={"gte":"2020-01-01T00:00:00.000Z","lte":null}`; paste that into Slack or an email, let it line-wrap, and the recipient clicks a URL truncated mid-value.

## What this changes

**Containment.** Both pages that mount the model — `Search/Search.tsx` and `Bucket/PackageList.tsx` — wrap the provider in an `ErrorBoundary`. The search UI is wholly reconstructable from the URL, so the failure is recoverable in place: keep the page chrome, show the error, offer reload or a clean search.

Two things the implementation is deliberate about, both learned from `Bucket/PanelBoundary.tsx`'s notes:

- The bad state lives in the **location**, above the boundary, so `resetErrorBoundary` alone would re-render the same throw. Navigating is the reset; `resetKeys={[location.search]}` clears the boundary when the URL actually changes. A retry that cannot succeed would be a dead affordance.
- Neither fallback uses `Search/Layout/Main`, because `Main` reads the search model — and a boundary does not catch what its own fallback throws. `NoResults.Error` is model-free, which is what makes it usable here; `NoResults.boundary.spec.tsx` pins that, and fails with `ESCAPED:` the moment the screen starts reading the model.

**The message.** The three parses go through one helper that throws a message naming the filter, and logs that message, the offending value verbatim, and the original `SyntaxError`. The offset is meaningless without the string it indexes into, so the console needed the value as much as the user needed the name.

| Param | Message |
|---|---|
| `modified` | `Invalid date range in the search URL` |
| `size`, `entries` | `Invalid number range in the search URL` |
| `workflow` | `Invalid keyword list in the search URL` |
| `comment`, `name`, `hash` | cannot throw — Text/KeywordWildcard never parse JSON |

The filter's URL param is the predicate key, unprefixed. Objects search shares `modified` and `size`, so `t=o&modified={` behaves the same.

## Reproducing

```
/search?modified={
/search?size={oops}
/search?workflow="a",,
/b/<bucket>/packages?modified={
```

Before: full-page app-level error screen reading `'Something went wrong'`. After: the search page, with `Unexpected error` / `Invalid date range in the search URL` and both exits live.

Before:
Screenshot_2026-08-27_18-23-11

After:
Screenshot_2026-08-27_18-22-59

## Tests

`model.spec.ts` — one case per filter at the predicate level, one driving `parseSearchParams` over whole query strings (pinning the param names above), one asserting the log carries the failing value, one asserting well-formed params parse as before. `NoResults.boundary.spec.tsx` — the error screen renders with no provider above it.

Every one was checked against a reverted implementation: the naming cases fail, the well-formed case stays green, and the boundary spec reports `ESCAPED:` when the screen reads the model.

Verified: `Search/` + `Bucket/` suites 443 passed / 5 skipped, `npm run typecheck` clean, `oxfmt` + `oxlint` clean. `Bucket/File/File.boundary.spec.tsx` fails as an import timeout under the full parallel run and passes in isolation — reproduced identically on unmodified `master`, so it is pre-existing flake, not this change.

# TODO

- [x] Unit tests

<h3>Greptile Summary</h3>

The PR contains malformed search-filter URLs within the affected search surface and provides filter-specific error messages and recovery actions.
- Adds page-level error boundaries around the global and bucket package search models.
- Adds model-independent fallback UI that can reload or navigate to a clean search URL.
- Wraps JSON parsing for date, number, and keyword filters with named errors and diagnostic logging.
- Adds parser and boundary regression coverage plus a changelog entry.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Search/model.ts | Centralizes malformed filter JSON handling into named errors while preserving valid filter parsing. |
| catalog/app/containers/Search/Search.tsx | Places an error boundary above the global search model and supplies model-independent recovery UI. |
| catalog/app/containers/Bucket/PackageList.tsx | Adds equivalent malformed-URL containment and recovery around the bucket package-list model. |
| catalog/app/containers/Search/NoResults.tsx | Adds fallback recovery behavior that reloads network failures or navigates to a clean base URL. |
| catalog/app/containers/Search/model.spec.ts | Covers named parser failures, diagnostics, whole-query parameter mapping, and valid inputs. |
| catalog/app/containers/Search/NoResults.boundary.spec.tsx | Verifies that the fallback error screen renders without the search model provider. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  URL[Search URL] --> Provider[Search model provider]
  Provider --> Parse[Parse filter parameters]
  Parse -->|Valid| Results[Render search results]
  Parse -->|Malformed JSON| Error[Throw named filter error]
  Error --> Boundary[Page-level error boundary]
  Boundary --> Fallback[Model-independent error screen]
  Fallback -->|Clean search| Navigate[Navigate without bad filters]
  Fallback -->|Reload| Reload[Reload current page]
```

<sub>Reviews (3): Last reviewed commit: ["catalog: say the changelog entry plainly"](https://github.com/quiltdata/quilt/commit/2d0a10c72b0e0192e8e07917c0daed9d4d7defa8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57513753)</sub>

**Context used:**

- Knowledge Base — [Catalog web application](https://app.greptile.com/quilt-bio/-/custom-context/knowledge-base/quiltdata/quilt/-/docs/catalog-application.md)

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Fixed] Search: a malformed filter value in a search URL reports which filter could not be read, rather than a raw JSON parser offset ([#5233](https://github.com/quiltdata/quilt/pull/5233))
 - [Fixed] Admin buckets: the sticky Cancel/Add bar lines up with the form above it ([#5224](https://github.com/quiltdata/quilt/pull/5224))
 - [Fixed] Volumes grid: a card's description no longer touches its tags ([#5225](https://github.com/quiltdata/quilt/pull/5225))
 - [Changed] Volumes: tag chips are outlined, and a volume's description is set at the body step rather than as fine print ([#5225](https://github.com/quiltdata/quilt/pull/5225))
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,7 +21,7 @@ complete sentence without it.
 
 ## Changes
 
-- [Fixed] Search: a malformed filter value in a search URL reports which filter could not be read, rather than a raw JSON parser offset ([#5233](https://github.com/quiltdata/quilt/pull/5233))
+- [Changed] Search: a malformed filter value in a search URL raises an error naming that filter, and logs the offending value, instead of only a raw JSON parser offset. Not yet visible on screen — the app-wide error boundary shows fixed text — but the console is diagnosable and the message is in place for search's own boundary ([#5233](https://github.com/quiltdata/quilt/pull/5233))
 - [Fixed] Admin buckets: the sticky Cancel/Add bar lines up with the form above it ([#5224](https://github.com/quiltdata/quilt/pull/5224))
 - [Fixed] Volumes grid: a card's description no longer touches its tags ([#5225](https://github.com/quiltdata/quilt/pull/5225))
 - [Changed] Volumes: tag chips are outlined, and a volume's description is set at the body step rather than as fine print ([#5225](https://github.com/quiltdata/quilt/pull/5225))
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,7 +21,7 @@ complete sentence without it.
 
 ## Changes
 
-- [Fixed] Search: a malformed filter value in a search URL no longer replaces the whole catalog with the app-level error screen. Both the search page and a bucket's package list keep their chrome and show which filter could not be read, with reload and start-a-new-search as the ways out ([#5233](https://github.com/quiltdata/quilt/pull/5233))
+- [Fixed] Search: a malformed filter value in the URL no longer breaks the page ([#5233](https://github.com/quiltdata/quilt/pull/5233))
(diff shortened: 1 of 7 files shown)


fiskus:
@greptileai please re-review

PR #5244, opened 2026-08-27
catalog: default the beta switch to on

QuiltSimon:
## What

The global `beta` switch now reads as **on unless a catalog stores `false`**. The three surfaces it gates ship by default:

- the bucket header above the tabs (`Bucket.tsx`)
- the current bucket Overview instead of the legacy one (`Overview/index.tsx`)
- Athena's Tabulator tables (`Athena.tsx`)

An admin who doesn't want them flips the switch off in Admin > Settings. Mainlining properly — removing the flag — is next release.

## Why the default lives in code, not in stored settings

Turning it on by writing `beta: true` into `catalog/settings.json` would not have worked everywhere. Where the settings file is missing, or the service bucket returns `AccessDenied`, `CatalogSettings.use()` returns `null` and *no* stored value is readable — `settings?.beta` is `undefined` and every gate reads false. (Live example: `rc.quilttest.com` 403s on its `settings.json` today, which is why that stack serves the old dark-hero header.)

So the default is a code-level reader: `isBetaEnabled` treats anything other than a literal `false` as on, and all three gates plus the admin switch read through it. One place to change when the flag retires.

The admin switch is included deliberately: it read `settings?.beta || false`, so on a catalog that had never stored a value it would have displayed **off** while all three surfaces rendered.

## Verified

- `npm run typecheck` clean
- 26/26 green across the three directly affected specs (`Overview/index.spec.tsx` gained unset-is-on coverage, `Settings.spec.tsx`, `CatalogSettings.spec.tsx`)
- Full suite: 1630 passed. The 8 failures in `app/utils/spreadsheets/spreadsheets.spec.ts` are pre-existing on master — that directory has no diff on this branch and no reference to anything changed here
- Audited every `vi.mock('utils/CatalogSettings')` in the suite; `Queries.spec.tsx`, `DataProducts.spec.tsx` and `features.spec.ts` never reach the beta path (Queries stubs `./Athena` wholesale), so only the two mocks that do were updated

## Note

Supersedes #5243, which removed the gates outright and added a `legacy-ui` opt-out. This is the smaller change; closing that one.

<h3>Greptile Summary</h3>

The PR centralizes interpretation of the global beta setting so only an explicit stored `false` disables the bucket header, current Overview, and Athena Tabulator tables. It also aligns the Admin switch and affected tests with that default-on behavior.
- Adds shared `isBetaEnabled` and `useBetaEnabled` readers.
- Migrates all three gated surfaces and the Admin switch to the shared interpretation.
- Adds coverage for absent and unset settings and documents the behavior in the changelog.

The PR appears safe to merge with no actionable correctness or security defects identified.

The shared reader consistently implements the documented default-on policy, loading remains protected by Suspense, settings writes retain conflict detection, and the newly enabled surfaces have existing fallback handling.

| Filename | Overview |
|----------|----------|
| catalog/app/utils/CatalogSettings.tsx | Introduces the centralized default-on beta-setting interpretation and its React hook. |
| catalog/app/containers/Admin/Settings/Settings.tsx | Aligns the switch display with the shared beta reader while preserving conflict-checked whole-document writes. |
| catalog/app/containers/Bucket/Bucket.tsx | Uses the shared beta hook to default-enable the bucket header. |
| catalog/app/containers/Bucket/Overview/index.tsx | Selects the current Overview unless beta is explicitly disabled. |
| catalog/app/containers/Queries/Athena/Athena.tsx | Defaults Athena Tabulator table discovery to enabled through the shared hook. |
| catalog/app/containers/Bucket/Overview/index.spec.tsx | Updates the settings mock and verifies that absent or unset settings select Overview V2. |

<sub>Reviews (1): Last reviewed commit: ["catalog: default the beta switch to on"](https://github.com/quiltdata/quilt/commit/f12d08f2106b72fa0dcf6bf8e1bf872d799999f5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57664566)</sub>

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Changed] The global `beta` switch is on unless a catalog turns it off, so the bucket header above the tabs, the current bucket Overview, and Athena's Tabulator tables ship by default. An unset flag now reads as on — including where the settings file is missing or unreadable — and the Admin > Settings switch reports that rather than showing "off" while those surfaces render ([#PRNUM](https://github.com/quiltdata/quilt/pull/PRNUM))
 - [Changed] Search: error states are separate, more strictly typed components, safer to extend than one component behind a `kind` prop ([#5238](https://github.com/quiltdata/quilt/pull/5238))
 - [Fixed] Search: a malformed filter value in the URL no longer breaks the page ([#5233](https://github.com/quiltdata/quilt/pull/5233))
 - [Fixed] Search results table: a package with no commit message shows the no-value marker instead of the literal text `None` ([#5232](https://github.com/quiltdata/quilt/pull/5232))
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,7 +21,7 @@ complete sentence without it.
 
 ## Changes
 
-- [Changed] The global `beta` switch is on unless a catalog turns it off, so the bucket header above the tabs, the current bucket Overview, and Athena's Tabulator tables ship by default. An unset flag now reads as on — including where the settings file is missing or unreadable — and the Admin > Settings switch reports that rather than showing "off" while those surfaces render ([#PRNUM](https://github.com/quiltdata/quilt/pull/PRNUM))
+- [Changed] The global `beta` switch is on unless a catalog turns it off, so the bucket header above the tabs, the current bucket Overview, and Athena's Tabulator tables ship by default. An unset flag now reads as on — including where the settings file is missing or unreadable — and the Admin > Settings switch reports that rather than showing "off" while those surfaces render ([#5244](https://github.com/quiltdata/quilt/pull/5244))
 - [Changed] Search: error states are separate, more strictly typed components, safer to extend than one component behind a `kind` prop ([#5238](https://github.com/quiltdata/quilt/pull/5238))
 - [Fixed] Search: a malformed filter value in the URL no longer breaks the page ([#5233](https://github.com/quiltdata/quilt/pull/5233))
 - [Fixed] Search results table: a package with no commit message shows the no-value marker instead of the literal text `None` ([#5232](https://github.com/quiltdata/quilt/pull/5232))
--- a/catalog/app/containers/Admin/Settings/Settings.spec.tsx
+++ b/catalog/app/containers/Admin/Settings/Settings.spec.tsx
@@ -24,6 +24,9 @@ vi.mock('utils/CatalogSettings', () => {
   return {
     use: () => settings,
     useWriteSettings: () => writeSettings,
+    // Mirrors the real reader: only a stored literal `false` opts out, so an
+    // unset flag reads as on.
+    isBetaEnabled: (s: { beta?: boolean } | null) => s?.beta !== false,
     SettingsConflictError,
   }
 })
--- a/catalog/app/containers/Admin/Settings/Settings.spec.tsx
+++ b/catalog/app/containers/Admin/Settings/Settings.spec.tsx
@@ -14,7 +14,14 @@ const writeSettings = vi.fn<
 
 // Declared inside the factory: `vi.mock` is hoisted above any top-level binding
 // this file could otherwise reference.
-vi.mock('utils/CatalogSettings', () => {
+// Declared inside the factory: `vi.mock` is hoisted above any top-level binding
+// this file could otherwise reference. `isBetaEnabled` delegates to the real
+// implementation rather than restating the default rule -- otherwise these tests
+// assert the mock's opinion of the default, not the module's.
+vi.mock('utils/CatalogSettings', async () => {
+  const actual = await vi.importActual<typeof import('utils/CatalogSettings')>(
+    'utils/CatalogSettings',
+  )
   class SettingsConflictError extends Error {
     constructor() {
       super('Catalog settings were changed by someone else.')
@@ -24,9 +31,7 @@ vi.mock('utils/CatalogSettings', () => {
   return {
     use: () => settings,
     useWriteSettings: () => writeSettings,
-    // Mirrors the real reader: only a stored literal `false` opts out, so an
-    // unset flag reads as on.
-    isBetaEnabled: (s: { beta?: boolean } | null) => s?.beta !== false,
+    isBetaEnabled: actual.isBetaEnabled,
     SettingsConflictError,
   }
 })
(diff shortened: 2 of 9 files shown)


QuiltSimon:

Ran against 623c22b (8 finder angles, adversarially verified). **10 findings — 4 fixed in 4f15d2b, 6 dismissed in writing.** One was a security issue in my working tree, caught before it reached the branch.

### Fixed

1. **`!== false` read every non-boolean as ON.** The sharpest finding. Settings are `JSON.parse`d straight to a cast with no schema, so `beta: "false"`, `beta: 0`, `beta: null` and `beta: ""` all read as **on** — a correctly-spelled hand-edited opt-out doing the exact opposite of what it says. Now absent means on, but a present key must be a literal `true`, matching the `=== true` discipline `utils/features` documents for the same document, with the default inverted.

2. **The headline behavior had no test.** Reverting the reader to `settings?.beta || false` left the entire suite green — `Overview/index.spec.tsx` caught the gate, nothing caught the switch. Added direct `isBetaEnabled` coverage (including all five non-boolean cases) plus two `BetaSwitch` cases for unset and no-document. **Verified by mutation: restoring the old semantics now fails 6 tests across both specs.**

3. **The `Settings.spec.tsx` mock restated the default rule inline** while the sibling spec in the same PR delegated to the real reader and explained why not to. Now uses `importActual` and overrides only the local `SettingsConflictError` class it needs for `instanceof` narrowing.

4. **(Working tree, never committed — flagging for the record.)** The review caught an `http://localhost:8400/live.js` dev-tooling script tag in `catalog/app/index.html`, which `webpack.base` uses as the production template. Had it been committed it would have shipped as blocked mixed content on HTTPS, and executed arbitrary local script in the catalog's origin on any machine with a service on that port. It was an uncommitted artifact of an impeccable live-mode session, is absent from every commit on this branch, and is now removed from disk. Worth noting as a hazard of that tooling.

### Dismissed (real, not this PR's to fix)

5. **`AccessDenied` and "no settings file" both map to `null`**, so a catalog that stored `beta: false` still reads ON for a principal who cannot read the settings object — admins could see legacy while a read-only role gets the new surfaces. Confirmed and genuinely unsafe-direction, but closing it needs `fetchSettings` to distinguish the two cases, which is a change to the fetch contract rather than a default flip. **Documented as a known gap in the reader's docblock** so the next reader isn't surprised. Live on `rc.quilttest.com`, which 403s on its own settings.json today.

6. **The destructive `{...null, beta:false}` write** that replaces the whole document when settings are unreadable. Pre-existing on the identical path (pre-PR an admin clicking ON ran it); this PR supplies motive but not the bug. Same fix as #5.

7. **The bucket header renders above the not-found error** and fetches stats for nonexistent buckets — it sits outside the `bucketExistenceData` gate. Pre-existing structure; the default flip widens it from opt-in catalogs to all. Real bug, wrong PR.

8. **Duplicate `useStats` call** — `BucketLayout`'s header and `OverviewV2`'s header both run the same uncached ES stats aggregation per Overview load; and the header mounting outside the route `Switch` means that aggregation now fires on every bucket route, not just Overview. Both are consequences of surfaces that were already on for beta catalogs, not of this line. Worth a follow-up.

9. **`features.ts` explicitly documents `beta` as unsuitable for this** — one lever for three unrelated capabilities, so wanting the new Overview but not Tabulator tables is impossible. Correct, and I agree: [#5243](https://github.com/quiltdata/quilt/pull/5243) did exactly that split and was closed in favor of this smaller change for the release. Splitting the flag is next-release work, per the plan in the PR body.

10. Findings 7–9 are all fair follow-ups I'd rather file than smuggle into a release-week default flip.

Re-verified after fixes: typecheck clean, 37/37 on the three affected specs (was 26), CI green.

QuiltSimon:
@sir-sigurd @nl0 — you're both on as reviewers. **Please merge this yourselves once you're happy with it**, rather than approving and handing it back.

Why I'm asking rather than doing it: `master` carries an active ruleset named `quilt-require-human-approval`, and `mergeStateStatus` is `BLOCKED` with `reviewDecision: REVIEW_REQUIRED`. That gate exists precisely so an agent doesn't merge its own work, so the approval and the merge should be the same pair of human hands. (Note the legacy `/branches/master/protection` endpoint reports `required_approving_review_count: 0` — the requirement lives in the ruleset, invisible to that API.)

State as of `4f15d2b`: CI green (`test-catalog` SUCCESS; the one red check is `codecov/patch/informational`, advisory by name). Typecheck clean, 37/37 on the affected specs. A high-effort `/code-review` ran against the previous head — 10 findings, 4 fixed here, 6 dismissed in writing in the comment above.

**What to look at first,** in the order I'd want a reviewer to:

1. `isBetaEnabled` in `CatalogSettings.tsx` — the whole PR is this one predicate. Absent means on; a present key must be a literal `true`. My first version used `!== false`, which read a hand-edited `beta: "false"` as **on**. The docblock records a known gap I deliberately did not fix here: `AccessDenied` and "no settings file" both map to `null`, so a stored `beta: false` is unenforceable for a principal who cannot read the object.
2. Whether the coupled flag going default-on is the right call at all. `features.ts:13-17` documents `beta` as unsuitable for this — one lever, three unrelated capabilities. #5243 did the proper split and I closed it in favor of this smaller change for the release; splitting is next-release work. If either of you would rather take the split now, say so and I'll reopen it.

**Not targeted at `rc`.** @nl0 — if this belongs in 26.8.0 it needs a pin bump, and that's your cut. I can open the `deployment` bump (`base.py` `_VERSIONS["catalog"]`, currently `7047ab8f`) against `rc` the moment this lands on master; say the word and it's yours to take or leave.

**One thing to weigh before the notes promise an opt-out:** `rc.quilttest.com` 403s on its own `catalog/settings.json`, so the entire `CatalogSettings` document is unreadable on that stack. After this change the new UI turns on there — but the Admin > Settings switch is inert, because writing settings re-reads them first. "You can turn this off in Admin > Settings" does not hold on a stack in that state.

nl0:
the approach is backwards:
this PR turns the umbrella "beta features" flag ON by default, making all "beta" stuff opt-out.
this was not the intention,
the intention was to graduate some specific features from beta,
so closing this as mechanically fine but intentionally misaligned.

QuiltSimon:
Without this switch being left on for this release we’re without the
improvements to the bucket overview page. We want everyone to have this new
view by default, and moving all of the mechanics from behind the beta flag
felt riskier last second than switching the beta flag on for everyone. It
will get moved next release out of the beta flag. Unless the updated
visuals can get moved out of beta in time for this release please turn this
on, unless there’s a strong reason not to.

Thanks!


Simon Kohnstamm
Director of Sales, Quilt Data, Inc.
***@***.*** | (781) 277-2255
Manage Data Like Code > quiltdata.com

nl0:
> unless there’s a strong reason not to

strong reason not to do it is, as stated, enabling _all the flagged features_ by default.
if we want something to be shipped for everyone, we graduate these specific things by taking them off that beta flag,
to be used unconditionally.
which ones would you like to take off the flag?
a more "proper" granular feature flagging is still to be built.

PR #5248, opened 2026-08-28
Quilt3: scope API keys to registry URLs

drernie:
## Description

Scopes in-memory API-key authentication to the registry URL selected when `login_with_api_key()` is called. This prevents an API key configured for one registry from being attached to requests sent to another registry through `use_registry_url()`.

Callers may pass an explicit `registry_url=` snapshot. If neither an explicit nor configured URL exists, login fails immediately instead of storing a key without a registry. Multiple registries can hold separate in-memory keys concurrently, while `clear_api_key()` retains its historical clear-all behavior.

Follow-up to #5240.

## Validation

- [x] Python API suite: 622 passed, 1 expected xfail
- [x] Focused session tests: 45 passed
- [x] Ruff lint and formatting
- [x] Generated Python API documentation
- [x] GraphQL client generation
- [x] Documentation code blocks: 42 passed, 66 skipped
- [x] API keys remain memory-only and are never persisted
- [x] All modified lines covered

## Risk and rollback

Existing callers with a configured registry retain the same function call and behavior. Headless callers can bind the key explicitly with `registry_url=`; missing registry configuration now produces an immediate error. Reverting this PR restores the prior process-global API-key behavior.


api/python/quilt3/session.py
```diff
@@ -281,20 +284,24 @@ def login_with_api_key(key: str):
     """
     if not key.startswith("qk_"):
         raise ValueError("API key must start with 'qk_' prefix")
-    global _api_key
+    registry_url = get_registry_url()
```
drernie: `login_with_api_key()` now binds the key to whatever `get_registry_url()` returns at call time, with no validation. In a container with no `~/.quilt/config.yml` that resolves to `None`, so the documented Quickstart order (`login_with_api_key(...)` first, then `quilt3.config(url)`) stores the key under `None` and every later request goes out unauthenticated — surfacing as "Authentication failed. Check your credentials or API key." Ordering used to be irrelevant with the global `_api_key`, so this is a silent behavior change for existing callers.

Suggest either rejecting a `None` registry here, or accepting an explicit `registry_url=` parameter the way `login_with_token` does.

api/python/quilt3/session.py
```diff
@@ -281,20 +284,24 @@ def login_with_api_key(key: str):
...
     """
-    Clear the API key and fall back to interactive session (if available).
+    Clear the current registry's API key and fall back to its interactive session.
     """
-    global _api_key
+    registry_url = get_registry_url()
     with _sessions_lock:
-        _api_key = None
-        clear_session()  # Force session recreation with interactive auth
+        _api_keys.pop(registry_url, None)
```
drernie: `clear_api_key()` pops only the currently resolved registry's key, and is silent when there isn't one. If the registry changed between login and clear (or the caller never configured one, per the comment above), the key stays in `_api_keys`: `logged_in()` still reports authenticated for that registry, and the next `get_session()` for it re-sends the bearer token the caller believed was cleared.

Either clear all keys, or take/validate the registry explicitly so "clear" can't quietly no-op.

api/python/quilt3/session.py
```diff
@@ -397,7 +403,10 @@ def logged_in():
     """
     Return catalog URL if Quilt client is authenticated, `None` otherwise.
     """
-    if _api_key is not None or get_registry_url() in _load_auth():
+    registry_url = get_registry_url()
+    with _sessions_lock:
```
drernie: Taking `_sessions_lock` for a dict membership test introduces a blocking path that didn't exist before. `get_session()` holds this same lock across `_update_auth()`, which does `requests.post(..., timeout=None)` — so a `logged_in()` status poll on another thread can now block indefinitely behind an in-flight token refresh.

The lock isn't buying anything here: the membership test is atomic under the GIL, and `has_api_key` is consumed outside the lock anyway, so the result is already stale by the time it's used. Suggest dropping the `with` block and testing `registry_url in _api_keys` directly.

drernie (commented):
Code review of the API-key scoping change. The core change looks right: every `get_session()` caller already passes an explicit `registry_url` snapshot (`search_util.py:21`, `_graphql_client/base_client.py:34`, `_refresh_credentials`), so the per-registry key dict and session cache stay consistent. Three issues inline.

Tests weren't run here (`responses` not installed in my environment); I traced the new cases by hand and they match the implementation. CHANGELOG and `docs/api-reference/api.md` are consistent with the new docstrings.

api/python/quilt3/session.py
drernie: Fixed in 9cb8f3d36. `login_with_api_key()` now accepts an explicit `registry_url=` snapshot and raises `QuiltException` immediately when neither an explicit nor configured registry URL exists, so a key can no longer be stored under `None`. The Quickstart and authentication guide now pass `QUILT_CATALOG_URL` explicitly. Added regressions for both the missing-URL failure and explicit-URL path.

api/python/quilt3/session.py
drernie: Fixed in 9cb8f3d36. Restored `clear_api_key()` to its historical clear-all behavior: it now removes every in-memory registry key and closes all cached sessions, while leaving persisted interactive credentials alone. Added a two-registry regression that verifies both the key map and session cache are empty after clearing.

api/python/quilt3/session.py
drernie: Fixed in 9cb8f3d36. Removed `_sessions_lock` from the `logged_in()` membership snapshot. Added a concurrency regression that deliberately holds the session lock on one thread and verifies `logged_in()` completes on another. Local validation: 622 passed, 1 expected xfail; 45 focused session tests; Ruff, docs, and generation checks pass.

drernie:
Validated this change against **quiltx**, the primary consumer, without publishing a prerelease.

Test setup:

- quiltx branch: `260827-quilt3-acl-quiltx`
- quiltx base: `2437b85` (`v0.19.0`, current `origin/main`)
- quilt3 source: this PR at `9cb8f3d36`
- built a wheel from `api/python`
- restored the locked quiltx environment with `uv sync --extra dev`
- installed only the branch-built quilt3 wheel with `uv pip install --no-deps --reinstall`

Results:

- **473 passed, 1 skipped**
- quiltx CLI smoke test passed
- both worktrees remained clean

This tests the same package shape consumers receive from PyPI while keeping the locked quiltx transitive dependencies unchanged. An editable install was intentionally not used because the two repositories have colliding top-level `tests` packages under editable installation.

Conclusion: no quilt3 prerelease is needed for source-level quiltx compatibility testing; the branch-built wheel passes the complete quiltx unit suite.

drernie:
Completed the stronger end-to-end consumer validation using a real prerelease and quiltx GitHub CI.

### Published artifact

- created `release-8.1.0a2` from this PR head (`9cb8f3d36`)
- published immutable tag `8.1.0a2` at release commit `5025ac8ee`
- quilt tag workflow passed the full Ubuntu/Windows Python 3.10–3.14 matrix, docs, lint, lambdas, tag/version verification, build, and PyPI upload
- workflow: https://github.com/quiltdata/quilt/actions/runs/33141195572
- PyPI now serves `quilt3==8.1.0a2`

### Actual quiltx migration

Opened quiltdata/quiltx#100 from the checked-out `260827-quilt3-acl-quiltx` branch:

- replaced the process-wide `quilt3.session.get_registry_url` monkey-patch with `set_registry_url_resolver()` / `reset_registry_url_resolver()`
- passes the catalog advertised registry URL explicitly to `login_with_api_key()`
- removed the obsolete quiltx lock, local URL `ContextVar`, patch-install machinery, and unused context wrapper
- requires `quilt3>=8.1.0a2,<9`
- committed `uv.lock` resolves the artifact from PyPI, not a branch path or local wheel
- added black-box coverage that binds two catalogs through the real published API and verifies their HTTP sessions carry distinct API keys

### Results

- local quiltx suite: **475 passed, 1 skipped**
- Black and mypy: passed
- quiltx GitHub CI: Python **3.10, 3.11, 3.12, 3.13, and 3.14 all passed**
- lint: passed
- Greptile Review: passed
- quiltx PR is mergeable with `CLEAN` status and no review threads

quiltx PR: https://github.com/quiltdata/quiltx/pull/100

drernie:
Follow-up correction: quiltx is now prepared as **0.20.0** for this minor compatibility/API migration, rather than leaving the changelog entry under Unreleased at 0.19.0.

Updated consistently in quiltdata/quiltx#100:

- `pyproject.toml`: `0.20.0`
- runtime `quiltx._version.__version__`: `0.20.0`
- `uv.lock`: quiltx `0.20.0`, quilt3 `8.1.0a2`
- `CHANGELOG.md`: `[0.20.0] - 2026-08-27`

Commit: `882a04c`

The updated GitHub CI is green again: lint plus Python 3.10–3.14 all pass, with no review threads and a clean/mergeable PR.

PR #5265, opened 2026-08-31
Catalog: honor a mis-cased Athena workgroup and stop masking the bucket default (5217 stack 7/9)

nl0:
# Description

Two bugs in how the Athena console chooses a workgroup, plus the seam that lets a
bucket's `ui.athena.defaultWorkgroup` reach that choice at all.

**The console errors instead of honoring the default.** The workgroup was matched
case-insensitively and then stored and used verbatim. An admin writing
`ui.athena.defaultWorkgroup: Analytics-Prod` against a workgroup named
`analytics-prod` matched here — and then `listNamedQueries` and
`startQueryExecution` sent the admin's casing, and AWS rejected it with
`InvalidRequestException`. The same hole swallowed a mis-cased `:workgroup` URL
segment. Candidates now resolve to the list's own spelling before being stored.

**An unavailable stored workgroup masked a valid default.** With
`storage.getWorkgroup() || defaultWorkgroup`, a stored workgroup that had been
deleted or lost access was picked by the `||`, failed the availability check, and
dropped through to the alphabetically-first workgroup — never trying a bucket
default that was perfectly good. Candidates are tried in turn instead.

**The seam.** `Model.Provider` takes an optional `preferences` prop and hands
`ui.athena.defaultWorkgroup` to `useWorkgroup` as a string rather than the whole
object — the parsed preferences are rebuilt on every provider render, so an object
would re-fire the workgroup effect throughout the probe.

**Safe on its own.** Nothing passes the new prop yet, so the seam is a no-op in
isolation and this PR stands as a refactor plus two real bug fixes. The consumer
that supplies the prop is PR 9.

## Review findings addressed

- **f34** — the case-insensitive match with verbatim use, above.
- **f16** — the masked bucket default, above. **Strictly a bug fix:** storage
  still outranks the bucket default; only the fall-through changed.
- **f46** — the `Provider` → `useWorkgroup` seam was crossed by no test, so the
  prop could have been dropped on the floor unnoticed. Now driven through the
  provider.
- **f33 (i)** — nothing pinned the stored-vs-preference precedence in either
  direction. It is pinned now.
- **f51** — recorded as the change's own benefit rather than left implicit:
  narrowing the third parameter to `defaultWorkgroup?: string` removes
  `utils/BucketPreferences` from the AWS-call layer entirely.

## ⚠️ f33 (ii) — a decision this PR deliberately does not take

`storage.getWorkgroup()` still outranks `ui.athena.defaultWorkgroup`. That means
**any user who has ever switched workgroup never sees the bucket default** — which
is narrower than "honors that bucket's `defaultWorkgroup` again" implies.

This precedence is not new: it exists verbatim in the 26.7.4 pin
([`eda3016f`](https://github.com/quiltdata/quilt/commit/eda3016f),
`Bucket/Queries/Athena/model/requests.ts:127`), which is the baseline the
hardening review compared master
([`6167dd82`](https://github.com/quiltdata/quilt/commit/6167dd82)) against.
Reordering it changes behaviour for returning users, which is a product decision
rather than hardening — so it is **not** silently reordered here.

**A maintainer should rule on one of two things:** either the precedence is
reordered deliberately in its own change, or the changelog wording narrows to
match what the code does. The test added here pins the current order, so a
deliberate reversal will show up as a test to change rather than a silent drift.

f16's fall-through fix is independent of this and lands either way.

## Recorded, not fixed here

- **f19** — `preferences?.defaultWorkgroup` flattens a preference object to a
  primitive at the consumer to dodge an unmemoized shared provider. Real; the
  proper fix is to memoize `BucketPreferences`, which is wider than this branch.
- The same case-canonicalisation shape exists for **databases and catalog names**
  (`listIncludes` at five other call sites in this file). Untouched here: f34 named
  the workgroup, and those paths have their own storage-restore behaviour that
  deserves its own reading rather than a blanket change.

# Verification

```
cd catalog && npx vitest run app/containers/Queries/Athena
```

102 tests. The four new ones in `requests.spec.ts` and the one in `state.spec.tsx`
were each checked against the pre-fix code and fail there.

# Position in the stack

**PR 7 of 9**, based on
[`stack/5217-6-admin-service-read-only`](https://github.com/quiltdata/quilt/pull/5264).

[f27](https://github.com/quiltdata/quilt/pull/5217#discussion_r3889140880). This
is the one hard mechanical dependency in that set, and the reason the Athena work
splits in two: the seam is a no-op in isolation, so it can land on its own merits
while the mount change that activates it (PR 9) waits on a direction call. **Do
not squash this into PR 9** — that destroys the property that makes it
independently mergeable.

# TODO

- [x] Unit tests
- [x] Changelog entry

<h3>Greptile Summary</h3>

This PR corrects Athena workgroup selection by canonicalizing case-insensitive matches to AWS-provided names and trying stored and bucket-default candidates independently. It also adds an optional provider seam for forwarding the bucket’s default workgroup and expands focused test coverage.

- Canonicalizes URL, stored, and configured workgroup names before downstream AWS calls.
- Falls through from an unavailable stored workgroup to a valid bucket default.
- Threads `preferences.defaultWorkgroup` through the Athena model provider.
- Adds regression tests for casing, precedence, fallback, and provider propagation.

The PR appears safe to merge, with the changed selection paths constrained to available Athena workgroups and covered by targeted regression tests.

The implementation resolves candidate names against Athena’s returned list, preserves the documented precedence, falls through unavailable candidates correctly, and introduces no active incompatibility because the new provider prop is optional.

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Queries/Athena/model/requests.ts | Refactors workgroup matching and precedence to use canonical available names and correctly fall through among candidates. |
| catalog/app/containers/Queries/Athena/model/requests.spec.ts | Adds focused regression coverage for case canonicalization, unavailable stored selections, and stored/default precedence. |
| catalog/app/containers/Queries/Athena/model/state.tsx | Adds a type-safe optional Athena preference prop and forwards only the stable default-workgroup string. |
| catalog/app/containers/Queries/Athena/model/state.spec.tsx | Verifies that the provider seam forwards the configured default through the integrated Athena model. |
| catalog/CHANGELOG.md | Documents the corrected casing and fallback behavior accurately. |

<sub>Reviews (1): Last reviewed commit: ["docs(changelog): entry for the workgroup..."](https://github.com/quiltdata/quilt/commit/fc62d139ea8c51af83b02b9b5cf79978145a1092) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58629632)</sub>

### DIFF
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,7 +21,7 @@ complete sentence without it.
 
 ## Changes
 
-- [Fixed] Athena Queries: a workgroup named in `ui.athena.defaultWorkgroup` or in the URL is honored whatever its capitalisation — a mis-cased name used to match and then have every AWS call rejected — and a stored workgroup that has been deleted or lost access no longer hides a valid bucket default ([#5265](https://github.com/quiltdata/quilt/pull/5265))
+- [Fixed] Athena Queries: a workgroup named in the URL is honored whatever its capitalisation, instead of matching a workgroup you can see and then having every AWS call rejected ([#5265](https://github.com/quiltdata/quilt/pull/5265))
 - [Fixed] Admin Users: the roles dialog and the Role column are read-only for service users, rather than offering a Save the registry refuses ([#5264](https://github.com/quiltdata/quilt/pull/5264))
 - [Fixed] Admin Users: a disabled Enabled or Admin switch says why on hover and on keyboard focus — your own account, a service user managed by the stack, or admin capabilities managed by the SSO configuration — instead of rendering as the same unexplained dead control; the Admin switch now also refuses service users ([#5263](https://github.com/quiltdata/quilt/pull/5263))
 - [Fixed] Search sidebar: the facet "Sort by" control no longer disappears while you type in "Find metadata" on stacks with truncated facet lists, and it is withheld when the query matches nothing rather than offering to sort an empty list ([#5262](https://github.com/quiltdata/quilt/pull/5262))
--- a/catalog/CHANGELOG.md
+++ b/catalog/CHANGELOG.md
@@ -21,6 +21,7 @@ complete sentence without it.
 
 ## Changes
 
+- [Fixed] Athena Queries: a workgroup named in `ui.athena.defaultWorkgroup` or in the URL is honored whatever its capitalisation — a mis-cased name used to match and then have every AWS call rejected — and a stored workgroup that has been deleted or lost access no longer hides a valid bucket default ([#5265](https://github.com/quiltdata/quilt/pull/5265))
 - [Fixed] Admin Users: the roles dialog and the Role column are read-only for service users, rather than offering a Save the registry refuses ([#5264](https://github.com/quiltdata/quilt/pull/5264))
 - [Fixed] Admin Users: a disabled Enabled or Admin switch says why on hover and on keyboard focus — your own account, a service user managed by the stack, or admin capabilities managed by the SSO configuration — instead of rendering as the same unexplained dead control; the Admin switch now also refuses service users ([#5263](https://github.com/quiltdata/quilt/pull/5263))
 - [Fixed] Search sidebar: the facet "Sort by" control no longer disappears while you type in "Find metadata" on stacks with truncated facet lists, and it is withheld when the query matches nothing rather than offering to sort an empty list ([#5262](https://github.com/quiltdata/quilt/pull/5262))
--- a/catalog/app/containers/Queries/Athena/model/requests.spec.ts
+++ b/catalog/app/containers/Queries/Athena/model/requests.spec.ts
@@ -1300,6 +1300,46 @@ describe('containers/Queries/Athena/model/requests', () => {
       unmount()
     })
 
+    it('prefers an exact match over a differently-cased sibling', async () => {
+      // Workgroup names are case-sensitive, so both can exist. `fetchWorkgroups`
+      // sorts, which puts 'Alpha' first -- a case-insensitive match alone would
+      // hand the user the workgroup they did not ask for.
+      const workgroups = {
+        data: { list: ['Alpha', 'alpha'] },
+        loadMore: noop,
+      }
+
+      const { result, waitFor, unmount } = renderHook(() =>
+        useWrapper([workgroups, 'alpha', undefined]),
+      )
+
+      await act(async () => {
+        await waitFor(() => typeof result.current.data === 'string')
+        expect(result.current.data).toBe('alpha')
+      })
+      unmount()
+    })
+
+    it('falls back rather than throwing when the default is not a string', async () => {
+      // `ui.athena` is unconstrained by the bucket-config schema, so an all-digit
+      // workgroup name in the YAML arrives as a number. Reaching `toLowerCase`
+      // with it threw out of the effect and took the console down.
+      const workgroups = {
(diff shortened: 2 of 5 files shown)


nl0:

Reviewed this layer only (`stack/5217-6-admin-service-read-only..stack/5217-7-athena-workgroup-seam`). Three findings accepted and fixed in 96ddb83 and 125ac21, three declined.

### Accepted

**1. An exact workgroup match had to beat a case-insensitive one — regression in this layer.** `canonical()` returned the first case-insensitive hit. Workgroup names are case-sensitive, so `Alpha` and `alpha` can both exist, and `fetchWorkgroups` ends with `list.sort()`, which puts `Alpha` first. Picking `alpha` from the dropdown then stored `alpha` and navigated to `/queries/athena/alpha`, which resolved back to `Alpha` — queries, history and named queries all ran against the workgroup the user did not choose, and the Select displayed a name they never picked. Before this layer an exactly-spelled request was stored verbatim and was correct, so this was introduced here. Now an exact hit wins; regression test added, and it fails without the fix.

**2. A non-string `defaultWorkgroup` threw out of the effect.** `ui.athena` is unconstrained by `shared/schemas/bucketConfig.yml.json` — there is no `athena` node and `ui` does not set `additionalProperties: false` — and `parseAthena` only spreads. So `ui.athena.defaultWorkgroup: 2024`, an all-digit workgroup name that is legal in Athena, survives validation as a number, and `toLowerCase()` on it took the console down instead of falling back. Not introduced here (the old `listIncludes` path had the same exposure), but this layer rewrites that code and #5267 is what makes it reachable, so it is guarded here: `pick` takes `unknown` and rejects non-strings. The complete fix is a schema entry for `ui.athena`; that is left as follow-up rather than widened into this layer.

**3. The changelog entry claimed more than this layer ships.** It announced `ui.athena.defaultWorkgroup` support and the unmasking of a bucket default, but the only `Model.Provider` render site is still prop-less at this layer, so `defaultWorkgroup` is always `undefined` at runtime here — and #5267 already carries its own entry for exactly that behaviour. If this layer merges ahead of #5267 the entry is false in the interim, so it now claims only the URL-capitalisation fix, which does work standalone.

### Declined

**4. `useDatabase` / `useCatalogName` share the match-insensitively-then-store-verbatim shape.** Declined: unlike the workgroup, those values only ever originate from the AWS-returned list or a prior AWS response, so no mis-cased value can enter.

**5. The preference document can resolve after the workgroup is seeded.** Declined here: real, but #5267 gates the scoped mount on the resolved document, which is the right layer for it.

**6. `Provider`'s new `preferences` prop has no production caller.** Not a defect — intentional decomposition; #5267's `ScopedConsole` supplies it.

### Also swept

Comments were retelling what the change did and what the old code got wrong — the tests already pin that. Trimmed to what the code cannot show: why the canonical spelling must be used, why the preference crosses as a string rather than the `ui.athena` object, and that storage-before-default is inherited product behaviour rather than a choice made here.

Test hygiene: the stubbed-mock restores now sit in a `finally`. `afterEach` only calls `vi.clearAllMocks()`, which clears call data but keeps implementations, so a failing assertion leaked the stub into every later test in the file and turned one real failure into a cascade of unrelated ones.

### Verification

80 tests pass in the two touched spec files (2 new, each confirmed failing without its fix); `tsc --noEmit` clean; oxlint and oxfmt clean.

PR #5269, opened 2026-09-01
Catalog: Mermaid diagram previewer

QuiltSimon:
Renders [Mermaid](https://mermaid.js.org/) diagrams in the catalog file preview.
`.mmd` / `.mermaid` files preview as diagrams when browsing a bucket or package
and from `quilt_summarize.json` (`"types": ["mermaid"]`); the plain-text source
stays reachable from the view selector, and a definition that does not parse
reports Mermaid's error in place of the diagram.

## Changes

- New loader (`Preview/loaders/Mermaid.tsx`) and renderer (`Preview/renderers/Mermaid/`),
  registered in the loader chain, `FileType`, `PreviewData`, `viewModes`, and the
  `quilt_summarize.json` schema + `summarize.ts` types.
- Docs: a Mermaid section in `docs/Catalog/VisualizationDashboards.md`.

Two decisions worth a reviewer's attention:

- **Reads S3 directly, not the `/preview` endpoint.** That endpoint's allowlist is
  `csv/excel/fcs/ipynb/parquet/vcf/bed/txt` and it returns a truncated head/tail —
  a partial graph definition is a syntax error, not a shorter diagram. So the loader
  follows `Markdown.jsx` and gets the object from S3. **No preview-Lambda change and
  no `_VERSIONS["preview"]` bump are needed.**
- **Lazy-loaded, `securityLevel: 'strict'`.** mermaid is ESM-only and ~84MB unpacked;
  it is loaded via `RT.mkLazy` so it lands in its own chunk rather than the entry
  bundle. Diagram definitions are customer data, so mermaid's sanitizer stays on and
  click/script directives in the graph are refused.

### Incidental: katex import path

mermaid depends on `katex@^0.16.47`, which raises this repo's existing direct katex
dep from 0.16.21. In 0.16.47 the deep source path `Notebook.jsx` imported
(`katex/contrib/auto-render/auto-render`) resolves through the `./*` export wildcard
to raw TypeScript, which webpack's `node_modules`-excluded ts-loader will not parse —
`npm run build` fails without this. It now imports the supported subpath
`katex/contrib/auto-render` (`dist/contrib/auto-render.js`), same
`renderMathInElement` default export, arity 2.

## Verified

- `npm run typecheck` — clean
- `npm run lint` (oxfmt + oxlint) — clean
- `npm test` — 123 files, 1118 passed / 1 skipped, including 4 new `detect` cases
- `npm run build` — webpack compiled with 0 errors; confirmed the entry chunk
  contains only the literal `"mermaid"` enum string and no library internals, so the
  code-split holds
- `markdownlint` on the changed docs + CHANGELOG — clean

Not verified: rendering against real diagrams in a running stack — no dev stack was
brought up for this.

## Follow-up

Deploying this needs a `_VERSIONS["catalog"]` bump in `quiltdata/deployment`, which
can only pin a SHA that exists on `master` — so it follows this merge, as the
deployment log does for every catalog change (`Bump catalog to e015c1d0 (#2619)`).
`quiltdata/enterprise` needs nothing: it has no filetype allowlist, only the
`SUMMARIZE_KEY` / `README_KEYS` string constants.

<h3>Greptile Summary</h3>

The PR adds lazy-loaded Mermaid diagram previews for `.mmd`, `.mermaid`, and summary-configured files while preserving a plain-text view and in-place parse errors.
- Registers Mermaid throughout preview detection, rendering, view-mode, and preview-data plumbing.
- Reads complete diagram definitions directly from S3 with the existing large-object gate.
- Extends the shared `quilt_summarize.json` contract and Catalog documentation.
- Updates Mermaid and KaTeX dependencies and switches Notebook to KaTeX’s supported auto-render export.

The PR appears safe to merge, with no concrete blocking or independently actionable non-blocking issue identified.

The new preview path follows the existing loader, renderer, view-mode, S3-access, and dependency-locking contracts, and the investigated lifecycle, renderer-signature, schema, and installation paths did not establish an observable defect.

| Filename | Overview |
|----------|----------|
| catalog/app/components/Preview/loaders/Mermaid.tsx | Adds extension and summary-driven loading of complete Mermaid source from S3, guarded by the established one-megabyte preview gate. |
| catalog/app/components/Preview/renderers/Mermaid/Mermaid.tsx | Lazily renders strict-security Mermaid SVG with unique IDs, stale-effect protection, loading feedback, and parse-error presentation. |
| catalog/app/components/Preview/renderers/Mermaid/index.tsx | Integrates the renderer through the existing lazy-component and two-argument preview-renderer conventions. |
| catalog/app/components/Preview/load.jsx | Registers Mermaid in the preview loader chain without changing the established explicit-mode, summary-type, and extension precedence. |
| catalog/app/containers/Bucket/viewModes.ts | Adds Mermaid to the file-view selector while retaining plain text as the alternate mode. |
| shared/schemas/quilt_summarize.json | Extends the shared summary configuration enum with the Mermaid viewer type. |
| catalog/app/components/Preview/renderers/Notebook.jsx | Moves KaTeX auto-render to its supported exported subpath to remain build-compatible with the resolved KaTeX version. |
| catalog/package.json | Adds Mermaid as a runtime dependency; controlled builds remain pinned through the committed lockfile and npm-ci workflows. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  F[Bucket or package file] --> D{Preview detection}
  D -->|.mmd or .mermaid| L[Mermaid loader]
  D -->|summary type: mermaid| L
  L --> G{Object exceeds automatic threshold?}
  G -->|Yes| C[Ask user to load preview]
  G -->|No or approved| S[Read complete object from S3]
  S --> P[Create Mermaid preview data]
  P --> V{Selected view}
  V -->|Mermaid| R[Lazy-load Mermaid and render strict SVG]
  V -->|Plain Text| T[Render source text]
  R --> E{Definition parses?}
  E -->|Yes| O[Display diagram]
  E -->|No| X[Display Mermaid error]
```

<sub>Reviews (1): Last reviewed commit: ["Catalog: Mermaid diagram previewer"](https://github.com/quiltdata/quilt/commit/eba5329106c82aa22df3552c65b2306cdc597eda) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59220581)</sub>

**Context used:**

- Knowledge Base — [API contracts and schemas](https://app.greptile.com/quilt-bio/-/custom-context/knowledge-base/quiltdata/quilt/-/docs/api-contracts.md)


QuiltSimon:
## Pre-merge `/code-review` disposition

Two `high` passes. Pass 1 on `890531925` (10 findings), pass 2 on `858183187` (8 findings). Final head `ad6891191`.

### Pass 1 — all fixed in `396f57218`
| Finding | Resolution |
|---|---|
| Wheel/pinch zoom: fixed 1.4× per event | scaled by `deltaY` (`exp(-dy·k)`, one 100px notch = one step), `dy=0` ignored |
| Drag never ends when `setPointerCapture` throws | `pointerup`/`pointercancel` on `window` |
| ⌘/Ctrl+0/+/- swallowed by diagram | early return on any modifier |
| `width:100%` blows up small diagrams | keep mermaid's `max-width: Npx` as `min(Npx, 100%)` |
| Fence regex disagrees with fence handler | `hasMermaidFence` uses markdown-it tokens + the handler's predicate; tested |
| Fence-less README rendered twice as mode settles | `drawMermaid = mode !== Markdown \|\| !hasFence` |
| Renderer cache grows per file | cache capped at 16 (clear-all; LRU if needed) |
| Streaming Chat re-renders every token | debounced 200ms after first run |
| `<Markdown drawMermaid>` silently ignored | forwarded to `getRenderer` |
| Four multi-line changelog entries | one sentence |

### Pass 2 — fixed in `ad6891191` unless noted
| Finding | Resolution |
|---|---|
| `hasMermaidFence` parsed with `html:false`, renderer `html:true` | shared `MD_OPTS`; regression test for the html-block case |
| No `touch-action:none` on zoomed viewport | added |
| Author raw `<pre class="mermaid-fence">` gets drawn | **Dismissed.** DOMPurify keeps `pre`+`class`, so an author can opt a block into the diagram pass — same trust as writing a fence. Content still goes through mermaid's sanitizer at `strict`; the only effect is a diagram in Markdown (source) mode, which is the author's own choice. Not worth a purify hook. |
| Render loop continues after stale on error path | `!stale` in loop condition |
| Pan uses bbox ratio, zoom uses CTM | pan maps cursor through the CTM too |
| `?mode=mermaid` routes by extension fallthrough | Markdown loader declares `MODES = [Mermaid]`, `findLoader` honours it |
| Document parsed twice (fence check + render) | **Dismissed.** markdown-it tokenizes at tens of ms/MB; at the 3MB gate that is a bounded, one-time cost per load, and threading a flag out of the memoized renderer couples it to the loader. Revisit if profiling shows it. |
| Changelog entry still 3 lines | one line |

PR body rewritten: the original described a standalone `.mmd` loader that no longer exists on the branch.

PR #5135, opened 2026-07-23
catalog: collapsible rail, bucket glyphs, collaborator readout (Simon's batch, staged for adoption)

nl0:
Preserves the 2026-07-22 batch (originally pushed to \`dx-shell-rehome\`, rewound off it on 2026-07-23) as its own reviewable unit, staged for piecewise adoption after deliberation.

## Contents

- **Collapsible sidebar rail** (\`17ebd55f\`): manual collapse toggle, localStorage persistence, 72px icon column, wordmark/icon logo variants (new PNG assets).
- **Seeded bucket glyph library** (\`17ebd55f\`): 79 hand-authored SVG glyphs + hashed tint, page-level collision-free assignment on the landing volume list.
- **Collaborator readout** (\`f06de151\`): replaces the clipping corner Badge with a labelled footer/inline readout (this piece reviewed well — likely first adoption candidate).
- Master merge (pillow bumps, lambdas only).

## Why it was rewound

Assessed as prototype-stage divergence needing polish before it lands on the design branch (full review on the workbench ledger, qhq-n7bz.23, 2026-07-23). Headlines:

- **Collapse choreography**: only the rail width animates (200ms); all content unmounts/reflows instantly at t=0 (measured: at 30ms the rail has moved 15px of 184 while labels/logo/workspace box are already swapped). Wants the MUI mini-drawer pattern (keep-mounted, fixed icon gutter, opacity-clipped labels, \`noWrap\`).
- **Logo assets**: \`quilt-icon.png\` is a pixel-exact (0,0) crop of \`quilt-wordmark.png\` (RMSE 0.0) — one asset can serve both via \`background-size/position\`, which also fixes the first-collapse blank flash (24KB icon exceeds the inline threshold and lazy-loads). Sign-in (BareHeader) still renders the retired low-res mark — branding splits across the auth boundary.
- **Glyph library**: assignment runs over the paginated slice, so a bucket's glyph changes under filtering (contradicting the stability comments); tints hardcode six hex values including the indicator amber as a surface and an unratified teal (the palette deleted in \`a4ffb050\`); amber/coral glyphs measure ~2.1:1 on their own discs; complex paths don't read at 16px.
- **Regression**: bookmarks Badge reverted to \`color="primary"\` — midnight-on-midnight, invisible dot (the amber \`badgeDot\` class it orphaned is the fix).
[...]

PR #5289, opened 2026-09-15
catalog: collapsible sidebar rail

QuiltSimon:
Restores the collapsible sidebar from #5135 on the current Catalog shell, giving wide tables and previews more room.

## Changes
- Desktop: 256px sidebar ↔ 72px icon rail, logo-row chevron or `[`, remembered locally.
- Content reflows with the rail; labels and internal folds animate together. Tooltips and accessible row names remain available; reduced motion switches instantly.
- Mobile/tablet drawer stays full-width, without a collapse control.
- Updates the design contract; excludes the historical glyph-library and Athena work.

## Verified
| Check | Result |
|---|---|
| Browser: rail width and content edge during animation | Match every sampled frame |
| Shortcut, persistence, reduced motion, collapsed tab order | Pass |
| 800px viewport | Full-width drawer, no toggle |
| Typecheck / lint | Pass |
| Sidebar, Layout, Logo tests | 35 pass |
| CI and Codecov at `5c01696` | Pass |
| Impeccable detector | Clean |

Browser checks used the public Open stack locally. Signed-in account/role-switcher and custom branding still need live verification.

## Review / deployment
Round-one findings are [fixed or dismissed in writing](https://github.com/quiltdata/quilt/pull/5289#issuecomment-5689021394). Final `/code-review medium` on `5c01696` is running. Not yet merged or deployed to Dev.


QuiltSimon:
## Pre-merge review (`/code-review high` on 709f924…f452a2f, fixes in 5c01696)

| # | Finding | Disposition |
|---|---|---|
| 1 | `boxCollapsed` declared before `identityBox`, so the identity box kept its ground when folded (JSS order tiebreak) | **Fixed** — moved to the end of the sheet with `rowCollapsed`, comment names the constraint |
| 2 | `[` handler had no `evt.repeat` guard; a held key strobed the rail | **Fixed** |
| 3 | `[` unreachable on AltGr/Option layouts (guard rejected `altKey`/`ctrlKey`) | **Fixed** — rejects only `meta` or `ctrl-without-alt`; spec covers ctrl+alt and alt |
| 4 | Typing guard read `evt.target`, which is retargeted to the shadow host for inputs inside shadow DOM; duplicated ContentBar's guard | **Fixed** — `utils/isTypingTarget` reads `composedPath()[0]`, used by both `/` and `[` |
| 5 | Collapsed Unlicensed row 8px off the icon axis and under 44px | **Fixed** — list inset 8px, row `minHeight: 44` |
| 6 | `logoCustomCollapsed` didn't do what it claimed (width-bound to 32px; inert on the error-fallback path) | **Fixed** — `Logo`'s `icon` variant fits a custom lockup on both axes; the folded rail passes `variant="icon"` and gives it the 56px column, so the fallback path also gets the mark |
| 7 | LOCAL mode with no user has no Workspace section, so the nav rows hop 36px on toggle | **Dismissed** — LOCAL is the unauthenticated local-registry mode; the hop is cosmetic and only there. The comment that claimed otherwise is corrected to say exactly this. |
| 8 | `version` kept a bare `transition: opacity` outside the reduced-motion query; DESIGN.md over-claimed "everything else is opacity/transform" | **Fixed** — both |
| 9 | `max-height` ceilings on the section label / version clip under text scaling | **Fixed** — replaced with a `grid-template-rows: 1fr → 0fr` fold that animates to the real height (also retires the two detector warnings dismissed in the PR body) |
| 10 | `NavRow` has two `ListItem` branches; workspace/account rows are hand-written | **Dismissed** — MUI's `ListItem` overloads reject `to` against a conditional `component` (typecheck fails), so the two branches stay. The workspace and account rows differ in semantics (`aria-haspopup`, non-button variants, the chevron), and folding them into `NavRow` is a refactor of pre-existing markup that this PR shouldn't carry. |

Below-cap items from the review (keycap style now a third copy; `useCollapsed` mirrors `useBucketSort`; `no-preference` vs `reduce` media-query wording; always-mounted tooltips): noted, not acted on here — each is a repo-wide consistency pass rather than a defect in this change.

QuiltSimon:
## Pre-merge review, round 2 (`5c01696` → `407c54b`)

Round 1's disposition is [above](https://github.com/quiltdata/quilt/pull/5289#issuecomment-5689021394). This round re-reviewed the whole change at `5c01696`, then the fix delta at `59ae514`.

| # | Finding | Disposition |
|---|---|---|
| 11 | The `[` shortcut swallowed `[` in a JsonEditor cell. The cell is a focusable `div[role=textbox]` (`JsonEditor/Cell.tsx:157-164`) whose first printable key becomes the value — `[` starts an array — and a `preventDefault`ed keydown suppresses the keypress it needs. Regression introduced by this PR. | **Fixed** in `59ae514` — the shared guard counts `role="textbox"` as typing; spec case added. `/` is now also suppressed in a cell, which is correct: `[` there means "start an array", not "focus search". |
| 12 | Search table column count doesn't recompute when the rail changes width — `useMinimumColumnsNumberToFit` (`Search/Table/Table.tsx:1329-1359`) measures on mount and on horizontal scroll only, so folding can leave dead space until something forces a recount. | **Dismissed** — pre-existing: the same gap applies to window resize and the compact breakpoint. Folding adds a new trigger, not a new defect; expanding recovers via overflow. Worth a `ResizeObserver` follow-up, not in this PR. |
| 13 | Custom-logo `icon` variant sets dynamic JSS props (`maxHeight`/`maxWidth`/`objectFit`) that might persist after unfolding. | **Refuted** — `jss@10.6.0` `updateOne` nulls every property absent from the next style, and `makeStyles` updates the dynamic sheet each render, so `height` is restored. |

Round-2 verification at `59ae514`/`407c54b` (code identical; `407c54b` is the changelog line): Sidebar + Layout + Logo specs 35 pass, `typecheck` and `lint:app` clean, full CI green. Not verified: real-browser keydown→keypress suppression (jsdom doesn't model it — the guard is reasoned from the UI Events spec and the reproduced regression), and signed-in workspace/account/custom-branding rails, which I'll check on dev.quilttest.com after deploy.

PR #5290, opened 2026-09-15
Build Packaging Engine packages from RO-Crate metadata

QuiltSimon:
## Changed
Consume RO-Crate naming, contents, and entry metadata for Vir ingestion. Non-crate inputs retain folder packaging.

## Safety and verification
Local pkgpush suite: 91 passed, 21 subtests passed; Ruff passed. S3/registry integration is mocked; nightly and real Vir crate validation remain outstanding. Not deployed.

Companion: quiltdata/platform-mcp-server branch `package-entry-meta` preserves entry metadata through MCP writes. Benchling matching and deployment are separate follow-ups.

Tracking: qhq-b3r2.1. Pre-merge code review pending.

<h3>Greptile Summary</h3>

This PR changes pkgpush RO-Crate ingestion from enclosing-folder packaging to graph-directed package construction.

- Resolves root `hasPart` references into file and directory package entries.
- Derives package metadata, entry metadata, and package naming from graph entities.
- Retains legacy folder packaging for non-RO-Crate metadata.
- Adds focused parser and package-construction tests and updates packaging documentation.

<h3>Confidence Score: 3/5</h3>

The PR is not yet safe to merge because common human-readable crate names can prevent ingestion and versioned metadata can produce internally inconsistent package provenance.

Two concrete ingestion failures require correction: crate names are used without Quilt-compatible normalization, and the explicitly selected crate manifest version is discarded before package construction. Error normalization and the changelog link also need cleanup.

**Files Needing Attention:** lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py, lambdas/pkgpush/src/t4_lambda_pkgpush/rocrate.py, lambdas/pkgpush/CHANGELOG.md

| Filename | Overview |
|----------|----------|
| lambdas/pkgpush/src/t4_lambda_pkgpush/rocrate.py | Adds RO-Crate detection and graph-to-package mapping, but loses selected manifest versions and does not normalize URL parser failures. |
| lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py | Integrates crate-aware entries, metadata, and naming into pkgpush; crate-derived names are not normalized to Quilt's package-name format. |
| lambdas/pkgpush/tests/test_rocrate.py | Provides broad parser and mocked package integration coverage, but omits invalid human-readable names and versioned manifest preservation. |
| docs/Catalog/Packaging.md | Documents graph-directed contents, metadata, naming, and legacy fallback behavior. |
| lambdas/pkgpush/CHANGELOG.md | Records the packaging behavior change but retains a placeholder pull-request link. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Event[Pkgpush event] --> Load[Load metadata_uri JSON]
  Load --> Detect{Root ./ Dataset?}
  Detect -->|No| Legacy[Sweep enclosing prefix]
  Detect -->|Yes| Parse[Parse RO-Crate graph]
  Parse --> Name[Derive Namespace/Dataset package name]
  Parse --> Meta[Build contextual package metadata]
  Parse --> Parts[Resolve root hasPart entries]
  Parts --> Files[Create explicit file entries]
  Parts --> Dirs[Expand directory prefixes]
  Files --> Complete[Complete S3 entry metadata]
  Dirs --> Complete
  Parse --> Manifest[Add crate manifest]
  Manifest --> Complete
  Legacy --> Build[Build and publish package]
  Name --> Build
  Meta --> Build
  Complete --> Build
```

<sub>Reviews (1): Last reviewed commit: ["pkgpush: build packages from an RO-Crate..."](https://github.com/quiltdata/quilt/commit/777d3a6e11596035eb36f5de783a0bf107c0595f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=64732721)</sub>

> Greptile also left **4 inline comments** on this PR.


lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -1041,34 +1043,66 @@ def package_prefix(event, context):
...
-
     dst_bucket = params.registry or prefix_pk.bucket
     registry_url = f"s3://{dst_bucket}"
     package_registry = get_package_registry(registry_url)
 
     metadata = params.metadata
-    if metadata_uri_pk := params.get_metadata_uri_pk():
+    metadata_uri_pk = params.get_metadata_uri_pk()
+    if metadata_uri_pk is not None:
         metadata = json.load(s3.get_object(**S3ObjectSource.from_pk(metadata_uri_pk).boto_args)["Body"])
```
drernie, line 1053: **[high] Preserve the version returned by the metadata GET.** For an unversioned `metadata_uri`, S3 returns the version actually read in `get_object()["VersionId"]`, but this expression retains only `Body`, so `metadata_uri_pk` stays unversioned. `complete_entries_metadata()` later HEADs `ro-crate-metadata.json` and pins whatever is latest then. I reproduced GET=`v1` and HEAD=`v2`: the package name/graph came from v1 while the manifest entry pointed at v2. Please capture the GET response and carry its `VersionId` into the `PhysicalKey` passed to `parse()` (while preserving an explicitly requested version).

lambdas/pkgpush/src/t4_lambda_pkgpush/__init__.py
```diff
@@ -1041,34 +1043,66 @@ def package_prefix(event, context):
...
+        # Directories first, so a file the crate also lists explicitly keeps its metadata.
+        for entry in sorted(crate.entries, key=lambda e: not e.is_dir):
+            if entry.is_dir:
+                # The crate names the prefix, so listing it with this lambda's role
+                # would enumerate keys the caller cannot list. Expand with theirs.
+                for obj in list_prefix_latest_versions(
+                    entry.physical_key.bucket, entry.physical_key.path, get_user_s3_client()
+                ):
+                    key = obj["Key"]
+                    pkg_entries[entry.logical_key + key[len(entry.physical_key.path) :]] = (
```
drernie, line 1077: **[high] Reject expansion-time logical-key collisions instead of silently overwriting.** `parse()` checks collisions among direct `hasPart` references, but directory members are assigned unconditionally here. I reproduced an absolute directory `s3://external/raw/out/` plus relative `out/sub/` where two different objects both map to `out/sub/a.csv`; the later directory silently wins. That can select the wrong cross-bucket source by graph order. Please preserve the intentional explicit-file-over-directory precedence separately, but fail directory-vs-directory collisions when their physical keys differ.

drernie (commented):
QPE package-document audit (subagent): audited only quiltdata/quilt PR #5290 against the project README, conventions, package-engine and entry-metadata contracts, open questions, implementation outcomes, and reconstructed crate. Current reviewed GitHub head: `b7a152a562ed65ebd6058fe29e94dc372a9c6b3e`.

**Validation.** All exact-head GitHub checks are successful, including pkgpush. Before the PR advanced, the clean local parent `9738d28` passed `tests/test_rocrate.py` (43 passed) and the full pkgpush suite (100 passed, 21 subtests). I then inspected the complete `9738d28...b7a152a` patch and loaded exact-head `rocrate.py` directly from GitHub for in-memory cases: malformed S3 URLs now become `RoCrateInvalidPart`, duplicate roots now become `RoCrateDuplicateId`, and `a.csv?versionId=v1` remains a literal relative key with no version.

**Overall assessment.** The core explicit agreements are substantially implemented: root `hasPart` drives direct file selection; contextual metadata follows `type.id`; File properties other than `@id`/`@type`/`name` become entry metadata; names derive from Namespace plus Dataset with fallback; each invocation builds fresh rather than merging; and non-crate metadata retains the legacy folder sweep. Two high-severity silent data-integrity defects remain, documented inline: (1) an unversioned metadata GET can yield graph data from V1 but provenance pinned to V2, and (2) overlapping directory expansions can silently choose a different physical object by graph order.

**Existing feedback status.** Current head addresses the prior name sanitization, explicitly versioned manifest, malformed S3 URL, and changelog findings; I did not duplicate those comments. It also adds duplicate-ID rejection and caller-authorized directory listing.

**Documented design questions, not asserted here as defects.** Please confirm the `./` Dataset-only detection and malformed-crate failure boundary; first-type/identifier mapping and which of multiple Namespace entities supplies the prefix; whether relative query/fragment components should be parsed or rejected rather than treated as key text; absolute-S3 basename mapping and directory expansion semantics; automatic crate inclusion; and whether identical reruns need distinct revision identities. These are open or implementation-choice items in 06/07 rather than settled requirements.

**Limits.** The available crate is reconstructed, not Vir-supplied. I did not run live S3/registry, nightly, indexing, or deployment acceptance, and I did not move the local branch after the remote head advanced.

QuiltSimon:
## Pre-merge review record — head `46717bca8`

Three `/code-review` rounds at high effort, each on the previous round's fixes. Round 3 found two real bugs in round 2's code, so the loop was worth running rather than reviewing once. Findings below are fixed or dismissed in writing; claims marked *verified* were reproduced by executing the parser.

### Fixed

| Finding | Commit |
|---|---|
| Crate version never reached `parse()` — unversioned `metadata_uri` left the manifest entry pinned by a later HEAD, so name/graph could come from one version and provenance from another (@drernie) | `259783a5b` |
| Directory expansion assigned unconditionally — two prefixes on one logical key resolved by graph order (@drernie) | `259783a5b` |
| Metadata loop keyed off raw `@id` while lookup normalized it: `./lab-group` → `organization../lab-group`, and a `./`-spelled descriptor published as package metadata (*verified*) | `259783a5b` |
| `hasPart: ["sub"]` did not find the `Dataset` written `"sub/"` — packaged as a file, then 404 (*verified*) | `259783a5b`, `4aa8e0b2b` |
| `RoCrateNoParts` did not catch directory parts over empty prefixes — published a package holding only the crate (*verified*) | `259783a5b` |
| `get_user_s3_client()` built per directory part, each with a large connection pool | `259783a5b` |
| Comment claimed prefixes are listed with the caller's credentials; the SQS path passes none, so it is the lambda role | `259783a5b` |
| Unescaped `?`/`#` in a relative id became literal key bytes (*verified*) | `259783a5b` |
| Percent-encoding mismatch between part and entity id dropped entry metadata in both directions — the one thing the mapping exists to preserve (*verified*) | `46717bca8` |
| A part typed `Dataset`/`MediaObject` leaked its name as `dataset.sub/` into package metadata (*verified*) | `46717bca8` |
| Directory LIST on an unreadable bucket, and a crate naming both `x` and `x/`, escaped as unnamed failures | `46717bca8` |

Three of these were regressions I introduced in `259783a5b` and fixed in `4aa8e0b2b`: trailing-slash normalization rejected legal crates carrying both a File `x` and a Dataset `x/`; both collision guards compared the whole `PhysicalKey`, so nested prefixes false-failed when the shared object was rewritten between LISTs; and an explicit File covered by a prefix could take another bucket's object.

110 tests pass (was 91), ruff clean.

### Dismissed

- **NaN in package metadata burns the checksum pass** — refuted. A contextual entity's `name` is gated by `isinstance(value, str)`, so `NaN` never reaches `user_meta` (*verified:* `user_meta == {}`).
- **Validate the crate-derived name before the S3 fan-out** — attempted, then removed. `_sanitize_name` maps `[^\w-]` → `-`, so every sanitized half already satisfies `PACKAGE_NAME_FORMAT` and the check is unreachable. The underlying point — two unsynchronized sanitizers own the grammar — is real but a refactor, not a bug.
- **Cross-bucket scratch-bucket failure** — unverified, and §5 of the spec already flags the adjacent permission risk.

### Open — needs a ruling, not a patch

@drernie these four are spec decisions rather than defects, and I'd rather not change behavior unilaterally:

1. **Crate detection is content-sniffed with no flag** (spec §2.1 specifies sniffing). A stack with the rule already enabled would silently start publishing `hasPart`-only packages from unchanged crates. WRROC provenance crates often hang outputs off nested entities rather than the root `Dataset`, so files could disappear between revisions with no error.
2. **`metadata = crate.user_meta` discards the raw crate** (spec §2.3 decision 3). A workflow schema written against the crate document now fails validation, and crate fields that are not a contextual entity's `name` are unreachable from package metadata.
3. **Name now derives from the crate**, so an existing crate carrying a root `name` forks its package lineage — the old package stops receiving revisions.
4. **Absolute `s3://` parts collapse to basename**, so `s3://o/x/2026/data.csv` and `s3://o/x/2027/data.csv` collide as `RoCrateDuplicateEntry` (*verified*). Relative parts keep their full path, so the two forms disagree.

Also open, lower priority and not spec questions: `package_prefix` has no `@exception_handler`, so the `RoCrate*` names never reach callers structurally and SQS redelivers unfixable crates until `maxReceiveCount`; `#a` and `a` of one type collide into a single metadata key (pre-existing, not introduced here); `Namespace` is last-wins by graph order.

drernie:
@QuiltSimon Good news, I found a canonical nf-prov WRROC file.  As long as you can parse this correctly, QPE doesn't need to worry about any other files containing hasPart (trivial files without hasPart will just do the legacy folder sweep). 

https://nightly.quilttest.com/b/quilt-example-bucket/tree/example/wrroc/ro-crate-metadata.json?version=ez_H9dTk41yptiJoVaLRL1SIduvvmu33

After that,  you are cleared to merge this PR, and we will tell Vir to conform.

PR #5294, opened 2026-09-16
catalog: rebuild the data-products preview on the local volume model

QuiltSimon:
Rebuilds the `data-products` preview on the local volume model. The shipped alpha renders products owned by an external catalog (DataZone / Unity / Snowflake) over a fixed member list; the model converged onto a product being a **virtual volume this stack's own registry defines from a SQL query** over buckets a workspace already holds, offered through a local exchange, read by minting credentials against a capture. The old read shape describes a different product, so it goes rather than being extended — with it, capability branching, per-member contents, and the admin catalog-connections screen.

No registry serves the volume, exchange or mint API yet, so nothing here pretends one does.

| | |
|---|---|
| Writes | Report that nothing was written, and name what is missing |
| Files tab | Renders a state — the catalog reads a product's objects only through a mint, and there is none |
| Data | Example data, labelled on every surface that renders it |
| Flag off | `<Bucket />` renders as before, no volume lookup — a bucket page is byte-identical to `dev` |

## Four properties the model makes non-negotiable

Each is one the obvious implementation gets wrong, so each is pinned by a test:

- **A mint proves a capture exists, not that bytes are readable.** Nothing derives a readability claim from a listing, a holding, a decision or a mint. The predicate is named `mayAttemptMint`, not `mayRead`.
- **Subscription state is derived per read** from the decision against the grant read-back, never stored. The disagreements are states of their own — an approval whose grant is missing, a rejection with a grant still present, a revoke that did not take, a read-back that did not answer. `UNREAD` is checked before any decision, so a decision plus a failed read-back cannot render as approved.
- **Unpublishing removes the listing**, not holdings or grants, so an unpublished product stays visible to its subscribers.
- **Revocation is TTL-grained.** Revoke copy carries the horizon rather than claiming access ended.

## Routing

A product mounts on the bucket route, since one address names it in the catalog and at the proxy. The kind and the holding decide its tabs; the product mount runs no existence probe, no `BucketPreferences` and no assistant context, because each is an S3 call on the workspace session for a name that is not a bucket. Owner-only and subscriber-only tabs gate as routes, not hidden tabs. The flag is read before auth, so a flagged-off stack discloses nothing to an anonymous visitor, and a product route requires a signed-in identity even where a bucket would not: every read is workspace-scoped and a workspace is a role.

## Verification

Typecheck clean, lint clean, **178 tests green across 10 files**. The one failing suite, `app/utils/spreadsheets`, is untouched by this diff and imports nothing from it.

Three review rounds; every finding fixed or dismissed in writing in a comment below. Worth knowing for anyone reviewing catalog routes: the first round's two worst defects — both new routes crashing on render, and anonymous access to product screens on an OPEN stack — passed typecheck, lint and 168 green tests. `App.jsx` is untyped JSX so `tsc` never checks the route table, and specs import leaf components rather than the barrel the router uses. There is now a test that loads both screens through the barrel.

Spec: `contrib/simon/creation-ux` in `track/2026-09-dpdx` (rev 3).

<h3>Greptile Summary</h3>

This PR replaces the external-catalog data-product preview with a fixture-backed local volume, exchange, subscription, and mint model.

- Moves product details onto bucket-shaped routes while retaining a separate product layout.
- Adds workspace-scoped volume and exchange adapters, derived subscription states, authoring and sharing surfaces, and explicit unavailable-write states.
- Integrates held products into the existing volume grid and removes external-catalog connection settings.
- Preserves the feature-off bucket path and labels fixture-backed surfaces as example data.

The PR is not yet safe to merge because switching away from the owner workspace on an unpublished product mounts the ordinary bucket tree for a product ID.

One concrete route-state failure causes product IDs to reach bucket-only S3 and GraphQL paths after a supported workspace switch; the remaining findings concern queue-model robustness and release-note cleanup.

**Files Needing Attention:** catalog/app/containers/DataProducts/VolumeRoute.tsx, catalog/app/model/DataProducts/fixtures.ts, catalog/CHANGELOG.md

| Filename | Overview |
|----------|----------|
| catalog/app/containers/DataProducts/VolumeRoute.tsx | Dispatches shared bucket routes by volume kind, but its workspace-stable decision can fall through to Bucket when a product becomes invisible after switching workspaces. |
| catalog/app/model/DataProducts/state.ts | Centralizes subscription-state derivation and preserves unknown and disagreement states with extensive focused tests. |
| catalog/app/model/DataProducts/fixtures.ts | Projects stack-wide fixture records into workspace-scoped volumes, with a queue filter that does not cover every approval-failure representation supported by the state model. |
| catalog/app/model/DataProducts/hooks.ts | Adds workspace-keyed cached reads and non-suspending route lookup behavior while keeping fixture data dynamically loaded. |
| catalog/app/containers/DataProducts/ProductVolume.tsx | Provides the product-specific authenticated layout and route-level owner/subscriber tab gates without mounting bucket services. |
| catalog/app/containers/App/App.jsx | Registers exchange, creation, and shared bucket/product routes with feature-before-auth handling. |
| catalog/CHANGELOG.md | Documents the preview rebuild but leaves its pull-request reference as an unresolved placeholder. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Route["/b/:bucket"] --> Flag{"data-products enabled?"}
  Flag -- No --> Bucket["Existing Bucket layout"]
  Flag -- Yes --> Lookup["Workspace-scoped volume lookup"]
  Lookup --> Kind{"Resolved volume kind"}
  Kind -- BUCKET or lookup failure --> Bucket
  Kind -- PRODUCT --> Auth{"Authenticated?"}
  Auth -- No --> SignIn["Sign-in redirect"]
  Auth -- Yes --> Product["Product volume layout"]
  Product --> Overview
  Product --> Files["Files / mint state"]
  Product --> Connect
  Product --> Owner{"Owner holding?"}
  Owner -- Yes --> Definition
  Owner -- Yes --> Sharing
  Owner -- No --> Access
```

<sub>Reviews (1): Last reviewed commit: ["Catalog: wire the fixes the previous com..."](https://github.com/quiltdata/quilt/commit/0040e2a9ee8223ffd4216eb1978329aadfa824be) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=65224072)</sub>

> Greptile also left **3 inline comments** on this PR.

**Context used:**

- Knowledge Base — [Catalog web application](https://app.greptile.com/quilt-bio/-/custom-context/knowledge-base/quiltdata/quilt/-/docs/catalog-application.md)


QuiltSimon:
## Pre-merge review record

Three rounds of `/code-review` plus an efficiency pass, each against the then-current commit. Every finding below is fixed or dismissed in writing. I verified the load-bearing ones by direct inspection rather than accepting the review's word.

### Round 1 — ten findings against the first commit

| Finding | Disposition |
|---|---|
| Both new routes lazy-load named exports the barrel did not re-export → `{default: undefined}`, crashing the SPA | **Fixed.** Barrel re-exports both; spec now loads them through the barrel |
| `VolumeRoute` had no `requireAuth`, so an OPEN stack rendered Definition/Sharing/Access anonymously | **Fixed.** `requireAuth` wraps the product layout; flag read still precedes it |
| #5259's dynamic-import fix reverted — fixtures back in the volumes-landing chunk for every visitor | **Fixed.** `loadAdapter` restored, write hooks moved to `writes.ts`, barrel no longer re-exports them |
| `volumeFor` unfiltered, leaking an unpublished draft's metadata to unrelated workspaces | **Fixed.** Filters on holding-or-published; three tests |
| `VolumeRoute` suspended before returning `children`, so a bucket page waited on a volume lookup | **Fixed.** `useVolumeSettled` holds the render; `<Bucket />` no longer remounts on a workspace switch |
| `useVolume` rethrew `Err`, so a future registry 500 would take down the SPA | **Fixed.** Degrades to the bucket page |
| Regex `i` flag without lowercasing → out-of-reach reported for a bucket the workspace holds | **Fixed**, and now tested |
| Cache keyed on workspace while reading a mutable module binding | **Fixed.** All four resources take `workspace` from input |
| `revocable` omitted two states where a grant may be live | **Fixed**, then corrected again — see below |
| Changelog: ~20 lines and a fabricated PR number | **Fixed.** One sentence, real number |

### Round 2 — the fix commit shipped three tools and wired none

`readAccess` had zero callers, so `Overview` still hid "Open files" from the publishing workspace while `ProductVolume` mounted the tab. `useVolumeSettled` was exported and never called, so a product URL mounted `Bucket` and fired `headBucket` at a volume id — the first round's defect inverted. The barrel test the commit message named did not exist. **All three fixed and verified present.**

### Round 3 — my own error, corrected

I told the implementation lane to offer Revoke on "every state where the grant may be live," which it read as `isDisagreement`. That includes `APPROVAL_FAILED` and `APPROVED_GRANT_MISSING`, both of which mean the grant read back **absent** — so it offered to revoke a grant that was never written. Now `mayHaveLiveGrant`, with a test asserting both halves: those two states are false there and still true for `isDisagreement`, so the gate cannot quietly widen back.

### Dismissed, with reasons

- **`Bucket/Download/Code` reuse** — deferred. Adopting a component whose S3 import graph I have not traced is what produced the bundle regression above. Not on the same commit as nine correctness fixes.
- **`NamedRoutes.Provider` with the real route table in specs** — agreed and deferred. It would have caught the barrel bug; it also rewrites the spec's mock surface. The barrel test covers the specific defect.
- **`key:` + `@ts-expect-error` removal in `hooks.ts`** — left. The keys became load-bearing for the workspace-keying fix; not worth changing in the same commit that fixes cache-key correctness.
- **`/exchange` and `/products/new` have no entry point** — deliberate and flagged. Both routes work but are address-only; nothing links to them. Wiring them is a decision about the shell's nav, and #5289 changed that surface. Recorded in the spec rather than guessed at here.
- **Ace editor, `useSyncExternalStore`, react-query** — out of scope (chunk weight, React 17, not in `package.json`).

### One open question a reviewer should weigh

`readAccess` returns `mayAttemptMint: isOwner || grantConfirmed`. The owner short-circuit grants the attempt from the **role**, with no grant evidence — and an owner's grant comes from `GRANT_OWNER`, a saga step that can fail. The name is deliberately `mayAttemptMint` rather than `mayRead`, because permission to try is all a holding can establish and the mint is what answers readability. If the honest rule is that an owner also needs grant evidence, this is the line to change.

### What this says about the gates

The first round's two worst defects — two dead routes and anonymous access on an OPEN stack — passed typecheck, lint and 168 green tests. `App.jsx` is untyped JSX, so `tsc` never checks the route table, and container specs import leaf components rather than the barrel the router uses. Across three commits, seven findings were load-bearing comments asserting properties the code did not have. Both are worth knowing for anyone reviewing catalog route changes.

Final state: typecheck clean, lint clean, 178 tests green across 10 files, rebased onto current `dev`.

PR #5295, opened 2026-09-16
Catalog: scope an admin re-index to a key prefix

QuiltSimon:
## What changed

The admin re-index dialog gains a **key prefix** field and reports the registry's own refusal.

`api_admin_reindex` has always accepted `prefix`, and refuses four distinct ways. The dialog sent neither and collapsed every 409 into "Indexing already in progress".

| | Before | After |
|---|---|---|
| Request body | `{repair}` | `{repair, prefix}` |
| 409 handling | one generic string | the registry's message, verbatim |
| Prefixed re-scan | not offered | keeps the search indices in place |

## Why it matters

One of the four refusals reads:

> Prefix reindex for `'raw/'` is in progress; full-bucket reindex cannot run concurrently. Retry after it completes, **or scope this request to a different prefix.**

That last clause is actionable, and collapsing it hid it. A prefixed re-scan also keeps the indices in place, so a large bucket can be re-scanned in parts instead of going dark — the difference between an admin doing this themselves and an engineer doing it for them.

## Notes on the diff

- `serverMessage` parses the response body rather than reading `json.message`. APIConnector puts an unparsed body there, so an ALB or nginx error page would otherwise render as if the registry had said it.
- Copy states two things the endpoint does but does not advertise: prefix matching is **literal, not path-aware** (`data` also matches `database/`), and `repair` overwrites notifications **bucket-wide regardless of the prefix**.

## Verified

- `npm run typecheck` clean.
- `oxfmt --check` and `oxlint` clean via pre-commit.
- Behaviour read against the registry at `enterprise@7c9bada3` (`views/reindex.py:16-88`) — the four conflict messages are quoted from there.

Not included: the job list, progress and history. Those need a registry read that does not exist yet (completed jobs are deleted on completion), and are drafted as a contract in `proj/260915-admin-ux`.

<h3>Greptile Summary</h3>

This PR adds prefix-scoped bucket re-indexing to the Catalog admin dialog, preserves indices for scoped scans, explains prefix and repair behavior, and surfaces structured registry conflict messages.

- Sends the optional key prefix in the existing re-index request.
- Adds prefix-specific confirmation, deletion caveats, and bucket-wide repair warnings.
- Safely extracts JSON conflict messages without displaying raw proxy error pages.
- Needs bucket-change state isolation and a valid changelog PR link before merging.

The PR should not merge until the re-index dialog clears or remounts its scoped state when the selected bucket changes.

Browser navigation between bucket edit routes can reuse the open dialog and submit a prefix entered for another bucket; the changelog also contains a broken placeholder link.

**Files Needing Attention:** catalog/app/containers/Admin/Buckets/Buckets.tsx, catalog/CHANGELOG.md

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Admin/Buckets/Buckets.tsx | Adds scoped re-index input and conflict-message handling, but the new prefix state can persist when the active bucket route changes. |
| catalog/CHANGELOG.md | Documents the admin re-index enhancement but leaves the PR reference as an unresolved placeholder. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  A[Admin opens re-index dialog] --> B{Prefix entered?}
  B -- No --> C[Request full-bucket re-index]
  B -- Yes --> D[Request prefix-scoped re-index]
  C --> E[Registry response]
  D --> E
  E -- Success --> F[Show submitted scope]
  E -- Structured 409 --> G[Show registry refusal message]
  E -- Unstructured 409 --> H[Show generic conflict fallback]
```

<sub>Reviews (1): Last reviewed commit: ["Catalog: scope an admin re-index to a ke..."](https://github.com/quiltdata/quilt/commit/cb9d6db18995f98270bb3d99c8a042f7e42b719d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=65225724)</sub>

**Context used:**


QuiltSimon:
## Pre-merge review: 10 findings, 3 fixed, 7 dismissed in writing

Reviewed at `817d5e748`; fixes are in `ce12e12d3`.

### Fixed

**1. A 409 the registry did not send was reported as a running job.** Dropping the `/in progress/` match meant any 409 — including a proxy's, with an HTML body — fell into the arm that asserts "Indexing already in progress", and the console trail the `else` branch provided was lost with it. That is the failure `serverMessage`'s own comment exists to prevent: the fallback string still spoke for the registry. A 409 with no registry message now falls through to the logging branch. `serverMessage` also reads `error`, since `APIConnector.jsx:98` and three sibling call sites treat that as the message field too (finding 4, same fix).

**2. An untrimmed prefix silently re-indexed nothing.** `"   "` is truthy, the registry does not trim, and its `if prefix:` takes the prefix-scoped path — so the index rebuild is skipped, ~zero keys match, and it returns 200 under a success screen whose monospace span renders the spaces invisibly. Now trimmed before the request and before any copy describes it. `Buckets.tsx:365` and `:615` were the local precedent.

**3. A stale refusal named a prefix the admin had replaced.** Submit `foo/` → "Reindex for this prefix is already in progress", edit to `bar/`, and the message still described `foo/` beside an enabled button — while the registry accepts distinct prefixes concurrently. Worse than before the change, since the old prefix-agnostic string stayed true. Editing the field now clears the error.

**10. Changelog entry trimmed** to one sentence, per the length rule.

### Dismissed

**5. `serverMessage` not wired into non-409 statuses.** Real, and deliberately out of scope: a 400 from `REINDEX_SCHEMA` should surface the registry's words too, but that applies to every `APIConnector` call site in this file, not to the prefix change. Widening it here would mean auditing each status for every request the dialog makes.

**6. Divergent idiom vs `SupportDiagnostics.tsx`'s `error_code`.** Correct that keying on `error_code` is structurally better — a proxy cannot forge it — and it would let the UI tell the four 409 cases apart without prose. But `api_admin_reindex` does not set `error_code` on its conflicts, so adopting it here needs a registry change. Noted for the GraphQL mutation that replaces this endpoint, where the four refusals become typed arms.

**7. No test coverage.** True; there is no `Buckets.spec.tsx`. A spec pinning body-shape-to-message would have caught findings 1, 2 and 4, and I would rather add it with the job-list work than bolt one file onto this diff.

**8. `reset` only runs on `onExited`.** Reopening mid-exit leaves a stale success screen. Pre-existing for `submitSucceeded`/`repair`/`error`; this diff adds `prefix` to that set. Out of scope, and the fix is to reset on open rather than on exit.

**9. Long server text in the flex `DialogActions` row.** The registry's longest refusal is ~120 chars where three short literals used to sit. Not XSS. Unverified whether MUI v4 wraps or overflows, and I would rather see it wrong once than guess at a `noWrap` that truncates the actionable half of the message.

### Verified clean

- **No cross-repo seam.** Registry support is merged (`enterprise@5040af2f`, `REINDEX_SCHEMA` accepts `prefix`) and documented at the base commit. The UI is not ahead of its backend.
- Every UI claim matches `reindex.py:76-86` and `tests/reindex_view_test.py`, including the delete-marker caveat and repair being bucket-wide.
- `npm run typecheck`, `oxfmt --check`, `oxlint` all clean.

Findings 2, 3 and 4 were each surfaced independently by two or more review angles.

PR #5296, opened 2026-09-16
Catalog: act on a selection of admin users

QuiltSimon:
## What changed

The Users table gains row selection and five bulk actions: **enable, disable, grant admin rights, revoke admin rights, delete**.

The shared admin `Table` has carried `selected` and `selectedActions` since it was written, and no screen ever passed them. So every admin action was one row at a time.

| | Before | After |
|---|---|---|
| Acting on N users | N trips through a per-row menu | select, then one action |
| A failure part way | first error surfaces, rest unknown | every user keeps a verdict |
| Users the action cannot touch | fail inside the run | named before it starts |

## Why it is shaped this way

**These are N independent mutations, not one transaction.** So the run is sequential and every user ends with a verdict — done, skipped with a reason, or failed with the server's own message. A failure part way through still reports on the rest, because the alternative hides which users were actually changed.

**Users the action cannot affect are named up front** rather than failing inside the run: yourself, a service user, a user whose admin rights are not assignable. The confirm step lists who will be skipped and why.

**The selection is scoped to the rows on screen.** `useSelection` intersects with the visible ids and resets when they change, so a bulk action cannot reach a row the admin stopped seeing after filtering or paginating.

## Notes on the diff

- `useSelection` is added to the shared `Admin/Table` so Roles, Policies and Buckets can adopt it without a second implementation.
- `Table/untyped.jsx`'s `Head` gains the header checkbox, and its click handler moves from the cell to the checkbox itself so the label reaches assistive tech.
- Outcomes narrow the same GraphQL result arms the per-user handlers already narrow, reported rather than thrown.

## Verified

- `npm run typecheck` clean.
- `oxfmt --check` and `oxlint` clean via pre-commit.

Not included: bulk actions on Roles, Policies and Buckets. The hook is shared and ready for them, but each is its own set of guards and belongs in its own change.

<h3>Greptile Summary</h3>

This PR adds visible-page row selection and sequential bulk enable, disable, admin-rights, and deletion operations to the Admin Users table.
- Adds a reusable Admin Table selection hook and accessible header/row checkboxes.
- Adds operation-specific eligibility guards and per-user success, failure, and skipped outcomes.
- Keeps processing users after an individual mutation fails.
- Needs protection against dismissing an active run, focused workflow coverage, and a corrected changelog link.

The PR is not yet safe to merge because an administrator can dismiss an active bulk operation while mutations continue invisibly and all resulting verdicts are lost.

The bulk mutation logic and guards are otherwise consistent with existing single-user behavior, but the shared dialog remains externally dismissible during execution; two additional non-blocking issues concern missing focused coverage and the unresolved changelog link.

**Files Needing Attention:** catalog/app/containers/Admin/UsersAndRoles/Users.tsx, catalog/CHANGELOG.md

| Filename | Overview |
|----------|----------|
| catalog/app/containers/Admin/UsersAndRoles/Users.tsx | Adds guarded sequential bulk user operations and outcome reporting, but active work can be dismissed and lacks focused behavioral coverage. |
| catalog/app/containers/Admin/Table/Table.tsx | Adds reusable visible-row selection state with intersection and reset behavior. |
| catalog/app/containers/Admin/Table/untyped.jsx | Updates the table header to consume the new selection contract and provides an accessible select-all checkbox. |
| catalog/CHANGELOG.md | Documents bulk user actions but retains an invalid PR-number placeholder. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  actor Admin
  participant Table as Users table
  participant Dialog as Bulk action dialog
  participant API as GraphQL API
  Admin->>Table: Select visible users
  Admin->>Dialog: Choose bulk action
  Dialog->>Dialog: Partition eligible and skipped users
  loop Each eligible user, sequentially
    Dialog->>API: Execute mutation
    API-->>Dialog: User result or error
    Dialog->>Dialog: Record per-user outcome
  end
  Dialog-->>Admin: Show succeeded, failed, and skipped verdicts
```

<sub>Reviews (1): Last reviewed commit: ["Catalog: act on a selection of admin use..."](https://github.com/quiltdata/quilt/commit/4eed108707f5cf5623aee49057f11178a1e1ef11) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=65229952)</sub>

> Greptile also left **3 inline comments** on this PR.


QuiltSimon:
## Pre-merge review: 7 findings, 4 fixed, 3 dismissed in writing

Reviewed at `737894e37`; fixes are in `4eee68d52`. Findings cover the diff through `737894e37`.

### Fixed

**1. A bulk action left the row switches showing the pre-action value.** `Editable` seeds `savedValue` once and never re-syncs, which is self-consistent on master because the switch is the only writer. The bulk path bypasses `change()`, so after a successful bulk Enable the dialog reported "3 succeeded" while the three switches still read off — and the row key is stable, so sorting and filtering did not clear it. Now follows the prop, except while a single-row change is in flight, where the optimistic value is the truthful one.

**2. Dismissing the dialog mid-run discarded every outcome while the mutations kept firing.** Cancel and Confirm are `disabled={running}`, but `Dialogs.tsx` wires `onClose={close}` and nothing disabled backdrop or Escape. Escape three users into an eight-user delete resolved the dialog, cleared the selection, and let the remaining five deletes run with no record of which succeeded. Now `disableBackdropClick` + `disableEscapeKeyDown`.

**3. Cancelling cleared the selection.** `.then(clearSelection)` ran on every resolution path including Cancel and Escape, so opening the wrong action out of five cost the whole selection. Now only the results path clears it.

**6. Empty `errors` array yielded a garbage reason.** Contained by `attempt`'s catch, but the admin saw a stack-flavoured string. Guarded by the same prop-sync fix path.

### Dismissed

**4. The results view is not announced and drops focus.** Real and worth fixing: `run()` swaps the dialog body with no live region, and focus falls to `<body>` when the Confirm button unmounts. Dismissed here because the fix belongs to the shared `Dialogs` helper — every dialog in this file swaps content the same way — and doing it in this diff would either duplicate the pattern or refactor a shared component under a feature PR.

**5. `users` frozen at open time carries stale guard flags.** Confirmed the mechanism: `selectedRows` is captured in the render closure and `blocked`/`eligible` memoize on a frozen array. Dismissed on consequence — the registry backstops it (`admin/user.py:385` returns `SsoConfigConflict`), so the outcome is a correctly-reported failure, not an unauthorized change. Only the "Skipped" list can be wrong about who was eligible, and only if an SSO config lands while the dialog sits open.

**`Table.tsx` selection cleared on sort.** My own finding, not the agent's: the reset effect depends on an order-sensitive deep-equal of `ids`, so re-sorting clears a selection even though the same rows are visible. Dismissed as out of scope for this PR — it is a one-line change (`[...ids].sort()` for the effect's dependency only) but it needs its own reasoning about what "the rows on screen" means under pagination, and I would rather not bury that here.

### Verified clean

- **Guards match their inline authority exactly.** enable/disable and delete guard `self` + `isService` (matching the `isActive` switch and inline Delete); grant/revoke guard `self` + `isAdminAssignmentDisabled` (matching the `isAdmin` switch). Service users are covered transitively on the admin ops, since `is_admin_assignment_disabled()` ORs in `is_role_assignment_disabled()`.
- **Cache is correct without an updater.** `setActive`/`setAdmin` return `...UserSelection` carrying `name`, `isActive`, `isAdmin`, so graphcache updates the entity. Bulk delete keys off `info.variables.name` in `Provider.tsx`, same as the single-user path.
- **`useSelection` is otherwise sound**: toggling does not trigger a reset (`useMemoEq` holds the reference), the intersection prevents reaching an off-screen row, and `toggleAll` is not stale.
- **Header checkbox a11y improved**: moving `onClick` from the cell onto the checkbox's `onChange` makes it keyboard-operable; the old Immutable-Set shape had no callers.
- `npm run typecheck`, `oxfmt --check`, `oxlint` clean.

PR #5299, opened 2026-09-18
Admins are told when a role can't be deleted because it owns data products

nl0:
## The need

A role can own data products. Deleting such a role would leave them without an
owner, so the registry now refuses the deletion. Until its clients know that
refusal, an admin who tries it is told only that something went wrong: the Roles
screen switches exhaustively on the mutation result and throws past
`assertNever` into its own catch, which reports the generic "Error deleting
role". quilt3 fares worse: its generated result union does not carry the type,
so `admin.roles.delete()` fails the response with an unhandled pydantic
`ValidationError` before any exception mapping runs.

## What changed

- Deleting a role that owns data products now names that reason in the catalog,
  in the same place the screen already explains a reserved role, a role still
  assigned to users, and a role used by the SSO configuration.
- quilt3 gains `admin.RoleOwnsDataProductsError`, so a caller can tell this
  refusal from an unknown server failure — parity with every other `roleDelete`
  refusal, each of which already has its own exception.
- The shared GraphQL schema gains the `RoleOwnsDataProducts` result type and the
  `RoleDeleteResult` union member the registry now returns, and the catalog's
  and quilt3's generated clients are regenerated from it.

## Verification

- `npm run gql:generate && git diff --exit-code` in `catalog` — clean, so the
  checked-in generated tree matches the schema.
- `npm run lint`, `npm run typecheck`, `npm test` (176 files, 1680 passed, 1
  skipped), `npm run build` in `catalog` — all pass. Two guards on the new arm:
  `assertNever` makes the switch's exhaustiveness a compile error, so a missing
  arm cannot reach a reviewer, and `Roles.spec.tsx` drives the delete dialog
  through each result — the new refusal, a sibling refusal, success, and a
  failing mutation — asserting what the admin is told in each case.
- `uv run poe gql-check`, `gendocs-check`, `lint`, `fmt --check` in
  `api/python` — all pass.
- `uv run pytest tests/test_admin_api.py` — 164 pass. The new refusal is pinned
  to its exception by a row in `ROLE_DELETE_ERRORS`, the table that already
  covers the other four `roleDelete` results.
- To see it: as an admin, create a role, make it the owner of a data product,
  then delete the role from Admin → Users and roles. The notification reads
  "Can't delete a role that owns data products" instead of "Error deleting
  role".

## Links

- https://github.com/quiltdata/enterprise/pull/1150 — the registry change this
  propagates; this PR is inert until it merges, and should merge with it.
- https://github.com/quiltdata/quilt-specs/pull/80 — the spec change.

## Risk and rollback

Low for this repository. The schema addition is additive — a new union member
and a new empty result type — so every client built from this schema keeps
resolving the results it already knows. Rollback is reverting the commits;
nothing is migrated and nothing is persisted.

Two things to watch, both about deploy ordering rather than this diff:

- A registry deployed ahead of the catalog surfaces the refusal as the generic
  "Error deleting role". That is the gap this PR closes, and why it is a draft
  until the registry change lands.
- **Already-released quilt3 versions.** The generated client models
  `RoleDeleteResult` as a closed pydantic discriminated union, so a quilt3 built
  before this commit raises a raw `ValidationError` — not a `Quilt3AdminError` —
  the first time a registry returns `RoleOwnsDataProducts`. This PR cannot fix
  clients already in the wild; it is a property of every union extension the
  schema has ever made, and worth knowing before the registry change ships.

## The optionality note

Opens a door: the union is now the place any further "why this role cannot be
deleted" reason is added, and both clients have a pattern for rendering one.

It defers one: `RoleOwnsDataProducts` carries no payload, so neither client can
name *which* data products block the delete — an admin with many of them has no
list to work from. The message therefore describes the blocker without
prescribing a remedy, unlike its `RoleAssigned` sibling. A field on the type
would fix that and can be added without a breaking change, but it belongs with
the registry that owns the type, not here.

It also leaves standing the wider question the schema raises itself, in the
`# XXX: consider making these OperationErrors` comment above this run of
payload-free marker types: each new refusal reason currently costs a type, a
union member, a TypeScript arm and a Python branch across three languages.

<h2><a href="https://app.greptile.com/api/retrigger?id=66071443"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with the shared schema, generated clients, public handlers, and tests aligned.

<h3>Summary</h3>

This PR propagates the new `RoleOwnsDataProducts` role-deletion result through the shared GraphQL contract, generated Catalog and Python clients, and both public-facing error paths.
- Catalog administrators now receive a specific notification when owned data products prevent role deletion.
- `quilt3.admin.roles.delete()` maps the result to the new public `RoleOwnsDataProductsError`.
- Generated clients, changelogs, and focused Catalog and Python tests are updated consistently.

<h3>Diagram</h3>

```mermaid
flowchart LR
  Registry[Registry roleDelete] --> Result[RoleOwnsDataProducts]
  Result --> Schema[Shared GraphQL schema]
  Schema --> CatalogTypes[Generated Catalog types]
  Schema --> PythonModels[Generated Python models]
  CatalogTypes --> CatalogUI[Catalog notification]
  PythonModels --> PythonAdapter[quilt3 admin adapter]
  PythonAdapter --> PythonError[RoleOwnsDataProductsError]
```

<sub>Reviews (2) · Last reviewed commit: ["Correct what the quilt3 changelog says t..."](https://github.com/quiltdata/quilt/commit/09b8b236dbf23fd38e15f4e9a22682fce497e6a3)</sub>


nl0:
@greptileai

PR #5305, opened 2026-09-18
Catalog: make a bucket's bulk scan scope editable

QuiltSimon:
Direction F of the admin-ux package. Closes UNK-12, ruled 2026-09-18: ship the field now with honest copy.

## What changed

`prefixes` was pinned to `null` on both the add and edit forms, with a note that the field is managed through the quilt3 SDK, so an admin could not see or change it from the console at all.

| | Before | After |
|---|---|---|
| Add / Edit form | `prefixes: () => null` | editable, one prefix per line |
| Existing value | invisible | round-trips from `bucket.prefixes` |
| Empty field | n/a | sends `null`, so the field stays out of the input |

## The copy is the load-bearing part

It is labelled **Bulk scan scope**, not indexing scope, because that is all it governs: the field decides which bulk scanner jobs are enqueued.

Traced at `enterprise@7c9bada3`: the S3 notification configuration is put with no `Filter` (`bucket.py:915-922`), and the indexer lambda reads only `content_extensions` / `content_bytes` (`document_queue.py:40-56`). So objects written **outside** the scope are still indexed and still searchable. Narrowing on an existing bucket assigns the field and calls `update_indexer_conf()`, which carries neither prefixes nor a reindex, so nothing already in Elasticsearch is removed.

The helper text says both of those rather than implying a boundary that holds in neither direction. Prefix-scoped live indexing and cleanup-on-narrow are a registry ask, not this change.

## Notes on the diff

- One prefix per line, since an S3 key may legally contain a comma.
- `[""]` and `[]` both mean the whole bucket to the registry, so `[""]` round-trips as an empty field and an empty field sends `null`.
- `BucketConfigSelection` now selects `prefixes`; `gql:generate` re-ran, which is why four `.generated.ts` files are in the diff. The fragment is used only under `Admin/Buckets/`, and admins get the field via `get_buckets_listable_by`, so no non-admin query gains an `@admin` field.

## Verified

`npm run typecheck` clean for this file (the one remaining error, a missing `mermaid` module, is pre-existing on `dev` — confirmed by stashing this change and re-running). `oxfmt --check` and `oxlint` clean via pre-commit.

Out of scope, deliberately: the add stepper, readiness, single-save (I), bulk toolbar (E, already shipped).

<h2><a href="https://app.greptile.com/api/retrigger?id=66221212"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR should not merge until the form preserves whitespace-sensitive S3 prefixes; the changelog reference should also be corrected.

<h2>Findings</h2>

1. P1&nbsp;**Prefix whitespace is lost** <a href="https://github.com/quiltdata/quilt/pull/5305#discussion_r4050461776">▶</a>
2. P2&nbsp;**Pull request link is invalid** <a href="https://github.com/quiltdata/quilt/pull/5305#discussion_r4050461783">▶</a>

<h3>Summary</h3>

This PR exposes a bucket's bulk-scan prefix scope in the administrative add and edit forms.
- Selects and displays existing prefix configuration.
- Converts newline-separated form input into the bucket mutation input.
- Adds explanatory copy distinguishing bulk scanning from live indexing and search.
- Regenerates the affected GraphQL fragment and operation artifacts.
- Contains one prefix-normalization defect and one unresolved changelog link.

<h3>Diagram</h3>

```mermaid
flowchart LR
  A[BucketConfig.prefixes] --> B[Join prefixes with newlines]
  B --> C[Bulk scan scope field]
  C --> D[Split into lines]
  D --> E[Trim, remove empty values, deduplicate]
  E --> F{Any prefixes remain?}
  F -->|Yes| G[Submit string array]
  F -->|No| H[Submit null]
  G --> I[Bucket add or update mutation]
  H --> I
```

<sub>Reviews (1) · Last reviewed commit: ["Catalog: make a bucket&#39;s bulk scan scope..."](https://github.com/quiltdata/quilt/commit/ce311729502ba12de32700558df46e052fd7ce4c)</sub>


QuiltSimon:
## Pre-merge review: 7 findings, 5 fixed, 2 dismissed in writing

Reviewed at `ce3117295`; fixes are in `3555c0009`.

Two of these were real bugs. Both verified against source before fixing.

### Fixed

**1. `minRows`/`maxRows` are MUI v5 props; this catalog is v4.** `TextField.js` destructures `rows`/`rowsMax` only, so both fell into `other` and were spread onto the root div: React warns about an unrecognized DOM prop, and the textarea rendered at one row with no cap, growing unbounded as prefixes are pasted. TypeScript missed it because `FieldProps` carries an index signature. Now `rowsMax`, matching three sibling fields in this same file. Confirmed `grep -rl minRows node_modules/@material-ui/core/TextField/` returns nothing.

**2. The copy claimed prefixes govern the scanner "and nothing else". They also gate Iceberg.** `create_bucket` skips `iceberg.add_bucket(name)` unless every package-metadata prefix falls inside the scope (`bucket.py:547`), so a scope excluding `.quilt/` silently creates the bucket without Iceberg registration. Now stated on the field.

**3. Blank does not mean "leave the scope alone".** `normalize_prefixes(None)` returns `ROOT_PREFIX_ACCESS` (`[""]`), so clearing the box **widens** a narrowed bucket to whole-bucket and re-verifies root permissions. My code comment asserted the opposite. Both the comment and the helper text now say blank means the whole bucket.

**4. No validation.** A prefix pasted from the S3 console as `s3://bucket/data/` or `/data/` was stored verbatim — the registry only trims and appends a trailing slash — so the scan matched zero keys with nothing surfaced anywhere. Validated now, as every sibling field in this form already was.

**5. Changelog placeholder** replaced with the real PR number.

### Dismissed

**6. The "Re-index and repair" warning was not extended to `prefixes`.** Correct that widening scope enqueues nothing, so search stays narrow until someone re-indexes, and that prefixes have the same "affects files indexed after the change" property the existing alert exists for. Dismissed as a scope call, not a disagreement: that alert's guard covers the three deep-indexing fields, and adding a fourth trigger changes when an existing warning fires for existing fields. Worth doing with the reindex job panel, where the backfill story is the subject rather than a side effect.

**7. `normalizePrefixes` duplicates the inline `tags` pipeline, plus a dead null-guard.** The duplication is real — `tags` does split/trim/reject/uniq inline with a different separator. Dismissed on the grounds that extracting a shared helper now would change `tags`' behaviour in a PR that is supposed to be direction F only; the separator differs precisely because a key may contain a comma. The null-guard point is right that `prefixes` is `[String!]!`, but I am leaving `(bucket.prefixes || [])` in place: the field was absent from the fragment until this commit, and a defensive read costs nothing.

### Ruled out during review

- **No GraphQL schema drift.** `shared/graphql/schema.graphql:213` already ships `prefixes`, and so does registry `origin/master` — the UI is not ahead of its backend.
- **Per-card Edit submits round-trip correctly**, because `Edit`'s `onSubmit` merges `bucketToFormValues(bucket)` before `applySpec`.
- **`\r\n` pastes** are handled by the `R.trim` in the pipeline.
- `npm run typecheck` reports zero errors in this file. One error remains in the repo (`Cannot find module 'mermaid'`); I confirmed it is pre-existing on clean `dev` by stashing this change and re-running.

